PR #3, opened 2015-07-02
Sample test for AliRoot

dberzano:
Defined with a recipe, like a package.

### DIFF
--- a/aliroot-test.sh
+++ b/aliroot-test.sh
@@ -1,0 +1,7 @@
+package: aliroot-test
+version: v1
+requires:
+  - aliroot
+---
+#!/bin/sh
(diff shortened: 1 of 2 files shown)


aliroot-test.sh
```diff
@@ -0,0 +1,7 @@
+package: aliroot-test
+version: v1
+requires:
+  - aliroot
+---
+#!/bin/sh
+aliroot -b BUILD/slc7-x86_64/aliroot/v5-06-28-1/aliroot/test/gun/sim.C
```
ktf: BUILD is not packaged and you should not be used. Use $ALICE_ROOT or $ALIROOT_ROOT.

PR #10, opened 2015-07-22
Add build recipes for ThePEG to master

ktf:


ktf:
Is there any reason why we should not have these in the master? They seems to be completely orthogonal to me, or you have a special reason for that?

Also, Predrag was confused about why the module files are different from the ones currently in CVMFS, what is the rationale about it?

dberzano:
Hi @ktf, the reason was only me being on vacation. I wanted to show them to Predrag before merging an agreed version on the master.

Modulefiles are different for two extremely good reasons (I'll explain them to Predrag tomorrow):

1. Our new Modulefiles read an environment variable allowing for _overriding_ what's on CVMFS. This is good if you want to test a build in an environment where you have both CVMFS components, and yet-to-be-published ones. I have used it in order to test ThePEG in a realistic configuration, without them I could not test them.

2. Current Modulefiles for ThePEG, YODA and Rivet miss important environment variables (notably, `PYTHONPATH`) that our new Modulefiles have.

There was a complaint against software names capitalization: the only thing I should amend is *FastJet* that should be called, for compatibility, *fastjet*. Also GSL has a version already published, but it is broken and it never worked, so I have actually proposed to eliminate it and keep our current convention.

ktf:
Ok, thanks.

PR #18, opened 2015-07-23
Recipes for Binutils and GCC

dberzano:
Note: for the moment no other recipe depends on GCC or Binutils.

### DIFF
--- a/gcc.sh
+++ b/gcc.sh
@@ -1,0 +1,40 @@
+package: GCC
+version: 4.9.3
+source: https://github.com/dberzano/gcc 
+tag: alice/4.9.3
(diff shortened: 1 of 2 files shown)


gcc.sh
```diff
@@ -0,0 +1,40 @@
...
+---
+#!/bin/bash -e
+
+case $ARCHITECTURE in
+  osx*) AdditionalLanguages=',objc,obj-c++' ;;
+esac
+
+cd "$SOURCEDIR"
+
+# TODO: maybe, have them in aliBuild?
```
ktf: definitely for alibuild

PR #32, opened 2015-07-30
Build ROOT in relocatable mode.

ktf:


ktf:
@dberzano this works for me.

PR #63, opened 2015-08-27
Several changes to build prerequisites for ThePEG on Ubuntu

qgp:
- add recipe for CGAL
- replace cvmfs installations with proper dependencies for fastjet, Rivet, YODA
- add module file for HepMC
- build boost with C++11 only when not on SLC 5
- add .gitignore


ktf:
Good for now. Will merge.

PR #64, opened 2015-08-27
updates for compilation on SLC 5

qgp:
Compilation on SLC 5 is a bit tricky, it requires special treatment in several packages, for pythia8 and yaml-cpp patching the build flow.

### DIFF
--- a/lhapdf.sh
+++ b/lhapdf.sh
@@ -3,12 +3,18 @@ version: v6.1.5
 source: https://github.com/alisw/LHAPDF
 tag: v6.1.5
 requires:
-  - yaml-cpp
+ - yaml-cpp
+ - boost
+ - "autotools:(slc5).*"
 ---
 #!/bin/sh
 # Does not work out of source.
 
 cd $SOURCEDIR
+
+# autoreconf if using our own build of autotools
+[[ -n ${AUTOTOOLS_ROOT} ]] && autoreconf -ivf
+
 ./configure --prefix=$INSTALLROOT \
             --with-boost=${BOOST_ROOT} \
             --with-yaml-cpp=${YAML_CPP_ROOT}
--- a/pythia8.sh
+++ b/pythia8.sh
@@ -4,14 +4,23 @@ source: https://github.com/alisw/pythia8
 requires:
   - LHAPDF
   - HepMC
+  - boost
 tag: alice/v8210
 ---
 #!/bin/sh
 cd $SOURCEDIR
+
 ./configure --prefix=$INSTALLROOT \
             --enable-shared \
             --with-hepmc2=${HEPMC_ROOT} \
-            --with-lhapdf6=${LHAPDF_ROOT}
+            --with-lhapdf6=${LHAPDF_ROOT} \
+            --with-boost=${BOOST_ROOT}
+
+if [[ $ARCHITECTURE =~ "slc5.*" ]]; then
+    ln -s LHAPDF5.h include/Pythia8Plugins/LHAPDF5.cc
+    ln -s LHAPDF6.h include/Pythia8Plugins/LHAPDF6.cc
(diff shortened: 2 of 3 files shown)


lhapdf.sh
```diff
@@ -3,12 +3,22 @@ version: v6.1.5
...
+ - "autotools:(slc5).*"
 ---
 #!/bin/sh
 # Does not work out of source.
 
 cd $SOURCEDIR
+
+# autoreconf is needed for SLC 5
+# because we have to use built autotools
+if [[ $ARCHITECTURE =~ "slc5.*" ]]; then
```
ktf: Why not key-ing on the existance of $AUTOTOOLS_ROOT, rather than the architecture? This way you do not need to keep the regexp in sync in two different places.

pythia8.sh
```diff
@@ -4,14 +4,24 @@ source: https://github.com/alisw/pythia8
...
 #!/bin/sh
 cd $SOURCEDIR
+
+# jkl: boost headers are required for LHAPDF6
 ./configure --prefix=$INSTALLROOT \
             --enable-shared \
             --with-hepmc2=${HEPMC_ROOT} \
-            --with-lhapdf6=${LHAPDF_ROOT}
+            --with-lhapdf6=${LHAPDF_ROOT} \
+	    --with-boost=${BOOST_ROOT}
```
ktf: Please respect indentation.

pythia8.sh
```diff
@@ -4,14 +4,24 @@ source: https://github.com/alisw/pythia8
...
+
+if [[ $ARCHITECTURE =~ "slc5.*" ]]; then
+    ln -s LHAPDF5.h include/Pythia8Plugins/LHAPDF5.cc
+    ln -s LHAPDF6.h include/Pythia8Plugins/LHAPDF6.cc
+    sed -i -e 's#\$(CXX) -x c++ \$< -o \$@ -c -MD -w -I\$(LHAPDF\$\*_INCLUDE) \$(CXX_COMMON)#\$(CXX) -x c++ \$(<:.h=.cc) -o \$@ -c -MD -w -I\$(LHAPDF\$\*_INCLUDE) \$(CXX_COMMON)#' Makefile
+fi
```
ktf: Any reason not to do this for each architecture?

lhapdf.sh
qgp: makes sense, I'll change this

pythia8.sh
qgp: Sorry, overlooked a tab, for me it looked ok.

pythia8.sh
qgp: This works around a compiler bug (feature?) that builds a precompiled header instaed of an object file if called on a .h file despite being told differently. I don't see why we should add workarounds on systems for which they are not needed. But that's up to you.

ktf: Leave it for now. We should try to see if upstream is willing to fix 
things in their code.

--
Ciao,
Giulio

qgp: It works with any half-way modern compiler, the problem really is that SLC 5 is so outdated and no one in the MC community is using and testing with such historic compilers.

ktf: Ok, hopefully we will forget about slc5 at some point.

lhapdf.sh
```diff
@@ -3,12 +3,19 @@ version: v6.1.5
 source: https://github.com/alisw/LHAPDF
 tag: v6.1.5
 requires:
-  - yaml-cpp
+ - yaml-cpp
+ - boost
+ - "autotools:(slc5).*"
+force_rebuild: true
```
ktf: Why force_rebuild?

lhapdf.sh
```diff
@@ -3,12 +3,19 @@ version: v6.1.5
...
 ---
 #!/bin/sh
 # Does not work out of source.
 
 cd $SOURCEDIR
+
+# autoreconf if using our own build of autotools
+[[ -n ${AUTOTOOLS_ROOT} ]] && autoreconf -ivf
```
ktf: This will exit with 1 if ${AUTOTOOLS_ROOT} is not defined, no? We run recipes with /bin/bash -e so I'm not sure how this can work.

qgp: Neither for commands after while, until, if or in tests, nor for commands in chain with && or || does bash -e exit if a command fails - except for the last one (here autoreconf). And this is what we want.

I had tested this with SLC 5 and Ubuntu.

ktf: Yeah, correct. My bad.

PR #70, opened 2015-09-01
Fix issues with the O2 builds not getting to the end.

ktf:


ktf:
@qgp Looks like we need autotools also on slc7 (centos7). Any objections to simply using it anywhere?

qgp:
No objections from my side.

PR #75, opened 2015-09-02
AliEn-Runtime needs CMake

dberzano:


ktf:
As discussed this breaks builds due to a bug in the topological sort of
dependencies. A proper solution will be provided next week but for the
moment you should be able to work it around by dropping cmake requires from
root.

PR #82, opened 2015-09-08
AliRoot with FastJet and various Modulefiles

dberzano:


dberzano:
@qgp: could you please check/adjust your ThePEG installation according to this? Thanks!

PR #84, opened 2015-09-10
compileable versions of ThePEG, JEWEL, CRMC:

qgp:
- fixes in deps (capitalization)
- adding JEWEL
- adding CRMC

### DIFF
--- a/crmc.sh
+++ b/crmc.sh
@@ -1,0 +1,41 @@
+package: CRMC
+version: v1.5.4
+requires:
+  - boost
+  - HepMC
+---
+#!/bin/bash -ex
+
+ARCHIVE="crmc-${PKGVERSION#v}-src.tgz"
+URL="http://service-spi.web.cern.ch/service-spi/external/MCGenerators/distribution/crmc/${ARCHIVE}"
+
+cd $SOURCEDIR
+curl -O "${URL}"
+tar xz --strip-components 2 -f ${ARCHIVE}
+
+cd $BUILDDIR
+
+cmake $SOURCEDIR -DCMAKE_INSTALL_PREFIX=$INSTALLROOT
+
+make ${JOBS+-j $JOBS} all
+make install
+
+# Modulefile
+MODULEDIR="$INSTALLROOT/etc/modulefiles"
+MODULEFILE="$MODULEDIR/$PKGNAME"
+mkdir -p "$MODULEDIR"
+cat > "$MODULEFILE" <<EoF
+#%Module1.0
+proc ModulesHelp { } {
+  global version
+  puts stderr "ALICE Modulefile for $PKGNAME $PKGVERSION-@@PKGREVISION@$PKGHASH@@"
+}
+set version $PKGVERSION-@@PKGREVISION@$PKGHASH@@
+module-whatis "ALICE Modulefile for $PKGNAME $PKGVERSION-@@PKGREVISION@$PKGHASH@@"
+# Dependencies
+module load BASE/1.0 boost/$BOOST_VERSION-$BOOST_REVISION HepMC/$HEPMC_VERSION-$HEPMC_REVISION
+# Our environment
+setenv CRMC_ROOT \$::env(BASEDIR)/$PKGNAME/\$version
+prepend-path PATH $::env(CRMC_ROOT)/bin
+prepend-path LD_LIBRARY_PATH $::env(CRMC_ROOT)/lib
+EoF
--- a/jewel.sh
+++ b/jewel.sh
@@ -1,0 +1,44 @@
+package: JEWEL
+version: v2.0.2
+requires:
+  - lhapdf5
+---
+#!/bin/bash -ex
+
+ARCHIVE="jewel-2.0.2.tar.gz"
+URL="https://www.hepforge.org/archive/jewel/${ARCHIVE}"
+
+cd $SOURCEDIR
+curl -O "${URL}"
+tar xz -f ${ARCHIVE}
+
+cd $BUILDDIR
+rsync -a $SOURCEDIR/ $BUILDDIR/
+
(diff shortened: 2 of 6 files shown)


dberzano:
Hello @qgp, could you change lhapdf5 -> lhapdf? We use Git branches to distinguish between versions, for instance we have the master branch using ROOT 5 and a "ROOT 6" branch with the same recipes but ROOT version 6. Thanks!

qgp:
The issue is depending on lhapdf v5, see alisw/alibuild#64 where @ktf suggested to use lhapdf and lhapdf5. The situtation is somewhat similar to pythia6 and pythia8. I can change this if there is another way to depend on a specific version of LHAPDF.

crmc.sh
```diff
@@ -0,0 +1,41 @@
...
+proc ModulesHelp { } {
+  global version
+  puts stderr "ALICE Modulefile for $PKGNAME $PKGVERSION-@@PKGREVISION@$PKGHASH@@"
+}
+set version $PKGVERSION-@@PKGREVISION@$PKGHASH@@
+module-whatis "ALICE Modulefile for $PKGNAME $PKGVERSION-@@PKGREVISION@$PKGHASH@@"
+# Dependencies
+module load BASE/1.0 yaml-cpp/$YAML_CPP_VERSION-$YAML_CPP_REVISION boost/$BOOST_VERSION-$BOOST_REVISION autotools/$AUTOTOOLS_VERSION-$AUTOTOOLS_REVISION
+# Our environment
+setenv LHAPDF_ROOT \$::env(BASEDIR)/$PKGNAME/\$version
```
dberzano: Watch out, here it should be `CRMC_ROOT` instead of `LHAPDF_ROOT`, could you fix it?

jewel.sh
```diff
@@ -0,0 +1,44 @@
...
+#%Module1.0
+proc ModulesHelp { } {
+  global version
+  puts stderr "ALICE Modulefile for $PKGNAME $PKGVERSION-@@PKGREVISION@$PKGHASH@@"
+}
+set version $PKGVERSION-@@PKGREVISION@$PKGHASH@@
+module-whatis "ALICE Modulefile for $PKGNAME $PKGVERSION-@@PKGREVISION@$PKGHASH@@"
+# Dependencies
+module load BASE/1.0
+# Our environment
```
dberzano: Here you should set `JEWEL_ROOT`.

lhapdf5.sh
```diff
@@ -0,0 +1,42 @@
...
+proc ModulesHelp { } {
+  global version
+  puts stderr "ALICE Modulefile for $PKGNAME $PKGVERSION-@@PKGREVISION@$PKGHASH@@"
+}
+set version $PKGVERSION-@@PKGREVISION@$PKGHASH@@
+module-whatis "ALICE Modulefile for $PKGNAME $PKGVERSION-@@PKGREVISION@$PKGHASH@@"
+# Dependencies
+module load BASE/1.0 yaml-cpp/$YAML_CPP_VERSION-$YAML_CPP_REVISION boost/$BOOST_VERSION-$BOOST_REVISION autotools/$AUTOTOOLS_VERSION-$AUTOTOOLS_REVISION
+# Our environment
+setenv LHAPDF_ROOT \$::env(BASEDIR)/$PKGNAME/\$version
```
dberzano: Since the package is `lhapdf5`, then `LHAPDF_ROOT` -> `LHAPDF5_ROOT`.

thepeg.sh
```diff
@@ -6,29 +6,33 @@ requires:
...
 
-Pythia8='/cvmfs/alice.cern.ch/x86_64-2.6-gnu-4.1.2/Packages/pythia/v8186'
-HepMC='/cvmfs/alice.cern.ch/x86_64-2.6-gnu-4.1.2/Packages/HepMC/v2.06.09'
-LhaPDF='/cvmfs/alice.cern.ch/x86_64-2.6-gnu-4.1.2/Packages/lhapdf/v5.9.1'
-Cgal='/cvmfs/alice.cern.ch/x86_64-2.6-gnu-4.1.2/Packages/cgal/v4.4'
-Boost='/cvmfs/alice.cern.ch/x86_64-2.6-gnu-4.1.2/Packages/boost/v1_53_0'
+export LD_LIBRARY_PATH="${LHAPDF5_ROOT}/lib:${CGAL_ROOT}/lib:${BOOST_ROOT}/lib:${LD_LIBRARY_PATH}"
+export LIBRARY_PATH="${LHAPDF5_ROOT}/lib:${CGAL_ROOT}/lib:${BOOST_ROOT}/lib:${LIBRARY_PATH}"
+export LHAPATH="${LHAPDF5_ROOT}/share/lhapdf"
+export PERLLIB=/usr/share/perl5
```
dberzano: The AliEn-Runtime package contains a version of perl5, but you can leave it like this, I'll take care of it.

qgp: But then we have to depend on AliEn-Runtime, don't we? And alien is not really required here. So, this seems like a strange way to pull in perl.

qgp:
Thanks for the comments, I'll update the pull request accordingly.

qgp:
Hi @dberzano, I updated the pull request and now also changed the dependencies in the module files to consistently use $PKG_VERSION-$PKG_REVISION. I assume you prefer this rather than having hardcoded versions. Thanks!

thepeg.sh
```diff
@@ -6,29 +6,33 @@ requires:
...
 
-find . \
-  -name configure.ac -or \
-  -name aclocal.m4 -or \
-  -name configure -or \
-  -name Makefile.am -or \
-  -name Makefile.in \
-  -exec touch '{}' \;
-autoreconf -ivf
+#find . \
```
ktf: Please try to avoid commenting code.

qgp: Understood. Here, I left it in on purpose because Dario said he needed that part, while I did not now (on Ubuntu). And we might indeed need it for SLC 5 which we can't build and test yet because of the missing dependencies (as a result of the lack of python).

PR #89, opened 2015-09-15
Append -0 to GEANT3 legacy version names

dberzano:


dberzano:
@ktf: we can avoid conflicts with packages produced by the old build system like this.

Issue #81, opened 2015-09-08
Various issues with Alien-runtime


ktf:
- [ ] repository should not be under dberzano.
- [ ] fails to build on mac with the following. It can be worked around by dropping the buildtime target from the Makefile.

```
DEBUG:AliPhysics:7bda97d1: svn: E155007: '/Users/ktf/work/active/sw/BUILD/55f057ced71040939e96078d881f5a7781302590/AliEn-Runtime/apps/tools/eugridpma-carep-classic' is not a working copy
DEBUG:AliPhysics:7bda97d1: date: illegal time format
DEBUG:AliPhysics:7bda97d1: usage: date [-jnu] [-d dst] [-r seconds] [-t west] [-v[+|-]val[ymwdHMS]] ...
DEBUG:AliPhysics:7bda97d1:             [-f fmt date | [[[mm]dd]HH]MM[[cc]yy][.ss]] [+format]
DEBUG:AliPhysics:7bda97d1: ERROR getting Makefile time
DEBUG:AliPhysics:7bda97d1: make[2]: *** [buildtime] Error 1
DEBUG:AliPhysics:7bda97d1: make[1]: *** [dep-../../..//apps/tools/eugridpma-carep-classic] Error 2
DEBUG:AliPhysics:7bda97d1: make: *** [dep-../..//apps/alien/alien-ca] Error 2
DEBUG:AliPhysics:7bda97d1:
```
- [ ] In various places `cp -d` is used in place of `cp -P` (or even just cp)
- [ ] Does not compile with clang due to the fact `-lgcc` is specified.

dberzano:
In general I think this recipe was never meant to be built on Mac. The AliEn-antidot repository has to be considered "transient" as it is in practice "a build system inside our build system" and we need to translate current "antidot" recipes to "our" recipes.

This is going to happen in the next weeks with the help of Miguel and Costin. For the time being, a smooth solution which allows us to keep having Mac builds is using the `alien-installer` command there.

ktf:
Well, given that I do want to run on mac, and that we were able to run on mac before this, we need to find a way to have everything back to a working state on mac as well, ASAP. I propose we resurrect the old alien recipe, just for the OSX case (we can use platform specific requires for that).

ktf:
I just noticed we meant the same thing. Trying that now.

dberzano:
Open the PR when ready, let me check it first :-)

Thanks!

ktf:
[...]

PR #93, opened 2015-09-15
Use old AliEn builds on OSX

ktf:
This is due to the fact that the new AliEn-runtime does not actually
work on OSX. This works around the problem by using the old recipe.

Since the old recipe did not bring in libxml2, we use our own libxml2 on
OSX. See also #81.


ktf:
@dberzano

PR #94, opened 2015-09-15
Add GMP and MPFR as CGAL dependencies

ktf:
Without this, they get picked up from the system and do not work in the
SLC7 docker images where gmp-devel is not installed. Notice I also need the new automake in order to compile MPRF.

### DIFF
--- a/autotools.sh
+++ b/autotools.sh
@@ -1,7 +1,7 @@
 package: autotools
-version: v1.0.0
+version: v1.1.0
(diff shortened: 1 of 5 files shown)


ktf:
@dberzano objections for merging this in the master?

dberzano:
Let me check.

dberzano:
Thanks for those recipes, my PR with Python & co. is coming soon.

PR #97, opened 2015-09-16
LHAPDF: fix bug with recent automake

### LINKED ISSUE #96: lhapdf does not compile with newer autotool
After #94, lhapdf does not compile anymore: the new automake does not set somehow `$(buiddir)`.

dberzano:
Fixes #96.


ktf:
GIven we have our own LHAPDF repository, it would be better to fix it there, rather than patching in the build recipe, so that we can contribute back.

PR #103, opened 2015-09-17
Merge master into IB/v5-06/prod

ktf:
This merges master in `IB/v5-06/prod`. Once this is done, we should change the default branch from master to `IB/v5-06/prod`.


dberzano:
Please do not merge this branch, and btw I was working on it in parallel... let me check it first :-)

PR #105, opened 2015-09-18
ThePEG and dependencies

### LINKED ISSUE #80: Package python 2.7
Python 2.4 or newer is required for YODA -> Rivet -> ThePEG (and other generators depending on Rivet). It is not available on SLC 5 and the lack of it is blocking the rebuild of Rivet. According to @ktf, it should be packaged.

dberzano:
@qgp: this should fix #80 and correct many small-to-large issues of #84.

@ktf: this autotools recipe has been tested with all SLCs and Ubuntu, but the original source code had to be modified, so I have created [alisw/autotools](https://github.com/alisw/autotools). I have realized that maybe star-externals is yours so we can just merge the changes there and drop ALICE's autotools.

This is the dependency plot for ThePEG (the dependency from AliEn-Runtime is because of openssl, it will go away as soon as we clean AliEn up).

dep_thepeg

@qgp: could you tell if the dependency tree is correct (at least for the Physics-related components)?

### DIFF
--- a/alien-runtime.sh
+++ b/alien-runtime.sh
@@ -8,7 +8,7 @@ prepend_path:
 env:
   GSHELL_ROOT: "$ALIEN_RUNTIME_ROOT/api"
   GSHELL_NO_GCC: "1"
-requires:
+build_requires:
   - CMake
 ---
 #!/bin/bash -e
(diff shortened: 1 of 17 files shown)


ktf:
We should introduce build_requires, which get merged to requires immediately (so they are not real build requires) but they would then get excluded from your diagram. Is there any reason why ThePeg depends on GSL and not CGAL? Is this because ThePeg itself uses GSL directly?

qgp:
@dberzano: Yes, the dependency tree looks fine to me.

@ktf: Yes, gsl is used by ThePEG.

dberzano:
OK, I am now testing alisw/alibuild#72, and this is the dependency plot:

dep_thepeg_simplified

which is a little bit simplified.

PR #108, opened 2015-09-22
Pick FastJet from source and tell AliPhysics about GMP and MPFR

dberzano:
This change should make IB/v5-06/next "useless" for now, as we can "override" via Jenkins the FastJet version.


dberzano:
Tested in [daily-aliphysics build number 36](https://alijenkins.cern.ch/job/daily-aliphysics/36).

PR #109, opened 2015-09-22
Patching lhapdf5 to avoid format security compilation error.

mpuccio:
I am trying to compile on Fedora 22 and on Mint 17.1 the O2 software using alibuild. 
On both the systems I got stuck due to a format security error in the compilation of lhapdf5.
This small patch solves this issue for me, but maybe you have cleaner solution.

### DIFF
--- a/lhapdf5.sh
+++ b/lhapdf5.sh
@@ -12,6 +12,7 @@ tar xz --strip-components 1 -f ${ARCHIVE}
 
 cd $BUILDDIR
 rsync -a $SOURCEDIR/ $BUILDDIR/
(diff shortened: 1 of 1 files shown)


ktf:
Thank you for your fix.

IMHO what we should do, though, is to create an `alisw/lhapdf5` repository and patch it there directly, so that people can se the actual code being built. Unless you need this immediately, I'm going to do it tomorrow and integrate your patch there.

@dberzano

dberzano:
Yes @ktf I second that, proceed.

@mpuccio, thanks for spotting this!

mpuccio:
Perfect! Thanks a lot, I close this pull request.

ktf:
Actually I will be in a training the whole day so it might take me a while.

PR #113, opened 2015-09-24
Drop tutorial 7 as it generates an ICE

ktf:
To be investigated. Is it due to memory issues?


ktf:
Being tested in 1​44​9

dberzano:
What do we do with this one?

ktf:
We should still fix the bug..

PR #114, opened 2015-09-25
Fix FastJet compilation error

ktf:
GMP is known to have troubles relocating when compiled as a shared
library.

This changes our build to use archive libraries for both GMP and MPFR so
that the problem does not happen when we reuse artifacts of a previous
build which was removed.


ktf:
@dberzano

ktf:
We should probably get rid of the module files as well, I guess.

PR #129, opened 2015-10-01
Use old recipe for slc6 / slc7 / ubuntu

ktf:
This is needed because the new one does not compile anymore and
bails out with a message:

```
DEBUG:AliRoot:0: + make install
DEBUG:AliRoot:0: ../../..//autodetect.sh: 32: shift: can't shift that many
```


dberzano:
Superseded by #144.

PR #132, opened 2015-10-03
Force build architecture to core2 on x86-64

ktf:
This is to avoid that GMP does too aggresive choices when compiled on
AMD, failing to work when is then used on Corei7 or other Intel CPUs.

This will hopefully fix fix the error:

```
checking for double-to-integer conversion bug... yes or failed to exec (exit status is 132)
configure: error: double-to-integer conversion is incorrect.
You need to use another compiler (or lower the optimization level).
```

and similar compilation errors which are seen at random during the builds of
MPFR / FastJet. The joys of cloud computing.

### DIFF
--- a/gmp.sh
+++ b/gmp.sh
@@ -4,7 +4,12 @@ source: https://github.com/alisw/GMP.git
 tag: v6.0.0
 ---
 #!/bin/sh
+case $ARCHITECTURE in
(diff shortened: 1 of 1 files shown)


ktf:
@dberzano this is potentially a problem for the production architecture as well. GMP by default is too aggressive in picking up compilation options.

PR #148, opened 2015-10-15
Working ThePEG on multiple platforms

dberzano:
Also:

- GMP with real destination arch (through `--host`)
- Parallel `make install` for AliRoot and AliPhysics (faster)


dberzano:
@qgp @mfloris we managed to get ThePEG compile on several platforms, tomorrow (Friday) we'll publish it.

PR #152, opened 2015-10-23
recipe for EPOS 3

qgp:
@dberzano, @ktf, @mfloris:
added a recipe for EPOS 3, using source from private gitlab repository

### DIFF
--- a/epos.sh
+++ b/epos.sh
@@ -1,0 +1,70 @@
+package: EPOS
+version: "%(tag_basename)s"
+tag: alice/v3.111
+source: ssh://git@gitlab.cern.ch:7999/ALICEPrivateExternals/epos.git
+requires:
+  - ROOT
+  - fastjet
+---
+#!/bin/bash -ex
+
+export EPOVSN=${PKGVERSION/./}
+export EPO=$PWD/
+export FASTSYS=$FASTJET
+export WORK=${EPO}Unu/
+export OBJ=${WORK}Lib/
+export CHK=${WORK}
+export OPT=${WORK}
+export HTO=${WORK}
+export DTA=${WORK}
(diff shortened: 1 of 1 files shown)


ktf:
Is it really so complicated the build procedure? It seems it could use quite some simplification. Merging since we do not build it by default.

qgp:
@ktf: I don't like this either. But the environment variables are used somewhere in the build process, and there is no install target. So, I don't see much room for simplification, unfortunately. I could try removing environment variables but that would be against what the instructions say and might result in breakage. I am open to suggestions, of course.

ktf:
I guess asking the author to learn cmake is not an option, right?

qgp:
I think we should give the feedback that the build process is not very suitable for non-dev installations. But as long as (also other) experiments are using it like this, there is little pressure on them. So, I don't think it's an option helping for the near future.

ktf:
Yeah, I guess we should provide the feedback. The main problem is that if you use common variable names like "WORK", the chances of having issues are close to 1.

dberzano:
@qgp, I have been thinking. If those variables are truly needed at runtime, maybe we could just have a wrapper script that sets them and calls the actual executable... this avoids polluting the environment with variable names that look pretty common...

qgp:
@ktf, @dberzano: agreed, this might hit us in the future. The idea of a wrapper script looks feasible to me (in fact, the executable usually called for EPOS by the user is just a shell script). So, I will see what is needed for at build and run time, respectively, and provide a wrapper script with the required settings.

Could you check if the build from gitlab works for you?

qgp:
filed pull request #154 with wrapper script and cleaner environment

PR #153, opened 2015-10-23
update modules file for PYTHIA8

qgp:
- add definition of PYTHIA8DATA to allow finding of the xml files
  also after the package has been moved
  (e.g. for ThePEG which otherwise remembers the path from the
  build server)


dberzano:
For consistency it should be set in the YAML header of the recipe too, I am going to fix it :-)

ktf:
We should have a utility which generates module files from environment.. ;)

PR #154, opened 2015-10-26
avoid pollution of environment variables:

qgp:
- remove unnecessary environment variables from modules file
- remove unncessary environment from build script
- add wrapper script setting the environment variables required at run time

which variables have to also go to the yaml header?


dberzano:
Merging for now, then I'll be working on it with some rebasing (like I did for #153) :-)

PR #177, opened 2015-11-25
Add debug defaults

ktf:


defaults-debug.sh
```diff
@@ -0,0 +1,10 @@
+package: defaults-release
```
dberzano: Shouldn't it be "defaults-debug"?

PR #179, opened 2015-11-26
Disable multithreading on Geant4

dberzano:
See ALIROOT-6439


dberzano:
Merging it right away, this is for a v5-05 of AliRoot (not the current one) and it has to be done. See #180 for the same request for the current prod branch.

PR #180, opened 2015-11-26
Disable multithreading on Geant4

dberzano:
See ALIROOT-6439


dberzano:
@ktf: this is required by AliRoot. Not sure about O2.

PR #201, opened 2015-12-08
Explicitly link CGAL in FastJet

dberzano:


dberzano:
Needed to make #200 work.

PR #242, opened 2016-03-09
Do not build GSL documentation

ktf:
Avoids texinfo requirement.


ktf:
Should fix alisw/alibuild#138.

PR #246, opened 2016-03-14
Integration of mesos-dds plugin

kvnnap:


ktf:
Merging this since it looks reasonable. If buggy, will open specific issues.

Issue #251, opened 2016-03-16
Missing dependencies when building O2 on slc6


vascobarroso:
While building O2 on slc6, I had to manually install the following RPMs:

- **texinfo** (missing makeinfo error while building MPFR)
- **perl-ExtUtils-Embed** (missing while building xalienfs)

ktf:
Regarding `perl-ExtUtils-Embed`, I am afraid that we will simply have to install it by hand. I will check with @dberzano if there is any simple solution, but I doubt. Notice also that you can disable alien with `--disable AliEn-Runtime`.

Regarding `texinfo` in MPFR, #253 should take care of it.

ktf:
BTW, why do you need MPFR? Because of GCC?

dberzano:
Hello @ktf, our GCC is built using its own MPFR, not the external one.

ktf:
That's what I meant. ;)

vascobarroso:
I'm just blindly following Giulio's O2 tutorial ;)

ktf:
Cool. :)

ktf:
Closing this in favour of #256 .

PR #253, opened 2016-03-16
Disable documentation in MPFR

ktf:


ktf:
Should solve at least part of #251

PR #254, opened 2016-03-17
Unset SIMPATH when building

ktf:
Not particularly good solution (SIMPATH might be picked up at runtime
accidentally) but already two people had this problem, so I try to limit
damage. Should really be cleaned up in O2 and FairRoot themselves.


ktf:
@vascobarroso

PR #258, opened 2016-03-23
GEANT4_VMC: improve examples location

dberzano:
Examples can be now accessed with:

    $G4VMCINSTALL/share/examples


dberzano:
@sawenzel: AliRoot scripts should now point to:

    $G4VMCINSTALL/share/examples/macro/g4libs.C

PR #260, opened 2016-03-23
No CXXFLAGS in AliRoot

dberzano:
Use CMAKE_BUILD_TYPE instead.


dberzano:
@aphecetche: this should solve your problem in [ALIROOT-6587](https://alice.its.cern.ch/jira/browse/ALIROOT-6587). In general you can use the defaults (via `alibuild --defaults release` for instance) to steer compilation options.

PR #262, opened 2016-03-28
GCC by default on SLC5 and ROOT with CMake

dberzano:


dberzano:
Note that 3165eba closes alisw/alibuild#162 (modulefiles logic for now is in the recipes).

PR #266, opened 2016-03-30
Set build and host in GCC-Toolchain

dberzano:


dberzano:
Currently testing before merging.

PR #278, opened 2016-04-06
add generators to meta package

qgp:

### DIFF
--- a/aligenerators.sh
+++ b/aligenerators.sh
@@ -1,0 +1,37 @@
+package: AliGenerators
+version: "v%(year)s%(month)s%(day)s"
+requires:
+  - CRMC
+  - JEWEL
+  - POWHEG
+  - pythia
+  - pythia6
+  - SHERPA
+  - ThePEG
+build_requires:
+  - ThePEG-test
+---
+#!/bin/bash -ex
+
+# Modulefile
+MODULEDIR="$INSTALLROOT/etc/modulefiles"
+MODULEFILE="$MODULEDIR/$PKGNAME"
+mkdir -p "$MODULEDIR"
+cat > "$MODULEFILE" <<EoF
+#%Module1.0
+proc ModulesHelp { } {
+  global version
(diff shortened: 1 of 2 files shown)


ktf:
Notice the generators package is currently used only by AliceO2.

qgp:
Hi @ktf, I added the generators required for ALICE here on explicit request by @dberzano. I leave it to you guys to figure out if this is useful for ALICE or not.

qgp:
you can also add sherpa.sh

dberzano:
Hello @qgp, thanks for this PR and sorry for having overlooked it.

I have two requests:

1. Could you, instead of changing the `generators` package (used by O2), create a new `AliGenerators` package?
2. This package should `require` the generators and `build_require` all the generator _tests_ we have.

dberzano:
As for the deployment: as discussed with @qgp the idea is to have to rebuild only `AliGenerators`: aliBuild is smart enough to rebuild only the things that have changed. `AliGenerators` would be a meta package but not like `AliEn-Runtime`: the single generators will be kept separated but this meta package would ease deployment in aliPublish and ease the rebuild procedure too.

dberzano:
As for versioning: the idea (again, as discussed with @qgp) is to use a date-based versioning for the `AliGenerators` package.

As we can see from [the documentation](http://alisw.github.io/alibuild/reference.html) this is already supported by aliBuild.

This allows for automatic versioning without manually changing the `version:` in the `aligenerators.sh` package.

Waiting for your refreshed PR :-)

dberzano:
For the record, @qgp: I have recently added a `DYLD_LIBRARY_PATH` thingy to all recipes, could you do the same for the generators while you refresh this PR? Thanks!

qgp:
@dberzano: here comes the udpated PR. There is still an issue with EPOS and the new gcc toolchain, I will look into this separately.

The DYLD_LIBRARY_PATH is already in place where relevant.

Let me know if you are fine with the PR now.

dberzano:
OK many thanks! Merging now.

PR #290, opened 2016-04-14
Ib/v5 08/next

sawenzel:
fixes to geant4.sh (necessary to run test/vmctest/gun with Geant4)


geant4.sh
```diff
@@ -6,13 +6,14 @@ build_requires:
   - CMake
 env:
   G4INSTALL: "$GEANT4_ROOT"
-  G4INSTALL_DATA: "$GEANT4_ROOT/share/Geant4-10.1.2"
+  G4INSTALL_DATA: "$GEANT4_ROOT/share/Geant4-10.1.3"
```
ktf: Any reason why you changed this? We would need to run this by the others using Geant4, I guess.

geant4.sh
```diff
@@ -6,13 +6,14 @@ build_requires:
...
   G4SYSTEM: "$(uname)-g++"
   G4LEVELGAMMADATA: "$G4INSTALL_DATA/data/PhotonEvaporation3.1"
   G4RADIOACTIVEDATA: "$G4INSTALL_DATA/data/RadioactiveDecay4.2"
   G4LEDATA: "$G4INSTALL_DATA/data/G4EMLOW6.41"
   G4NEUTRONHPDATA: "$G4INSTALL_DATA/data/G4NDL4.5"
-  G4NEUTRONXS: "$G4INSTALL_DATA/data/G4NEUTRONXS1.4"
+  G4NEUTRONXSDATA: "$G4INSTALL_DATA/data/G4NEUTRONXS1.4"
+  G4SAIDXSDATA: "$G4INSTALL_DATA/data/G4SAIDDATA1.1"
```
ktf: How big is this?

geant4.sh
sawenzel: See issue 289. The old value of $GINSTALL_DATA did not point to a valid location.

geant4.sh
sawenzel: Big as in size of installed data? I believe this data is already part of the installation. There was only the corresponding environment variable missing (in case of AIDXS).

PR #294, opened 2016-04-18
Separate python from its modules

ktf:


ktf:
Using #305 in place of this.

PR #297, opened 2016-04-19
Add libperl as system requirement

ktf:


ktf:
Requires latest version of alibuild to be effective.

PR #302, opened 2016-04-19
Make texinfo a system_requirement for GCC

ktf:


ktf:
Not needed anymore actually, thanks to #303.

Issue #282, opened 2016-04-07
Texinfo needed to compile root on SLC6


Barthelemy:
Building GCC-Toolchain@v4.9.3 it fails : 

> 
WARNING: `makeinfo' is missing on your system.  You should only need it if
         you modified a `.texi' or `.texinfo' file, or any other file
         indirectly affecting the aspect of the manual.  The spurious
         call might also be the consequence of using a buggy `make' (AIX,
         DU, IRIX).  You might want to install the `Texinfo' package or
         the `GNU make' package.  Grab either from any GNU archive site.
make[3]: *** [bfd.info] Error 1

It might be related to #251 where the same dependency was mentioned.

ktf:
Can you attach the full log?

Barthelemy:
> 
[log.txt](https://github.com/alisw/alidist/files/208442/log.txt)

Barthelemy:
to test : eulisse@amorebuild1   
make sure that texinfo is not installed.

Barthelemy:
still open

ktf:
I managed to compile GCC using #303 on `aliamore01`. Closing this. Reopen if still problematic.

PR #303, opened 2016-04-19
Disable makeinfo in GCC

ktf:
Since we do not want to depend on makeinfo / texinfo, we disable it everywhere.


ktf:
Should fix #282.

PR #305, opened 2016-04-20
Separate python from its modules

ktf:
This allow us to use the system python and only build the few modules
which might be missing on the system (i.e matplotlib, certifi, bumpy). This also should fix the case in which `freetype` or `libpng` come from the system.


ktf:
slc5, slc6, slc7 and ubuntu compiled fine. Rivet-test fails due to unrelated issue. Merging.

PR #306, opened 2016-04-20
add Rivet directory to PYTHONPATH

qgp:
Required to make the python-based rivet utilities work.

@ktf: is it ok to use python2.7 hardcodedly in the path?


ktf:
At the moment we either require python 2.7 from the system or we compile it ourself, so we should be fine. We need to think of a better way, of course.

PR #308, opened 2016-04-20
add test for Rivet

qgp:
first draft for testing Rivet using ThePEG for producing an input HepMC file 
(this includes testing the python installation, we refrain from running rivet-mkhtml to avoid requiring latex)

@ktf: there is still a problem with rivet-cmphistos as the PYTHONPATH is not set when Rivet is run from the test. When I load the environment interactively, it works fine. Am I overlooking something obvious?

### DIFF
--- a/rivet-test.sh
+++ b/rivet-test.sh
@@ -1,0 +1,47 @@
+package: Rivet-test
+version: v1
+force_rebuild: 1
+requires:
+  - ThePEG
+  - Rivet
+---
+#!/bin/bash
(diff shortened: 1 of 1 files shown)


ktf:
Can I ask you to comply to the guidelines at:

https://github.com/alisw/alidist#guidelines-for-commit-messages

I.e. can you make "add" with capital initial? Thanks.

ktf:
Jenkins tested this and it dies because of the issue fixed #306. Since nothing depends on it, I merge this, now that we have automated tests.

PR #311, opened 2016-04-20
Add missing python path to yaml header

qgp:


ktf:
Rivet seems to be fixed, however yoda python path seems to be broken as well. Given this is self consitent I will merge it, though.

PR #314, opened 2016-04-20
Fix yaml headers for YODA, Rivet

qgp:
- Rivet: remove appended $PYTHONPATH
- YODA: replace $(yoda-config ...) with explicit path

@ktf: I had copied the line with PYTHONPATH from python-modules.sh (https://github.com/ktf/alidist/blob/python-split/python-modules.sh), also there still is the second PYTHONPATH appended again.

### DIFF
--- a/rivet.sh
+++ b/rivet.sh
@@ -7,7 +7,7 @@ requires:
   - HepMC
   - boost
 prepend_path:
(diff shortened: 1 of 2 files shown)


ktf:
slc5, slc6 and slc7 now works correctly (including Rivet-test). Ubuntu MIA, but unrelated. Merging.

PR #316, opened 2016-04-21
Add Sherpa

qgp:
Adding SHERPA:
- building of documentation disabled for ALICE, anyway fails because of missing files for makeinfo

### DIFF
--- a/sherpa.sh
+++ b/sherpa.sh
@@ -1,0 +1,40 @@
+package: SHERPA
+version: "%(tag_basename)s"
+source: https://github.com/alisw/SHERPA
+tag: "alice/v2.1.1"
+build_requires:
+  - autotools
+  - GCC-Toolchain
+---
+#!/bin/bash -e
+
+rsync -a --delete --exclude '**/.git' --delete-excluded $SOURCEDIR/ ./
+
+autoreconf -ivf
+./configure --prefix=$INSTALLROOT \
+            --with-sqlite3=install
+
+make ${JOBS+-j $JOBS}
+make install
+
+# Modulefile
+MODULEDIR="$INSTALLROOT/etc/modulefiles"
+MODULEFILE="$MODULEDIR/$PKGNAME"
+mkdir -p "$MODULEDIR"
+cat > "$MODULEFILE" <<EoF
+#%Module1.0
+proc ModulesHelp { } {
+  global version
+  puts stderr "ALICE Modulefile for $PKGNAME $PKGVERSION-@@PKGREVISION@$PKGHASH@@"
+}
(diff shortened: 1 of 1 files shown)


sherpa.sh
```diff
@@ -0,0 +1,40 @@
+package: SHERPA
+version: "%(tag_basename)s"
+source: https://github.com/alisw/sherpa
```
ktf: The name of the repository is SHERPA (capital)

sherpa.sh
```diff
@@ -0,0 +1,40 @@
...
+  - autotools
+  - GCC-Toolchain
+---
+#!/bin/bash -e
+
+rsync -a --delete --exclude '**/.git' --delete-excluded $SOURCEDIR/ ./
+
+autoreconf -ivf
+./configure --prefix=$INSTALLROOT \
+            --with-sqlite3=install
```
ktf: Do you really need sqlite? If yes, we probably should have a separate package for that (eventually picking it up from the system).

sherpa.sh
```diff
@@ -0,0 +1,40 @@
...
+
+make
```
ktf: Can't it build with `-j`?

sherpa.sh
qgp: Fixed

sherpa.sh
qgp: I did not manage to build without it. I can investigate further but to my understanding it's used for configuration handling.

sherpa.sh
qgp: My bad, forgot to put it back after looking into compilation problems. Fixed.

sherpa.sh
ktf: Ok, then we should package our own version in case of grid production. SQLITE API is known to be incompatible between slc5 / slc6 / slc7  .

qgp: Ok, what we do here is we build the required sqlite stuff within SHERPA. So also here we avoid the problem of incompatibilities, no?

Sure, a separate package is more flexible and re-usable also for other things. Who takes care of this?

ktf: Yeah, I guess you are right. Let's continue as is for now, then we see. Do you have anything to test this?

qgp: Not yet. In principle for all generators, we could try and run a Rivet analysis on their output. We could even come up with something to validate the Rivet output. But this would still take me a moment for Sherpa to figure it out.

ktf: Ok, let me (well, jenkins) just check it compiles everywhere and then I will merge this.

ktf:
Compiles fine on SLC5, SLC6, SLC7. Ubuntu MIA as usual for some reason. Merging.

PR #321, opened 2016-04-21
Get matplotlib to compile on aidrefflp01

ktf:


ktf:
slc6 and slc7 worked. slc5 died because of an unrelated network glitch. Ubuntu MIA. Merging.

@vascobarroso, this should get python-modules to build using the new pip. Unless you have objections I'll revert back to the old version and try to fix that as well.

PR #325, opened 2016-04-22
GEANT_VMC

sawenzel:
Providing GEANT3_VMC-EXAMPLES AND PYTHIA6_ALISW.
This is WIP; Pull request to start discussing the code.

### DIFF
--- a/geant3_vmc-examples.sh
+++ b/geant3_vmc-examples.sh
@@ -1,0 +1,50 @@
+package: geant3_vmc-examples
+version: "1"
+requires:
+  - ROOT
+  - GEANT4_VMC
+  - GEANT3
+  - pythia6
+build_requires:
+  - CMake
+---
+
+case $ARCHITECTURE in
+  osx*)
+    LDSUFFIX="dylib"
+  ;;
+  *)
+    LDSUFFIX="so"
+  ;;
+esac
+
+
+#!/bin/bash -e
+cmake -DVMC_WITH_Geant3=ON \
+      -DVMC_WITH_Geant4=OFF \
+      -DGeant3_DIR=${GEANT3_ROOT}/lib64/Geant3-2.0.0 \
+      -DPythia6_LIBRARY=${PYTHIA6_ROOT}/lib/libPythia6.${LDSUFFIX} \
+      -DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE} \
+      -DCMAKE_INSTALL_PREFIX="$INSTALLROOT" \
+      ${GEANT4_VMC_ROOT}/share/examples
+
+# make them
+make ${JOBS+-j $JOBS} -i
+make install
+
+# Modulefile
+MODULEDIR="$INSTALLROOT/etc/modulefiles"
+MODULEFILE="$MODULEDIR/$PKGNAME"
+mkdir -p "$MODULEDIR"
+cat > "$MODULEFILE" <<EoF
+#%Module1.0
+proc ModulesHelp { } {
+  global version
+  puts stderr "ALICE Modulefile for $PKGNAME $PKGVERSION-@@PKGREVISION@$PKGHASH@@"
+}
+set version $PKGVERSION-@@PKGREVISION@$PKGHASH@@
+module-whatis "ALICE Modulefile for $PKGNAME $PKGVERSION-@@PKGREVISION@$PKGHASH@@"
+# Dependencies
+module load BASE/1.0 ROOT/$ROOT_VERSION-$ROOT_REVISION GEANT4_VMC/$GEANT4_VMC_VERSION-$GEANT4_VMC_REVISION GEANT3/$GEANT3_VERSION-$GEANT3_REVISION PYTHIA6/$PYTHIA6_VERSION-$PYTHIA6_REVISION
+EoF
+
--- a/geant4_vmc.sh
+++ b/geant4_vmc.sh
@@ -5,6 +5,7 @@ source: https://github.com/alisw/geant4_vmc
 requires:
   - ROOT
   - GEANT4
+  - VGM
 build_requires:
   - CMake
 env:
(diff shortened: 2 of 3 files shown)


ktf:
Do we really need to have two pythia6? I would simply keep the one which works for you.

sawenzel:
Ok. In this case I am going to remove the old pythia6.sh (which made several static libraries).

sawenzel:
Updated according to suggestion.

pythia6.sh
```diff
@@ -1,59 +1,77 @@
+# a pythia6 recipe based on the one from FairROOT
```
ktf: How about we create alisw/pythia6 and patch it there, rather than doing everything in the recipe?
That would better mark our changes.

sawenzel:
Updated; now pythia is taken from alisw; added also VGM so that the Geant4_VMC examples run fine.

vgm.sh
```diff
@@ -0,0 +1,35 @@
+package: VGM
+version: "%(tag_basename)s%(defaults_upper)s"
+tag: "4.3"
+source: https://github.com/alisw/vgm.git
```
ktf: VGM is uppercase.

pythia6.sh
ktf: Nice cleanup  :). However you should not touch the master with alice code, but rather create a alice/416 branch and have ALICE specific commits there. This way we can make sure we always know what is ALICE and what is upstream.

pythia6.sh
```diff
@@ -1,59 +1,35 @@
...
-    F77="`which gfortran` -fPIC"
-  ;;
-  *) 
-    PLATF_CONF_OPTS="--disable-shared --enable-static"
-    PLATF_LDFLAGS=""
-    PLATF_LD=""
-    F77="`which gfortran` -fPIC"
-  ;;
-esac
+cmake ${SOURCEDIR} \
```
ktf: Bonus points if you align the "\" ...;) `:'<,'>Tabularize / \` on vim + tabularize plugin.

pythia6.sh
sawenzel: Done

vgm.sh
sawenzel: Fixed ( Although what is the convention here? Many *.sh contain package names in capital letters ).

ktf: > Fixed ( Although what is the convention here? Many *.sh contain 
> package names in capital letters ).

The best practice would be that the repository name matches the package 
name as specified in the .sh (or viceversa). The rationale for that is 
that if you want to develop a package locally, you do not need any extra 
directory name for the git clone.

The cases where this is not the case are historical naming of the 
packages wrt their original repository (e.g. ROOT was already ROOT on 
CVMFS, while the repository is root because that’s what SFT used).

Anyways, I tried to make the policy more explicit in #336.

sawenzel:
I rebased this pull request so that it can be merged "ff" without conflict.

ktf:
vgm builds fine. Other tests were skipped because of an issue in the Jenkinsfile. Merging.

ktf:
Ah, no, wait. I think there is an issue in one of the modulefiles. Will comment inline.

geant4_vmc.sh
```diff
@@ -32,7 +36,7 @@ proc ModulesHelp { } {
 set version $PKGVERSION-@@PKGREVISION@$PKGHASH@@
 module-whatis "ALICE Modulefile for $PKGNAME $PKGVERSION-@@PKGREVISION@$PKGHASH@@"
 # Dependencies
-module load BASE/1.0 GEANT4/$GEANT4_VERSION-$GEANT4_REVISION ROOT/$ROOT_VERSION-$ROOT_REVISION
+module load BASE/1.0 GEANT4/$GEANT4_VERSION-$GEANT4_REVISION ROOT/$ROOT_VERSION-$ROOT_REVISION VGM
```
ktf: Shouldn't this be `VGM/$VGM_VERSION-$VGM_REVISION`?

geant3_vmc-examples.sh
```diff
@@ -0,0 +1,50 @@
...
+cat > "$MODULEFILE" <<EoF
+#%Module1.0
+proc ModulesHelp { } {
+  global version
+  puts stderr "ALICE Modulefile for $PKGNAME $PKGVERSION-@@PKGREVISION@$PKGHASH@@"
+}
+set version $PKGVERSION-@@PKGREVISION@$PKGHASH@@
+module-whatis "ALICE Modulefile for $PKGNAME $PKGVERSION-@@PKGREVISION@$PKGHASH@@"
+# Dependencies
+module load BASE/1.0 ROOT/$ROOT_VERSION-$ROOT_REVISION GEANT4_VMC GEANT3 pythia6
```
ktf: Shouldn't this be:

```
+module load BASE/1.0 ROOT/$ROOT_VERSION-$ROOT_REVISION GEANT4_VMC/$GEANT4_VMC_VERSION-$GEANT4_VMC_REVISION GEANT3/$GEANT3_VERSION-$GEANT3_REVISION pythia6/$PYTHIA6_VERSION-$PYTHIA6_REVISION
```

geant4_vmc.sh
```diff
@@ -14,7 +15,10 @@ prepend_path:
 ---
 #!/bin/bash -e
 cmake "$SOURCEDIR" \
+  -DCMAKE_CMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE} \
```
ktf: While you are at it, please align the `\`.

geant4_vmc.sh
sawenzel: It worked like this but I will change anyway.

ktf: Yes, please. It will work fine in a single installation, but I suspect problems will arise when you have multiple ones like on the grid (I think it takes the latest one with what you have).

sawenzel:
Implemented all suggestions.

ktf:
There is still an issue with this. I think there is a cyclic dependency.

ktf:
#353 fixes the issue. alisw/alibuild#234 tracks the actual problem.

PR #330, opened 2016-04-22
Add modulefiles for AliceO2 packages

ktf:
This adds modulefiles for the following packages:

- AliceO2
- DDS
- FairRoot
- sodium
- ZeroMQ

so that alienv works with them.


ktf:
Compiles fine. On SLC6. On SLC7 there was an unrelated out of memory error when compiling AliROOT. Merging.

PR #369, opened 2016-05-02
Add missing c++11 support in debug defaults

mkrzewic:


dberzano:
@mkrzewic: thanks, we've definitely overlooked that :-)

PR #370, opened 2016-05-03
Disable shadow password support in root

mkrzewic:
this fixes a build problem on centos 7


dberzano:
OK thank you very much for this PR! For the record, this is a weird ROOT problem that has never been solved and for some reason "shadow passwords" support is enabled by default on some systems. Anyway the workaround (which does not affect any of the functionalities we use) [comes from the ROOT developers](https://sft.its.cern.ch/jira/browse/ROOT-4091).

PR #384, opened 2016-05-11
Do not compile dijet when fastjet is disabled

dstocco:
Dear all,
this patch avoids a crash in the compilation of POWHEG when fastjet is disabled.
Can you please push it?

Thanks in advance,
best regards,
Diego

### DIFF
--- a/powheg.sh
+++ b/powheg.sh
@@ -15,7 +15,7 @@ install -d ${INSTALLROOT}/bin
 
 export LIBRARY_PATH="$LD_LIBRARY_PATH"
 
-PROCESSES="dijet hvq W Z"
+PROCESSES="${FASTJET_VERSION:+dijet }hvq W Z"
 for proc in ${PROCESSES}; do
     mkdir ${proc}/{obj,obj-gfortran}
     make -C ${proc}


qgp:
@alisw/pwgmm

ktf:
How about:

```
PROCESSES="${FASTJET_VERSION:+dijet }hvq W Z"
```

dstocco:
Ciao Giulio,
yep, that could be more elegant...
Can you do it directly or should i prepare a new patch?
Thanks,
ciao,
Diego

ktf:
I can do it. @qgp / @alisw/pwgmm please object if you do not agree on the change.

ktf:
actually, thinking again, how about you do it, so that you can modify it if there is other comments from @qgp or alike?

ktf:
(we also avoid having a new PR)

dstocco:
Will do it.
Just a silly question, since I'm not so used to github.

Should I simply push another commit on my branch, and you take care of squashing it when applying to alisw, or should I rather amend the previous commit (and thus rewrite the history)?

Thanks,
ciao,
Diego

ktf:
Now it's actually possible to squash the history at merge time, but it's best practice / kind if PR proponents do it. ;-)

ktf:
I've added a remark about it at @17aab49f0373df33e6b3408fa3d05f2f00c550a4

dstocco:
Ciao Giulio,
ok thanks. The commit is amended.
Ciao,
Diego

ktf:
Merging in next, I assume this is not a problem for prod either, but in any case it's not going in for a few days.

PR #400, opened 2016-05-18
"Preserve module files ..." in o2 recipe too

graffaele:
I am not sure if this is the right branch where to pull this. It solves the isssue that alienv works only after the first aliBuild. I have seen that this was already done for root, aliroot and aliphysics, probably o2 was left out just because it didn't have module files yet.


ktf:
Is this still needed also with the latest version of aliBuild?

ktf:
I assume not. Closing this.

PR #401, opened 2016-05-19
GSL recipe modifications

akubera:
Added requirement that system GSL must be less than version 2.0.
This fixes an error generated while building ROOT on OSX (10.10&10.11) with GSL 2.1 (the current stable version) installed.

Also changed use of `brew` to `pkg-config` to determine proper system gsl cflags.

The use of pkg-config is up for discussion; I'd like to hear why it's not being used.

### DIFF
--- a/gsl.sh
+++ b/gsl.sh
@@ -8,7 +8,7 @@ requires:
   - "GCC-Toolchain:(?!osx)"
 prefer_system: (?!slc5)
 prefer_system_check: |
-  printf "#include \"gsl/gsl_version.h\"\n# if (GSL_MAJOR_VERSION * 100 + GSL_MINOR_VERSION < 116)\n#error \"Cannot use system's gsl.\"\n#endif\nint main(){}" | gcc  -I$(brew --prefix gsl)/include -xc++ - -o /dev/null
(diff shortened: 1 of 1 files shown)


ktf:
I'd rather not have pkg-config dependency at this point, because that would require proper integration between whatever pkg-config comes from the user and the one which we potentially have built as part of the autotools package. Moreover, there is the additional problem that `pkg-config --prefix gsl`, might give a path (e.g. /usr/local) which might have more than one package in it and that is potentially dangerous because it could end up overriding the lookup of other externals.

Say for example that you have something like:

```
cmake -DGLS_ROOT=`pkg-config --prefix gsl` -DROOT_ROOT=$ROOT_ROOT
```

this gets resolved to:

```
cmake -DGLS_ROOT=/usr/local -DROOT_ROOT=/alice/sw/osx_x86-64/ROOT/v5.34.0
```

and effectively becomes:

```
c++ -I/usr/local -I/alice/sw/osx_x86-64/ROOT/v5.34.0
```

now if you have root6 also installed in /usr/local, that would get precedence over the one we compile and you have a problem.

We support brew simply because we are guaranteed that `brew --prefix` will give a path which only contains that external.

ktf:
The other change is fine, however please make sure you do a pull request to the `IB/v5-08/next` branch.

PR #406, opened 2016-05-19
Require system GSL version to be in range [1.16, 2.0)

akubera:
The current stable version of GSL is 2.1. With this installed on my
system (OSX 10.10), ROOT does not compile due to incompatible GSL headers.

This change adds the extra requirement that system GSL is less than v2.0 when preferring system packages.


ktf:
Thanks.

PR #412, opened 2016-05-25
Avoid variable order issue in Geant4

ktf:


ktf:
@fuhlig1 this is what you need for your issue.

PR #416, opened 2016-06-01
Move to latest version of Mesos & Protobuf

ktf:


ktf:
Unrelated error when building.

PR #419, opened 2016-06-02
Fix autotools emacs and termcap problem

dberzano:
Checks if termcap is available on the system and compiles autotools with emacs disabled.
Prevents autotools recipe from hanging when checking emacs or failing when trying to compile .elc files.


ktf:
This fails in the tests for slc7 because of:

```
ERROR: Please install the ncurses development package on your system (usually ncurses-devel or libncurses-dev)
```

are you sure the test is correct there?

ktf:
Nevermind that was the slc7-daq machine. I will update the container.

PR #423, opened 2016-06-07
Check if Perl libs can be linked

dberzano:


dberzano:
@ktf: this should fix some of the problems seen on the mailing list. Not only it checks for the headers, but for the usability of `-lperl`. Does that work for you? I have tested it with `aliDoctor`.

ktf:
The problem is that libperl.so might not be in the search path of the compiler, like on mac:

```
→ aliDoctor libperl
ERROR: Package libperl is a system requirement and cannot be found.
ERROR: This is due to the fact that the following script fails:
ERROR:
ERROR: printf "#include <EXTERN.h>\n#include <perl.h>\n" | gcc -xc++ -lperl -I`perl -MConfig -e 'print $Config{archlib}'`/CORE - -o /dev/null && perl -MExtUtils::Embed -e 1
ERROR: with the following output:
ERROR:
ERROR: libperl: ld: library not found for -lperl
ERROR: libperl: clang: error: linker command failed with exit code 1 (use -v to see invocation)
ERROR: libperl:
ERROR: Please install libperl and perl-ExtUtils-Embed development package on your system
ERROR:
```

ktf:
If you add 

```bash
-L`perl -MConfig -e 'print $Config{archlib}'`/CORE
```

it should actually be fine.

ktf:
I've a solution for this which I am testing.

dberzano:
Superseded by #452.

PR #425, opened 2016-06-07
Make the list of boost libraries consistent

matthiasrichter:
FairRoot reqires now also boost chrono and atomic to enable building of
FairMQ, which is needed for O2


ktf:
I guess we can have this for AliRoot / AliPhysics as well. They are tiny libraries and the only need for boost there is in fastjet. @dberzano @pzhristov objections?

PR #436, opened 2016-06-14
Improve check for system's GCC-Toolchain

### LINKED ISSUE #208: Avoid requiring autotools on modern platforms
We should evaluate whether or not we should require autotools on modern platforms where we can rely on the system ones.

dberzano:
Fixes alisw/alibuild#208.

### DIFF
--- a/gcc-toolchain.sh
+++ b/gcc-toolchain.sh
@@ -9,7 +9,7 @@ build_requires:
  - autotools
 prefer_system: .*
 prefer_system_check: |
-  which cc && printf "#define GCCVER ((__GNUC__ << 16)+(__GNUC_MINOR__ << 8)+(__GNUC_PATCHLEVEL__))\n#if (GCCVER < 0x040800) || ((GCCVER >= 0x050000) && (GCCVER < 0x050300)) || (GCCVER >= 0x060000)\n#error \"System's GCC cannot be used: we need 4.8, 4.9 or 5.X (with the exception of 5.0 to 5.2). We are going to compile our own version.\"\n#endif\n" | cc -xc++ - -c -o /dev/null && which gfortran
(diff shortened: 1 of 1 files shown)


dberzano:
@ktf: what do you think of the last commit? I want to make sure that `cc` and `c++` come from the same location.

ktf:
+1

ktf:
I just realized this will fail if c++ is a symbolic link?

dberzano:
No, I have tested it - `test -f` should work in this case. Did it fail for you?

rqwa:
I talked to the admins of our cluster and they suggested to set the compiler explicitly in the configuration. I.e.: `CC=gcc CXX=g++ cmake` 
Is this also a possible solution?

rqwa:
After the link to cc  have been added, the output of `ls -l $(which cc) $(which c++)` is the following:
```
-rwxr-xr-x 4 root root 3116106 Jan 31  2015 /cm/shared/apps/gcc/4.8.4/bin/c++
lrwxrwxrwx 1 root root       3 Jan 28  2015 /usr/bin/cc -> gcc
```
Also compiling works now and a dry-run gives the following output:
```
[...]
We will build packages in the following order: defaults-release AliEn-CAs autotools CMake OpenSSL ApMon-CPP UUID GSL GEANT4 gSOAP MonALISA-gSOAP-client XRootD xalienfs AliEn-Runtime ROOT vgm GEANT3 GEANT4_VMC AliRoot AliPhysics
--dry-run / -n specified. Not building.
```
But after switching to IB/v5-08/next with:
```
cd ali-master/alidist
git checkout IB/v5-08/next
git pull
```
The output of the dry-run includes the the GCC-Toolchain again:
```
[...]
We will build packages in the following order: defaults-release AliEn-CAs autotools GCC-Toolchain ApMon-CPP CMake UUID GSL OpenSSL GEANT4 gSOAP MonALISA-gSOAP-client XRootD xalienfs AliEn-Runtime ROOT vgm GEANT3 GEANT4_VMC AliRoot AliPhysics
--dry-run / -n specified. Not building.
```

ktf:
`CC=gcc CXX=g++ cmake` is not a solution as it would require changes everywhere and have a special setup for clang. Given that the official `cmake` expects `cc` and `c++` I would prefer this is fixed on your side (or that you use the compiler we provide, as it now happens in next).

rqwa:
I haven't changed anything in the cmake config.
As I already wrote a link was set from cc to the right gcc, and the compilation works with the already installed gcc version. But in next the link is not detected and alibuild wants to build the GCC-toolchain even it is not necessary.

ktf:
Ah, I see now. Your link setup is wrong. You should have:

```
/cm/shared/apps/gcc/4.8.4/bin/c++
/cm/shared/apps/gcc/4.8.4/bin/cc -> gcc
```

aliBuild will check that c++ and cc come from the same directory.

rqwa:
Thanks, that solved the problem.

ktf:
Great.

PR #438, opened 2016-06-15
Add support for detector algorithms in AliRoot

dberzano:
* A new default (daq) sets some environment variables required for DAQ software.
* Using `--defaults daq` ROOT will be built statically (using `configure`
  instead of CMake, since the latter does not support static builds).
* AliRoot also produces RPMs for the detector algorithms if `--defaults daq`.

### DIFF
--- a/aliroot.sh
+++ b/aliroot.sh
@@ -7,6 +7,7 @@ requires:
   - GEANT4_VMC
 build_requires:
   - CMake
+  - DAQ:slc6.*
 env:
   ALICE_ROOT: "$ALIROOT_ROOT"
 source: http://git.cern.ch/pub/AliRoot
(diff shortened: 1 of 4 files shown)


dberzano:
@Barthelemy: with this recipes I have managed to produce DA RPMs.
@ktf: let's review before merging (and make sure this does not break all the rest!)

Barthelemy:
Once it is merged, you could present it to Sylvain. Thank you for the hard work. Please also comment on the JIRA tickets.

dberzano:
For reference, [ALIROOT-5801](https://alice.its.cern.ch/jira/browse/ALIROOT-5801) and [ALIROOT-5936](https://alice.its.cern.ch/jira/browse/ALIROOT-5936).

PR #441, opened 2016-06-16
Allow boost up to v1.55.00

ktf:


ktf:
Nevermind. We actually need 1.59 for cgal.. :-/

Issue #433, opened 2016-06-13
Provide a list of generators / packages used in production


ktf:
@qgp this is what we discussed in the Offline board. Can you provide  a list of packages which you consider "PWG-MM" approval required? We can then write a test which will ask for your sign off if changes enter in one of them.

qgp:
@ktf: thanks for the follow up, here comes the list of recipes which are of direct concern for @alisw/pwgmm:
- aligenerators
- crmc
- epos
- epos-test
- hepmc
- jewel
- lhapdf
- lhapdf5
- powheg
- pythia
- pythia6
- rivet
- rivet-test
- sherpa
- thepeg
- thepeg-test
- yoda

Of course, these packages have dependencies and changes to them could affect the packages; yet I didn't list the dependencies here.

ktf:
Wouldn't "aligenerators and all it's dependencies" be enough of a description? That could be automated very easily. Any reason why aligenerators does not include all of them?

qgp:
@ktf: that's a good idea. `aligenerators.sh` contains all the generators and their tests. The things not covered would be hepmc, lhapdf(5), rivet, yoda, which are not technically generators, yet they are @pwgmm responsibilities. hepmc and lhapdf(5) show up as dependencies of the generators, yoda would if rivet was included. So, if we say "aligenerators + rivet + all (also second order) dependencies" we would have everything covered, I think. But like this, we also include plenty of packages which appear as dependency but are not a particular concern to PWG-MM, e.g. gcc-toolchain.

So, the best I can see would be a bit more explicit:
- aligenerators + deps
- rivet
- hepmc
- yoda
- lhapdf(5)

But I am glad if you find a more elegant formulation :-)

PR #445, opened 2016-06-17
Add notice about PWGMM owned packages

ktf:


ktf:
Fixes #433 .

Of course the check should be automated, but this is good enough for now.

ktf:
@alisw/pwgmm

PR #454, opened 2016-06-22
Add path for OpenSSL headers

ktf:


dberzano:
@ktf, @pzhristov: actually the problem of the correct OpenSSL detection has to be fixed on the ROOT side (this will make the compilation work, but it will still prefer the system's OpenSSL). For what concerns ROOT 5 it works because I have patched the CMake files in our branch.

PR #462, opened 2016-06-24
Add support for mesos-dds

kvnnap:
This adds the mesos-dds recipe which also depends on the Microsoft C++ REST SDK (Casablanca).


ktf:
Please rebase and squash your changes, push them again and I will merge it. Also make sure you do not have unneeded changes like the `simulation.sh` ones.

PR #463, opened 2016-07-06
Fix ROOT package name

dberzano:
This prevents packages from being called with `alice` appearing twice such as in `v5-34-30-alice5-alice-1`. With this patch, the next built package will be called `v5-34-30-alice5-1`.


ktf:
Any reason for not merging this?

dberzano:
I was having a look at why the test failed. It looks there's a problem in the Rivet test on SLC7, but from ROOT's perspective it's fine. I am merging it now.

PR #470, opened 2016-07-07
Update Pythia6 to version 6.4.28

qgp:
- add path for Pythia6 to AGILE_GEN_PATH

@sawenzel: This should conclude the Pythia6 update. Any objections?

Signed-off-by: @qgp for @alisw/pwgmm


pythia6.sh
```diff
@@ -33,6 +33,7 @@ module load BASE/1.0
 # Our environment
 setenv PYTHIA6_ROOT \$::env(BASEDIR)/$PKGNAME/\$version
 prepend-path LD_LIBRARY_PATH \$::env(PYTHIA6_ROOT)/lib
+prepend-path AGILE_GEN_PATH \$::env(PYTHIA6_ROOT)
 $([[ ${ARCHITECTURE:0:3} == osx ]] && echo "prepend-path DYLD_LIBRARY_PATH \$::env(PYTHIA6_ROOT)/lib")
```
sawenzel: what is this path?

sawenzel:
I believe we still have to update the alisw/pythia repo?
Also: Did you modify the library name?

pythia6.sh
qgp: It adds it to the environment variable which is used by AGILe to search for generators (AGILe is an interface to run (old) generators and produce HepMC output). The corresponding PR is in preparation (https://github.com/qgp/alidist/tree/add_agile) but there's no point adding this before we have PYTHIA.

qgp:
@sawenzel: Update is in alisw/pythia6 (branch alice/428), the library is called libpythia6.so.
This is what we had discussed by email, no?

sawenzel:
Yes fine. Was just wondering whether this was done. I must have missed the merge request.

qgp:
@sawenzel: You didn't miss it, I committed the changes directly.

@dberzano: So, I think, this can be merged to next. Then, I will also make the other PRs. Any estimate when the next next -> prod merging will happen?

sawenzel:
We then still need to change the recipe of the geant3_vmc-examples to use the new library name and check if it compiles fine ( run aliBuild build geant3_vmc-exampes ). This should be done together with this merge request (ideally).

see
```
cmake -DVMC_WITH_Geant3=ON                                          \
      -DVMC_WITH_Geant4=OFF                                         \
      -DGeant3_DIR=$Geant3_DIR                                      \
      -DPythia6_LIBRARY=${PYTHIA6_ROOT}/lib/libPythia6.${LDSUFFIX}  \
      -DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE}                        \
      -DCMAKE_MODULE_PATH=${GEANT4_VMC_ROOT}/                       \
      -DGeant4VMC_DIR=${Geant4VMC_DIR}                              \
      -DCMAKE_INSTALL_PREFIX="$INSTALLROOT"                         \
      ${GEANT4_VMC_ROOT}/share/examples
```

qgp:
@sawenzel: Ok, can you make the PR for that recipe change?

sawenzel:
I would recommend or favour if you do it as part of this commit even. Then we have a nice and clean atomic change.

qgp:
ok, PR updated

sawenzel:
Thanks. Looks good.

dberzano:
Hello, let's wait for Jenkins before doing the merge, to check if everything's sound.

@qgp: we tend to be very conservative in changing the prod branch (last merge happened after almost two months!), however since this does not touch anything in our main software chain (AliPhysics, AliRoot, ROOT) I guess that we can just cherry-pick the changes to prod (and rebase next) when we feel they're ready.

Is this needed for the analysis tutorial next Friday?

qgp:
Sure, we wait for jenkins.

@dberzano: It's not crucial for the tutorial on Friday, we have other generators which are already usable, but it would add a bit of spice as one could run more generators. Otherwise, we would need them afterwards when people start with the Rivet implementations. So, we surely don't want to wait months for this to go to prod. But we have enough time to first collect everything in next and then do a coordinated cherry-pick.

qgp:
Jenkins seems to fail in AliRoot (FairMQ stuff), or do I misread something? How can this be solved if it's not actually related to what we changed here.

pythia6.sh
dberzano: Hello @qgp, I am seeing this `AGILE_GEN_PATH` too. Since it is used by AGILe, shouldn't be added in the AGILe recipe instead (and then make AGILe depend on Pythia, see #471)?

qgp: Ok, apparently I explained it badly. AGILe uses an environment variable to know where to look for generators (in addition to standard locations), this variable is AGILE_GEN_PATH. The path comes from the generator, and AGILe can work with different versions of the generator. So, every (relevant) generator should add itself to AGILE_GEN_PATH such that it can be picked up by AGILe. It's a bit like every package that wants to be executable adds itself to PATH.

If we added this to AGILe, and made it depend it on Pythia, we would have fixed it to one specific version to Pythia, or the other way around we would need another AGILe version for every generator.

Does this make sense now?

ktf:
pythia6 is also a dependency of O2, that's why it's getting tested. I agree this is unrelated to this PR.

pythia6.sh
dberzano: Hello @qgp, thanks for clarifying (indeed you are _prepending_ to the variable and not _setting_ it, sorry for overlooking it). It makes sense to me now 👍

PR #476, opened 2016-07-15
Add HepMC and LHAPDF support to Sherpa

qgp:


qgp:
@dberzano: rebased as discussed ...

dberzano:
This fails because of:

```
DEBUG:SHERPA:8234828a: checking for SQLite3 library >= 3.3.9... not found
DEBUG:SHERPA:8234828a: configure: error: Sqlite3 required but not found.
DEBUG:SHERPA:8234828a: To install a local copy, use the configure option --with-sqlite3=install.
```

The problem occurs on SLC7.

I think we should add a recipe for SQLite3. SQLite can be used with a single big C file called "amalgamation", I am wondering why/if SHERPA can be configured like that?

qgp:
Hi @dberzano, at the moment the recipe contains --with-sqlite3=install, maybe the sqlite3 compilation fails on CC 7. Can I have a look at the logs from the test somewhere?

Cheers,
   Jochen

dberzano:
Now it fails with:

```
DEBUG:SHERPA:8234828a: checking for LHAPDF installation directory... configure: error: Unable to use LHAPDF from specified path.
```

I guess it's because you're referring to it as `LHAPDF_ROOT` instead of `LHAPDF5_ROOT`. Could you fix this?

Thanks!

qgp:
@dberzano: thanks for spotting this, I'll look into it. In any case, I still have to do push the wget -> curl substitution (still need to finish testing it though).

dberzano:
@qgp I have approved the test, let's wait.

PR #478, opened 2016-07-20
Fix build options for DAQ DAs

dberzano:
* No SSL in ROOT
* RPM target in AliRoot

### DIFF
--- a/aliroot.sh
+++ b/aliroot.sh
@@ -54,6 +54,7 @@ if [[ $GIT_TAG == master && ! $ALICE_DAQ ]]; then
   make -k ${JOBS+-j $JOBS} install || true
 else
   make ${JOBS+-j $JOBS} install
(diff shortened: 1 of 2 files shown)


dberzano:
RPM creation for DAs has been streamlined in AliRoot. These recipes:

* disable unneeded SSL in ROOT (which cause problems during linking)
* create all DA RPMs (whose creation is optional and requires invoking an extra target, which is invoked in single core mode to avoid `rpmbuild` problems)

This follows up [ALIROOT-5801](https://alice.its.cern.ch/jira/browse/ALIROOT-5801).

dberzano:
@ktf: since we're about to release v5-08-15, if the Jenkins test succeeds for this PR I suggest we cherry-pick this commit and 50ecbe7349 (the one that fixes ROOT's package naming) into `IB/v5-08/prod` (and then we rebase `IB/v5-08/next`).

dberzano:
PR test fails for unrelated reason (compiler error in building Rivet).

PR #486, opened 2016-07-27
Automate processing of coverage information

ktf:


ktf:
Merging since nothing actually uses it.

PR #489, opened 2016-08-01
Move to tagged version of ZeroMQ

ktf:


ktf:
Error in AliceO2 is actually unrelated and will be fixed with a separate PR.

PR #494, opened 2016-08-10
add variables needed for simulation

MohammadAlTurany:


ktf:
Closing this in favour of alisw/alidist#495:

- No reason to do it only for O2, we can have it in the "next" branch 
- We still need the -j 2 when compiling with Jenkins, due to some memory issue.
- Commits squashed.

PR #495, opened 2016-08-11
Fix FairRoot simulation setup

ktf:
- Add module_old to cmake module path in O2
- Add missing G4 data files
- Add missing variables for simulation with FairRoot


ktf:
Same as #494 , just squashed in a single commit and pull request done to `IB/v5-08/next` since there is no reason for not having it also in the production setup. @sawenzel can you have a look as well?

ktf:
This was failing tests because we are using the `next` branch, rather than the O2 branch which has the AliRoot + ZeroMQ fixes.

PR #496, opened 2016-08-15
Use defaults to specify O2 peculiarities

ktf:
This disables AliEn-Runtime and moves to ROOT 6.

This requires alisw/alibuild#352 .


ktf:
Nothing uses this. Merging.

PR #497, opened 2016-08-16
Do not depend on c++ to check for perl

ktf:


ktf:
@dberzano can you cross check and merge? Without this we have an unneeded implicit dependency on g++.

dberzano:
Checked on all SLCs, Ubuntu and OSX. Makes sense. Merging.

PR #498, opened 2016-08-16
Add defaults for the O2 DAQ builds

ktf:
- No simulation / generators packages.
- Use ROOT6
- Use new CMake
- AliRoot does not need to depend on Fastjet

### DIFF
--- a/defaults-o2-daq.sh
+++ b/defaults-o2-daq.sh
@@ -1,0 +1,33 @@
+package: defaults-o2-daq
+version: v1
+env:
+  CXXFLAGS: "-fPIC -g -O2 -std=c++11"
+  CFLAGS: "-fPIC -g -O2"
+  CMAKE_BUILD_TYPE: "RELWITHDEBINFO"
+disable:
+  - AliEn-Runtime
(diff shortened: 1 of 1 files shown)


ktf:
@Barthelemy @vascobarroso : this should simplify quite a lot building O2 in a DAQ like environment (i.e. without the simulation / analysis parts). When coupled with the `next` branch of aliBuild (or the soon to be released alibuild 1.4.0), it will allow you to do:

```
aliBuild  --defaults o2-daq build O2
```

vascobarroso:
Looks good! 
Will it also be part of the o2 branch ?

ktf:
Actually I was planning to remove completely the o2 branch once we release aliBuild 1.4.0, since that would be equivalent to either:

- `aliBuild  --defaults o2 build O2`
- `aliBuild  --defaults o2-daq build O2`

depending on wether or not you want simulation stuff. Does that sound OK?

vascobarroso:
Perfect!

Barthelemy:
yes, that is ok.

PR #501, opened 2016-08-17
Improve ROOT dependency checks

dberzano:
* Use own FreeType in ROOT
* Check for X development packages

### DIFF
--- a/opengl.sh
+++ b/opengl.sh
@@ -1,8 +1,8 @@
 package: opengl
 version: "1.0"
-system_requirement_missing: "Please install OpenGL (and X11) development packages on your system"
+system_requirement_missing: "Please install the OpenGL development packages on your system"
 system_requirement: ".*"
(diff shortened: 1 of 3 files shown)


dberzano:
@ktf as discussed. Seems to work OK on Mac too (El Capitan at least).

ROOT does not depend on FreeType in our case as we're compiling it with Cocoa. In such a case ROOT's CMake automatically switches to `builtin_freetype` no matter if you have some external FreeType you want to link.

dberzano:
Fails because of a missing bit in a container. Fixing the container first, see alisw/docks#117.

dberzano:
One of the errors relates to the testing (`aliBuild build Xdevel` returns nonzero because there is nothing to do, but it is OK in this case and it should be fixed elsewhere), the other is an ICE unrelated to this PR. Merging...

PR #502, opened 2016-08-17
Add check for system curl

### LINKED ISSUE #460: More required externals
- [x] xft
- [x] xpm
- [x] curl (required to be able to fetch non git stuff. Maybe we should fix the recipes instead)

dberzano:
Fixes #460.

### DIFF
--- a/cgal.sh
+++ b/cgal.sh
@@ -6,6 +6,7 @@ build_requires:
   - GMP
   - MPFR
   - CMake
+  - curl
 ---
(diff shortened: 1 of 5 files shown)


dberzano:
@ktf pls have a look

ktf:
How about we redirect the output to `>/dev/null`?

dberzano:
@ktf right, done (stderr is not redirected but if `curl` exists no output is produced there).

PR #505, opened 2016-08-22
Fix curl check on all platforms

dberzano:
Notably, `curl --help|--version` returns 2 on SLC5. Here we check whether the
shell returns 127 (command not found) explicitly.


dberzano:
Reported by @qgp while testing it on SLC5. @ktf, please have a look :-)

qgp:
@dberzano: fixes my problem, thanks!

PR #506, opened 2016-08-23
Add DPMJET package for use by AliRoot

qgp:
This PR is up for discussion on how to factor out DPMJET. It adds a package DPMJET providing the library libDPMJET_ext.so and adds the dependency on DPMJET to AliRoot. As the name of the library is different (the AliRoot-internal version is called libDPMJET.so) it remains unused for now. Once the new recipes are in prod, we can switch AliRoot to use the external library instead of the integrated one (already tried locally, of course) and then remove DPMJET from AliRoot. Thoughts?

For the AliRoot compilation I needed to add the export of LIBRARY_PATH. I don't know if this is your preferred way of handling it (just assigning LD_LIBRARY_PATH). Also this is up for suggestions.


qgp:
@ktf, @dberzano: the test fails with fetching from gitlab:
> Unable to fetch from https://gitlab.cern.ch/ALICEPrivateExternals/DPMJET.git

Can you have a look, please? Can you also comment on the questions above?

dberzano:
We are looking into the GitLab issue. For what concerns the rest, I don't think it's a good idea to use `LIBRARY_PATH` or this `_ext` workaround.

Can we just drop at some point DPMJET from AliRoot, then use it from here and add it in `AliGenerators` instead? AliRoot does not need to depend on DPMJET (it includes it for the moment, but this is a different story) AFAIK.

qgp:
@dberzano: Thanks for your answer, comments below:

LIBRARY_PATH was necessary to find the library.

How do we make the transition if both libraries are called libDPMJET.so? It's unclear which one gets used.

AliRoot is linked against DPMJET and this is not changed (as of now), we have just factored out the source code. So, AliRoot compilation wouldn't work without DPMJET and, thus, it is needed as dependency. What do I overlook?

dberzano:
We plan to bump alidist next to prod for the forthcoming AliRoot release, so we'll migrate directly from DPMJET inside AliRoot to DPMJET outside AliRoot in one go. Having two libraries increases confusion IMHO.

I did not know that AliRoot was linked against DPMJET. From a brief look (`git grep DPMJET | grep txt:`, I see that there is, as a matter of fact, only one module of AliRoot which uses DPMJET, and it is `TDPMjet`. In turn, I could not find any other module using `TDPMjet` itself.

I am wondering whether we can do a single DPMJET package "for AliRoot" which comprises both DPMJET and TDPMjet. The package would be completely independent from AliRoot (I repeat: from a first brief look, therefore I might be wrong) and it can be made part of AliGenerators.

Shall we discuss it tomorrow if you can?

qgp:
@dberzano: no objections to this if it works - I have my doubts, however (at least the headers are used, libTDPMjet.so is usually loaded at run-time). I'll have a closer look and, yes, I'd be happy to discuss tomorrow (would have to be morning though).

Thanks!

qgp:
Updated based on discussion with @dberzano, using FindDPMJET of AliRoot introduced in AliRoot with commit 32a0198.

Transition to new recipe still to be discussed.

dberzano:
Our policy is: if your current AliRoot master can be built on your laptop, do not update alidist. With the current backwards-compatible modification we've made in AliRoot, this is still the case (as AliRoot will fall back on builtin DPMJET).

Therefore I suggest, as we've partially discussed already:

 - to merge this one (into `next`, but it will soon end in `prod`), so that whoever downloads aliBuild for the first time will get external DPMJET
 - to eliminate DPMJET from AliRoot later on, in a v5-09 version, as a change in a minor version as per our policy means that you must update alidist

@ktf comments?

qgp:
@dberzano, @ktf: In fact the change to AliRoot could be done such that it would still build with an old alidist (just not having DPMJET support). I agree that bumping the minor version makes sense, but we would not even break something for those who don't update alidist at that moment.

dberzano:
OK I had to rebase it so I have squashed it manually to next, see 73d301fe. Closing!

PR #508, opened 2016-08-23
Fix modulefile for FairRoot

ktf:
Wrong case for GEANT3 and GEANT4.


ktf:
Trivial and no tests. Merging.

PR #515, opened 2016-08-30
Add automation for coverage builds

ktf:
- Correct GCOV_PREFIX_STRIP. Given we have to use development packages
  for AliRoot and AliPhysics in any case, we revert back the changes to
  those which work in that case.
- Add lcov as a dependency.
- Fix various typos


ktf:
Somehow this got lost.

PR #521, opened 2016-09-07
Change geant3.sh recipe to use alisw/geant3.git tag alice/v2-1

sawenzel:
* Change necessary following a patch applied to geant3


dberzano:
Superseded by #535.

PR #528, opened 2016-09-18
Fix GCC build with GCC 6 and modulefile

dberzano:
Metapackages are used to load the correct GCC version based on the current
platform. They only have the GCC `MAJ.MIN.PATCHES` version in their name, as in
`GCC-v4.9.3`, therefore any extra version number added after a dash will be
stripped from the current package version when determining which metapackage to
load. This fix solves a potential catastrophe on the Grid, where jobs might end
up not having a GCC runtime environment and failing abruptly.


dberzano:
@rqwa this goes together with alisw/gcc-toolchain#1, thanks!

PR #540, opened 2016-10-01
Update AGILe to v1.4.1-alice1

dberzano:


dberzano:
@qgp this is done with a note: since some months we've decided that the only action that triggers a recompilation with `aliBuild` is updating the recipes. We've actually changed our original policy. In practice we don't want users to silently get an AGILe update _without_ updating the recipe because the old `alice/v1.4.1` branch received an update.

To work around this change of conventions,

* Old `alice/v1.4.1` branch has been force-restored to the previous state.
* New `alice/v1.4.1-patches` branch has been created with your fix on top.
* Tag `v1.4.1-alice1` has been created on that branch and has been updated in this recipe.

PR #543, opened 2016-10-05
xalienfs - Add brew --prefix openssl flags

aaniin:
Homebrew refuses to link openssl, even with the --force flag. Thus, the full preprocessor and linker flags have to be provided as suggested by homebrew itself. This PR adds the openssl flags for xalienfs which otherwise fails to compile for me.
(sierra, clang 8.0, brew 1.0.6)


dberzano:
Hello @aaniin, we've created/merged #547 which should implement this change in a way that's more uniform with respect to other similar recipes. Please check if the new `IB/v5-08/next` works for you!

Thanks for reporting!

PR #550, opened 2016-10-20
Improve ROOT 6 and Python support

dberzano:


dberzano:
@ktf to be double-checked before merging.

libpng.sh
```diff
@@ -33,7 +33,7 @@ proc ModulesHelp { } {
 set version $PKGVERSION-@@PKGREVISION@$PKGHASH@@
 module-whatis "ALICE Modulefile for $PKGNAME $PKGVERSION-@@PKGREVISION@$PKGHASH@@"
 # Dependencies
-module load BASE/1.0 ${ZLIB_ROOT:+zlib/$ZLIB_VERSION-$ZLIB_REVISION}
+module load BASE/1.0 $([[ $ALIEN_RUNTIME_VERSION ]] && echo "AliEn-Runtime/$ALIEN_RUNTIME_VERSION-$ALIEN_RUNTIME_REVISION" || echo "${ZLIB_VERSION:+zlib/$ZLIB_VERSION-$ZLIB_REVISION}")
```
ktf: Wouldn't:

```
${ALIEN_RUNTIME_VERSION:+AliEn-Runtime/$ALIEN_RUNTIME_VERSION-$ALIEN_RUNTIME_REVISION} ${ALIEN_RUNTIME_VERSION:-${ZLIB_VERSION:+zlib/$ZLIB_VERSION-$ZLIB_REVISION}}
```

Actually do the same thing? Why is this actually needed?

python-modules.sh
```diff
@@ -7,21 +7,35 @@ requires:
...
+# Force pip installation of packages found in current PYTHONPATH
+unset PYTHONPATH
 
-# Install matplotlib (very tricky)
+# The X.Y in pythonX.Y
+export PYVER=$(python -c 'import distutils.sysconfig; print(distutils.sysconfig.get_python_version())')
+
+# Install as much as possible with pip
+export PYTHONUSERBASE=$INSTALLROOT
+for X in "mock==1.0.0"         \
```
ktf: Why not:

```
pip install --user mock==1.0.0 numpy==1.9.2 certifi==2015.9.6.2 ...
```

Why the for loop?

root.sh
```diff
@@ -99,7 +101,7 @@ else
         -Dsoversion=ON                                            \
         -Dshadowpw=OFF                                            \
         -Dvdt=ON                                                  \
-        -DCMAKE_PREFIX_PATH="${FREETYPE_ROOT}"
+        -DCMAKE_PREFIX_PATH="$FREETYPE_ROOT;$SYS_OPENSSL_ROOT;$GSL_ROOT;$ALIEN_RUNTIME_ROOT;$PYTHON_ROOT;$PYTHON_MODULES_ROOT"
```
ktf: Is this the part which fixes the issue we discussed yesterday?

root.sh
```diff
@@ -141,11 +143,16 @@ proc ModulesHelp { } {
 set version $PKGVERSION-@@PKGREVISION@$PKGHASH@@
 module-whatis "ALICE Modulefile for $PKGNAME $PKGVERSION-@@PKGREVISION@$PKGHASH@@"
 # Dependencies
-module load BASE/1.0 ${ALIEN_RUNTIME_ROOT:+AliEn-Runtime/$ALIEN_RUNTIME_VERSION-$ALIEN_RUNTIME_REVISION} ${GSL_VERSION:+GSL/$GSL_VERSION-$GSL_REVISION} ${FREETYPE_VERSION:+FreeType/$FREETYPE_VERSION-$FREETYPE_REVISION}
+module load BASE/1.0 ${ALIEN_RUNTIME_ROOT:+AliEn-Runtime/$ALIEN_RUNTIME_VERSION-$ALIEN_RUNTIME_REVISION} \\
```
ktf: Please align `\\`

python.sh
```diff
@@ -72,7 +71,9 @@ proc ModulesHelp { } {
 set version $PKGVERSION-@@PKGREVISION@$PKGHASH@@
 module-whatis "ALICE Modulefile for $PKGNAME $PKGVERSION-@@PKGREVISION@$PKGHASH@@"
 # Dependencies
-module load BASE/1.0 ${ALIEN_RUNTIME_VERSION:+AliEn-Runtime/$ALIEN_RUNTIME_VERSION-$ALIEN_RUNTIME_REVISION} ${ZLIB_VERSION:+zlib/$ZLIB_VERSION-$ZLIB_REVISION}
+module load BASE/1.0 $([[ $ALIEN_RUNTIME_VERSION ]] && echo "AliEn-Runtime/$ALIEN_RUNTIME_VERSION-$ALIEN_RUNTIME_REVISION" || echo "${ZLIB_VERSION:+zlib/$ZLIB_VERSION-$ZLIB_REVISION}") \\
```
ktf: Same as above. BTW, is this really needed? Doesn't the current implementation do exactly the same (at the cost of an extra path)? As usual, I do not particularly like, mixing cleanups and features additions.

swan.sh
```diff
@@ -1,44 +0,0 @@
-package: SWAN
```
ktf: The reason why I had a separate package for the swan modules is that doing it as part of the Python-modules will force a rebuild of a bunch of generators for no particular good reason. Is this actually something we want? If yes, we should be extremely careful when merging in production and announce it properly to the various generators users.

python-modules.sh
```diff
@@ -7,21 +7,35 @@ requires:
...
   - curl
 env:
-  SSL_CERT_FILE: "$(export PYTHONPATH=$PYTHON_MODULES_ROOT/lib/python2.7/site-packages:$PYTHONPATH; export PATH=$PYTHON_ROOT/bin:$PATH; export LD_LIBRARY_PATH=$PYTHON_ROOT/lib:$LD_LIBRARY_PATH; python -c \"import certifi; print certifi.where()\")"
+  SSL_CERT_FILE: "$(env PYTHONPATH=$PYTHON_MODULES_ROOT/lib/python$(python -c \"import distutils.sysconfig; print(distutils.sysconfig.get_python_version())\")/site-packages:$PYTHONPATH python -c \"import certifi; print certifi.where()\")"
 prepend_path:
   PYTHONPATH: $PYTHON_MODULES_ROOT/lib/python2.7/site-packages:$PYTHONPATH
 prefer_system: (?!slc5)
 prefer_system_check:
-  python -c 'import matplotlib; import numpy; import certifi'
+  python -c 'import matplotlib; import numpy; import certifi; import IPython; import ipykernel; import yaml'
```
ktf: This should probably become:

```
import matplotlib,numpy,certify,...
```

python-modules.sh
```diff
@@ -47,15 +60,16 @@ if [[ $FREETYPE_ROOT ]]; then
 fi
 perl -p -i -e "s|'darwin': \['/usr/local/'|'darwin': ['$INSTALLROOT'|g" setupext.py
 
-mkdir -p $INSTALLROOT/lib64/python2.7/site-packages $INSTALLROOT/lib/python2.7/site-packages
+mkdir -p $INSTALLROOT/lib/python$PYVER/site-packages \
```
ktf: How about:

```
mkdir -p $INSTALLROOT/{lib,lib64}/python$PYVER/site-packages
```

?

python.sh
```diff
@@ -16,11 +16,10 @@ prefer_system_check:
 ---
 #!/bin/bash -ex
 
-case $ARCHITECTURE in 
+case $ARCHITECTURE in
   slc5*) ;;
   *)
-    echo "Building our own python. If you want to avoid this, please install python >= 2.7"
```
ktf: This is probably more accurate for python itself, however wouldn't it make more sense to move the message about the missing modules to the `python-modules.sh` recipe, rather than dropping it completely?

libpng.sh
dberzano: Yep I think you're right.

python-modules.sh
dberzano: Yep this is smarter.

python-modules.sh
dberzano: Unfortunately `pip` does not fail (exitcode != 0) if at least one of the packages installed correctly. So we have to separate the `pip`s in order to catch potential errors.

python-modules.sh
dberzano: You were against it some time ago because you said it makes harder to do a "find" in the recipes :-)

root.sh
dberzano: Yes - and this actually fixes the problem without patching ROOT. (ftr there was a problem in my private config that prevented it from working)

python.sh
dberzano: Agreed.

swan.sh
dberzano: Yes that's the idea. I would not add extra packages, I would prefer to rebuild the generators. In any case tests are being performed on the nightlies part because we don't want to pollute the production environment.

dberzano:
@ktf update to ROOT v6.09.01 (in spite of the fact that the latest stable is v6.06.08) was explicitly requested by @pzhristov for testing. We cannot use "pure" v6.09.01 as the tagged version had a missing header. A couple of commits after the tag the missing header was fixed.

dberzano:
As agreed we are merging this to next. Some tests are failing apparently in Rivet, so the caveat is that the standard AliPhysics chain will work if using `--defaults root6` in aliBuild (not sure about the rest for now).

PR #551, opened 2016-10-20
Ib/v5 08/dev

mfasDa:
While installing aligenerators I came across several issues:
- I provided BOOST as external installation in my defaults file, still aliBuild wanted to build it and ignored my existing BOOST_ROOT -> test adapted to allow for external BOOST_ROOT
- CRMC: Compiler error in ROOT interface as root was compiled under c++14, CRMC however under c++98. Module adapted in the way that it handles both c++14 and c++11, depending on which standard was used in the root installation. This part is only handled if root-config is found. Also the interface to fastjet is fixed.
- JEWEL: tag from defaults was included in the filenames of the jewel files in install, making the install fail because the files it was looking for do not exist. It can probably be fixed in a more elegant way by the experts, but for the moment it is necessary to remove the defaults tag from the version tag.

### DIFF
--- a/boost.sh
+++ b/boost.sh
@@ -8,7 +8,7 @@ build_requires:
  - "bz2"
 prefer_system: (?!slc5)
 prefer_system_check: |
-  printf "#include \"boost/version.hpp\"\n# if (BOOST_VERSION < 105900)\n#error \"Cannot use system's boost. Boost > 1.59.00 required.\"\n#endif\nint main(){}" | gcc -I$(brew --prefix boost)/include -xc++ - -o /dev/null
+  printf "#include \"boost/version.hpp\"\n# if (BOOST_VERSION < 105900)\n#error \"Cannot use system's boost. Boost > 1.59.00 required.\"\n#endif\nint main(){}" | gcc -I$(brew --prefix boost)/include -I$BOOST_ROOT/include -xc++ - -o /dev/null
 ---
 #!/bin/bash -e
 
--- a/crmc.sh
+++ b/crmc.sh
@@ -16,9 +16,22 @@ case $ARCHITECTURE in
   ;;
 esac
 
+# In case root is available determine std
+std=
+if [ "x$(which root-config)" != "x" ]; then
+  if [ $(root-config --has-cxx11) == "yes" ] ; then
+    std="c++11"
+  fi
+  if [ $(root-config --has-cxx14) == "yes" ] ; then
+    std="c++14"
+  fi
+fi
+
 cmake $SOURCEDIR                               \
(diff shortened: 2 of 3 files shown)


boost.sh
```diff
@@ -8,7 +8,7 @@ build_requires:
  - "bz2"
 prefer_system: (?!slc5)
 prefer_system_check: |
-  printf "#include \"boost/version.hpp\"\n# if (BOOST_VERSION < 105900)\n#error \"Cannot use system's boost. Boost > 1.59.00 required.\"\n#endif\nint main(){}" | gcc -I$(brew --prefix boost)/include -xc++ - -o /dev/null
+  printf "#include \"boost/version.hpp\"\n# if (BOOST_VERSION < 105900)\n#error \"Cannot use system's boost. Boost > 1.59.00 required.\"\n#endif\nint main(){}" | gcc -I$(brew --prefix boost)/include -I$BOOST_ROOT/include -xc++ - -o /dev/null
```
ktf: System provides are not supposed to have $BOOST_ROOT defined. If it cannot be found in the usual paths, it should not be considered a system default. Only exception is brew on mac, which is the de facto standard in terms of package management.

crmc.sh
```diff
@@ -16,9 +16,22 @@ case $ARCHITECTURE in
   ;;
 esac
 
+# In case root is available determine std
```
ktf: ROOT is not a dependency of `CRMC`, AFAICT. If it can be, it should be properly declared. Moreover, why do you set only `c++XX`? Would't it make sense to have something like:

```
STD="-std=c++XX"
```

? Also I  think you can cleanup a bit the if condition to make it more linear.

crmc.sh
```diff
@@ -16,9 +16,22 @@ case $ARCHITECTURE in
...
+    std="c++14"
+  fi
+fi
+
 cmake $SOURCEDIR                               \
       ${BOOST_ROOT:+-DBOOST_ROOT=$BOOST_ROOT}  \
-      -DCMAKE_INSTALL_PREFIX=$INSTALLROOT
+      -DCMAKE_INSTALL_PREFIX=$INSTALLROOT \
+      ${std:+-DCMAKE_CXX_FLAGS="$(printf "\-std=%s" "$std")"} \
+      ${FASTJET_ROOT:+-DFastjet_DIR=$FASTJET_ROOT}
```
ktf: Fastjet is not a dependency. It it can be, can you declare it as such?

dberzano:
This one looks quite obsolete now - I am closing it and invite you to create new PRs based on the upstream code.

PR #561, opened 2016-11-03
Update DDS / Mesos versions

ktf:


dberzano:
@ktf I think this can be merged

ktf:
Actually no. There is some sort of conflict between our version of protobuf and the one mesos 0.28.x expects.

dberzano:
OK I have actually picked the DDS update alone for another PR (I did not push it yet).

ktf:
Sounds good actually.

PR #562, opened 2016-11-04
nanomsg: bump to version 1.0

sbinet:
This CL bumps the nanomsg version to the first production-grade
release of nanomsg: 1.0

### DIFF
--- a/nanomsg.sh
+++ b/nanomsg.sh
@@ -1,13 +1,33 @@
 package: nanomsg
-version: v0.8
+version: v1.0.0
 source: https://github.com/nanomsg/nanomsg
-tag: 0.8-beta
+tag: 1.0.0
 build_requires:
-  - autotools
+  - CMake
 ---
 #!/bin/sh
-rsync -a --delete --exclude '**/.git' $SOURCEDIR/ ./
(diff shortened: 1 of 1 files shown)


nanomsg.sh
```diff
@@ -1,13 +1,34 @@
...
+version: v1.0.0
 source: https://github.com/nanomsg/nanomsg
-tag: 0.8-beta
+tag: 1.0.0
 build_requires:
-  - autotools
+  - CMake
 ---
 #!/bin/sh
 rsync -a --delete --exclude '**/.git' $SOURCEDIR/ ./
```
ktf: Sorry for taking long to review this. cmake should be able to build out of sources. Can you please check this is the case and use something like:

```
cmake $SOURCEDIR -DCMAKE_INSTALL_PREFIX:PATH="${INSTALLROOT}"
```

and then drop the rsync?

sbinet: done.
PTAL

(I'll squash when LGTM'd)

sbinet:
squashed

ktf:
Mesos DDS issue unrelated. Merging this as `o2-daq` still compiles.

PR #564, opened 2016-11-08
Make libxml2 dependency explicit

ktf:
We now have a ROOT6 dependency on libxml2, without actually
making it explicit. This should fix the issue.


dberzano:
See #596.

PR #571, opened 2016-11-14
Use FastJet v3.2.1_1.024 with CINT fix

dberzano:


dberzano:
We had to hide some C++14 and C++11 constructs from the view of `rootcint`. Note that the unpatched FastJet version does not show any problem with our own compiler (v4.9.3). C++14-enabled compilers are triggering the problem. The result is that AliPhysics cannot compile on modern platforms (_e.g._ Ubuntu 16.10 with its GCC v5.4.0). This patch fixes the problem.

PR #575, opened 2016-11-21
Use $IGNORE_ERRORS to compile with -k

ktf:
Now that defaults allow us to do so, let's use an environment variable to
control wether or not the build should fail on error.


ktf:
@dberzano can you merge this for the next release?

dberzano:
Makes perfect sense to me. This also solves a problem for some users. The variable `IGNORE_ERRORS` is supposed to be set manually by the user, right?

sawenzel:
@dberzano: Yes, from my understanding we should catch errors by default and ignore them only upon request.

dberzano:
Ported in `IB/v5-08/next` (this PR is to prod). Closing, will soon merge to prod.

PR #576, opened 2016-11-22
Synchronize ROOT 6 and O2 defaults

dberzano:
* ROOT 6 updated to the latest stable version (v6.08.00) for `root6`, `o2` and `o2-daq`

### DIFF
--- a/defaults-o2.sh
+++ b/defaults-o2.sh
@@ -8,16 +8,29 @@ disable:
   - AliEn-Runtime
   - AliRoot
 overrides:
-  ROOT:
-    version: "%(tag_basename)s"
-    tag: "v6-06-04"
+  ROOT@969984f5f25c5c5326d6b4d4f20e72b0ffad164b:
+    tag: "v6-08-02"
+    source: https://github.com/root-mirror/root
+    requires:
+      - AliEn-Runtime:(?!.*ppc64)
+      - GSL
+      - opengl:(?!osx)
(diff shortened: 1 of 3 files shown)


dberzano:
@ktf note that this requires an update to O2 to support the new DDS API...

dberzano:
For the record it compiles on macOS Sierra with `--defaults o2 --disable DDS` (DDS 1.4 does not work yet for the reasons mentioned above).

dberzano:
@matthiasrichter please try this one - note that this does not include any DDS update, so if you use a "too recent" version of FairRoot (which now should work with DDS v1.4) it should _not_ compile.

dberzano:
I have changed the PR to get rid of what @matthiasrichter has already done in #581, and to include `o2-daq`. @ktf can we merge this?

ktf:
I'd like to see `build/AliPhysics/root6` to succeed, first. ;-)

dberzano:
Merged as AliPhysics builds with it. Uses ROOT 6, v6-08-02.

dberzano:
For the record the problem's in AliPhysics and has to be fixed there.

dberzano:
@ktf ...and I cannot reproduce it on my Ubuntu 16.04 for the record, where all libraries are loaded flawlessly.

ktf:
Smells like some platform issue with fastjet, though.. We should try to keep it in mind.

PR #577, opened 2016-11-22
Add the recipes for the Configuration module

Barthelemy:
New attempt directly to prod branch


ktf:
Now builds correctly while testing with O2-daq. Merging. The other issues are unrelated and will have to be fixed separately.

PR #581, opened 2016-11-29
Find DDS o2 and fairroot setups. Fixes AliceO2Group/AliceO2#186

matthiasrichter:


ktf:
Most likely unrelated. I am checking. Is DDS now mandatory for both FairRoot / O2? If not, how about:

```
${DDS_ROOT:+-DDS_PATH=$DDS_ROOT}
```

so that people can disable it with `--disable DDS` if they do not need it?

matthiasrichter:
@ktf 
No, DDS is not mandatory for neither FairRoot nor O2. However, we have it as a fixed dependency in the recipes. But that can be disabled, true, I ignored this.

I think its in any case much better to use the conditional flag as you propose. Should I add another commit to the PR which then can be squashed? Or do you want to do it after checking what happend with the tests?

ktf:
I'd prefer one single commit. You can either add one more (and I will do the squash at merge time) or do the rebase yourself. Your call.

matthiasrichter:
I have added a correction now, squashed the two commits, and did a forced push to my branch. So its ready to be merged depending on the outcome of the tests.

ktf:
Given `O2-DAQ` compiles, I am merging this.

PR #585, opened 2016-12-01
Move defaults root6 to v6-08-00

ktf:


ktf:
@dberzano any objection to this? Given we are not affected by the persistency bug in ROOT recently discovered any reason for not using v6-08-00? Do we have any other thing in our own branch which prevents the usage of the official tag?

dberzano:
Subset of #576, closing.

PR #587, opened 2016-12-05
Add missing DDS_ROOT variable for CMake

Barthelemy:
On CC7 (at least) not passing DDS_ROOT to CMake prevent DDS from being found and thus a number of modules to be built (e.g. QC).


Barthelemy:
It is already fixed here : https://github.com/alisw/alidist/commit/999d722a9cb71acd2a69aa557e882889638648f7

PR #591, opened 2016-12-08
Declare Vc as external dependency to AliRoot

sawenzel:
* AliRoot supports building against external Vc
 * This mode is needed in future in order to build other dependencies against Vc


dberzano:
The rebased version of this is #594. Closing, continuing on that one. Thanks @sawenzel!

Issue #536, opened 2016-09-28
FairRoot does not compile on Mac Sierra


dstocco:
Dear all,
I've been trying to install O2-daq on Mac Sierra following the instructions [here](https://alisw.github.io/alibuild/o2-tutorial.html)
but I cannot compile FairRoot.

I'm using the branch "next" of alibuild and "IB/v5-08/prod" of alidist
I build with the command:
`aliBuild -z -w $ALICE_WORK_DIR --defaults o2-daq build O2`

But I keep getting this error:
[log.txt](https://github.com/alisw/alidist/files/497987/log.txt)

Looks like it's searching for sodium in INSTALLROOT instead searching for it in the proper install directory.

Am I doing something wrong?

Thanks in advance,
best regards,
Diego

dstocco:
Hi all,
is there any news on this issue?
As of today, FairRoot is still not working.
Thanks in advance,
best regards,
Diego

dstocco:
Closing the issue since it was solved elsewhere at last.

PR #592, opened 2016-12-08
Adaptions needed for ROOT6

mfasDa:
- ROOT source set to official root mirror
- GSL prefer_system_check accepts GSL versions >= 2

### DIFF
--- a/defaults-o2.sh
+++ b/defaults-o2.sh
@@ -11,6 +11,10 @@ overrides:
   ROOT:
     version: "%(tag_basename)s"
     tag: "v6-06-04"
+    source: https://github.com/root-mirror/root
+  GSL:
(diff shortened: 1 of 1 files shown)


mfasDa:
Following issues were found:

- ROOT tag specified not available on the alice mirror (alice mirror out of sync), consequently it has to be obtained from the official root mirror
- As ROOT6 v6-06-04 is used, GSL versions >= 2 are supported by root. The prefer_system_check has been ported from defaults-root6.sh

Without system GSL I was running in a linker error similar to the one reported in #536 (referencing library path in INSTALLROOT)

dberzano:
Could you check #576? I opened it a long time ago, and just merged it.

PR #597, opened 2016-12-10
Use $IGNORE_ERRORS to compile with -k

dberzano:
Now that defaults allow us to do so, let's use an environment variable to
control wether or not the build should fail on error.


dberzano:
Same as #575 but on `next`.

PR #607, opened 2016-12-21
make o2 depend on Vc

wiechula:


ktf:
Do you really need Vc 1.3.0? That would force us to rebuild AliRoot and I think there is some C++11 issue there which we cannot afford. How about you modify `defaults-o2.sh` and `defaults-o2-daq.sh` to update it only for O2? You basically simply need to add:

```
  Vc:
    tag: "v1.3.0"
```

ktf:
@sawenzel do you remember the story for Vc 1.3.0?

wiechula:
I don't know the details of the Vc version history. I only saw that there is a newer version and took the change to update the version to 1.3.0.
It might well be that 1.2.0 is good for O2. @sawenzel  can comment.

sawenzel:
I think we should be able to live with Vc 1.2.0. For the moment. I am not aware of any particular reason to move to 1.3. (for the moment).

ktf:
Ok, I changed this to 1.2.0. Let's see the tests.

ktf:
Test did not run because the PR was for the next branch. Given This only changes O2, I've moved the PR to be directly in /prod

ktf:
Rivet error unrelated. `build/O2/o2-daq` is green and that's what matters.

PR #610, opened 2017-01-10
Add DDS to the defaults for o2-daq

Barthelemy:
It is needed !


ktf:
Yeah, I think it got dropped while I was checking the memory exhaustion problem.

PR #612, opened 2017-01-11
Use AliRoot and AliPhysics from GitHub

dberzano:
* Introduce external AliRootData and AliPhysicsData packages containing "big"
  and "binary" files only
* Use AliRoot and AliPhysics from GitHub
* Install AliRoot and AliPhysics data inside AliRoot and AliPhysics installation
  directories (data repos are just "build time" dependencies)


dberzano:
@qgp, @mpuccio, @ktf (I'd like to tag Jan Fiete but I don't know his gh account name).

This is one possible way, to be discussed with you, of splitting the packages.

For AliRoot and AliPhysics there are two "data" repositories (hosted on CERN GitLab) called AliRootData and AliPhysicsData. Those two packages have corresponding "build recipes" which are seen as "build time dependencies" by AliRoot and AliPhysics.

Eventually, the AliRoot and AliPhysics installation directory will also contain data from the AliRootData and AliPhysicsData directories.

The user workflow would be to have four packages in "development mode". To be done only once:

    aliBuild init AliRoot AliRootData AliPhysics AliPhysicsData

For analyzers

qgp:
Hi @dberzano,

maybe I overlook something but this means that AliRootData and AliPhysicsData are always pulled in for the build, right? In what we discussed before, we considered the option not to require these (e.g. for analysis OCDB is not needed). Was this a deliberate decision to drop this idea?

Cheers,
   Jochen

dberzano:
Hello @qgp, thanks for your feedback.

* For what concerns AliPhysicsData (which is basically the OADB) it is dangerous not to pull it, so I would not make it optional
* For what concerns AliRootData, you should have a look at the [repository](https://gitlab.cern.ch/alisw/AliRootData): it does not only contain the OCDB, but also plenty of ROOT/binary/data/whatever files which _might_ or _might not_ be needed, who knows.

Personally (but again we can discuss it) I think that we simply have no time to put under scrutiny what ROOT/binary/data files are needed and what are not :-(

O2 will be the real cleanup of AliRoot. The objective of this migration campaign is:

* to have a code repository which is leaner,
* to introduce more control over the code by introducing pull request tests, clearer error feedback and a more transparent approval procedure.

What I propose is that for the "common" user every piece of data will always be downloaded, by default. One can always `--disable` unwanted packages. We can have a `--default` that disables AliRootData.

qgp:
Hi @dberzano,

thanks for the explanation. If we take a deliberate decision to handle it this way fine with me, it was just different from what we had talked about some time last year. And then my understanding was still that we use that occasion to clean things up a bit. If we don't do it now, it won't happen before O2 - but that might be the pragmatic way to go.

Cheers,
   Jochen

dberzano:
Hello @qgp, yes, given the expected lifetime of AliRoot, which will be superseded by O2, it might not be worth/pragmatic to clean things up. Even having AliRootData and AliPhysicsData as separate packages, with separate environment variables, would require quite some work, therefore the proposal to maintain the expected installation directory structure. My experience with AliRoot -> AliRoot+AliPhysics taught me that one cannot simply `sed` to replace paths :-)

In my last commit I have allowed for disabling data packages.

One should also consider that the new data repository does not have the full history preserved, so the clone is thinner than before. For use cases which allow it, one can also download it and never update it.

qgp:
Hi @dberzano, fine with me then, thanks!

jgrosseo:
The OADB updates are very frequent. I think it is not a good solution that the update of AliPhysicsData triggers a full rebuild of AliPhysics? Is there a reason for it? I mean a non-technical one? The binaries should not affect the build...
Will we then daily tag AliPhysics and AliPhysicsData?

jgrosseo:
Hi,
Thinking further about this, I believe it is a bad idea to have part of the repository in github and part in gitlab. In this way we depend on two services. Two pulls have to be done which is error prone.
If we cannot have binaries in github, it would be better to put the whole AliPhysics in gitlab and keep it together.
Cheers, Jan Fiete

dberzano:
Hello @jgrosseo, thanks for your feedback. Let me try to answer to your points.

* First off, this PR is a proposal - a starting point for a discussion. Nothing has been decided yet.

* **Data files need to be separated from the repositories whether we stay on GitHub or GitLab.** Keeping the code repository lean is of utmost importance (and this work will be probably inherited by O2 too): we had absolutely no guarantee from CERN IT that so many large forks are handled efficiently. In fact, they suggest not to do that.

* **We already depend on two services (actually, more than two).** Notably, we strongly depend on GitHub already. This repository, for instance, is on GitHub. If you look at the `source:` line in most recipes, it will point to GitHub. If we don't want to depend on two services then 

* **Rebuilding AliPhysics if AliPhysicsData changes can be fixed in aliBuild.** It was already somehow in the plans to have dependencies that don't cause rebuilds.

* **We do not plan to implement the workflow on the Data repositories.** The number of persons who change the OADB is small. We can restrict them at GitLab level. We can even decide not to have pull/merge requests for Data at all. TBD.

* **Our other options for storing data in a versioned fashion are unsatisfactory or unstable.** We have discussed (via email, in mid-August, on the public mailing list):

  * CVMFS (but currently does not support a clear way to use versions, and people will have to install it, which can be very problematic).
  * Git LFS (but instead of making people doing `git pull` twice it will involve installing an extra tool and learning new commands. We have found it quite complicated to use for the average user).
  * Git Annex (same points of LFS, in addition it is even more complicated to use).
  * FTP/HTTP/rsync (they don't offer versioning - in fact, Git can be seen as "versioning" over HTTP, and more).

One potential solution would be the GEANT4 one. The `make` phase downloads data files from an external place, and it will download the correct version. We will however use the "caching" capabilities of aliBuild, which will make subsequent builds much faster. But if people frequently use incremental builds of AliPhysics only, the impact would be limited.

Still, if a change needs to be made, committers will need to clone the Data package separately. Keep in mind that the list of users that made at least 10 commits under OADB is 16 people long:

```
29	atoia
24	cholm
12	cvetan
19	dberzano
229	ddobrigk
91	ekryshen
35	gconesab
34	hristov
10	jgrosseo
22	mfiguere
37	mfloris
134	morsch
73	prsnko
16	shahoian
15	wiechula
22	zconesa
```

All of them are, reasonably, experts and capable of dealing with both proposals, in my opinion. Notably, the last proposal involves them only (other users won't need a second clone at all).

qgp:
Hi @dberzano,

as this is a proposal, it would make sense to discuss it, I think. And for this I would consider the git working group the right forum. I think it makes sense to open the discussion now instead of waiting  until we don't have time to discuss any more.

Cheers,
   Jochen

PR #613, opened 2017-01-12
Provide recipes for O2 code checks

sawenzel:
Providing initial recipes to build and execute a basic O2 code checking tool.
Updating existing clang recipe (bump version + modulefile support).

### DIFF
--- a/clang.sh
+++ b/clang.sh
@@ -1,12 +1,23 @@
 package: Clang
-version: "3.6"
+version: "3.9"
 source: https://github.com/alisw/clang
 tag: master
 ---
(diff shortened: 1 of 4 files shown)


sawenzel:
Jenkins fails because the recipes are related to O2 and need o2 defaults.

ktf:
Nothing really depends on this. Merging.

ktf:
Actually, how about you do this in prod? Why next?

sawenzel:
Just thought that it is normal procedure to introduce things in next which are released into prod.

PR #621, opened 2017-02-02
Add HIJING fortran code

ktf:


ktf:
@sawenzel @wiechula Since we discussed this over lunch, a first attempt at having at least HIJING itself available as a separate package.

PR #624, opened 2017-02-06
Add an O2 dependency on HIJING 

brinick:


ktf:
Thanks. Can you also disable it in `defaults-o2-daq.sh`?

PR #626, opened 2017-02-07
Drop ZeroMQ dependency on sodium

ktf:
Sodium is hard to build and not really needed.


ktf:
@sawenzel

ktf:
Errors are unrelated. Only O2 / FairRoot depend on zeromq right now. Merging.

PR #629, opened 2017-02-14
Do not build Vc tests

sawenzel:
By default, do not build the Vc tests
 which takes a very long time


ktf:
o2-daq, builds. Merging.

PR #632, opened 2017-02-14
Allow to build pythia without lhapdf support.

fuhlig1:


ktf:
I think this can safely go into prod. Changing the destination branch.

ktf:
Mmm... I actually need to rebase it.

PR #637, opened 2017-02-16
Add defaults for GitHub dailies

dberzano:


ktf:
How about we put the AliRoot version here as well?

dberzano:
We cannot do that unless we want all builds to have the default name `_GITHUB` appended in production if you see what I mean. This default is meant to have `_GITHUB` appended for test dailies.

ktf:
You could change the name in the default as well.

dberzano:
I have to change the name of quite a number of packages, which I find cumbersome, time-consuming and error-prone. I prefer to use the release default from a separate branch, which is clean and simple...

ktf:
Once we merge alisw/alibuild#415 that should not be an issue, as only the packages affected by a default will be applied the new default, not all of them.

dberzano:
Just to clarify the scope. Note that this is a PR to the `git-migration` branch and must enable us, as of now, to publish daily builds on the Grid from GitHub. This part is not intended to live after the migration to GitHub and will never enter `IB/v5-0X/prod`. We will deal with a better way of doing this after the priority item (Git migration) is done smoothly.

Issue #642, opened 2017-02-21
Add support for boost_python


Barthelemy:
In recipe for boost, python_boost should be enabled (needed for flp prototype).

Barthelemy:
I am preparing a PR for this. Could @ktf or @dberzano tell me why the recipe for python depends on AliEn-Runtime ? that does not seem to make sense.

ktf:
I agree we should reevaluate this.

ktf:
Because we package a few alien specific dependencies like openssl inside
alien itself. Indeed it makes no much sense but it was done to minimize the
number of externals we publish for AliRoot on CVMFS.

Barthelemy:
ok thank you for your explanation. It is not a problem for me, I was just curious.

ktf:
Comments arrived in the wrong order, by the way... ;-)

Barthelemy:
I noticed :) but I deciphered your messages nonetheless

Barthelemy:
@dberzano I think that this is already fixed. Can you confirm ?

PR #643, opened 2017-02-23
Add boost_python in boost recipe

Barthelemy:
See ticket #642


Barthelemy:
@dberzano isn't this covered by your recent work ? if so we can kill this PR.

dberzano:
Indeed.

PR #644, opened 2017-03-01
Generate compile_commands.json for O2

ktf:


aphecetche:
Hi,

Just curious : is that by chance to allow better integration with editors for tools like e.g. YouCompleteMe ? ;-)

o2.sh
```diff
@@ -57,6 +57,17 @@ if [[ $GIT_TAG == master ]]; then
...
+DEVEL_SOURCES="`readlink $SOURCEDIR || echo $SOURCEDIR`"
+# This really means we are in development mode. We need to make sure we
+# use the real path for sources in this case. We also copy the
+# compile_commands.json file so that IDEs can make use of it directly, this
+# is a departure from our "no changes in sourcecode" policy, but for a good reason
+# and in any case the file is in gitignore.
+if [ "$DEVEL_SOURCES" != "$SOURCEDIR" ]; then
+  perl -p -i -e "s|$SOURCEDIR|$DEVEL_SOURCES|" compile_commands.json
+  ln -sf $BUILDDIR/compile_commands.json $DEVEL_SOURCES/compile_commands.json
+fi
```
sawenzel: I am bit confused here since you don't actually seem to produce `compile_commands.json` using the "CMAKE_EXPORT_COMPILE_COMMANDS" option.

sawenzel:
@aphecetche : This is primarily so that our static code checker knows how to compile the source correctly. But yes ... it can be beneficial to many setups actually.

o2.sh
ktf: Got lost. Added back.

ktf:
@aphecetche indeed YouCompleteMe was my personal motivator for this (and it works like a charm).

PR #648, opened 2017-03-03
Add Python and ROOT_INCLUDE_PATH

dberzano:


aliroot.sh
```diff
@@ -11,6 +11,8 @@ build_requires:
   - DAQ:slc6.*
 env:
   ALICE_ROOT: "$ALIROOT_ROOT"
+prepend_path:
+  ROOT_INCLUDE_PATH: "$ALIROOT_ROOT/include"
```
sawenzel: what is the purpose of ROOT_INCLUDE_PATH?

dberzano: So, first off, this of course has not to be merged yet. This is required for the QA on SWAN, see [ATO-383](https://alice.its.cern.ch/jira/browse/ATO-383) (and the addition of this export can be limited to ROOT 6-only builds because of that).

dberzano:
OK after a discussion with the relevant persons we have clarified that those changes are needed only for ROOT6. Closing this.

PR #659, opened 2017-03-22
Add AliRoot-OCDB optional package

dberzano:


dberzano:
See also alisw/alisw.github.io#50

PR #661, opened 2017-03-25
openssl: Added header check on linux

aaniin:
Hi Dario, hi Giulio,

Aaron and Antonin reported on the task force list that the openssl system availability check is not enough on Ubuntu. In fact, there was no check at all. I added the openssl/bio.h check which is done for mac OS with the brew prefix for other OS as well. Just checked on my Ubuntu 16.04 and the aliDoctor correctly reports that openssl is present. 

Cheers,
Hans

### DIFF
--- a/openssl.sh
+++ b/openssl.sh
@@ -4,7 +4,7 @@ tag: "v0.9.8_1.2.4"
 source: https://github.com/alisw/alice-openssl.git
 prefer_system: (?!slc5|slc6)
 prefer_system_check: |
-  if [ `uname` == Darwin ]; then test -d `brew --prefix openssl || echo /dev/nope` || exit 1; echo '#include <openssl/bio.h>' | c++ -x c++ - -I`brew --prefix openssl`/include -c -o /dev/null || exit 1; else exit 0; fi
+  if [ `uname` == Darwin ]; then test -d `brew --prefix openssl || echo /dev/nope` || exit 1; echo '#include <openssl/bio.h>' | c++ -x c++ - -I`brew --prefix openssl`/include -c -o /dev/null || exit 1; else echo '#include <openssl/bio.h>' | c++ -x c++ - -c -o /dev/null || exit 1; fi
 build_requires:
  - zlib
  - "GCC-Toolchain:(?!osx)"


maireiphc:
I second that pull-request by Hans.
It seem such a problem re-appears from time to time, whenever an aliBuild is installed on a freshly installed linux machine (Fedora, Ubuntu at least...).

See for instance 
https://groups.cern.ch/group/alice-project-analysis-task-force/Lists/Archive/DispForm.aspx?ID=17588, Jul 2016

marcovanleeuwen:
I also ran into this; gSOAP compilation did not work on a fresh Ubuntu 16.04 system, because openssl/bio.h was missing:

```
g++ -DWITH_OPENSSL -DWITH_GZIP -DWSDL_IMPORT_PATH="\"/opt/alice/sw/INSTALLROOT/db207d34ac1384665605c511f0fb99c78a79de29/ubuntu1604_x86-64/gSOAP/v2.7.13-1/share/gsoap/WS\"" -fPIC -g -O2   -I../../gsoap -L.  -o wsdl2h wsdl2h-wsdl2h.o wsdl2h-wsdl.o wsdl2h-schema.o wsdl2h-types.o wsdl2h-service.o wsdl2h-soap.o wsdl2h-mime.o wsdlC.cpp ../../gsoap/libgsoapssl++.a -lssl -lcrypto -lz
In file included from includes.h:38:0,
                 from wsdlStub.h:11,
                 from wsdlH.h:10,
                 from wsdlC.cpp:13:
../../gsoap/stdsoap2.h:630:26: fatal error: openssl/bio.h: No such file or directory
compilation terminated.
```

On Ubuntu 16.04, installing libssl-dev solves this.

aaniin:
Hello @ktf hello @dberzano 

Can we start tests for this pull request in alidist which adds a check for openssl on Linux? Several people are affected and support this pull request.

Thanks for considering!

Cheers,
Hans

dberzano:
Hi @aaniin, first off - can you make the pull requests to the `dev` branch, not to the `master` one? We use `master` for building our production releases, and we tend to aggregate the changes and sync the updates with AliRoot releases to avoid rebuilding the existing ones with new configurations.

In addition, I think the check can be simplified a lot by simply taking out the `if` clause on Darwin entirely. The check (on multiple lines) would become like this:

```
if [ `uname` == Darwin ]; then test -d `brew --prefix openssl || echo /dev/nope` || exit 1; fi
echo '#include <openssl/bio.h>' | c++ -x c++ - -I`brew --prefix openssl`/include -c -o /dev/null || exit 1
```

This would do the following.

* On macOS it checks if the path returned by `brew --prefix openssl` (which is where OpenSSL from Homebrew is supposed to be found) really exists. In many cases, the path is returned but OpenSSL is not there for real. If it exists, the test proceeds.
* On all platforms, it attempts to compile a dummy file including an OpenSSL header. The `brew` command is substituted with `true` on non-macOS platforms if I recall correctly (can @ktf confirm?)

ktf:
Regarding the command brew on non-macOS platforms, you are correct.

https://github.com/alisw/alibuild/blob/b784346f02dbfda687437336fb7083cc0d4a5a54/aliDoctor#L135

aaniin:
I created a new PR #706

PR #665, opened 2017-03-30
Allow to build pythia without lhapdf support.

ktf:


ktf:
@qgp Would you agree?

qgp:
fine for me

PR #670, opened 2017-04-05
Adapt to new FairRoot version

ktf:
- Bump CMake version for O2 related packages
- Bump googletest and add it as requirement
- Cleanup googletest recipe to use make install

### DIFF
--- a/defaults-o2-daq.sh
+++ b/defaults-o2-daq.sh
@@ -32,6 +32,11 @@ overrides:
     tag: "v3.0.2"
   O2:
     version: "%(short_hash)s%(defaults_upper)s"
+  CMake:
+    version: "%(tag_basename)s"
+    tag: "v3.7.2"
+    prefer_system_check: |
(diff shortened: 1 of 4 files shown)


ktf:
@sawenzel this is a start, I think they also bumped boost requirements, but I need to test.

defaults-o2-daq.sh
```diff
@@ -32,6 +32,11 @@ overrides:
     tag: "v3.0.2"
   O2:
     version: "%(short_hash)s%(defaults_upper)s"
+  CMake:
+    version: "%(tag_basename)s"
+    tag: "v3.7.2"
```
matthiasrichter: What is the reason to jump to version 3.7.2 while FairRoot requires 3.6?

defaults-o2-daq.sh
```diff
@@ -32,6 +32,11 @@ overrides:
...
+    prefer_system_check: |
+      which cmake && case `cmake --version | sed -e 's/.* //' | cut -d. -f1,2,3 | head -n1` in [0-2]*|3.[0-5].*) exit 1 ;; esac
```
matthiasrichter: considered that 3.7 is used as tag, the system check should probably be `3.[0-6].*`

PR #673, opened 2017-04-05
Use tagged version of FairRoot for builds

ktf:


ktf:
Giving up on compiling the current dev. I am fixing to the working tag, for the moment, then we see. One can still build dev by using a development package, of course, but at the moment it will fail, I think due to https://github.com/FairRootGroup/FairRoot/commit/8b6114ee51a538b12764fa5504822ebe0ba2f564

PR #679, opened 2017-04-07
Move to GCC v6.2, C++14 and stable builds

dberzano:


dberzano:
@ktf @sawenzel @mkrzewic @awegrzyn not yet to be merged, but almost - I am doing the final moves to finally enable C++14 and stabilizing the FairRoot builds we use.

awegrzyn:
Hi @dberzano,
This PR still prevents from using gcc 6 from devtoolset-6 - see [gcc-toolchain.sh:15](https://github.com/dberzano/alidist/blob/wp3/gcc-toolchain.sh)

dberzano:
Hi @awegrzyn, nope - look at the `defaults-o2.sh` (maybe you need the `o2-daq` ones?)

awegrzyn:
Hi,
Indeed if I run it with `--defaults o2` it picks gcc correctly!
Thanks

awegrzyn:
@dberzano Could you apply following change in `libpng.sh:7`:
`
< source: git://git.code.sf.net/p/libpng/code
---
> source: https://git.code.sf.net/p/libpng/code
`
In O2 Lab we can only access outside via HTTP.

dberzano:
Please pull. I have also changed the recipe for `libxml2` which featured a `git://` URL as well.

dberzano:
Hi all and in particular @mkrzewic, I have added full C++14 support for VGM and GEANT4_VMC. Note that in order to avoid problems with software versions requiring validations, those two components are exactly the same as before plus the additional patches required for C++14, so their functionalities are unchanged.

dberzano:
I am closing it in favor of #679 (which is wrt/`master`).

PR #685, opened 2017-04-12
Add module file generation to cmake recipe

matthiasrichter:
I added to module file generation to the cmake recipe, so that the version installed with alibuild can be used more easily.

Assuming that the new `dev` branch is now the branch for development.


dberzano:
Hi, this is in #695 already which is going to be merged shortly (~minutes).

PR #695, opened 2017-04-27
Move O2 to GCC v6.2, C++14 and stable FairRoot

dberzano:
Supersedes #679.


dberzano:
Currently building [AliPhysics/AliRoot](https://alijenkins.cern.ch/job/build-any-ib/6396) with it to ensure backwards compatibility.

PR #697, opened 2017-04-28
Fix nanomsg path passed to FairRoot

dberzano:


dberzano:
@mkrzewic this fixes your problem and lets you use our own nanomsg. This one, FairRootGroup/FairRoot#548, will make FairRoot recognize system-wide nanomsg installations.

PR #698, opened 2017-04-28
ZeroMQ: check for system static lib

dberzano:
This is required by FairRoot. If not found we will compile our own.


dberzano:
@mkrzewic this fixes your problem on CC7 (ZeroMQ installed via `yum`, but FairRoot does not find it as the static library does not come with the package). The change works on macOS too.

PR #702, opened 2017-04-30
Bump Vc to v1.3.1 and fix libxml2 URL

dberzano:


dberzano:
* Vc 1.3.1 should solve some warnings with GCC 6
* Git URL for libxml2 must be HTTPS as HTTP does not work any longer

PR #711, opened 2017-05-04
Correctly load alo path

dstocco:
Fix typos that prevented the correct loading of the alo binaries and libraries in the PATH.

PR #712, opened 2017-05-04
Correctly load alo path

dstocco:
Fix typos that prevented the correct loading of the alo binaries and libraries in the PATH


dstocco:
This is like the pull-request #711 but I correctly merge to the master.
Not sure why `hub pull-request` decided to rebase to IB/v5-08/prod....

Anyways, this is the right commit.
Cheers,
Diego

PR #713, opened 2017-05-04
bump nanomsg version

dennisklein:
* fixes a broken test on macOS
 * fixes missing LC_RPATH in dylib on macOS

This is part of the resolution of AliceO2Group/AliceO2#343.

### DIFF
--- a/nanomsg.sh
+++ b/nanomsg.sh
@@ -1,7 +1,7 @@
 package: nanomsg
-version: v1.0.0
+version: 1.0.0+git_%(short_hash)s
 source: https://github.com/nanomsg/nanomsg
-tag: 1.0.0
+tag: c52f1bedca6b72fb31b473929d99f2fe90a13445
 build_requires:
   - CMake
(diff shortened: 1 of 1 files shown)


dberzano:
Hi, I have pushed it to master (0357c84c1f434942442a3c6314c555f9debe9efe), it's a slightly modified version of this commit (in particular I am maintaining the convention of having the `v` at the beginning of the version).

Out of curiosity: what is the advantage of using `cmake --build` over `make`? :-)

dennisklein:
> Out of curiosity: what is the advantage of using cmake --build over make? :-)

`make` relies on a (hidden) cmake default (I mean we usually do not explicitely call `cmake -G "Unix Makefiles"`). `cmake --build` will always work, even if they would change the default generator. Or maybe in the future we could decide to make the `-G` argument to cmake configurable for some reason, then you need to use `cmake --build`.

dberzano:
OK that's what I thought, good to know!

PR #725, opened 2017-05-09
FLP prototype recipe and dependencies

Barthelemy:


ktf:
This will have to be coordinated with an AliRoot release, since you modify APMon. A way to shortcut that is:

- Change ApMon in the dev branch
- Use ApMon-cxx@dev in the o2 default

Barthelemy:
Ok, I will add a commit for the shortcut.

Barthelemy:
ap-mon change removed

PR #727, opened 2017-05-10
Temporary disable root history when performing ctest

dstocco:
The ctest runs a series of load-libraries commands that are stored in the root history.
Since there are lots of library, the user history basically goes outside the buffer.
The patch disable the root history before running the ctest, and turn it on again when the ctest is over

### DIFF
--- a/aliphysics.sh
+++ b/aliphysics.sh
@@ -10,7 +10,9 @@ prepend_path:
   ROOT_INCLUDE_PATH: "$ALIPHYSICS_ROOT/include"
 incremental_recipe: |
   make ${JOBS:+-j$JOBS} install
+  [[ ! -z $ROOT_HIST ]] && tmpROOT_HIST="$ROOT_HIST"; export ROOT_HIST=0
   ctest -R load_library --output-on-failure ${JOBS:+-j $JOBS}
+  ROOT_HIST=$tmpROOT_HIST && [[ -z $tmpROOT_HIST ]] && unset ROOT_HIST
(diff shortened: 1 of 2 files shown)


ktf:
Can't we simply `export ROOT_HIST=0` at the beginning of the recipe?

dstocco:
Hi Giulio,
but in that case the ROOT_HIST will be permanently set to 0 in the terminal, right? So, if after compiling the user runs root or aliroot in that terminal, the history will not be saved anymore.
My idea is to disable the history only during compilation, and then recover the original value.
Now, for most of users ROOT_HIST is not set, so a simple unset ROOT_HIST at the end of the recipe would do the job. However, I tried to be more general for the (few?) users who have a custom ROOT_HIST.
Maybe I can try to check if, setting ROOT_HIST=0 *without* export can do the trick.

Cheers,
Diego

ktf:
No. The recipes run in their own separate process, so your terminal environment is preserved.

dberzano:
Hi @dstocco, thanks for pointing out this problem. This should be fixed in AliRoot and AliPhysics and not in the recipe in my opinion, can you please follow up there? The code invoking `root` for the tests is in the toplevel `CMakeLists.txt` in both cases.

dstocco:
Hi @dberzano ,
sure, this is an option. I found the code and added the ROOT_HIST=0 there.
Ok, so I close this pull request and open one for AliRoot and one for AliPhysics.

Thanks!
Ciao,
Diego

PR #729, opened 2017-05-10
Add missing cxx14 feature check

Barthelemy:


ktf:
This will have to be coordinated with a release.

Barthelemy:
It is not urgent.

PR #735, opened 2017-05-11
Use apmon-cpp recipe from dev branch and fix monitoring dependency on apmon

Barthelemy:


ktf:
Can you also modify the o2-dev-fairroot default? Or you do not plan to use it?

Barthelemy:
Sure, I can. I was not aware of it. Let's be consistent and change it as well.

Barthelemy:
Done !

PR #743, opened 2017-05-16
Adding ZeroMQ dependency for AliRoot build in defaults-o2

### LINKED ISSUE #742: ZeroMQ installation parameters not forwarded to AliRoot recipe
Because of https://github.com/AliceO2Group/AliceO2/issues/369, AliRoot needs to be build with ZeroMQ. The installation parameters need to be forwarded.

TODO:
- [x] Make AliRoot recipe depending on ZeroMQ, but *only* in the o2 defaults (this will generate the `ZEROMQ_ROOT` variable)
- [x] forward parameter to AliRoot cmake configuration

matthiasrichter:
Fixes #742
This handles https://github.com/AliceO2Group/AliceO2/issues/369

Note: AliRoot is by default disabled in the defaults, but might be
included in the build by removing this. Then the correct dependencies
must be applied.

In the override section, all dependencies have to be specified, simply
adding to the already defined dependencies is not yet supported.

### DIFF
--- a/defaults-o2.sh
+++ b/defaults-o2.sh
@@ -30,6 +30,14 @@ overrides:
       - Xdevel:(?!osx)
       - FreeType:(?!osx)
       - Python-modules
+  AliRoot:
+    requires:
+      - ROOT
+      - fastjet:(?!.*ppc64)
+      - GEANT3
+      - GEANT4_VMC
+      - Vc
+      - ZeroMQ
   GSL:
     prefer_system_check: |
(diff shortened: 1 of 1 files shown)


aliroot.sh
```diff
@@ -48,6 +48,7 @@ cmake $SOURCEDIR                                                     \
       ${ALIEN_RUNTIME_ROOT:+-DALIEN="$ALIEN_RUNTIME_ROOT"}           \
       ${FASTJET_ROOT:+-DFASTJET="$FASTJET_ROOT"}                     \
       ${DPMJET_ROOT:+-DDPMJET="$DPMJET_ROOT"}                        \
+      ${ZEROMQ_ROOT:+-DZEROMQ=$ZEROMQ_ROOT}                          \
```
ktf: This one will force people to rebuild AliRoot. We usually allow that only when we do a new release. Can you split this PR in two and do this change in dev, so that @dberzano can pick up the change next time he builds AliRoot?

matthiasrichter:
Note: this effectively only applies to defaults-o2.sh, the normal AliRoot build is not effected otherwise.

matthiasrichter:
@ktf, done. This PR now only contains the changes to defaults-o2

Issue #741, opened 2017-05-15
ROOT fails to build on Ubuntu 16.04 with alidist 26fbafc221cf59e80671a305182


sawenzel:
calling `alibuild build O2 --defaults=o2`, I get during compilation of  ROOT:

DEBUG:O2:f4ddebc4: Running /home/sandro/alisw_new/sw/SOURCES/ROOT/v6-08-02_O2/v6-08-02/build/unix/compiledata.sh
DEBUG:O2:f4ddebc4: Making include/compiledata.h
DEBUG:O2:f4ddebc4: -- Configuring done
DEBUG:O2:f4ddebc4: -- Generating done
DEBUG:O2:f4ddebc4: -- Build files have been written to: /home/sandro/alisw_new/sw/BUILD/d1fa5dfa39e25f3ad2bbfeb43bf187adc7048e3c/ROOT
DEBUG:O2:f4ddebc4: + FEATURES='builtin_pcre mathmore xml ssl opengl minuit2
DEBUG:O2:f4ddebc4:             pythia6 roofit soversion vdt  cxx14 
DEBUG:O2:f4ddebc4:             
DEBUG:O2:f4ddebc4:             '
DEBUG:O2:f4ddebc4: + NO_FEATURES=
DEBUG:O2:f4ddebc4: + bin/root-config --features
DEBUG:O2:f4ddebc4: asimage astiff builtin_afterimage builtin_ftgl builtin_gl2ps builtin_glew builtin_pcre builtin_unuran builtin_lzma builtin_llvm cxx14 cling exceptions explicitlink fortran genvector http mathmore memstat minuit2 opengl pch pythia6 pythia6_nolink python roofit root7 shared soversion thread tmva vdt xft xml x11
DEBUG:O2:f4ddebc4: + for FEATURE in '$FEATURES'
DEBUG:O2:f4ddebc4: + bin/root-config --has-builtin_pcre
DEBUG:O2:f4ddebc4: + grep -q yes
DEBUG:O2:f4ddebc4: + for FEATURE in '$FEATURES'
DEBUG:O2:f4ddebc4: + bin/root-config --has-mathmore
DEBUG:O2:f4ddebc4: + grep -q yes
DEBUG:O2:f4ddebc4: + for FEATURE in '$FEATURES'
DEBUG:O2:f4ddebc4: + bin/root-config --has-xml
DEBUG:O2:f4ddebc4: + grep -q yes
[...]

PR #745, opened 2017-05-16
Fix ROOT->GCC dependency and allow building AliRoot+O2 together

sawenzel:
* make the dependency of ROOT on GCC more explicit; fixes an issue
   where ROOT was not compiled with the required version of GCC
   (see #741) when compiling with o2 defaults
 * allow for the possibility to build AliRoot with the o2 defaults
 * disable the old explicit dependency of o2 on AliRoot which is no
   longer needed


sawenzel:
this PR is based on the dev branch. We could cherry-pick this commit into master directly.

ktf:
Given it only touched O2, I rebased this on top of master and fixed the conflict with @matthiasrichter recent commit in `defaults-o2.sh`. What about the other defaults? Can you check if the need the fix as well?

defaults-o2.sh
```diff
@@ -6,7 +6,6 @@ env:
   CMAKE_BUILD_TYPE: "RELWITHDEBINFO"
 disable:
   - AliEn-Runtime
-  - AliRoot
```
matthiasrichter: also `AliEn-Runtime` should be removed for consistency

sawenzel:
Updated this PR, added the GCC dependency for ROOT to more defaults. Also implemented suggestion by @matthiasrichter.

sawenzel:
Please squash the commits (I edited online).

sawenzel:
Closing this. Another PR was made with same content.

PR #754, opened 2017-06-01
Check copyright notice and fix Clang recipe

dberzano:


o2checkcode.sh
```diff
@@ -10,26 +9,54 @@ version: "1.0"
...
+//
+// In applying this license CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+EOF
+)"
+COPYRIGHT_LINES=$(echo "$COPYRIGHT" | wc -l)
+set +x
+while read FILE; do
+  [[ "$(head -n$COPYRIGHT_LINES "$FILE")" == "$COPYRIGHT" ]] || { printf "$FILE:1:1: error: missing or malformed copyright notice\n" >> error-log.txt; }
```
ktf: Are you sure about the `"$FILE"`?

dberzano: Yes, and it works, see [here](https://ali-ci.cern.ch/repo/logs/AliceO2Group/AliceO2/407/2c7a526eae44df727ec2f4c059dd8e68be52be79/fullLog.txt). Why?

ktf: Probably I am wrong, but don't you have wrongly nested quotes? Did you try with `$FILE` with a space in it?

dberzano: I see what you mean now :-) Yes, those two pairs of quotes are inside two different shells, so [this is fine](https://stackoverflow.com/questions/6611648/nesting-quotes-in-bash).

PR #757, opened 2017-06-06
use master of flpproto

Barthelemy:


Barthelemy:
PLEASE WAIT BEFORE MERGING< I WANT TO CHECK VERSIONS OF OTHER PACKAGES (config and monitoring)

Barthelemy:
Ready for review and merge.

configuration.sh
```diff
@@ -1,5 +1,5 @@
 package: Configuration
-version: master
+#version: master
```
ktf: Please remove commented lines

flpproto.sh
```diff
@@ -1,5 +1,5 @@
 package: flpproto
-version: alibuild-compilable
+version: master
```
ktf: I suggest you use an actual tag for the version, and if you need to develop things you use a devel package. While we added the possibility to specify branch early in alibuild development, this was source of issues, because it triggers a rebuild every time.

monitoring.sh
```diff
@@ -1,5 +1,5 @@
 package: Monitoring
-version: master
+#version: master
```
ktf: Remove commented code.

flpproto.sh
Barthelemy: I agree, let me see with my colleagues

Barthelemy:
@ktf  should be ok now.

ktf:
We should probably have a linter for recipes, since I find it unnatural that "version" comes so low, however   since we don't I will simply merge this.

Barthelemy:
thank you, indeed, I don't mind where tags end up my recipes could be linted on commit.

PR #759, opened 2017-06-08
Update O2 defaults

dberzano:
* Bump ROOT 6 to v6-08-06-80-ge2951b4
* Allow optional builds of AliRoot using O2 defaults

### DIFF
--- a/defaults-o2-dev-fairroot.sh
+++ b/defaults-o2-dev-fairroot.sh
@@ -27,7 +27,8 @@ overrides:
       which gfortran || { echo "gfortran missing"; exit 1; }
       which cc && test -f $(dirname $(which cc))/c++ && printf "#define GCCVER ((__GNUC__ << 16)+(__GNUC_MINOR__ << 8)+(__GNUC_PATCHLEVEL__))\n#if (GCCVER < 0x060200)\n#error \"System's GCC cannot be used: we need at least GCC 6.X. We are going to compile our own version.\"\n#endif\n" | cc -xc++ - -c -o /dev/null
   ROOT:
-    tag: "v6-08-02"
+    version: "v6-08-06+git_%(short_hash)s"
+    tag: e2951b4cc49ee88651a69a0b54a583333c0ae263
     source: https://github.com/root-mirror/root
     requires:
       - GSL
--- a/defaults-o2-dev-fairroot.sh
+++ b/defaults-o2-dev-fairroot.sh
@@ -5,8 +5,6 @@ env:
   CFLAGS: "-fPIC -g -O2"
   CMAKE_BUILD_TYPE: "RELWITHDEBINFO"
 disable:
-  - AliEn-Runtime
-  - AliRoot
   - simulation
   - generators
   - GEANT4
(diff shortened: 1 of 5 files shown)


dberzano:
@sawenzel this also implements the addition of `GCC-Toolchain` as a dependency, but note that - in order for that to be truly effective - we also need to update `root.sh` (another pr is coming, but we cannot merge it to master until we release the next AliRoot as it will cause a rebuild).

Can you check if this is OK, and really supersedes #745?

dberzano:
Also note that I deliberately did not touch `defaults-o2-daq.sh`.

dberzano:
As agreed with @sawenzel next time we will work on a single PR using the "Allow edits from maintainers" option (I have never used it).

defaults-o2-dev-fairroot.sh
```diff
@@ -29,14 +27,25 @@ overrides:
...
     source: https://github.com/root-mirror/root
     requires:
       - GSL
       - opengl:(?!osx)
       - Xdevel:(?!osx)
       - FreeType:(?!osx)
       - Python-modules
+      - GCC-Toolchain:(?!osx)
+  AliRoot:
+    requires:
```
sawenzel: I am wondering: why do we need to list these things? Isn't this listed in the AliRoot recipe?

dberzano: AliRoot dependencies for O2 must include ZeroMQ, which is not listed in the default `aliroot.sh` recipe. Does this answer your question?

sawenzel: Partially. Why do we need ZeroMQ for AliRoot?

dberzano: That is a good question that we should ask @mkrzewic maybe. Note that I would separate the ZeroMQ dependency problem from this PR, which just synchronizes `defaults-o2.sh` (that always had ZeroMQ) and `defaults-o2-dev-fairroot.sh`. If this is controversial I'll refactor the PR by removing this synchronization, no problem with me.

dberzano:
Note that I have:

* removed the addition of ZeroMQ to AliRoot (let's do it in a different PR if required)
* added the (accidentally) removed AliEn-Runtime dependency to ROOT

Basically, ROOT has to be built _with_ AliEn-Runtime... even if you will not build AliRoot.

I am not sure if this is exactly desirable... maybe, once again, two different defaults are really what we need?

pzhristov:
Since for the moment we do not have any replacement for AliEn, it is needed and has to be build. Lets proceed with the merge.

defaults-o2-dev-fairroot.sh
mkrzewic: Hi, the HLT part of aliroot needs zeromq, also the event display part that speaks to the HLT for online display.  In general it makes sense to have this anywhere we have the zmq library installed - which is the case for O2 installs.

sawenzel:
What is blocking this PR to be merged?

sawenzel:
Is there any news on this? Without this PR (or related fixed) ROOT is still not compiled with the correct gcc compiler and AliRoot is still strictly disabled in o2-defaults.

ktf:
@dberzano do you have time to converge on this?

defaults-o2-dev-fairroot.sh
```diff
@@ -5,8 +5,6 @@ env:
   CFLAGS: "-fPIC -g -O2"
   CMAKE_BUILD_TYPE: "RELWITHDEBINFO"
 disable:
-  - AliEn-Runtime
-  - AliRoot
```
matthiasrichter: @dberzano why would you remove this, effectively enable the packages? Probably wrong for this default configuration.

sawenzel: No no ... removing this line just does not prohibit building AliRoot with this default. It does not make AliRoot a dependency (at least with #817).

ktf:
Is this still relevant? Can we close it if not?

sawenzel:
I think it is still relevant. It allows building AliROOT with o2 defaults and correctly makes COMPILTER_TOOLCHAIN a prerequisite for ROOT.

sawenzel:
@ktf: This is the PR I was talking about (which correctly makes GCC_TOOLCHAIN a requirement to build ROOT).

sawenzel:
This PR can be closed I believe.

PR #762, opened 2017-06-15
Check if HTTP is enabled in ROOT

dberzano:
Required since alisw/AliRoot#248


dberzano:
To be merged along with the new AliRoot tag.

PR #756, opened 2017-06-02
Enable code checks only on a set of changed files

sawenzel:
We are now able to run the static analysis only on  files touched by a PR!!

PR #766, opened 2017-06-19
Revert "Enable code checks only on a set of changed files"

ktf:
Reverts alisw/alidist#756


ktf:
@sawenzel this is currently broken. I think the problem is that you key on  `-I<something>` but this might actually not be there as cmake decides to use -isystem instead. I've no time to look into this, so I am reverting this to avoid breaking current tests.

PR #770, opened 2017-06-20
Fix YODA installation w/o python from system

pkarczma:
- adding Cython which is needed for YODA
  (unclear how installation could work without so far if not provided by the system)
- PYTHON_VERSION creates confusion for YODA and Rivet,
  when set searching for pythonv2.7.10
  (still to be understood)

This PR is up for discussion and not yet to be merged before the issues are understood.

Przemek and Jochen

### DIFF
--- a/python-modules.sh
+++ b/python-modules.sh
@@ -12,8 +12,8 @@ prepend_path:
   PYTHONPATH: $PYTHON_MODULES_ROOT/lib/python2.7/site-packages:$PYTHONPATH
 prefer_system: (?!slc5)
 prefer_system_check:
-  python -c 'import matplotlib,numpy,certifi,IPython,ipywidgets,ipykernel,notebook.notebookapp,metakernel,yaml';
-  if [ $? -ne 0 ]; then printf "Required Python modules are missing. You can install them with pip (better as root):\n  pip install matplotlib numpy certifi ipython ipywidgets ipykernel notebook metakernel pyyaml\n"; exit 1; fi
+  python -c 'import matplotlib,numpy,certifi,IPython,ipywidgets,ipykernel,notebook.notebookapp,metakernel,yaml,cython';
+  if [ $? -ne 0 ]; then printf "Required Python modules are missing. You can install them with pip (better as root):\n  pip install matplotlib numpy certifi ipython ipywidgets ipykernel notebook metakernel pyyaml cython\n"; exit 1; fi
 ---
 #!/bin/bash -ex
 
@@ -31,6 +31,7 @@ If you want to avoid this please install the following modules (pip recommended)
   - notebook
   - metakernel
   - pyyaml
+  - cython
 
 EoF
 fi
@@ -52,7 +53,8 @@ for X in "mock==1.0.0"         \
          "ipykernel==4.5.0"    \
          "notebook==4.2.3"     \
          "metakernel==0.14.0"  \
-         "pyyaml"
+         "pyyaml"              \
+         "cython==0.25.2"
 do
   pip install --user $X
 done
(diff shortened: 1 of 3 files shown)


dberzano:
News on this one?

pkarczma:
Yes, I applied all the changes that were requested. As from now, `PYTHON_VERSION` variable should be unset just for `autoreconf,` `configure,` and `make,` so it should not affect modulefile generation. I added `cython` to the list of packages and an error message to that.

Also, I updated `cython` to even newer version (0.25.2) because after recent updates it was required to do so. Wouldn't it be better not to provide the version of `cython` and install the newest one available?

dberzano:
Hi @pkarczma, the point is that we don't know what's the latest version available _on the Grid_. With aliBuild we have the nice option of:

* building our own (so that we're on the safe side on the Grid),
* but use the one you have locally, if any (so that you save lots of build time locally)

I believe this covers all cases.

dberzano:
Uh as for using the latest version available: we have to to pinpoint versions if we want to be able to reproduce the installation in the future.

ktf:
Can you please rebase and address @dberzano comment?

ktf:
@dberzano can you check if this is fine for you?

dberzano:
Checking in a second.

pkarczma:
Hi @ktf and @dberzano, I applied all the changes you asked for. Please let me know if you require something more.

dberzano:
Hi @pkarczma, @qgp: I cannot make this PR work on my system. It breaks at:

```
Error compiling Cython file:
------------------------------------------------------------
...
include "include/Bin2D.pxi"
include "include/ProfileBin2D.pyx"
include "include/HistoBin2D.pyx"
include "include/Histo2D.pyx"
include "include/Profile2D.pyx"
include "include/Point.pyx"
^
------------------------------------------------------------

core.pyx:41:0: 'include/Point.pyx' not found
```

Any guess?

dberzano:
Plus, from what I see, cython is absolutely not required for compiling YODA. It is used instead to generate C++ files, which will be in turn used by YODA, but those C++ files are already in our YODA source distribution. So, this recipe always worked as this is not needed.

@pkarczma: can you please clarify what are the cases when `cython` is invoked on your system?

By looking at the relevant YODA Makefile (generated by autotools, so not in the sources) I see only three `cython` calls:

```
core.cpp core.h: $(CYTHON_SRC) .made_pyx_templates
        cython ...

util.cpp: util.pxd util.pyx .made_pyx_templates
        cython ...

rootcompat.cpp: util.pxd rootcompat.pxd rootcompat.pyx .made_pyx_templates
        cython ...
```

(Also note that those `cython` calls are automatically commented out in the generated Makefile in case a proper version of `cython` is not found on the system.)

All those files (including the expanded `$(CYTHON_SRC)`) are in the source distribution, including `.made_pyx_templates`, which is touched when C++ files are generated to prevent useless regeneration.

As a matter of fact, changing the recipe to get rid of this placeholder forces `cython` to be invoked at every build, but this is not what we want.

There is only one case that comes to my mind when `cython` is executed. Makefile rules are invoked when the left-hand file does not exist (needs to be generated) *or it's older than the right-hand files*.

Now, all the necessary files are already pre-generated in the YODA distribution, with the correct timestamps. However, timestamps are not preserved when importing to Git (see what [Linus Torvalds says](https://www.spinics.net/lists/git/msg24942.html) about adding timestamp support).

This means that, quite randomly, it might happen that for instance `rootcompat.pxd` is created on your computer with a timestamp which is more recent than `rootcompat.cpp`. It also seems that this does not happen in the majority of cases.

The solution is to save the timestamps after unpacking in a separate file, and tell the recipe to apply the timestamps back when compiling. This requires editing the YODA recipe and the YODA repo as well. We have done the same for GCC-Toolchain for instance.

Do you mind if we close this PR and I open a new one with the proper fix?

dberzano:
Superseded by #811, closing this for now.

dberzano:
Superseded by #811, closing this for now.

PR #777, opened 2017-06-22
Bump ROOT 6 to v6-10-02 (O2 and ROOT6 defaults)

dberzano:


dberzano:
@sawenzel we're testing it and maybe we'll merge it before going home... in order not to trigger a full rebuild of the currently running tests.

PR #778, opened 2017-06-22
Add simple virtual MC AliRoot test with G3/G4

sawenzel:
…o gun test

The idea is to use this recipe in automatic regression test for AliRoot


ktf:
Why not using aliroot-test?

sawenzel:
Because aliroot-test is generic and does not call finaltest.sh which is specific to vmctest/gun. I am trying to get this testing set up quickly. Feel free not to use the new recipe, but I would like to have these tests actually run as soon as possible.

PR #780, opened 2017-06-28
Improve check changes only in O2 check code

dberzano:
Rationale: speed up automatic checks by checking only what effectively changed, and what was affected by such changes.

* By default it checks all files
* If `ALIBUILD_BASE_HASH` is given, checks current working area (including
  untracked files) with respect to that commit: useful for local tests
* If both `ALIBUILD_BASE_HASH` and `ALIBUILD_HEAD_HASH` are given, check only
  what changed between them: useful for centralized PR tests
* If more than 50 files changed, revert to checking all files: computing the
  affected files would be too expensive

### DIFF
--- a/o2checkcode.sh
+++ b/o2checkcode.sh
@@ -19,20 +19,21 @@ cp "${O2_ROOT}"/compile_commands.json .
 O2_SRC=$(python -c 'import json,sys,os; sys.stdout.write( os.path.commonprefix([ x["file"] for x in json.loads(open("compile_commands.json").read()) if not "G__" in x["file"] and x["file"].endswith(".cxx") ]) )')
 [[ -e "$O2_SRC"/CMakeLists.txt && -d "$O2_SRC"/.git ]]
 
-BASECOMMIT=${O2CHECKCODE_BASECOMMIT:=${ALIBUILD_PR_BASE}}
-HEADCOMMIT=${O2CHECKCODE_HEADCOMMIT:=${ALIBUILD_PR_HEAD}}
-
-# We query Git if a base commit was given, which we treat as an indication for a delta check. With
-# more than 50 files changed we run the checks on all files.
-if [[ $BASECOMMIT && $(cd "$O2_SRC" && git diff $HEADCOMMIT $BASECOMMIT --name-only | wc -l) -le 50 ]]; then
-  O2_CHECKCODE_CHANGEDFILES=$(cd "$O2_SRC" && git diff $HEADCOMMIT $BASECOMMIT --name-only               | \
-                              grep -E '\.cxx$|\.h$'                                                      | \
-                              while read FILE; do [[ -e "$O2_SRC/$FILE" ]] && echo "$FILE" || true; done | \
-                              xargs echo | sed -e 's/ /:/g')
-  if [[ ! $O2_CHECKCODE_CHANGEDFILES ]]; then
-    echo "Nothing changed with respect to base commit: not checking anything"
-    exit 0
-  fi
+# We have something to compare our working directory to (ALIBUILD_BASE_HASH). We check only the
+# changed files (including the untracked ones) if the list of relevant files that changed is up to
+# 50 entries long
+if [[ $ALIBUILD_BASE_HASH ]]; then
+  pushd "$O2_SRC"
+    ( git diff --name-only $ALIBUILD_BASE_HASH${ALIBUILD_HEAD_HASH:+...$ALIBUILD_HEAD_HASH} ; git ls-files --others --exclude-standard ) | grep -E '\.cxx$|\.h$' | sort -u > $BUILDDIR/changed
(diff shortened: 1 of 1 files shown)


dberzano:
This change explicitly requires both commits to be specified and computes a diff list containing only the actual changes between the specified head and the PR base (the "three dots", `...`). [Here](https://stackoverflow.com/questions/462974/what-are-the-differences-between-double-dot-and-triple-dot-in-git-com) there is an excellent explanation (with Venn diagrams) to see what it does.

This goes together with alisw/ali-bot#464, where we also pass the correct "head hash" along with the PR "base hash".

@sawenzel do you agree?

dberzano:
Just updated according to what we've discussed. Note that the variable naming has changed to avoid confusion between what is head and base.

o2checkcode.sh
```diff
@@ -19,20 +19,21 @@ cp "${O2_ROOT}"/compile_commands.json .
...
+# changed files (including the untracked ones) if the list of relevant files that changed is up to
+# 50 entries long
+if [[ $ALIBUILD_UPSTREAM_HASH ]]; then
+  pushd "$O2_SRC"
+    ( git diff --name-only $ALIBUILD_UPSTREAM_HASH${ALIBUILD_PR_HASH:+...$ALIBUILD_PR_HASH} ; git ls-files --others --exclude-standard ) | grep -E '\.cxx$|\.h$' | sort -u > $BUILDDIR/changed
+    if [[ $(cat $BUILDDIR/changed | wc -l) -le 50 ]]; then
+      O2_CHECKCODE_CHANGEDFILES=$(while read FILE; do [[ -e "$O2_SRC/$FILE" ]] && echo "$FILE" || true; done < <(cat $BUILDDIR/changed) | \
+                                  xargs echo | sed -e 's/ /:/g')
+      if [[ ! $O2_CHECKCODE_CHANGEDFILES ]]; then
+        echo "Nothing changed with respect to base commit: not checking anything"
```
sawenzel: probably this message needs to be changed, too (since you exchanged HEAD<->BASE).

o2checkcode.sh
```diff
@@ -19,20 +19,21 @@ cp "${O2_ROOT}"/compile_commands.json .
...
-                              while read FILE; do [[ -e "$O2_SRC/$FILE" ]] && echo "$FILE" || true; done | \
-                              xargs echo | sed -e 's/ /:/g')
-  if [[ ! $O2_CHECKCODE_CHANGEDFILES ]]; then
-    echo "Nothing changed with respect to base commit: not checking anything"
-    exit 0
-  fi
+# We have something to compare our working directory to (ALIBUILD_UPSTREAM_HASH). We check only the
+# changed files (including the untracked ones) if the list of relevant files that changed is up to
+# 50 entries long
+if [[ $ALIBUILD_UPSTREAM_HASH ]]; then
```
sawenzel: How can a user now use the recipe? Does she have to define `ALIBUILD_UPSTREAM_HASH`? In the old version it was quite clear that she had to define more meaningful variables `O2CHECKCODE_...`

dberzano: The rationale was to have a single variable to trigger the behaviour, but now that I am writing this comment I realize that:

* for central use cases, the recipe stays independent and should use `ALIBUILD_UPSTREAM_HASH`
* for local use cases, it's more intuitive to export `O2CHECKCODE...`

So... I think you are right. Let me update this.

o2checkcode.sh
dberzano: You are right. Update coming.

dberzano:
@sawenzel can you check whether the variable names `O2CHECKCODE_UPSTREAM_HASH` and `O2CHECKCODE_WORK_HASH` work for you? As said I would like to avoid the names "base" and "head".

(See, more than 5 minutes work...)

o2checkcode.sh
```diff
@@ -19,26 +19,33 @@ cp "${O2_ROOT}"/compile_commands.json .
...
 
-BASECOMMIT=${O2CHECKCODE_BASECOMMIT:=${ALIBUILD_PR_BASE}}
-HEADCOMMIT=${O2CHECKCODE_HEADCOMMIT:=${ALIBUILD_PR_HEAD}}
+# We have something to compare our working directory to. We check only the
+# changed files (including the untracked ones) if the list of relevant files
+# that changed is up to 50 entries long. User can trigger a check by setting the
+# upstream reference hash in the O2CHECKCODE_UPSTREAM_HASH. User can also export
+# O2CHECKCODE_WORK_HASH if she does not want to compare the current head plus
+# the working directory.
+ALIBUILD_UPSTREAM_HASH=${ALIBUILD_UPSTREAM_HASH:=$O2CHECKCODE_UPSTREAM_HASH}
```
sawenzel: I find `UPSTREAM` still a bit confusing. (But I don't want to block this any longer either).

PR #781, opened 2017-06-29
Bump boost to v1.64.0 and CMake to v3.8.2

dberzano:


dberzano:
@vascobarroso @dennisklein I am still testing it (I have found out that CMake had to be updated too otherwise `FindBoost.cmake` does not find the new boost), will merge it when done.

Once this is merged, you'll need to update alidist, and a quite full rebuild will be triggered, unfortunately.

PR #790, opened 2017-07-11
Add Support for G4PY, XercesC and add automated version detection for GEANT4 data files

PMunkes:
This patch adds support for the Python bindings for GEANT4 and it also allows the versions of the GEANT4 data files to be automatically detected. 
Previously the version numbers had to be manually set in the GEANT4 recipe, which effectively restricted the recipe to a single version of GEANT4.

### DIFF
--- a/g4py.sh
+++ b/g4py.sh
@@ -1,0 +1,55 @@
+package: G4PY
+version: "%(tag_basename)s%(defaults_upper)s"
+tag: v4.10.01.p03
+source: https://github.com/alisw/geant4
+requires:
+  - GEANT4
+  - Python-modules
+  - ROOT
+  - boost
+  - XercesC
+  - opengl
+build_requires:
+  - CMake
+  - "Xcode:(osx.*)"
+env:
+  G4PYINSTALL: "$G4PY_ROOT"
+---
+#!/bin/bash -e
+rsync -a $SOURCEDIR/environments/g4py/* $INSTALLROOT
+
+cmake $INSTALLROOT                               \
+      -DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE}     \
+      -DBoost_NO_SYSTEM_PATHS=TRUE               \
+      -DCMAKE_INSTALL_PREFIX="$INSTALLROOT"      \
+      -DBOOST_ROOT=${BOOST_ROOT}                 \
+      -DXERCESC_ROOT_DIR=${XERCESC_ROOT}         \
+      -DCMAKE_VERBOSE_MAKEFILE=TRUE              \
+      -DBoost_NO_BOOST_CMAKE=TRUE
+
+make ${JOBS:+-j $JOBS}
+make install
+ctest
+
+ln -s lib64 $INSTALLROOT/lib
+
+
+# Modulefile
+MODULEDIR="$INSTALLROOT/etc/modulefiles"
+MODULEFILE="$MODULEDIR/$PKGNAME"
+mkdir -p "$MODULEDIR"
+cat > "$MODULEFILE" <<EoF
+#%Module1.0
+proc ModulesHelp { } {
+  global version
+  puts stderr "ALICE Modulefile for $PKGNAME $PKGVERSION-@@PKGREVISION@$PKGHASH@@"
+}
+set version $PKGVERSION-@@PKGREVISION@$PKGHASH@@
+module-whatis "ALICE Modulefile for $PKGNAME $PKGVERSION-@@PKGREVISION@$PKGHASH@@"
+# Dependencies
+module load BASE/1.0 ROOT/$ROOT_VERSION-$ROOT_REVISION ${GEANT4_VERSION:+GEANT4/$GEANT4_VERSION-$GEANT4_REVISION} XercesC/$XERCESC_VERSION-$XERCESC_REVISION boost/$BOOST_VERSION-$BOOST_REVISION
+# Our environment
+setenv G4PY_ROOT \$::env(BASEDIR)/$PKGNAME/\$version
+setenv PYTHONPATH \$::env(G4PY_ROOT)/lib:\$::env(G4PY_ROOT)/lib/g4py:\$::env(G4PY_ROOT)/lib/Geant4:\$::env(G4PY_ROOT)/lib/examples:\$::env(G4PY_ROOT)/lib/tests
+$([[ ${ARCHITECTURE:0:3} == osx ]] && echo "prepend-path DYLD_LIBRARY_PATH \$::env(G4PY_ROOT)/lib")
+EoF
--- a/geant4.sh
+++ b/geant4.sh
@@ -1,16 +1,16 @@
 package: GEANT4
 version: "%(tag_basename)s%(defaults_upper)s"
-source: https://github.com/alisw/geant4
 tag: v4.10.01.p03
+source: https://github.com/alisw/geant4
 requires:
   - "GCC-Toolchain:(?!osx)"
 build_requires:
   - CMake
   - "Xcode:(osx.*)"
 env:
-  G4INSTALL: "$GEANT4_ROOT"
-  G4INSTALL_DATA: "$GEANT4_ROOT/share/Geant4-10.1.3"
-  G4SYSTEM: "$(uname)-g++"
+  G4INSTALL:                "$GEANT4_ROOT"
+  G4INSTALL_DATA:           "$GEANT4_ROOT/share/Geant4-10.1.3"
+  G4SYSTEM:                 "$(uname)-g++"
   G4LEVELGAMMADATA:         "$GEANT4_ROOT/share/Geant4-10.1.3/data/PhotonEvaporation3.1"
   G4RADIOACTIVEDATA:        "$GEANT4_ROOT/share/Geant4-10.1.3/data/RadioactiveDecay4.2"
   G4LEDATA:                 "$GEANT4_ROOT/share/Geant4-10.1.3/data/G4EMLOW6.41"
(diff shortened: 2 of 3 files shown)


xercesc.sh
```diff
@@ -0,0 +1,33 @@
+package: XercesC
+version: v3.1.4
+source: https://github.com/PMunkes/XercesC
+tag: v3.1.4
+env:
+  XERCESC_INST_DIR: "$XERCESCINST"
```
ktf: Why do you need this?

g4py.sh
```diff
@@ -0,0 +1,51 @@
...
+  - opengl
+build_requires:
+  - CMake
+  - "Xcode:(osx.*)"
+env:
+  G4PYINSTALL: "$G4PY_ROOT"
+---
+#!/bin/bash -e
+cp -a $SOURCEDIR/environments/g4py/* $INSTALLROOT
+cmake $INSTALLROOT                               \
```
ktf: Can't it be built via `cmake $SOURCEDIR`?

geant4.sh
```diff
@@ -40,11 +40,30 @@ cmake $SOURCEDIR                                    \
   -DCMAKE_STATIC_LIBRARY_C_FLAGS="-fPIC"            \
   -DGEANT4_USE_G3TOG4=ON                            \
   -DGEANT4_INSTALL_DATA=ON                          \
-  -DGEANT4_USE_SYSTEM_EXPAT=OFF
+  -DGEANT4_USE_SYSTEM_EXPAT=OFF                     \
+  ${XERCESC_ROOT:+-DGEANT4_USE_OPENGL_X11=ON -DGEANT4_USE_GDML=ON -DXERCESC_ROOT_DIR=$XERCESC_ROOT}
```
ktf: Why do you need the OpenGL bindings? what happens if they are not found? What happens if this is built on mac? Why do you need GDML?

g4py.sh
```diff
@@ -0,0 +1,51 @@
+package: G4PY
+version: "%(tag_basename)s%(defaults_upper)s"
+tag: v10.3.1
+source: https://github.com/PMunkes/geant4
```
ktf: Why do you need to use your own repository?

ktf:
Thank you for your contribution. I've a few questions to be answered (and maybe some changes, after you answer them), however given Geant4 is actually meant to be part of production AliRoot, bear with me. ;-)

xercesc.sh
PMunkes: This Package is a dependency of G4PY, the python bindings of GEANT4.

g4py.sh
PMunkes: This is a patched version of G4PY that was left in here by accident. It should work with the normal GEANT4 repository. I'll change it.

g4py.sh
PMunkes: Unfortunately with `cmake $SOURCEDIR`G4PY installs itself into `$SOURCEDIR.` This seems to be intentional behaviour, and as such I am installing the minimal amount of source files necessary to build G4PY in `$INSTALLROOT` to install into the proper directory. Otherwise the package depends on the source directory to always be available.

geant4.sh
PMunkes: These are necessary options to build G4PY. These options only become active if XercesC has been specified in the `requires:` section, otherwise they are not set. If somebody is using XercesC, then they are most likely trying to build G4PY and know to also set the OpenGL bindings as a requirement. 
Unfortunately I cannot test this recipe on a mac, due to a lack of hardware, but Fons has been using G4PY on his mac and this is directly taken from the makefile he is currently using.

g4py.sh
ktf: Ok.

geant4.sh
ktf: Ok, I get your point. Yes, this sounds reasonable and given it's controlled by an override it should not be an issue.

xercesc.sh
ktf: No, I meant, XERCESC_INST_DIR

PMunkes: I am currently testing a build without this variable, as I think this is no longer needed. It was needed in a previous version of the patch, but has since been superseded by the use of `$XERCESC_ROOT` at all relevant positions.

ktf: Ok, great. I think this is the only controversial point, then. I would still like to give it a try myself, but I am not sure I can do it before the 25th of July. Are you blocked by this? If not, would you mind waiting?

geant4.sh
```diff
@@ -40,11 +40,30 @@ cmake $SOURCEDIR                                    \
...
 
 make ${JOBS+-j $JOBS}
 make install
 
+#Get data file versions:
+. $INSTALLROOT/bin/geant4.sh
```
dberzano: Can you use `source` please. We use `bash`, no need to be POSIX-compliant and this is consistent with most of the other recipes.

xercesc.sh
ktf: Ah, I actually forgot. Do you really need your own repository for xercesC?

geant4.sh
```diff
@@ -40,11 +40,30 @@ cmake $SOURCEDIR                                    \
...
+
+G4LEVELGAMMADATA_NAME=`basename "$G4LEVELGAMMADATA"`
```
dberzano: Also here, can you use `$(...)` instead of the backticks.

g4py.sh
```diff
@@ -0,0 +1,51 @@
...
+  - XercesC
+  - opengl
+build_requires:
+  - CMake
+  - "Xcode:(osx.*)"
+env:
+  G4PYINSTALL: "$G4PY_ROOT"
+---
+#!/bin/bash -e
+cp -a $SOURCEDIR/environments/g4py/* $INSTALLROOT
```
dberzano: Hi, you are `cmake`ing and building in the installation directory, this is wrong. Can you do, instead:

```
rsync -a $SOURCEDIR/environments/g4py/ .  # copies everything in the current directory, which is the build one
cmake . -D... -DCMAKE_INSTALL_PREFIX=$INSTALLROOT
```

As said, the "current" directory is the scratch, build directory. If you need to refer to it by variable, its name is `$BUILDDIR`.

The use of `rsync` instead of `cp -a` is consistent with the other recipes.

Note that, contrary to the other comment from @ktf, I am telling you to run `cmake $BUILDDIR`, not `cmake $SOURCEDIR`.

xercesc.sh
PMunkes: This is not really blocking, as most of the changes are already in shipdist. I am still working on a number of other recipes needed for FairShip, and this would only be blocking that.

PMunkes: As there is no upstream git repository we currently need this, unless you were willing to fork it for alisw. This is also the case for a few of the other packages for which I am working on recipes.
Eventually we will fork these repositories for ShipSoft and I'll update the links then.

geant4.sh
PMunkes: Okay, I'll perform the change.

g4py.sh
PMunkes: Like I already mentioned, this would just install G4PY to `$BUILDDIR`, as it installs itself into the source directory. I spend roughly 3 days on this problem and I have not found a better solution, that works with an unmodified source.

I will test your suggestion and notify you of the results.

dberzano: We are frequently in contact with GEANT4 developers - quite evidently, the behaviour is wrong, and if you have a modified GEANT4 source that works, the best option is to open a pull request. They are very, very responsive (especially if you tag us!)

Anyway we'll see the results from your test... thanks!

dberzano: Haha alright, great!

PMunkes: I have tested your suggestion and it did not work. It is still installing into the directory of the source files, now located in `$BUILDDIR`. 

I have checked the source code and am now fully convinced, that this is working as intended:
`# installation prefixes for libraries
set(CMAKE_INSTALL_PREFIX ${PROJECT_SOURCE_DIR})
`
In an excerpt from the build log the cmake command is as follows:
`cmake . -DCMAKE_BUILD_TYPE=RELWITHDEBINFO -DBoost_NO_SYSTEM_PATHS=TRUE -DCMAKE_INSTALL_PREFIX=/opt/SHiP/alibuild/sw/INSTALLROOT/f3c690189ff37e165859e5aec89962d6a2c1bf8b/ubuntu1604_x86-64/G4PY/v10.3.1-ship2-2 -DBOOST_ROOT=/opt/SHiP/alibuild/sw/ubuntu1604_x86-64/boost/v1.64.0-alice1-1 -DXERCESC_ROOT_DIR=/opt/SHiP/alibuild/sw/ubuntu1604_x86-64/XercesC/v3.1.4-2 -DBoost_NO_BOOST_CMAKE=TRUE`
While a few lines later the following is logged:
`CMAKE_INSTALL_PREFIX: /opt/SHiP/alibuild/sw/BUILD/f3c690189ff37e165859e5aec89962d6a2c1bf8b/G4PY`
Furthermore nearly all python files are installed in either source/ or site-modules/ in `$BUILDDIR.`

As this is clearly intended behaviour, I would stick with the original idea of copying the source files to '$INSTALLDIR` and building inside `$BUILDDIR`. It would probably be a good idea to use rsync for this, if only for consistency

ktf:
Sorry for neglecting this. Looks fine to this. @dberzano are you happy about the changes?

PMunkes:
Actually a few changes were still needed, I will push the commits later today. Also I am not sure about merging this changeset, while the repository-URLs are still on my own account. 
If you merge this pull request at this point, I will need to create another one to correct the URLs. 
However, if that is fine with you, then I would not object.

ktf:
Can you please address the (newly introduced) travis report? It should be enough to move `tag` to be right after `version`.

PMunkes:
Barring objections, this is the final version of the patches. However, as the ShipSotf/XercesC does not yet exist, I would still not merge it, as is. Once the repository exists, I will provide notice here.

ktf:
I think these have been moved to the `shipdist` of whatever that is called. Closing this.

PR #794, opened 2017-07-12
PDA recipe

Barthelemy:
Also its dependencies (which have to be installed with yum) and its use in flpproto. 
Note that it should not be built outside of CC7.

### DIFF
--- a/flpproto.sh
+++ b/flpproto.sh
@@ -4,12 +4,13 @@ requires:
   - Configuration
   - O2
   - "GCC-Toolchain:(?!osx)"
-build_requires:
+  - "PDA:slc7.*"
+build_requires: 
   - CMake
   - MySQL
 source: https://github.com/AliceO2Group/FlpPrototype
 version: "%(tag_basename)s"
-tag: v0.4.4
+tag: v0.4.6
 incremental_recipe: |
   make ${JOBS:+-j$JOBS} install
   mkdir -p $INSTALLROOT/etc/modulefiles && rsync -a --delete etc/modulefiles/ $INSTALLROOT/etc/modulefiles
@@ -42,8 +43,9 @@ cmake $SOURCEDIR                                              \
     -DFAIRROOTPATH=$FAIRROOT_ROOT \
     -DFairRoot_DIR=$FAIRROOT_ROOT                               \
     -DROOTSYS=$ROOTSYS \
-    ${Configuration_ROOT:+-DConfiguration_ROOT=$Configuration_ROOT} \
-    ${Monitoring_ROOT:+-DMonitoring_ROOT=$Monitoring_ROOT}
+    ${CONFIGURATION_VERSION:+-DConfiguration_ROOT=$CONFIGURATION_ROOT} \
+    ${MONITORING_VERSION:+-DMonitoring_ROOT=$MONITORING_ROOT} \
+    ${PDA_VERSION:+-DPDA_ROOT=$PDA_ROOT} 
 
 make ${JOBS+-j $JOBS} install
 
@@ -63,7 +65,8 @@ module load BASE/1.0                                                            
             ${GCC_TOOLCHAIN_ROOT:+GCC-Toolchain/$GCC_TOOLCHAIN_VERSION-$GCC_TOOLCHAIN_REVISION} \\
             O2/$O2_VERSION-$O2_REVISION                                                         \\
             Monitoring/$MONITORING_VERSION-$MONITORING_REVISION                                 \\
-            Configuration/$CONFIGURATION_VERSION-$CONFIGURATION_REVISION                        
+            Configuration/$CONFIGURATION_VERSION-$CONFIGURATION_REVISION                        \\
+            PDA/$PDA_VERSION-$PDA_REVISION  
 
 # Our environment
 setenv FLPPROTO_ROOT \$::env(BASEDIR)/$PKGNAME/\$version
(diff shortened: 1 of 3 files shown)


Barthelemy:
@dberzano @ktf A review might be worth it.

kernel-devel.sh
```diff
@@ -0,0 +1,8 @@
+package: kernel-devel
+version: "1.0"
+
+system_requirement_missing: "Please install kernel-devel pciutils-devel kmod-devel"
+system_requirement: ".*"
+system_requirement_check: "ls /usr/src/kernels/*/include/linux/kernel.h > /dev/null 2>&1 && ls /usr/include/libkmod.h  > /dev/null 2>&1 && ls /usr/include/pci/pci.h > /dev/null 2>&1"
```
ktf: Why not compiling something including it, actually?

pda.sh
```diff
@@ -0,0 +1,42 @@
+package: PDA
+version: "%(tag_basename)s"
+tag: 11.0.7
+source: https://github.com/PascalBoeschoten/pda
```
ktf: Why not moving this to AliceO2 or alisw? One should use the devel package support of alibuild for development.

pda.sh
```diff
@@ -0,0 +1,42 @@
...
+requires:
+  - "GCC-Toolchain:(?!osx)"
+build_requires:
+  - kernel-devel
+  - autotools
+
+--- 
+#!/bin/sh
+
+case $ARCHITECTURE in
```
ktf: Can't you simply put `PDA:slc7.*` where you actually use this? You would then be guaranteed that you are on slc7 and don't need the case.

Barthelemy: oh, actually I do it (cf flpproto.sh). I will remove it then.

kernel-devel.sh
Barthelemy: we could. What is the benefit though ?

pda.sh
Barthelemy: Correct. 

How should we proceed ? This is a modified fork of https://github.com/cbm-fles/pda which is a fork of https://github.com/kdiff3/pda.

Barthelemy: @PascalBoeschoten

ktf: We can either for one of the two above (whichever you prefer) or have our own standalone `AliceO2Group/pda` where we sync the changes by hand. Whatever is more convenient to you. Once you decide let me know and I can create the repo and give you push rights.

kernel-devel.sh
ktf: You know that your compiler can compile the headers. Not that important actually, I guess.

pda.sh
Barthelemy: @PascalBoeschoten : here are the instructions to transfer your pda repo to the AliceO2Group organization. Could you try it ? I think that you have the right to create a repo there so it should work.

kernel-devel.sh
Barthelemy: ok, so let's keep it as is for now if ok for you.

dberzano: I agree with @ktf, but we can keep it like this for now given the scope. I assume the paths are correct on CC7, right?

flpproto.sh
```diff
@@ -43,7 +44,8 @@ cmake $SOURCEDIR                                              \
     -DFairRoot_DIR=$FAIRROOT_ROOT                               \
     -DROOTSYS=$ROOTSYS \
     ${Configuration_ROOT:+-DConfiguration_ROOT=$Configuration_ROOT} \
-    ${Monitoring_ROOT:+-DMonitoring_ROOT=$Monitoring_ROOT}
+    ${Monitoring_ROOT:+-DMonitoring_ROOT=$Monitoring_ROOT} \
```
dberzano: Hmm how can this work or have ever worked. We expose uppercase variables, so it has to be:

```
${Configuration_ROOT:+-DConfiguration_ROOT=$CONFIGURATION_ROOT}
${Monitoring_ROOT:+-DMonitoring_ROOT=$MONITORING_ROOT}
```

kernel-devel.sh
Barthelemy: @dario Yes.

flpproto.sh
Barthelemy: ok, well, it worked ;) I am going to change it. thank you for spotting it

Barthelemy: Shouldn't it read : 
```
${CONFIGURATION_ROOT:+-DConfiguration_ROOT=$CONFIGURATION_ROOT}
${MONITORING_ROOT:+-DMonitoring_ROOT=$MONITORING_ROOT}
```

flpproto.sh
```diff
@@ -4,12 +4,13 @@ requires:
   - Configuration
   - O2
   - "GCC-Toolchain:(?!osx)"
-build_requires:
+  - "PDA:(?!osx)"
```
Barthelemy: How would I restrict to CC7 only ? this is not restrictive enough I guess as it would still build PDA on Ubuntu.

flpproto.sh
dberzano: You are absolutely right, yes. Of course the way you call your CMake variable is up to you, mine was a typo.

flpproto.sh
dberzano:
```
- "-PDA:slc7*"
```

should work.

ktf: Isn't it `PDA:slc7.*`?

dberzano: Yes. I keep making typos today, sorry!

Barthelemy:
When Pascal will have transfered the repo I will update the recipe of Pda and I think that it will be good to be merged. 
Thank you for your help

PascalBoeschoten:
Repo has been transferred:
https://github.com/AliceO2Group/pda

Barthelemy:
Ok, good to merge in my opinion.

Barthelemy:
@dberzano all good now,l you can proceed

PR #796, opened 2017-07-17
Add Support for multiple generators

PMunkes:
This pull request adds support for the generators GENIE, Tauola++ and PHOTOS++.
It also adds support for log4cpp, a dependency of GENIE.
It adds some PDF files to lhapdf5, that are needed by FairShip.
Finally it adds Support for MillepedeII.

### DIFF
--- a/lhapdf5.sh
+++ b/lhapdf5.sh
@@ -1,6 +1,6 @@
 package: lhapdf5
-tag: alice/v5.9.1
 version: "%(tag_basename)s%(defaults_upper)s"
+tag: alice/v5.9.1
 source: https://github.com/alisw/LHAPDF
 env:
   LHAPATH: "$LHAPDF5_ROOT/share/lhapdf"
@@ -16,7 +16,7 @@ rsync -a --exclude '**/.git' $SOURCEDIR/ ./
 make ${JOBS+-j $JOBS} all
 make install
 
-PDFSETS="cteq6l cteq6ll CT10 CT10nlo MSTW2008nnlo EPS09LOR_208 EPS09NLOR_208"
+PDFSETS="cteq6l cteq6ll CT10 CT10nlo MSTW2008nnlo EPS09LOR_208 EPS09NLOR_208 cteq66a cteq66a0 cteq4m"
 pushd $INSTALLROOT/share/lhapdf
   $INSTALLROOT/bin/lhapdf-getdata $PDFSETS
   # Check if PDF sets were really installed
--- a/log4cpp.sh
+++ b/log4cpp.sh
@@ -1,0 +1,38 @@
+package: log4cpp
+version: "%(tag_basename)s%(defaults_upper)s"
+tag: REL_1_1_1_Nov_26_2013
+source: https://github.com/PMunkes/log4cpp
+requires:
+  - GCC-Toolchain:(?!osx)
+build_requires:
+  - autotools
+---
+#!/bin/bash -ex
+rsync -a $SOURCEDIR/ .
+./autogen.sh
+./configure          --prefix=$INSTALLROOT  \
+		     --enable-shared        \
+		     --enable-static        \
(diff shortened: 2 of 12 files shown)


ktf:
Sorry for the late reply, looks reasonable to me. Can you please squash commits a bit? E.g. all the PHOTOS ones into one?

lhapdf5.sh
```diff
@@ -16,7 +16,7 @@ rsync -a --exclude '**/.git' $SOURCEDIR/ ./
 make ${JOBS+-j $JOBS} all
 make install
 
-PDFSETS="cteq6l cteq6ll CT10 CT10nlo MSTW2008nnlo EPS09LOR_208 EPS09NLOR_208"
+PDFSETS="cteq6l cteq6ll CT10 CT10nlo MSTW2008nnlo EPS09LOR_208 EPS09NLOR_208 cteq66a cteq66a0 cteq4m"
```
ktf: How big are these? @qgp, any objection to include them?

pythia6.sh
```diff
@@ -16,6 +16,11 @@ make ${JOBS+-j$JOBS}
 make install
 tar -c lib include | tar -x -C $INSTALLROOT 
 
+cd $INSTALLROOT/lib/
```
ktf: Why not simply:

```
ln -s libpythia6.so $INSTALLROOT/lib/libPythia6.so
```

millepedeii.sh
```diff
@@ -0,0 +1,30 @@
...
+version: V04-03-01
+source: https://github.com/PMunkes/MillepedeII
+tag: V04-03-01
+---
+#!/bin/bash
+rsync -a $SOURCEDIR/* .
+
+make  pede 
+./pede -t
+rsync pede $INSTALLROOT/bin/
```
ktf: Why not `cp pede $INSTALLROOT/bin/`?

log4cpp.sh
```diff
@@ -0,0 +1,35 @@
...
+version: "%(tag_basename)s%(defaults_upper)s"
+tag: REL_1_1_1_Nov_26_2013
+source: https://github.com/PMunkes/log4cpp
+build_requires:
+  - autotools
+---
+#!/bin/bash -ex
+rsync -a $SOURCEDIR/* .
+./autogen.sh
+./configure          --prefix=$INSTALLROOT  \
```
ktf: You should align `\`. If you use `vim` I suggest you install the [tabular](https://github.com/godlygeek/tabular) plugin or something similar.

photospp.sh
```diff
@@ -0,0 +1,41 @@
...
+  - pythia
+  - Tauolapp
+---
+#!/bin/sh
+
+export  HEPMCLOCATION="$HEPMC_ROOT"
+
+rsync -a $SOURCEDIR/* .
+
+./configure --enable-debug --with-hepmc=$HEPMC_ROOT --with-pythia8=$PYTHIA_ROOT --with-tauola=$TAUOLAPP_ROOT --prefix=$INSTALLROOT CFLAGS="$CFLAGS" CXXFLAGS="$CFLAGS"
```
ktf: I suppose you did try running `$SOURCEDIR/configure`, right? Also, please use `rsync -a $SOURCEDIR/  .` (no `*`).

lhapdf5.sh
PMunkes: These PDF-files are approximately 500 kB in size.

PMunkes:
I have rebased all of the commits and adressed most of the comments. At the moment there is only one blocker from my side, most of the repositories for these recipes are still on my Github account. Once they have been transferred, I would need to change all the URLs from ../PMunkes/.. to ../ShipSoft/... . Until then I am submitting these changes for review.

ktf:
I think these have been moved to `shipdist`.

PR #798, opened 2017-07-18
Improve DIM dependencies and installation

dberzano:


dberzano:
@Barthelemy with this installation method, this is what we get in the installation directory:

```
.
./test_server
./etc
./etc/modulefiles
./etc/modulefiles/dim
./etc/modulefiles/dim.unrelocated
./etc/profile.d
./etc/profile.d/init.sh.unrelocated
./etc/profile.d/init.sh
./did
./dim
./dim/dim_core.hxx
./dim/tokenstring.hxx
./dim/dic.h
./dim/dim.h
./dim/dim.hxx
./dim/dim_jni.h
./dim/dic.hxx
./dim/dicLorenzo.hxx
./dim/dim_common.h
./dim/dis.h
./dim/sllist.hxx
./dim/dllist.hxx
./dim/dis.hxx
./lib
./lib/libdim.so
./lib/libdim.a
./testServer
./testClient
./dim_send_command
./checkDns
./.build-hash
./dim_get_service
./webDid
./DimBridge
./checkDimServers
./.original-unrelocated
./test_client
./relocate-me.sh
./dns
```

which means that:

* headers go in `<prefix>/dim`
* executables in `<prefix>`
* libraries in `<prefix>/lib`

The directory structure is very questionable... but it's there for compatibility :-(

What do you think?

dim.sh
```diff
@@ -15,21 +15,22 @@ curl -L -O $URL
...
-# compile
-export OS=Linux
-source setup.sh
-gmake realclean
-gmake X64=yes
+# Build (needs special environment)
+( export OS=Linux
+  source setup.sh
+  gmake realclean
+  gmake X64=yes )
```
Barthelemy: what are those parentheses ?

dim.sh
```diff
@@ -15,21 +15,22 @@ curl -L -O $URL
...
 
-# copy to destination
-cp -r dim $INSTALLROOT
-cp -r linux/* $INSTALLROOT  
+# Installation
+rsync -av dim/ $INSTALLROOT/dim/                                                # headers
+rsync -av --exclude "*.o" --exclude "*.so" --exclude "*.a" linux/ $INSTALLROOT/ # executables
+mkdir $INSTALLROOT/lib && cp -v linux/*.{a,so} $INSTALLROOT/lib/                # libraries
```
Barthelemy: if we have any ways a messy installation folder, I would not move the libs in lib and keep them in the top dir along with the binaries.

dim.sh
dberzano: Those parentheses are like the Fight Club. Whatever happens there (environment-wise), stays there :-)

Jokes aside: it's an inner shell, so whatever `export`, etc. will not be available outside of them.

Barthelemy: ok thank you !

dim.sh
dberzano: No problem for me, but you told me:

```
cp -r linux/*.a linux/*.so $INSTALLROOT/lib
```

What would your preferred way be then? Toplevel?

Barthelemy: toplevel, sorry I mislead you.

dberzano: OK update coming.

Barthelemy:
can be merged

PR #799, opened 2017-07-26
Prefer system VC if available

ktf:


davidrohr:
Hi,

that doesn't work for me. Vc is not even considered as a package that can be taken from the system.
It works if I modify it as follows ("prefer_system" taken from the boost recipe, I also added a Vc version check):

prefer_system: (?!slc5)
prefer_system_check: |
  printf "#include <Vc/version.h>\n#if Vc_VERSION_CHECK(1,3,2) > Vc_VERSION_NUMBER\n#error Incorrect Vc version\n#endif" | c++ -xc++ - -c -M 2>&1

Not sure if this is the right thing to do, and whether the version treatment is as you would want it.

ktf:
Yes, you are right, I forgot. We should imply `prefer_system: .*` if `prefer_system_check` is specified. I've corrected as you suggested, which looks sane to me. Can you verify the PR matches what you have?

davidrohr:
works for me.
I saw that you are missing the newline before the ---, don't know whether that makes a difference.
and with prefer_system: .* it works as well, probably that is the correct solution then.
thanks
David

ktf:
Ok, great. Assigning this to @dberzano to include it in the release.

PR #800, opened 2017-07-28
Avoid conflict with local LHAPDF installation

ktf:
Turns out LHAPDF is particularly stubborn in trying to pick up system
installation, if available. This results in errors when we are trying to
install our own PDF sets. This commit forces installation of PDF pdf
sets to use our own version of lhapdf, lhapdf-config (which was not the
case before) and forces usage of our own `share` path.

### DIFF
--- a/lhapdf.sh
+++ b/lhapdf.sh
@@ -45,7 +45,10 @@ make ${JOBS+-j $JOBS} all
 make install
 
 PDFSETS="cteq6l1 MMHT2014lo68cl MMHT2014nlo68cl"
-$INSTALLROOT/bin/lhapdf install $PDFSETS
+PATH=$INSTALLROOT/bin lhapdf               \
+      --pdfdir=$INSTALLROOT/share/LHAPDF   \
+      --listdir=$INSTALLROOT/share/LHAPDF  \
+      install $PDFSETS
 # Check if PDF sets were really installed
 for P in $PDFSETS; do
   ls $INSTALLROOT/share/LHAPDF/$P


ktf:
ping?

dberzano:
Makes sense to me. Was it tested? Does it solve AliceO2Group/AliceO2#423? Nobody has confirmed that.

ktf:
@urasantonio can you confirm what we did last week works fine for you?

urasantonio:
Hi, I confirm that after the modification done by Giulio, lhapdf was built correctly. However, in order to fully test the commit, this morning I restarted the full AliceO2 installation from scratch, using the alidist/lhapdf.sh in attachment (I applied by hand the patch #800 - I had to change the extension to .txt since .sh attachments are not allowed
[lhapdf.txt](https://github.com/alisw/alidist/files/1208019/lhapdf.txt)
) and I find a new problem, see the second attachment 
[log.txt](https://github.com/alisw/alidist/files/1208021/log.txt)
. This time it seems to be a python-related issue, no idea if any link exists with the modification of lhapdf.sh.

urasantonio:
Hi, shall I start a new thread for the python-related issue reported in my last message? Is it better to use the mailing list?

dberzano:
Hello @urasantonio, we have merged both #800 (which admittedly has a problem that might lead to your error message too), and #804 (that fixes the above problem too). Please use the master then.

urasantonio:
Thanks @dberzano, since I'm supposed to develop some code for the MFT, I should use the dev branch, isn't it?

dberzano:
I am not sure I have understood your question. This fix to alidist has been merged, so I suggested you use the _master branch of alidist_. You are probably talking about O2 where we don't really have a master branch, and the target one is, as you've said, dev.

urasantonio:
Hi, yes, I misunderstood your message... I updated the alidist master and relaunched the O2 building: lhapdf has just been built without any problem. Thanks again for the support!

PR #804, opened 2017-08-08
Fix LHAPDF missing system PATH

dberzano:


dberzano:
Fixes problem introduced with #800

Issue #395, opened 2016-05-13
Check certificate expiration date


mpuccio:
Ciao @dberzano,

`alice-env.sh` was checking the certificate expiration date giving useful alert messages.
Would it be possible to implement something similar, maybe directly in alien-token-init?

Thanks,
Max

aaniin:
Hi @mpuccio !

You requested to check the validity of the grid certificate some time ago. I translated Dario's alice-env.sh check into tcl. Can you try out #808 please? 

Cheers,
Hans

dberzano:
So, after what was said in #808: why not having it in `alien-token-init`, which is quite simply a shell script?

aaniin:
Hi @dberzano , @mpuccio ,

I wonder whether one needs the check at all. I get enough emails from CERN about my certificate expiration. There's one at -30days, -7days, and -1day. 

Cheers,
Hans

dberzano:
The reason why I put the check in the shell back in the old days was that updating the certificate also implies a step for having it converted to PEM and under `~/.globus`, possibly on multiple machines. Anyway I have added the check to `alien-token-init`.

PR #808, opened 2017-08-09
Add check on certificate expiration in AliEn-Runtime

aaniin:
This incorporates the check whether one's grid certificate is still valid as suggested by Max in issue #395. The check is a direct tcl translation of the former bash check in alice-env.sh by Dario.


dberzano:
Hi @aaniin, I understand the convenience of checking the certificate's validity but doing it in the Modulefile is a bad idea and it has an impact on Grid jobs too. Can we close this and think of a better way?

Sorry...

aaniin:
Sure :) I see everything failed :(

PR #811, opened 2017-08-14
Fix YODA and Rivet issues with Python

dberzano:
* Unset `PYTHON_VERSION` as the same variable is used internally by autotools
  and causes weird behaviours
* Restore autogenerated files timestamps for YODA to allow building using
  generated C templates instead of requiring Cython to regenerate them

### DIFF
--- a/rivet.sh
+++ b/rivet.sh
@@ -44,6 +44,8 @@ fi
 
 [[ "$CXXFLAGS" != *'-std=c++11'* ]] || CXX11=1
 
+(
+unset PYTHON_VERSION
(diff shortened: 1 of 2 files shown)


dberzano:
@pkarczma @qgp can you give it a try? This should fix YODA's Cython problem (without requiring Cython!) and supersedes #770.

pkarczma:
@dberzano Yes I will check if it is working.

pkarczma:
@dberzano I confirm that I successfully compiled YODA and Rivet on my laptop (CERN CentOS 7).

PR #817, opened 2017-08-16
o2 recipe: Remove AliRoot dependency and set CMake build type

sawenzel:


sawenzel:
This solves a compilation problem on Mac and removes an obsolete requirement on AliRoot (which was anyway on contrast to the setting in the defaults).

o2.sh
```diff
@@ -75,6 +74,7 @@ cmake $SOURCEDIR -DCMAKE_INSTALL_PREFIX=$INSTALLROOT
       -DPROTOBUF_LIBRARY=$PROTOBUF_ROOT/lib/libprotobuf.$SONAME   \
       ${GSL_ROOT:+-DGSL_DIR=$GSL_ROOT}                            \
       ${PYTHIA_ROOT:+-DPYTHIA8_INCLUDE_DIR=$PYTHIA_ROOT/include}  \
+      -DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE}                      \
```
matthiasrichter: probably better to use `${CMAKE_BUILD_TYPE:+-DCMAKE_BUILD_TYPE=$CMAKE_BUILD_TYPE}`

sawenzel:
@matthiasrichter : Thanks for the suggestion. Implemented.

dberzano:
Hi @sawenzel, I have pushed an additional commit making the linter happy. Your PR looks OK to me, and I agree with it. Especially now that we are testing the new aliBuild, it should be possible for those who need AliRoot in O2 for development to add it in a default without rebuilding all the other packages too.

I will be adding just one more conditional line in the `o2.sh` recipe and merge.

Barthelemy:
@dberzano @sawenzel I get a compilation error due to line 22 : https://alijenkins.cern.ch/job/build-any-ib/6491/console

dberzano:
@Barthelemy this is fixed now, it was my bad :-(

PR #819, opened 2017-08-17
Propagating CMAKE_BUILD_TYPE to cmake in FairRoot

matthiasrichter:
Concerns https://github.com/alisw/alibuild/issues/400
Instead of using hardcoded build types, the type should be taken
from the global settings or default overrides

### DIFF
--- a/fairroot.sh
+++ b/fairroot.sh
@@ -1,7 +1,7 @@
 package: FairRoot
 version: "%(tag_basename)s"
-source: https://github.com/alisw/FairRoot
 tag: "alice-dev-20170711"
+source: https://github.com/alisw/FairRoot
 requires:
   - generators
   - simulation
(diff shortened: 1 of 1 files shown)


matthiasrichter:
@dberzano @ktf what is the reason for the failed check, I don't see anything obvious in the log

dberzano:
```
tag should always be right after version, if the latter is present
```

There is a linter checking for "proper" formatting of the recipes on Travis. You should simply try with your local alidist to run your recipe with the linter script inside the `scripts/` subfolder to see if any warning is spotted.

dberzano:
The point being: you did not do anything wrong, the recipe was already like this, but you have triggered the check as you have changed it :-)

dberzano:
@matthiasrichter thanks for making the PR editable by the maintainers :-) I have made the change myself haha. Merging.

matthiasrichter:
good, thanks

PR #820, opened 2017-08-18
Fix #818: remove -g from global compiler flags in o2 defaults

### LINKED ISSUE #818: Make sure LLVM Interpreter is not build with debug symbols in ROOT compilation
This is a follow up to https://github.com/alisw/alibuild/issues/434

ROOT 6 compilation requires >8 GB disk space when build with debug symbols. This is due to the compilation of LLVM interpreter with debug symbols. Patches are proposed to change this behavior.

This issue is intended to keep track of this and switch to a suitable Root version whenever possible.

matthiasrichter:
The `-g` can be removed from the global compiler flags because cmake will add it automatically for the build type RELWITHDEBINFO. Having it globally set leads to some problems as described in #818 and https://github.com/alisw/alibuild/issues/434

Fixes #818
Partially addresses also https://github.com/alisw/alibuild/issues/400

### DIFF
--- a/defaults-o2-daq.sh
+++ b/defaults-o2-daq.sh
@@ -1,8 +1,8 @@
 package: defaults-o2-daq
 version: v1
 env:
-  CXXFLAGS: "-fPIC -g -O2 -std=c++14"
-  CFLAGS: "-fPIC -g -O2"
+  CXXFLAGS: "-fPIC -O2 -std=c++14"
+  CFLAGS: "-fPIC -O2"
   CMAKE_BUILD_TYPE: "RELWITHDEBINFO"
 disable:
   - AliEn-Runtime
--- a/defaults-o2-dev-fairroot.sh
+++ b/defaults-o2-dev-fairroot.sh
@@ -1,8 +1,8 @@
 package: defaults-o2-dev-fairroot
 version: v1
 env:
-  CXXFLAGS: "-fPIC -g -O2 -std=c++14"
-  CFLAGS: "-fPIC -g -O2"
+  CXXFLAGS: "-fPIC -O2 -std=c++14"
+  CFLAGS: "-fPIC -O2"
   CMAKE_BUILD_TYPE: "RELWITHDEBINFO"
 disable:
   - AliEn-Runtime
--- a/defaults-o2.sh
+++ b/defaults-o2.sh
@@ -1,8 +1,8 @@
 package: defaults-o2
 version: v1
 env:
-  CXXFLAGS: "-fPIC -g -O2 -std=c++14"
-  CFLAGS: "-fPIC -g -O2"
+  CXXFLAGS: "-fPIC -O2 -std=c++14"
+  CFLAGS: "-fPIC -O2"
   CMAKE_BUILD_TYPE: "RELWITHDEBINFO"
 disable:
   - AliEn-Runtime


sawenzel:
Nice. Please note that it only partially fixes alisw/alibuild#400 because next to "-g" there is the question if we should set *any* optimization flag globally.

sawenzel:
If there are no comments, I think we can go on with the merge?

dberzano:
Here's the résumé of a discussion with @matthiasrichter: the problem we should solve is clearly separate what packages should be built with "debug information" and which ones can live without them.

Currently, the defaults define `CXXFLAGS`, `CFLAGS` and `CMAKE_BUILD_TYPE` in a consistent way, so that packages considering `CXXFLAGS` will have a `-g` appended, packages using CMake will use whatever `RELWITHDEBINFO` means, which is `-g` in 99% of the cases.

There are two problems we are trying to solve.

* ROOT builds are currently too large because of the internal LLVM's debug symbols. This PR solves this problem as this is due by the `-g` in `CXXFLAGS`.
* We would like to compile only selected "relevant" packages with `-g`, and only when appropriate. This PR _accidentally_ solves it by removing `-g` from everywhere except CMake-based projects, and our "relevant" packages are CMake-based. I do not think this PR solves the problem in a clean way.

We have however a solution underway, which allows (with the newest aliBuild) to use multiple defaults, and override compile flags on a per-package basis. The new stable release is due now, but we still need to solve alisw/alibuild#435.

Can we hold this until the problem is solved? For the record, solving this problem is a priority.

sawenzel:
I think we should still merge this PR due to following reasons:
a) it provides a quick solution to a real problem and will be beneficial to users now (faster compile, less disc space)
b) it does not affect production as only done for o2
c) We should, IMHO, or even must not set -g or -O2 flags globally as they interfere with other configurations 
d) This PR does not prevent or is in contradiction with the solution that is proposed in your comment.

matthiasrichter:
Let me just add to @dberzano 's comment that we disagree on the fact whether the `-g` should be in the global flags or not. I think in the same way as Sandro that this is simply wrong. It overrides proper behaviour of package build. The memory consumption during ROOT compilation is just a symptom.

Let me also add that even with a more fine-grained flag handling on the level of packages the `-g` is wrong in the global flags.

This PR only concerns O2 compilation, the mainstream developer wants to compile O2, FairRoot, probably AliRoot and ROOT with debug information. All those use CMake and there is no difference, while we speed up compilation and reduce memory consumption. 

The case is open for too long time, much longer than the PR, and has consumed too many work hours.  @sawenzel has nicely summarized why the PR should be merged, I fully agree with that.

ktf:
My two cents:

1) This PR means that we split production and O2 packages again, because the two defaults become incompatible.
2) If the above is acceptable until "Dario's proper fix", which I agree is what is really needed, I see no reason for not simply merging this, since in any case that's what we will have to do in the proper fix in any case (as we want per package tuning, and therefore the option will have to go from everywhere).

I therefore suggest that @matthiasrichter and @sawenzel state that 1 is ok and then we merge this and let Dario do the proper fix with less rush.

matthiasrichter:
Point 1) is ok for me. I expect that the production packages anyhow need to be adjusted after the proposed fix in alibuild.

sawenzel:
@dberzano : I think we reasonably discussed this issue to the end and there should be no more delay or reason to keep this in the pipeline. If the decision is not to accept, the PR should be closed.

dberzano:
Hi @sawenzel, I am merging this for now - it's reasonable to think as @ktf said that it gets to the result we want _for now_, even if it's not the best, and in the meanwhile I keep working on fixing it properly.

dberzano:
By the way if you don't mind I would leave the o2-daq defaults alone, would it be OK for you?

matthiasrichter:
The question is whether the DAQ build requires any other packages than ROOT, FairRoot and O2 to be build with debug symbols. Those packages use cmake and the build type will set `-g`.
@Barthelemy what do you think? Should that patch also be applied for the DAQ O2 default (as already in the PR)?

Barthelemy:
Hi everyone, @matthiasrichter @dberzano @ktf @sawenzel 
We can't build on the openstack VMs due to the memory consumption of the ROOT compilation. Therefore, and after chatting with Dario, I propose to modify the o2-daq defaults. 
Thank you all

dberzano:
OK thanks for your feedback! Merging...

(Please note this will cause quite the delay on all O2 pull request checks as full rebuilds are needed.)

sawenzel:
Thanks

Am 01.09.2017 12:51 nachm. schrieb "Dario Berzano" <notifications@github.com
>:

OK thanks for your feedback! Merging...

(Please note this will cause quite the delay on all O2 pull request checks
as full rebuilds are needed.)

—
You are receiving this because you were mentioned.

Reply to this email directly, view it on GitHub
<https://github.com/alisw/alidist/pull/820#issuecomment-326551104>, or mute
the thread

PR #821, opened 2017-08-18
 O2 recipe fix for the modulefile 

Barthelemy:


Barthelemy:
@matthiasrichter Indded. Better ?

Barthelemy:
@ktf sure. By the way, we should start using a version number rather than a tag like "dev".

matthiasrichter:
thanks, Barth!

PR #822, opened 2017-08-21
Fix o2.sh by removing spurious $

Barthelemy:
Bug introduced in #817 : extraneous $ in ALIROOT_VERSION cmake line


Barthelemy:
@dberzano @sawenzel

ktf:
Previously it was obviously wrong. Merging.

PR #826, opened 2017-08-23
Add Support for more modern OpenSSL (v1.0.2l)

PMunkes:
The OpenSSL-Version currently shipped in aliBuild/alidist is more than ten years old, and is not supported by some packages already in aliBuild/alidist. This pull request adds support for the current OpenSSL 1.0  version, which requires ECC.


dberzano:
#1091 implements this.

PR #828, opened 2017-08-24
Improve recipes linter

dberzano:
* Add debug option to show recipe tree
* Do not report, _e.g._, `version:` misplacement inside `overrides:`

### DIFF
--- a/scripts/lint-recipes
+++ b/scripts/lint-recipes
@@ -27,16 +27,29 @@ IN_ERROR=10000
 def keyState(key):
     return globals().get("IN_" + key.upper() + "_KEY", IN_GENERIC_KEY_VALUE)
 
+def stateKey(state):
+    for x in globals():
+        if type(globals().get(x)) != int: continue
+        if x.startswith("IN_") and globals().get(x) == state:
+            return x
+    return "<none_%d>" % state
+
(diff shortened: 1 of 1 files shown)


dberzano:
@ktf the problem I am trying to solve is that currently the linter reports errors whenever a `version:` field is found in a defaults under the `override` mapping. This commit solves the problem and only checks for "toplevel" tags in the document tree.

More extended checks will include:

* checking that `overrides` and `env` are mappings,
* checking that `overrides` is not present in a recipe which is not a default
* ...

For the moment we need #823 to be green :-)

ktf:
Not sure I understand the code, but if it works around the problem, this is ok for me. The proper check would be, IMHO, that we push the context when entering overrides and enforce also there that version and tag are the first two properties overridden. Good enough for now.

dberzano:
@ktf yes, ensuring proper ordering in the overrides has to be added to the list too, I agree!

What the code does is:

* counting only toplevel properties (property count is used to ensure correct positioning which we only do at toplevel for the moment)
* setting the state to something different from "generic" only when I have a scalar event at toplevel

and "toplevel" is defined by counting the number of items in the states stack, which identify the nesting level.

Thanks for merging!

PR #831, opened 2017-08-25
Common-O2 goes to v0.3.0

Barthelemy:
This is to be compatible with the latest repositories extraction from FlpPrototype.


Barthelemy:
@ktf as discussed

Barthelemy:
There is one more thing to fix. Closing for jnow.

PR #835, opened 2017-08-28
Fix boost check and DDS boost lookup

dberzano:


dberzano:
@shahor02 this should, finally, fix your problem

PR #836, opened 2017-08-30
Addition of cmake option to flpproto related recipes

Barthelemy:
Addition of -DCMAKE_EXPORT_COMPILE_COMMANDS=ON to cmake command to be able to use the o2codechecker.


ktf:
Now you see why I think one single repo is better... ;-)

PR #844, opened 2017-09-06
Set PYTHONPATH in readoutcard.sh

PascalBoeschoten:
This lets the user more easily access the Python interface

### DIFF
--- a/readoutcard.sh
+++ b/readoutcard.sh
@@ -10,6 +10,8 @@ requires:
   - "dim:(?!osx)"
 build_requires:
   - CMake
+prepend_path:
+  PYTHONPATH: $READOUTCARD_ROOT/lib
 source: https://github.com/AliceO2Group/ReadoutCard
 incremental_recipe: |
   make ${JOBS:+-j$JOBS} install
--- a/readoutcard.sh
+++ b/readoutcard.sh
@@ -55,6 +55,7 @@ module load BASE/1.0                                                          \\
 setenv READOUTCARD_ROOT \$::env(BASEDIR)/$PKGNAME/\$version
 prepend-path PATH \$::env(READOUTCARD_ROOT)/bin
 prepend-path LD_LIBRARY_PATH \$::env(READOUTCARD_ROOT)/lib
+prepend-path PYTHONPATH \$::env(READOUTCARD_ROOT)/lib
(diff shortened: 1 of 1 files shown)


readoutcard.sh
```diff
@@ -55,6 +55,7 @@ module load BASE/1.0                                                          \\
 setenv READOUTCARD_ROOT \$::env(BASEDIR)/$PKGNAME/\$version
 prepend-path PATH \$::env(READOUTCARD_ROOT)/bin
 prepend-path LD_LIBRARY_PATH \$::env(READOUTCARD_ROOT)/lib
+prepend-path PYTHONPATH \$::env(READOUTCARD_ROOT)/lib
```
dberzano: Out of curiosity, is there any reason why Python modules for readout are installed in a rather non-standard path? Normally they will go in a `site-packages` or `dist-packages` directory or something (look at the `yoda.sh` recipe for instance).

PascalBoeschoten: Well, there is actually no separate module. The libReadoutCard.so C++ library is also directly usable by Python.

dberzano: OK, but isn't `LD_LIBRARY_PATH` sufficient then? Pardon my ignorance!

PascalBoeschoten:
I had a look at the `yoda.sh` recipe, but I'm not sure I understand.
Is the `prepend_path:` in the header supposed to replace the `prepend-path PYTHONPATH ...` in the modulefile section? Or is it in addition to it? 
Also, it doesn't seem to be done for the PATH and LD_LIBRARY_PATH variables?

dberzano:
Hi, we actually need both. `PATH` and `LD_LIBRARY_PATH` don't need it (unless you use paths different than `bin` and `lib`), this is done automatically by aliBuild.

The reason is:

* modulefile is used at runtime (i.e. with `alienv`)
* `prepend_path` at build time (and has an effect is a package depending on your readout one wants to use some functionality during the build requiring `PYTHONPATH` to be exported, which is probably not the case, but for consistency I would add it)

Hope this clarifies.

readoutcard.sh
PascalBoeschoten: Unfortunately not! We need PYTHONPATH for the import statement to work

PascalBoeschoten:
I added the prepend_path. Is this what you had in mind?

PR #845, opened 2017-09-07
Remove tests from nanomsg recipe

ktf:


ktf:
Apparently there is some test which currently fails.

PR #849, opened 2017-09-11
Add support for external compilers

fuhlig1:


protobuf.sh
```diff
@@ -11,6 +11,12 @@ prefer_system_check: |
 
 rsync -av --delete --exclude="**/.git" $SOURCEDIR/ .
 autoreconf -ivf
+
+# Set the environment variables CC and CXX if a compiler is defined in the defaults file 
+# In case CC and CXX are defined the corresponding compilers are used during compilation  
+[[ -z "$CXX_COMPILER" ]] || export CXX=$CXX_COMPILER
```
ktf: Why do you need the export? If it's defined in the defaults environment it should be always exported already. Is that not the case?

pythia6.sh
```diff
@@ -1,20 +1,25 @@
 # a pythia6 recipe based on the one from FairROOT
 package: pythia6
 version: "%(tag_basename)s%(defaults_upper)s"
-tag: "428-alice1"
+tag: "alice/428"
```
ktf: No, this is not ok. We should use a tag, not a branch in the recipes, because otherwise fixes to pythia will trigger a rebuild for everyone. If you really want to have a branch, please do so in a separate default.

nanomsg.sh
```diff
@@ -1,4 +1,4 @@
-package: nanomsg
+#package: nanomsg
```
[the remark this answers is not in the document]
fuhlig1: Hi @ktf,

this must be a typo. There is no reason to comment it.

protobuf.sh
fuhlig1: Because CXX is not defined in the defaults. I tried to change the recipes in such a way that they still work with default files which does not define the compiler. This should allow to gradually change the recipes. Only if all recipes support the new scheme one would change the default files. For the compilation of O2 all recipes support an external compiler, but I did not touch any recipe which is not needed for the 02 recipe.

That's why I defined variables like CXX_COMPILER in the defaults file which are then used to setup the needed variables like CXX. If there is a possibility to do so in the defaults file this would be the cleaner solution.

pythia6.sh
fuhlig1: I agree that one should use a tag instead of a ranch, but there was a reason why I had to use the branch, which I can't recall in the moment. I will try to find out why I switched to the branch.

dberzano:
Hi @fuhlig1 thanks for your PR, I'll add my review as soon as possible.

dberzano:
Hi @fuhlig1, do you mind making this PR editable by maintainers? Thanks!

fuhlig1:
Hi @dberzano, sure, but I think the PR already allow edits from the maintainers.

dberzano:
OK sorry I hadn't checked before actually!

fuhlig1:
Since no progress happened I close the issue.

PR #850, opened 2017-09-13
FLP prototype checkcode recipe

awegrzyn:


flpprotocheckcode.sh
```diff
@@ -0,0 +1,92 @@
...
+  - flpproto
+  - o2codechecker
+build_requires:
+  - CMake
+force_rebuild: 1
+---
+#!/bin/bash -e
+
+# Runs a set of (selected) code checks on the O2 code-base. Assumes the
+# compile_commands.json file is available under $O2_ROOT
```
sawenzel: probably this should be adapted? Is $O2_ROOT still correct?

awegrzyn: Indeed, I'll make appropriate changes

sawenzel: This file is probably 99% the same as o2checkcode.sh; I am wondering whether one could reuse it for multiple repos. The only thing that changes is the location of the compile_commands.json location which could be given as some argument?

awegrzyn: That's true the biggest difference is it depends on `flpproto` instead of `O2`, but I agree there should be a single recipe for this.

sawenzel: I see. Indeed it's probably hard making the dependencies a runtime behaviour ...
Maybe one could keep the actual script inside a separate file and include it in different recipes.

ktf: Actually, it should be trivial if we had two defaults, one for `flpprototype` and one for `O2` (which we already have). That would allow dependencies override, like we do for other stuff.

awegrzyn: Okay, so how do we proceed? I'd like to make it available to people developing asap.

ktf: I would go ahead and finish what you have done, then when we have time we merge the two. @dberzano up to you.

awegrzyn:
Is there anything that stops from merging?

awegrzyn:
@dberzano ...

PR #851, opened 2017-09-14
Provide C++11 Guideline Support Library; Use in O2

sawenzel:
This provides the C++ guidelines support library with useful
features such as span (array_view) on containers.
We can use this until we are fully C++17 compliant.

### DIFF
--- a/ms_gsl.sh
+++ b/ms_gsl.sh
@@ -1,0 +1,12 @@
+package: ms_gsl
+version: "1"
+source: https://github.com/Microsoft/GSL.git
+tag: b014508
+---
+#!/bin/bash -e
+
+# recipe for the C++ guidelines support library (Microsoft implementation)
+# can be deleted once we are fully C++17 compliant
+
+# just rsync into the installdir since header only
+rsync -a $SOURCEDIR/include $INSTALLROOT
--- a/ms_gsl.sh
+++ b/ms_gsl.sh
@@ -1,7 +1,7 @@
 package: ms_gsl
 version: "1"
-source: https://github.com/Microsoft/GSL.git
 tag: b014508
+source: https://github.com/Microsoft/GSL.git
 ---
 #!/bin/bash -e
 
--- a/o2.sh
+++ b/o2.sh
@@ -6,6 +6,7 @@ requires:
   - DDS
   - Vc
   - hijing
+  - ms_gsl
 source: https://github.com/AliceO2Group/AliceO2
 prepend_path:
   ROOT_INCLUDE_PATH: "$O2_ROOT/include"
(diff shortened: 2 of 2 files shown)


ms_gsl.sh
```diff
@@ -0,0 +1,31 @@
+package: ms_gsl
+version: master
```
ktf: Please fix the linting and pick an actual tag / commit. We should not depend on master.

o2.sh
```diff
@@ -76,6 +77,7 @@ cmake $SOURCEDIR -DCMAKE_INSTALL_PREFIX=$INSTALLROOT
       ${GSL_ROOT:+-DGSL_DIR=$GSL_ROOT}                            \
       ${PYTHIA_ROOT:+-DPYTHIA8_INCLUDE_DIR=$PYTHIA_ROOT/include}  \
       ${CMAKE_BUILD_TYPE:+-DCMAKE_BUILD_TYPE=$CMAKE_BUILD_TYPE}   \
+      -DMS_GSL_INCLUDE_DIR=$MS_GSL_ROOT/include                   \
```
ktf: Is O2 already ready to accept this?

ms_gsl.sh
```diff
@@ -0,0 +1,31 @@
...
+source: https://github.com/Microsoft/GSL.git
+tag: master
+---
+#!/bin/bash -e
+
+# recipe for the C++ guidelines support library (Microsoft implementation)
+# can be deleted once we are fully C++17 compliant
+
+# just rsync into the installdir
+rsync -a $SOURCEDIR/ $INSTALLROOT/
```
ktf: I think you need to exclude `.git`, no? Also mention the fact that this is header only.

ms_gsl.sh
```diff
@@ -0,0 +1,31 @@
...
+
+# Modulefile
+MODULEDIR="$INSTALLROOT/etc/modulefiles"
```
ktf: Why do you need this? `gsl` is header only, so there is no need to for the modulefile.

ms_gsl.sh
sawenzel: fixed

ms_gsl.sh
sawenzel: ok

ms_gsl.sh
sawenzel: fixed

o2.sh
sawenzel: No test will be broken. This is just a definition of some variable ... which if not used will do nothing.

ktf: ok. That's what I meant with my first question.

sawenzel: I have the version that used this ready. Just needs some cleanup.

sawenzel: For the record: Here is the pull request that used it on the O2 side: https://github.com/AliceO2Group/AliceO2/pull/553

PR #852, opened 2017-09-15
Make linter check for Modulefile creation

dberzano:
Does not check defaults

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,5 +8,7 @@ script: |
   for X in $(git diff --name-only $TRAVIS_COMMIT_RANGE | grep .sh); do
     echo Linting $X
     scripts/lint-recipes $X || ERR="$ERR $X"
+    [[ $X != defaults-* ]] || continue
+    grep -q '#%Module1.0' $X || { echo "$X: Missing Modulefile"; ERR="$ERR $X"; }
   done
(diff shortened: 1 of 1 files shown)


dberzano:
#851 was merged without a proper Modulefile and it's causing publisher problems. We should make sure that all recipes have a Modulefile creation. (Given that the "proper" solution would be to make aliBuild create a default one in case it's absent.)

sawenzel:
The modulefile in #851 was there initially but we took it out since this is a header only library.

ktf:
@sawenzel We should have made `ms_gsl` a build_requires as well. I forgot to mention.

dberzano:
Hi @ktf, there was a request for aliBuild/alienv to also load the environment you would use for building: in principle in that case Modulefiles should be there, but also in my opinion the clean solution would be to automatically create them with aliBuild. Closing it.

@sawenzel @ktf Could please someone fix #851 then?

dberzano:
I see it's done in #853.

ktf:
I agree on creating them in alibuild, but then again, if build_require is specified, I see no reason for checking if a modulefile is there. What should also probably happen is that the linter uses the same parser as alibuild and is able to lint the whole chain of dependencies, not single packages.

dberzano:
@ktf Yes this is why I have closed the PR ;-)

PR #853, opened 2017-09-15
Make ms_gsl a build_requires

ktf:


ktf:
This should fix the current problem with ms_gsl publication, no?

PR #857, opened 2017-10-03
Fix XCode9 issues in ApMon

ktf:


dberzano:
Hi @ktf, I have tagged the fix you've added (alisw/apmon-cpp#1) and updated this PR accordingly, so that we pinpoint the tag and not the head of a branch (this is a policy we've discussed together some time ago). OK for you?

ktf:
Yes, you are right.

PR #860, opened 2017-10-03
Fix defaults

sawenzel:
This PR supersedes old PR #759.

It achieves:
* correct compilation of ROOT6 when we use defaults (which needs to depend on GCC_TOOLCHAIN)
* does not forbid the compilation of O2 and AliRoot using the same default
   - this is very important to do cross comparisons/tuning of simulation for instance


sawenzel:
@dberzano : Can this be merged, please?

dberzano:
Checking now.

PR #869, opened 2017-10-04
Bump ROOT 6 to a hash compatible with Xcode 9

dberzano:


dberzano:
Note that I am testing it now before merging.

PR #872, opened 2017-10-06
Update tagged FairRoot version

sawenzel:
Updating to status 05 October 2017;
Using official repository from FairRoot


sawenzel:
This should quickly fix #872; One could go back to an alice tagged version in alisw/FairRoot later.

ktf:
Now I remember why we have a separate repository... We did not want to have the hash as a name... Good enough for now, we can fix it later...

PR #873, opened 2017-10-06
Enable HepMC3 in O2 and avoid conflicts with packages requiring HepMC2

preghenella:
This PR is to modify the defaults-o2 and o2 receipts in order to add HepMC3 as a requirement and at the same time removing the requirements on HepMC2 loaded from other modules.

It also contains a second commit to allow building AliRoot using defaults-o2 receipt, enabling one to have AliRoot compiled with the same package list as O2 and to be coupled in case it is needed.


dberzano:
Hi @preghenella, what's the status of this one? Can it be closed?

preghenella:
Ciao @dberzano,

sorry for the late reply, although I don't understand why the issue has been closed.
It was opened, but never merged.
What went wrong in the PR?

Thanks,

dberzano:
The PR was closed as no feedback was given and the tests are red. Let's reopen it then. I have additional comments, I'll put them inline.

defaults-o2.sh
```diff
@@ -53,6 +52,15 @@ overrides:
     tag: "v3.8.2"
     prefer_system_check: |
       which cmake && case `cmake --version | sed -e 's/.* //' | cut -d. -f1,2,3 | head -n1` in [0-2]*|3.[0-7].*) exit 1 ;; esac
+  simulation:
```
dberzano: Please modify directly `simulation.sh`. We use it only for O2, no need to override it in the defaults.

o2.sh
```diff
@@ -82,6 +83,7 @@ cmake $SOURCEDIR -DCMAKE_INSTALL_PREFIX=$INSTALLROOT
       -DPROTOBUF_LIBRARY=$PROTOBUF_ROOT/lib/libprotobuf.$SONAME   \
       ${GSL_ROOT:+-DGSL_DIR=$GSL_ROOT}                            \
       ${PYTHIA_ROOT:+-DPYTHIA8_INCLUDE_DIR=$PYTHIA_ROOT/include}  \
+      ${HEPMC3_ROOT:+-DHEPMC3_DIR=$HEPMC3_ROOT}                    \
```
dberzano: Fix alignment of `\`.

preghenella: ok, please check updated PR

defaults-o2.sh
preghenella: ok, I modified as requested, please see updated PR.

It was although not clear to me that simulation.sh package is only used in O2, according to the structure. I mean, why then the dependencies tracked in simulation.sh are not explicitly included in the o2.sh receipt? That looked to me like someone was using this as a package collection.

Anyway, the change is done.

preghenella:
Ciao @dberzano ,

thanks for your message. Previously I did not see the tests running and failing here.
I will keep an eye and make sure the test is successful.

Cheers

defaults-o2.sh
```diff
@@ -53,6 +53,10 @@ overrides:
     tag: "v3.9.4"
     prefer_system_check: |
       which cmake && case `cmake --version | sed -e 's/.* //' | cut -d. -f1,2,3 | head -n1` in [0-2]*|3.[0-8].*|3.9.[0-3]) exit 1 ;; esac
+  pythia:
```
dberzano: So, for my understanding: for O2 builds, in `pythia` you disable `HepMC` in order to avoid conflicts with `HepMC3`. Is this intended? Did you check with @sawenzel if this is OK?

dberzano: (This is just to say that changes are fine for me, approving. Let's just wait for @sawenzel's feedback.)

preghenella: Yes, this is intended. The use of HepMC in pythia is related to standalone event generation where the pythia output is translated in HepMC2 format. I will discuss (or rediscuss in case we did already one month back) with @sawenzel if there is any issue with that.

dberzano:
ping @sawenzel: do you have any objection to this PR?

Issue #876, opened 2017-10-07
GCC / boost version


davidrohr:
Hi,

the recipes defaults-o2, defaults-o2-daq, and gcc-toolchain impose minimal and maximum versions for boost and GCC. This is a bit annoying when you have an up to date linux installation.

Currently, I locally have to modify all the recipes and remove maximum version limits, and yet I have not seen any problems.
Could we perhaps introduce an experimental or unstable environment variable, or any other means, which would disable all the checks for a maximum version?
(We could leave those checks in, where we know that a new version is buggy, and then throw a detailed error like 'GCC > 6.2 leads to miscompilation with O2 package'.)

But that would enable people to use an up to date system in a general and clean way.

Cheers
David

ktf:
PRs to improve the detection / messages are of course welcome. ;-) For GCC I am surprised that the latest works with ROOT, because that's acknowledged not to be the case by the ROOT team itself, so that's why we are in general a bit cautious when updating. For boost I agree an upper bound is probably less critical. @dberzano shall we discuss this in the next WP3?

PR #878, opened 2017-10-07
Use system lhapdf if available

davidrohr:


dberzano:
Hi @davidrohr, LHAPDF is a small package which does not take so much time for being built. Do you have a particular reason for preferring a system-installed one? If we merge this, we might get into weird issues with our users having LHAPDF "accidentally" installed system-wide. I'd prefer not to merge this, but I am open to listen to your use case :-)

ktf:
I agree with @dberzano. My experience is that using system installations for "physics" packages leaves the door open to can of worms.

davidrohr:
Hi, OK, I see, might make sense from your standpoint. I usually prefer system packages as general principle, but I can keep that patch locally.

ktf:
Alternatively we could have some special default where we are much more aggressive with this.

ktf:
Something like `--defaults full-system-dependencies-use-at-your-own-risk`

davidrohr:
Hi,

that sounds like a good idea for me. Can you provide multiple defaults in a row?
Then, one could perhaps also use that for what I described here: https://github.com/alisw/alidist/issues/876.

ktf:
No, you cannot, by design. We want to avoid people specifying multiple defaults on the command line, in order to avoid huge command lines, to avoid having to check if they are compatible with each other, and to avoid to have default order dependent builds.

What we discussed at some point was to allow defaults to inherit from each other, but unless I am mistaken (@dberzano correct me if not the case) no work went into it.

dberzano:
@ktf you're correct, no work's been done so far.

PR #879, opened 2017-10-07
fix usage of system libprotobuf

davidrohr:
Recreate pull request after rebase on master (somehow the forced push closed it...):

Compilation with system protobuf does not work because PROTOBUF_ROOT is empty, but the strings PROTOBUF_LIBRARY and PROTOBUF_PROTOC_EXECUTABLE contain absolute paths, which in that case lead e.g. to /bin/protoc, but that executable does not exist.

This patch checks for the empty string, and removes the absolute patch.
For the binary, it uses which to find the correct patch.

For O2, there is the additional problem that O2 needs an absolut patch to the library in the build process.
The patch treats this by modifying the which protoc patch, which is probably not ideal, but I do not see what the script should do better. Ideally, one could improve the O2 build process to use a proper FindProtobuf cmake module or so.

### DIFF
--- a/fairroot.sh
+++ b/fairroot.sh
@@ -44,6 +44,14 @@ esac
 
 [[ $BOOST_ROOT ]] && BOOST_NO_SYSTEM_PATHS=ON || BOOST_NO_SYSTEM_PATHS=OFF
 
+if [ -n "$PROTOBUF_ROOT" ]; then
+    PROTOBUF_EXECUTABLE=$PROTOBUF_ROOT/bin/protoc
+    PROTOBUF_LIB=$PROTOBUF_ROOT/lib/libprotobuf.$SONAME
+else
+    PROTOBUF_EXECUTABLE=`which protoc`
+    PROTOBUF_LIB=libprotobuf.$SONAME
+fi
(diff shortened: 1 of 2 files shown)


ktf:
How about we change this to:

- Always use `which protoc` to pick up `protoc`
- Assume the headers are in `$(basename $(which protoc))/../include`.
- Assume the library is in `$(basename $(which protoc))/../lib`.

davidrohr:
Makes totally sense to me, will update the PR accordingly

ktf:
@davidrohr can you resolve the conflicts?

davidrohr:
@ktf : I have rebased on master and solved the conflicts, and verified that it builds correctly on my box with the system libprotobuf.

ktf:
Assigned to @dberzano for integration.

dberzano:
By the way, the `FindProtobuf` in FairRoot is "old-schooled" like some others we've already contributed back to (nanomsg, zeromq). I am fixing it in FairRoot, more news soon.

dberzano:
This can be closed in favor of #892.

PR #880, opened 2017-10-09
Add O2 tracking library to O2 build

davidrohr:
This add a build recipe for the O2 tracking library only: It checks out AliRoot from the github fork of the library, and builds the library only. No AliRoot dependencies are needed.

When the AliRoot patch is merged into master, I will change this to use the official AliRoot repo.
Later, this could also be changed to take the library from the normal AliRoot build process, if AliRoot would become a dependency.

Currently, the change to o2.sh only passes a -D cmake parameter where to find the library. This parameter is currently unused.
As a second pull request after this one is accepted, I will push the patch to O2 that uses the cmake parameter to build versus the tracing library.


hlto2catracking.sh
```diff
@@ -0,0 +1,34 @@
+# Package for the HLT TPC O2 CA Tracking library in standalone build (This package is to be used if AliRoot is not in the dependency)
+package: HLTO2CATracking
+version: "%(tag_basename)s"
+tag: hlt_o2_ca_tracking-v1.0
+source: https://github.com/davidrohr/AliRoot
+build_requires:
+  - CMake
+
```
sawenzel: to ensure consistent compilers being used, we should require here "GCC_TOOLCHAIN" as a dependency.

hlto2catracking.sh
```diff
@@ -0,0 +1,34 @@
...
+---
+#!/bin/sh
+cmake ${SOURCEDIR}/HLT/TPCLib/tracking-standalone/cmake_Standalone/ \
```
sawenzel: what about the build type? Should we pickup the globally set CMAKE_BUILD_TYPE?

hlto2catracking.sh
```diff
@@ -0,0 +1,36 @@
+# Package for the HLT TPC O2 CA Tracking library in standalone build (This package is to be used if AliRoot is not in the dependency)
+package: HLTO2CATracking
+version: "%(tag_basename)s"
+tag: hlt_o2_ca_tracking-v1.0
+source: https://github.com/davidrohr/AliRoot
+build_requires:
+  - CMake
+  - GCC-Toolchain:(?!osx)
```
sawenzel: maybe this even needs to be a real requirement (as opposed to build_requires) since the C++ libraries might be needed at runtime?

davidrohr:
Pushed a new version, all comments are included

sawenzel:
@davidrohr : I tested building which works fine. For consistency you should probably add the GCC-TOOLCHAIN in the module file as dependency.

davidrohr:
@sawenzel : Thanks for testing, GCC-TOOLCHAIN is already added in the last update from yesterday.

sawenzel:
Note that the package does not build on MAC:

/Users/swenzel/alisw_new/sw/SOURCES/HLTO2CATracking/hlt_o2_ca_tracking-v1.0/hlt_o2_ca_tracking-v1.0/HLT/TPCLib/tracking-standalone/cmake_Standalone/../include/TStopwatch.h:200:17: error: use of undeclared identifier 'CLOCK_REALTIME'
  clock_gettime(CLOCK_REALTIME, &t);
                ^
1 warning and 1 error generated.

ktf:
I guess we should use `std::chrono` everywhere now..

davidrohr:
Hi,

since this compiles also for the HLT, I cannot rely on C++11.
I will add cases for C++11 with std::chrono, and also a fallback for Mac.

Cheers
David

davidrohr:
Hi @sawenzel , I have update the recipe and the library. Could you retry on Mac?

sawenzel:
@davidrohr : Still getting the same error.

sawenzel:
@ktf , @dberzano , @davidrohr : I confirm that this builds now correctly on Linux + Mac. This can be merged now.

o2hltcatracking.sh
```diff
@@ -0,0 +1,35 @@
+package: O2HLTCATracking
+version: "%(tag_basename)s"
+tag: hlt_o2_ca_tracking-v1.1
```
matthiasrichter: I think it's not acceptable to make everyone dependent on a private fork

o2hltcatracking.sh
```diff
@@ -0,0 +1,35 @@
...
+source: https://github.com/davidrohr/AliRoot
```
matthiasrichter: I think it's not acceptable to make the project depending on another fork of AliRoot

ktf: Yes, agreed.

dberzano: You are right - it was meant to be temporary, as a matter of fact, until it's included in upstream AliRoot. Unfortunately as you know it has to be used both in AliRoot and O2.

A mitigation could be working on a branch of _our_ AliRoot fork to prevent aliBuild to create yet another full clone of AliRoot.

Also, from what I understand: one of my suggestions was to have a separate default for that, but this part is really needed everywhere.

If @davidrohr agrees (but I sense he will not!) an educated solution would be... to duplicate the code.

davidrohr:
I agree this is definitely not the best approach and also not a permanent solution.
I have the feeling everyone has a different idea how to handle this, and they are kind of not compatible.
We should definitely converge to a proper solution, but I also have the feeling going forth and back over and over again only wastes time.

To the ideas:
1. One of the first implementation ideas was a dedicated standalone library, i.e. take the code out of AliRoot. This is totally fine with me, but others complained that would introduce another dependency for AliRoot, if we want to continue to use the code there.
2. The private fork is only a temporary solution. We want to merge this into alisw/AliRoot, but the TPC is still investigating the implications of the other changes in that branch. Since people wanted to start working with the tracking now, we chose this approach (and it was actually discussed with @dberzano @sawenzel  and @ktf ). I am totally fine of moving this to a branch of alisw/AliRoot, but we can also just wait few weeks until it makes it into master. It is definitely not going to stay like it is now.
3. Duplicating the code is in my opinion principally a bad solution. In this case, we would have the addiitonal problem that AliRoot forbits C++11 while O2 uses it. Anyway, if you prefer so, I can keep a copy of the code in the O2 repo and keep it in sync with the version in the AliRoot repo.

PR #883, opened 2017-10-09
Allow protobuf 2.6 as system package

ktf:
Current version only works for protobuf 3+, while we compile 2.6 in our own recipe.


dberzano:
Checking it now.

PR #885, opened 2017-10-10
Add build type to alo cmake invocation

aphecetche:


dberzano:
Linter's not happy with the recipe:

```
tag should always be right after version, if the latter is present
  in "<string>", line 10, column 6:
    tag: master
```

aphecetche:
Hi Dario,

I actually missed the fact that there's a linter for the recipes now (that information should may be be added to http://alisw.github.io/alibuild/reference.html ?). Anyway, it's fixed now.

PR #886, opened 2017-10-10
Bump ROOT 6 to fix for macOS MonaLisa writer

dberzano:


dberzano:
@masera this fixes your problem (and your students'). My patch was approved upstream, see root-project/root#1163 and root-project/root#1164.

dberzano:
It is also possible that ROOT "7" features can be re-enabled after their fix, so I am holding on before merging, we might need to change that too and it's better to do it in one go.

dberzano:
Turns out we still need to keep ROOT "7" disabled. Merging as it is.

PR #889, opened 2017-10-11
Version bump to v1.2, fix header that makes o2 compilation on MacOS f…

davidrohr:
…ail, use new version of track fit, some bug fixes and improvements

@sawenzel : Can you test this on Mac. We need a version bump because the AliRoot header breaks the compilation of O2 on Mac.


sawenzel:
Can we merge this?

PR #890, opened 2017-10-12
Fix system check for protobuf, check for both the headers and the bin…

davidrohr:
…ary, because some systems might only have the library installed

### DIFF
--- a/protobuf.sh
+++ b/protobuf.sh
@@ -6,7 +6,7 @@ build_requires:
  - "GCC-Toolchain:(?!osx)"
 prefer_system: "(?!slc5)"
 prefer_system_check: |
-  printf "#include \"google/protobuf/message.h\"\nint main(){}" | c++ -I$(brew --prefix protobuf)/include -Wno-deprecated-declarations -xc++ - -o /dev/null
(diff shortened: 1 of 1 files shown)


dberzano:
I am having a look at it right now.

protobuf.sh
```diff
@@ -6,7 +6,7 @@ build_requires:
  - "GCC-Toolchain:(?!osx)"
 prefer_system: "(?!slc5)"
 prefer_system_check: |
-  printf "#include \"google/protobuf/message.h\"\nint main(){}" | c++ -I$(brew --prefix protobuf)/include -Wno-deprecated-declarations -xc++ - -o /dev/null
```
dberzano: Hi, I would use `protoc -h` (which brings up a help screen and exits with 0) instead of `type protoc`. Can you do this change?

dberzano:
Actually if @davidrohr agrees I can do the change myself.

davidrohr:
@dberzano : sure, go ahead

dberzano:
@ktf I believe you are commenting as @alibuild :-)

PR #896, opened 2017-10-18
Bump FairRoot for fixing protobuf lookup

dberzano:


dberzano:
Note that tomorrow's date is not a typo, it is intended.

PR #898, opened 2017-10-19
Add dependency to common-o2

Barthelemy:
Configuration needs a dependency to common-o2

### DIFF
--- a/configuration.sh
+++ b/configuration.sh
@@ -7,6 +7,7 @@ requires:
   - "GCC-Toolchain:(?!osx)"
   - protobuf
   - grpc
+  - Common-O2
 build_requires:
   - CMake
 source: https://github.com/AliceO2Group/Configuration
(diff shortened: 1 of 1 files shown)


matthiasrichter:
Hi @Barthelemy , can we maybe find a better name for the `Configuration` project? I had problems understanding the setup, as file names like `configuraton.sh` are very general and cause a wrong guidance.

Barthelemy:
Hi @matthiasrichter , 
I understand the issue. It is not easy as the pcakge is really taking care of configuration. 

This should perhaps be discussed with @vascobarroso and @PascalBoeschoten .

dberzano:
@matthiasrichter remember we're using a piece of software called ROOT :-) (just kidding!)

vascobarroso:
@matthiasrichter no problem in changing if a better naming can be found. What did you have in mind ?

PR #900, opened 2017-10-20
Drop perl support from xrootd

ktf:


ktf:
Several people are reporting issues when building xrootd due to perl bindings. Do we actually need them?

PR #902, opened 2017-10-23
Follow recent FairRoot requirement: cmake 3.9.4

matthiasrichter:
Latest update in FairRoot breaks the O2 CI, newer cmake is now
required (was at 3.8.2), see
https://github.com/FairRootGroup/FairRoot/pull/659
Boost is already at 1.64 in our recipies


matthiasrichter:
@ktf, I did a rebase to avoid the merge commit, so the approval is needed again, sorry ...

PR #903, opened 2017-10-23
Version bump of O2 HLT tracking library

davidrohr:


davidrohr:
@ktf @dberzano : Hi, just for my education: is there any standard validation procedure to verify whether this version bump breaks anything? I built it on my Linux system and it works there.

Could you please accept the pull request if you have no concerns about it, since it is required for a pending pull request with an O2 tracking update.

ktf:
1 day of delay to update externals is more than reasonable, IMHO. We should really do the effort to move this in proper O2 if faster development cycle is needed. alidist is for slowly paced external packages, not for packages which should be part of the main development.

Issue #905, opened 2017-10-24
Problem building ROOT


kouzinopoulos:
I am trying to build a fresh installation of the O2 stack using AliBuild (and the master branch of AliDist):

`aliBuild --defaults o2-daq build O2`

However the compilation stops at the ROOT recipe, without a descriptive message of what went wrong/ what is missing. I am attaching sw/BUILD/ROOT-latest/log:

[log.txt](https://github.com/alisw/alidist/files/1411090/log.txt)

matthiasrichter:
does aliDoctor report anything when you run it with the same default?

sawenzel:
@kouzinopoulos , @ktf: I just had the same problem with a fresh installation on a new machine.
AliDoctor does not complain but then ROOT compilation fails at cmake stage because it does not find
things like `xml` or `ssl`. It seems to me that these are actual dependencies which should be expressed in the root recipe and reported by aliDoctor if not found.

@kouzinopoulos : In your case it suffices to install openssl-devel as a system package.

ktf:
The problem is that both OpenSSL and XML are picked up from alien runtime, which is not there in the `o2-daq` default. I am thinking on how to solve this.

sawenzel:
In my case it did not compile with the ordinary `o2` default. So the problem is wider than just `o2-daq`.

ktf:
Can you have a look to https://github.com/alisw/alidist/pull/907 to see if it fixes the problem? Apparently we disable alien also in O2, for whatever reason.

ktf:
Actually, let's close this and move the discussion in #907.

PR #907, opened 2017-10-25
Make ROOT build also when Alien-Runtime is not there

ktf:
Right now I think it simply works by accident, since the required
dependencies on libxml2 and openssl are very likely to be found on the
system. This should make sure that we build our own version in case
that's not the case.


kouzinopoulos:
Referring to issue [#905](https://github.com/alisw/alidist/issues/905)

Installing libssl-dev fixed the issue as per the instructions of @sawenzel 

AliDoctor O2 did not report any error  on the missing library dependency

defaults-o2-daq.sh
```diff
@@ -42,6 +42,7 @@ overrides:
       - FreeType:(?!osx)
       - Python-modules
       - libxml2
+      - OpenSSL
       - "GCC-Toolchain:(?!osx)"
```
sawenzel: libxml + openssl also need to be added to `defaults-o2.sh`

ktf: Actually, those should come from `Alien-Runtime` in the case of `default-o2.sh`.

sawenzel: Yes, but it is usually better to be explicit about dependencies ... and not rely in implicit stuff which is hard to track? In any case, if it already works no problem ...

awegrzyn:
Hi, what's the status of this PR?

dberzano:
Let me have a look: I have an independent change in the ROOT recipe to be merged too, I'll open a separate PR so that we can merge them together and avoid too many rebuilds.

dberzano:
OK so at the moment I am refactoring it by making it possible to use ROOT without AliEn-Runtime at all. Some packages, now part of AliEn-Runtime, are in any case going to be taken out after the migration to JAlien. This is the list of packages:

* OpenSSL
* XRootD
* libxml2

dberzano:
I have by the way found an interesting loophole in CODEOWNERS. This PR was opened by @ktf but edited by me (I am an allowed maintainer). Now, I can review _my own_ code just because the PR was originally opened by @ktf :-)

That said, this should address the problem of compiling ROOT without AliEn-Runtime available, and should pave the way for JAlien as well (@costing @mimarpe).

dberzano:
Well: this iteration does not work, at least for ROOT 5 on slc5 (sic). More commits to come...

dberzano:
I have managed to successfully build Run2 and Run3 software with this recipe. Note that:

* I added Python as a dependency for Run2 software as well
* I have fixed Python's recipe in order to _really_ pick our OpenSSL version

dberzano:
I would merge it if there's no objection so that the next O2 nightly will be based on it.

PR #910, opened 2017-10-26
Make defaults-o2-dev-fairroot build simulation and take true dev branch

sawenzel:
We have the o2-dev-fairroot default in order to compile against the head of the FairRoot dev branch, -- in particular to detect problems with the dev branch early before we adopt it.

This commit extends the scope of such builds to include simulation/reco which otherwise would not be checked.

This commit also makes sure that we actually compile the true dev branch and not a potentially outdated branch of our fork.

### DIFF
--- a/defaults-o2-dev-fairroot.sh
+++ b/defaults-o2-dev-fairroot.sh
@@ -7,14 +7,6 @@ env:
 disable:
   - AliEn-Runtime
   - AliRoot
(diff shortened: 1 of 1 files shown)


dberzano:
Is it ready to be merged?

sawenzel:
Yes, I believe so.

PR #917, opened 2017-10-30
Handle corner case for builtin freetype

ktf:
Apparently this combination is actually possible.

### DIFF
--- a/root.sh
+++ b/root.sh
@@ -113,9 +113,14 @@ else
         -DCMAKE_PREFIX_PATH="$FREETYPE_ROOT;$SYS_OPENSSL_ROOT;$GSL_ROOT;$ALIEN_RUNTIME_ROOT;$PYTHON_ROOT;$PYTHON_MODULES_ROOT"
   FEATURES="builtin_pcre mathmore xml ssl opengl minuit2 http
             pythia6 roofit soversion vdt ${CXX11:+cxx11} ${CXX14:+cxx14} ${XROOTD_ROOT:+xrootd}
-            ${ALIEN_RUNTIME_ROOT:+alien monalisa}
-            ${ENABLE_COCOA:+builtin_freetype}"
-  NO_FEATURES="${FREETYPE_ROOT:+builtin_freetype} root7"
+            ${ALIEN_RUNTIME_ROOT:+alien monalisa}"
+  NO_FEATURES="root7"
+
+  if [[ "$ENABLE_COCOA" != "" ]]; then
+    FEATURES="$FEATURES builtin_freetype"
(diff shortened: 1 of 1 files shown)


root.sh
```diff
@@ -113,9 +113,14 @@ else
...
-            ${ALIEN_RUNTIME_ROOT:+alien monalisa}
-            ${ENABLE_COCOA:+builtin_freetype}"
-  NO_FEATURES="${FREETYPE_ROOT:+builtin_freetype} root7"
+            ${ALIEN_RUNTIME_ROOT:+alien monalisa}"
+  NO_FEATURES="root7"
+
+  if [[ "$ENABLE_COCOA" != "" ]]; then
+    FEATURES="$FEATURES builtin_freetype"
+  elif [[ "$FREETYPE_ROOT" != "" ]]; then
+    NO_FEATURES="$FEATURES builtin_freetype"
```
dberzano: This should probably read `NO_FEATURES="$NO_FEATURES builtin_freetype"`.

root.sh
```diff
@@ -113,9 +113,14 @@ else
         -DCMAKE_PREFIX_PATH="$FREETYPE_ROOT;$SYS_OPENSSL_ROOT;$GSL_ROOT;$ALIEN_RUNTIME_ROOT;$PYTHON_ROOT;$PYTHON_MODULES_ROOT"
   FEATURES="builtin_pcre mathmore xml ssl opengl minuit2 http
             pythia6 roofit soversion vdt ${CXX11:+cxx11} ${CXX14:+cxx14} ${XROOTD_ROOT:+xrootd}
-            ${ALIEN_RUNTIME_ROOT:+alien monalisa}
-            ${ENABLE_COCOA:+builtin_freetype}"
-  NO_FEATURES="${FREETYPE_ROOT:+builtin_freetype} root7"
+            ${ALIEN_RUNTIME_ROOT:+alien monalisa}"
+  NO_FEATURES="root7"
+
+  if [[ "$ENABLE_COCOA" != "" ]]; then
```
dberzano: This can simply be: `if [[ $ENABLE_COCOA ]]` (quotes are superfluous when using `[[`, even if variable contains spaces or other special characters, Bash does the correct thing internally).

ktf:
The purpose of this is to make sure that:
- If we are on mac, builtin_freetype is used
- If we are on linux and we do have to compile freetype, i.e. FREETYPE_ROOT is set, we do not use builtin_freetype.

IIRC the previous code had "builtin_freetype" in both "NO_FEATURES" and "FEATURES" under certain circumstances.

dberzano:
@ktf yes understood 👍

PR #897, opened 2017-10-18
Bump Configuration to v1.1.0

PascalBoeschoten:
Adding MySQL support and command-line utilities

PR #918, opened 2017-11-01
Add protections and dependencies to configuration.sh, system option for grpc.sh, protobuf to grpc.sh modulefile

PascalBoeschoten:
PR for addressing the comments in #897 

1. in grpc.sh you say you depend on protobuf, but then protobuf is not mentioned in the modulefile below. If someone picks up only grpc he will be in trouble.
2. grpc.sh is not pickable from the system, which would be nice.
3. If the above is done, then you need to protect configuration.sh against it.

I think I've done points 1 and 3, but I'm not sure how to do 2.

### DIFF
--- a/configuration.sh
+++ b/configuration.sh
@@ -9,6 +9,7 @@ requires:
   - grpc
   - Common-O2
   - MySQL
+  - RapidJSON
 build_requires:
   - CMake
 source: https://github.com/AliceO2Group/Configuration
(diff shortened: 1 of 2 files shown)


ktf:
Please fix the linter.

configuration.sh
```diff
@@ -47,8 +47,8 @@ module load BASE/1.0                                                          \\
             ${BOOST_VERSION:+boost/$BOOST_VERSION-$BOOST_REVISION}            \\
             ${GCC_TOOLCHAIN_ROOT:+GCC-Toolchain/$GCC_TOOLCHAIN_VERSION-$GCC_TOOLCHAIN_REVISION} \\
             ${PROTOBUF_VERSION:+protobuf/$PROTOBUF_VERSION-$PROTOBUF_REVISION} \\
-            grpc/$GRPC_VERSION-$GRPC_REVISION                                 \\
-            Common-O2/$COMMON_O2_VERSION-$COMMON_O2_REVISION
+            ${GRPC_VERSION:+grpc/$GRPC_VERSION-$GRPC_REVISION}                \\
+            ${COMMON_O2_VERSION:+Common-O2/$COMMON_O2_VERSION-$COMMON_O2_REVISION}
```
ktf: Common will always built via AliBuild, no?

PascalBoeschoten: I'm not sure. Should I change it back?

ktf: Yes, I think so.

ktf:
Regarding 2, you can have a look at `prefer_system` and `prefer_system_check` [in the reference manual](http://alisw.github.io/alibuild/reference.html).

PascalBoeschoten:
Is the prefer_system stuff OK like this?

ktf:
Thanks, however maybe you should change the title of the PR to reflect the change? I will than squash all your commits and merge.

PR #922, opened 2017-11-08
Lower CMake requirement on SHUTTLE

dberzano:
SHUTTLE nodes only require v2.8.12, let's not build our own.


dberzano:
This has been fully tested on the SHUTTLE test node.

Issue #921, opened 2017-11-08
Use different defaults on user's laptop and centralized builds


dberzano:
Since aliBuild 1.5.0 we have the ability to share builds done with different defaults (as the `defaults-whatever` package is no longer added as a dependency).

This opens the ability to disable, by default, many packages normally unused on user laptops (short for "personal installations") resulting in an installation speedup, 

The proposal is to:

* Move current `defaults-release.sh` to `defaults-prod.sh` and use it on centralized builds (our build servers) by explicitly building with `--defaults prod`
* Use `defaults-release.sh` for laptop installations and disable `DPMJET`, `GEANT3`, `GEANT4`, and `GEANT4_VMC`

Among other things, this would also allow us to do things such as disabling our custom and very outdated `OpenSSL` on user laptops (unfortunately we need to keep it) and always get it from the system.

This issue is meant to discuss that. What do you think?

cc: @brinick @ktf @rbertens

ktf:
adding also @sawenzel . While things like DPMJET can be removed and for sure OpenSSL we can pick from the system in most cases, I am not sure we should go as far as removing Geant3 / Geant4 by default. That said, why not starting by drafting a "defaults-user` which is optimised for the laptop case and advertise that as the "preferred default for laptops"?

ktf:
We could also use ROOT6, rather than ROOT5 and try to make sure that the same default can be used for AliPhysics and O2.

rbertens:
Hi @dberzano ,  I support this proposal as by also providing a --defaults prod, users do not lose any functionality, we just have to make them aware of this. It would be good to do a very quick investigation to see how many use cases there are for DPMJET (likely a very small number) 

@ktf  I _think_ the same holds for Geant3/Geant4, where the use case for personal builds is probably limited to some small scale testing which is usually not carried out by beginners (and Geant* takes up a lot of disk space) 

For root6 one has to keep the daily use of interpreted and not CLING ready macros in mind - changing  that default would need some proper instructions to users of which changes they need to make to their macros

dberzano:
@ktf what we've observed during our tutorial week is that users forget to type options. So if they run `aliBuild build AliPhysics --defaults user`, next time they'll forget. I would use a sensible default working with no options passed.

[...]

PR #924, opened 2017-11-15
Add new defaults for prod and user installation

dberzano:
This is a first implementation of #921. It has to be discussed before merging.

Three new defaults are introduced (thus, for the moment, not affecting the `release` one):

* `prod`: to be used in place of `release` for centralized Run 2 builds
* `user`: to be used on user laptops (has GEANT* and DPMJET disabled)
* `prod-latest`: to be used for continuous integration (for testing AliPhysics PRs against the latest tagged version of AliRoot instead of, or along with, master)

In a future iteration, and after proper discussion, `user` will be renamed to `release`, becoming the default for every installation. Central installations in the meanwhile should be using `prod`.

One more TODO enhancement the usage of ROOT 6 on platforms not supporting ROOT 5 any longer (such as macOS High Sierra).

### DIFF
--- a/defaults-prod-latest.sh
+++ b/defaults-prod-latest.sh
@@ -1,0 +1,23 @@
+package: defaults-prod-latest
+version: v1
+env:
+  CXXFLAGS: "-fPIC -g -O2 -std=c++11"
+  CFLAGS: "-fPIC -g -O2"
+  CMAKE_BUILD_TYPE: "RELWITHDEBINFO"
+overrides:
+  AliRoot:
+    version: v5-09-19
+    tag: v5-09-19
+  AliPhysics:
+    version: v5-09-19-01
+    tag: v5-09-19-01
+---
+# To be used with aliBuild option `--defaults prod-latest`.
+#
+# This is the default set used to produce centralized builds of AliRoot and
+# AliPhysics, with pinpointed versions.
+#
+#   * It is recommended on user laptops to use `--defaults user` instead, which
+#     pinpoints AliRoot and AliPhysics versions.
+#   * Continuous integration tests use this defaults to test AliPhysics changes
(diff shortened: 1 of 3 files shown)


dberzano:
OK nothing changed, I have just rebased. Merging.

defaults-prod-latest.sh
```diff
@@ -0,0 +1,23 @@
+package: defaults-prod-latest
+version: v1
+env:
+  CXXFLAGS: "-fPIC -g -O2 -std=c++11"
+  CFLAGS: "-fPIC -g -O2"
```
matthiasrichter: probably the `-g` should be removed right away

dberzano: Hi @matthiasrichter, since `prod` aims to be the new `release` for central Run 2 builds. Those flags are correct in that context. Please ignore them for O2.

dberzano: ...and since you quote #820: what you say is not true, we have `CXXFLAGS` and `CFLAGS` because _all_ recipes need to work with them. The build type (`CMAKE_BUILD_TYPE`) only applies to CMake projects.

matthiasrichter: Fine for `prod`, however `user` is exposed to users. The flags as they are right now keep on producing problems in that context. Adding new files seems to me the right moment to clean this up. Just my personal take. I'm not using AliRoot anyhow.

dberzano: `user` for now is, again, exposed to Run 2 users, but it hasn't been announced or advertised. The idea is, indeed, to have some convergency for local builds between Run 2 and Run 3. Since #820 to me is a workaround to the for-now non-existing ability to override build flags per package (remember our discussion about overriding `env`?), can you remind what are the problems being produced in case `-g` is used?

PR #926, opened 2017-11-17
Relocate path for dependent boost libs on macOS

dberzano:
boost libraries depending on other boost libraries on macOS need to know where to look for them. We tell them to look in the same directory as the main library which translates to `@loader_path` to the macOS dynamic loader.


dberzano:
@matthiasrichter @ktf @mpuccio this should fix build problems on macOS, including "red" builds in our CI.

PR #927, opened 2017-11-20
Disable AliEn explicitly for DAQ DA builds

dberzano:
AliEn is not needed. We make sure we don't accidentally pick it up if available from the system.


dberzano:
@Barthelemy

Issue #724, opened 2017-05-09
Python-modules failure on Mac


Barthelemy:
On my mac when building O2 I get the following error when building Python-modules : 
```Finished processing dependencies for matplotlib==1.4.3
+ rm -rvf /Users/bvonhall/dev/alice/sw/INSTALLROOT/15178333586bb815d2bebed05b0b005f28f53de1/osx_x86-64/Python-modules/1.0-1/share '/Users/bvonhall/dev/alice/sw/INSTALLROOT/15178333586bb815d2bebed05b0b005f28f53de1/osx_x86-64/Python-modules/1.0-1/lib/python*/test'
/Users/bvonhall/dev/alice/sw/INSTALLROOT/15178333586bb815d2bebed05b0b005f28f53de1/osx_x86-64/Python-modules/1.0-1/share/jupyter/nbextensions/jupyter-js-widgets/extension.js
/Users/bvonhall/dev/alice/sw/INSTALLROOT/15178333586bb815d2bebed05b0b005f28f53de1/osx_x86-64/Python-modules/1.0-1/share/jupyter/nbextensions/jupyter-js-widgets/extension.js.map
/Users/bvonhall/dev/alice/sw/INSTALLROOT/15178333586bb815d2bebed05b0b005f28f53de1/osx_x86-64/Python-modules/1.0-1/share/jupyter/nbextensions/jupyter-js-widgets
/Users/bvonhall/dev/alice/sw/INSTALLROOT/15178333586bb815d2bebed05b0b005f28f53de1/osx_x86-64/Python-modules/1.0-1/share/jupyter/nbextensions
/Users/bvonhall/dev/alice/sw/INSTALLROOT/15178333586bb815d2bebed05b0b005f28f53de1/osx_x86-64/Python-modules/1.0-1/share/jupyter
/Users/bvonhall/dev/alice/sw/INSTALLROOT/15178333586bb815d2bebed05b0b005f28f53de1/osx_x86-64/Python-modules/1.0-1/share/man/man1/ipython.1.gz
/Users/bvonhall/dev/alice/sw/INSTALLROOT/15178333586bb815d2bebed05b0b005f28f53de1/osx_x86-64/Python-modules/1.0-1/share/man/man1
/Users/bvonhall/dev/alice/sw/INSTALLROOT/15178333586bb815d2bebed05b0b005f28f53de1/osx_x86-64/Python-modules/1.0-1/share/man
[...]

Issue #704, opened 2017-05-03
Build fails in building Python@v2.7.10


matthiasrichter:
After updating to most recent master 45f5a408be25287577fc0ce57dee7d7f8e8d4a58, my build on Ubuntu 16.04.1 LTS fails in building Python@v2.7.10 with the error

```bash
+ pip install -U pip
/home/richterm/src/alisw/sw/SPECS/ubuntu1604_x86-64/Python/v2.7.10-1/Python.sh: /home/richterm/src/alisw/sw/INSTALLROOT/506bc692e9aa7d244bdf2315afd018f3e45fef62/ubuntu1604_x86-64/Python/v2.7.10-1/bin/pip: /home/richterm/src/alisw/sw/INSTALLROOT/506bc692e9aa7d244bdf2315afd018f3e45fe: bad interpreter: No such file or directory
```

Command:
```bash
alibuild/aliBuild --defaults o2 build O2
```

ktf:
I guess we can close this.

PR #931, opened 2017-11-23
Fixes for integrating O2 nightlies on CVMFS

dberzano:
Those fixes are necessary for such integration. For CVMFS builds, we pick all packages from our recipes and we don't use "system" packages at all, and this causes some problems: this is one of the points being addressed.

**Note:** this is a WIP at the moment, do not merge!

The only recipe depending on Python should be Python-modules itself.

### DIFF
--- a/agile.sh
+++ b/agile.sh
@@ -7,7 +7,7 @@ requires:
   - boost
   - lhapdf5
   - HepMC
-  - Python
+  - Python-modules
 build_requires:
   - autotools
 ---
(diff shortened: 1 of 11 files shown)


agile.sh
```diff
@@ -7,7 +7,7 @@ requires:
   - boost
   - lhapdf5
   - HepMC
-  - Python
+  - Python-modules
```
matthiasrichter: this reminds me about the issue that the compilation of `Python-modules` does not work on mac, so there it's a system requirement. Could that be reflected in the defaults? #704, #724

dberzano: You have a point. This _is_ a system requirement everywhere but on slc5 and slc6 and should be treated as such.

dberzano:
Let's start with these fixes we have, they are minimizing problems and make O2 effectively usable on lxplus :-)

PR #935, opened 2017-11-24
Generalize G4 recipe; Upgrade sim environment for O2

sawenzel:
This is a complex commit with the goal to bump the versions
of the simulation infrastructure for O2. In order to do so, the
following things are done here:

 * make the Geant4 recipe autodiscover the versions installed
   data files
   (previously they were hard-coded, preventing to override
    the package in defaults)

 * make sure the G4 simulation packages are compiled in c++14 mode
    for O2
   (previously this was not the case as some packages ingore CXXFLAGS
    or override the flags based on some internal logig; the only reliable
    way to set the C++ standard is via the CMAKE_CXX_STANDARD flag; for
    this purpose a CXXSTD environment variable was added to the defaults
    and some recipes)

 * actually bump the version numbers for G3, G4, VMC, VGM for O2
   and use the respective official github resources

This commit does not affect the production environment.


sawenzel:
This PR is necessary for https://github.com/AliceO2Group/AliceO2/pull/684. Please review as soon as possible.

sawenzel:
@ihrivnac : I bumped the versions as discussed (please review). This should also unblock the work on the MT simulation.

sawenzel:
@PMunkes: I just saw that you asked to partly similar changes in # . Sorry for the overlap but I think we can easily modify your PR.

PR #940, opened 2017-11-28
Make O2 depend on Monitoring and Configuration

Barthelemy:
Needed in order for piotr and myself to use the Configuration and Monitoring package within AliceO2.


Barthelemy:
Although it is not ideal for the reasons you pointed out, there are several important points to keep in mind. 
1. we need to be able to use the Configuration and Monitoring from our code in AliceO2 and we need it now
2. most of these dependencies will disappear once the evaluation of Configuration's backend is done and we keep only 1 
3. using plugins is in the plan, as far as I know, but not ready yet and will not be ready in a reasonable time for us
4. we could get out of the problem by moving back to a separate repository `DataSampling` but the code seemed to fit quite well in the AliceO2 repo (and it was going in your direction of having less repos)

Looking at the specific dependencies of Configuration : 
 - ppconsul : trivial, we have many of such dependencies
 - protobuf : already in fairroot recipe
 - mysql : already in root recipe
 - RapidJSON : already in alo recipe
 - grpc : this one is a bit more cumbersome

monitoring : 
 - ApMon-CPP : already used in several recipes

Given that alibuild hides the complexity of dependencies management, I would not say that adding these 2 extra dependencies is excessive.

Please reconsider.

@knopers8 @vascobarroso @awegrzyn @PascalBoeschoten @sy-c @dberzano

dberzano:
Any problem in making this only for `o2-daq`? This would make everybody happy, is it? @ktf

MohammadAlTurany:
Is it really a good idea to add extra dependancies? I remember that we put the plugins as high priority and we finished it months ago to avoid such dependancies.

vascobarroso:
@MohammadAlTurany plugin usage doesn't fundamentally change things because if I want to put the plugin code in AliceO2 (which for me seems a legitimate request) then I still need the dependencies, right ?

MohammadAlTurany:
@vascobarros : Right if the plugins will be build within AliceO2. 
And In any case if this is only needed by the monitoring why should the simulation and reconstruction (Which could run on a laptop or grid) depend on that? 
P.S: In the o2-daq  we remove all simulation stuff because you do not need it!

ktf:
@vascobarroso the way I would do it is to have the plugin reside inside Monitoring / Configuration or in a package which depends on them + FairMQ. Then we can load the plugins as needed, as part of the environment setup. A depedencies-less implementation e.g. command line options for Configuration (which is already there), or a simple "key: value" UDP packets for Monitoring, could be provided for the "vanilla" AliceO2 case.

@Barthelemy what is the actual showstopper for Piotr? There is already a way (actually two, if you count DDS) to configure your devices and the Monitoring part can probably come later (or simply have a mock interface), so I do not see such an urgent need for them, compared to the urgency of getting a DataSampling device up and running. Or are you saying that DataSampling will not be a Device?

@dberzano I would not do it in o2-daq as IMHO components which we plan to expose as plugins should not be part of the main dependency chain, IMHO.

In general, since the plugin API has been there since a while, is there any showstopper to actually have plugins for Monitoring or for Configuration? Are we missing anything which we need to ask @MohammadAlTurany & co for?

vascobarroso:
@ktf @MohammadAlTurany the Monitoring and Configuration libraries should be treated as first class citizens in AliceO2 as any other module. They should be considered as basic functionality available everywhere (from readout to simulation, on the O2 cluster or on a laptop) and not as an optional task that can be disabled for some use cases (like Geant for DAQ). 

It seems that we are reverting the logic here by having the build system put constraints on the technologies used by the O2 software. IMHO it should be the contrary.

MohammadAlTurany:
@vascobarroso I do not remember that we start classifying any first or second class citizens here! We are only concerned about modularity and keeping the dependancy as simple as possible. 

> They should be considered as basic functionality available everywhere (from readout to simulation, on the O2 cluster or on a laptop) and not as an optional task that can be disabled for some use cases (like Geant for DAQ).

First of all this sounds like a classification of first and second class citizens! Are you sure about this statement? we survived the last twenty years in simulation and reconstruction without making them depends on similar libraries.

What is wrong with suggestion made by @ktf:

> the way I would do it is to have the plugin reside inside Monitoring / Configuration or in a package which depends on them + FairMQ. Then we can load the plugins as needed, as part of the environment setup. A depedencies-less implementation e.g. command line options for Configuration (which is already there), or a simple "key: value" UDP packets for Monitoring, could be provided for the "vanilla" AliceO2 case.

and finally:

> It seems that we are reverting the logic here by having the build system put constraints on the technologies used by the O2 software. IMHO it should be the contrary.

The logic is to reduce the dependancies and deliver the functionality needed by all groups in a convenient way

awegrzyn:
@MohammadAlTurany If the goal is to reduce heavy dependencies why `O2HLTCATracking` was added recently to the `AliceO2` dependency list? Apart it's a **fork** of `AliRoot` it weights ~500 MB...

ktf:
@vascobarroso I fully agree that the *interfaces* for monitoring and configuration should be first class citizens in any O2 related thing. I also agree that a barebone implementation for both (e.g. one that uses simple UDP and simple text files) should be part of the core of our software, either hardcoded or provided as a plugin. What I object to is the fact that every backend has to be included in our distribution, rather than being provided as a separate plugin which one might or might not load. Given the plugin mechanism is there, was presented multiple times, and advertised as one of the features of ALFA, I think it's a pity we cannot take advantage of it and have a coherent design. It feels like having Firefox depend on flash, because the flash plugin you don't use needs it.

The Geant4 / Geant3 case are actually different (and so is the `O2HLTCATracking`) because while they are admittedly a pain to have (and for the latter I was against its inclusion, by the way), we have no plugin interface for them, so the only thing we can do is to add them as dependencies and disable when we do not want them. I personally think that for Monitoring and Configuration we can and should do better. Again, I do want to be able to use Configuration and Monitoring, I do not want to depend on every possible backend they support in order to be able to use them.

Notice also that the build system is not putting any constraint here, build system can do both the solution proposed in this PR, and can do what I am proposing. The issue is really if we want to have a plugin based design for Configuration and Monitoring, and if we do how we implement it so that we do not end up loading all the dependencies in any case.

Shall we discuss tomorrow at some point how to go forward?

MohammadAlTurany:
@awegrzyn  maybe because there was no other way to do it but this! and it have to disappear as soon as the tracking get ported to O2.

ktf:
@vascobarroso and myself had a discussion on the topic today. The following is a solution which we both think is reasonable and allows us go forward in the right design direction.

* @Barthelemy should rework this PR, so that by default Monitoring and Configuration have all but the trivial backends removed in the `o2` default. This should allow Piotr to do his work and it will be what is advertised to "casual" users.
* We introduce a separate `O2-full` default with all the backends enabled which experienced users can use.
* @awegrzyn and @PascalBoeschoten will work on integrating the plugin mechanism (the priority of this task has been bumped, if I understand correctly) and once this is ready we will rework dependencies to have something similar to the "cleaner way" we were discussing.

@vascobarroso did I miss anything?

Barthelemy:
@ktf Ok, I am preparing the modifications.

Barthelemy:
@ktf Could you check ?

ktf:
This works for me. I would also disable mysql, but we can do it later, I guess.

vascobarroso:
@ktf your summary is clear and complete, thanks
@ktf @Barthelemy thanks a lot for the fast action

Barthelemy:
Sorry, I messed up the rebase with master, thus the new commit hash for your commit Giulio. 

By the way, we need ROOT compiled with MySQL for the QualityControl. This is why `root.sh` depends on MySQL but in a better way compared to what was in `configuration.sh` : `  - "MySQL:slc7.*"`).

defaults-o2-full.sh
```diff
@@ -0,0 +1,63 @@
+package: defaults-o2-full
+version: v1
+env:
+  CXXFLAGS: "-fPIC -O2 -std=c++14"
+  CFLAGS: "-fPIC -O2"
+  CMAKE_BUILD_TYPE: "RELWITHDEBINFO"
```
sawenzel: please set the `CXXSTD` environment variable as in defaults-o2.sh (just `-std=c++14` is sometimes not enough)

Barthelemy: Done! that was added after my copy of defaults-o2.sh I guess. Thanks for catching it.

ktf:
Having QC depend on MySQL is really a bummer. :-/ It would be great if work could be allocated to use one of the many DBI interfaces out there and abstract away the direct dependency. Anyways, your workaround is enough for me.

Barthelemy:
As a matter of fact, I agree with you. It actually has an interface and it will not compile the classes that use mysql if mysql is not available or ROOT was not compiled with it. However, at the moment, it means that qc will have no repository thus making it useless. 

I will use the workaround for now and created a JIRA ticket to improve the situation (https://alice.its.cern.ch/jira/browse/QC-56).

dberzano:
Since this became quite a hot topic I have decided to wait until today before pitching in. I'll try to sum up my take on it.

First off, it has been decided that Configuration and Monitoring are tightly a part of the O2 ecosystem, and it has been decided that those packages exist as they are, as externals. We can always backtrack on our decisions if appropriate (things change), but this needs to be discussed in the context of WP3 meetings, not this PR.

As for the very essence of the Configuration and Monitoring packages, and their back-ends currently implemented as potentially undesired library dependencies, this is again a discussion that I would move to the context of their respective WPs (given that constructive criticism is paramount in every project).

Speaking as WP3 convener, we are going now in the direction of making the full O2 stack available in the most accessible way to everybody in the collaboration. I welcome the Configuration and Monitoring packages on board, they will be part of the defaults called `o2-prod.sh`, used by our nightly builds. @Barthelemy actually asked me before opening the PR and made a good point about the urgency of such deployment: thanks to it, we can test the full stack more easily on clusters (even off-site) and we can also test what works/does not work in the very Configuration and Monitoring packages. We won't need to choose what urgency is more urgent :-)

Since the changes to `defaults-o2` are instead minimal, the impact on the laptop users is minimal as well. They won't see any potentially unwanted dependency, and still they will be able to use the most simple monitoring interface implemented as well (stdout).

Pending a test I am currently running to spot potential pitfalls, the "full" version of O2 will be available every night on CVMFS with the full set of dependencies.

p.s. we will need to provide an _actual_ MySQL package rather than a simple system dependency as we do now, this is duly noted...

defaults-o2-prod.sh
```diff
@@ -58,6 +58,17 @@ overrides:
...
+    tag: v10.3.3
+    source: https://gitlab.cern.ch/geant4/geant4.git
+  GEANT4_VMC:
+    tag: "v3-5"
+    source: https://github.com/vmc-project/geant4_vmc
+  vgm:
+    tag: "v4-4"
+    source: https://github.com/vmc-project/vgm
+  GEANT3:
+    tag: v2-5
```
sawenzel: Thanks for adding these.

PR #941, opened 2017-11-29
Harden support for spaces in paths

ktf:
While it's reasonable to ask users not to have spaces in the `--work-dir`, they might be not in control of PATH / LD_LIBRARY_PATH / DYLD_LIBRARY_PATH, so we should probably protect ourself at least for those.


ktf:
@dberzano I think we do need to protect against these, as it's not user's decision to have a space in the path in this particular case.

aphecetche:
Hi Giulio,

Sorry for the late answer. 
Actually your change is not enough as it only touches the incremental part. 

A modified version is https://github.com/alisw/alidist/pull/944

dberzano:
I suggest we continue on #944.

PR #943, opened 2017-11-30
Bump Ali* to v5-09-20*

dberzano:


dberzano:
@brinick this is how we change the base AliRoot for AliPhysics daily tags (it is sufficient to switch defaults).

PR #945, opened 2017-11-30
Adding recipe for Google Benchmark framework

aphecetche:
Nothing fancy, a mere copy of the google test recipe, but it works ;-)


aphecetche:
Good question. I must admit I'm not 100% clear about when a module file is/is not needed, but I would argue that for google benchmark we don't need one as it's really a build dependency only ? (and the benchmark lib itself is a static one)

dberzano:
Whatever we produce as RPM, or we deploy on CVMFS, requires a Modulefile. If your recipe is a dependency to another recipe we publish, same story. It's in general safer to have a Modulefile, I'd say.

dberzano:
...as a matter of fact, you also need to load the package locally with `alienv enter`, but if it's only a build dependency, no, you don't need it.

PR #948, opened 2017-12-05
Deactivate ppconsul on Mac

Barthelemy:
It does not compile on Mac at the moment and is used by Configuration for the tests with Consul. If we use consul in the future we need to use a different interface.

### DIFF
--- a/configuration.sh
+++ b/configuration.sh
@@ -9,7 +9,7 @@ requires:
   - grpc
   - Common-O2
   - RapidJSON
-  - Ppconsul
+  - "Ppconsul:(?!osx)"
   - "MySQL:slc.*"
 build_requires:
   - CMake


dberzano:
I see that the future plan is to get rid of Ppconsul in any case (@PascalBoeschoten, @Barthelemy please correct me if I am wrong). This should be done, of course, quite quickly as we aim to reach recipes stability.

This fixes a problem of building O2 with `o2-daq` and `o2-prod`.

Barthelemy:
@dberzano yes, agreed. As soon as the benchmark is done, things will clarify

dberzano:
I just wanted to check with @ktf if the problem he briefly mentioned to me was related to Ppconsul and `o2-prod` defaults. In such a case, this will fix it.

PR #953, opened 2017-12-06
Cannot assume boost is installed with brew

aphecetche:


ktf:
@PascalBoeschoten @awegrzyn @Barthelemy can you please fix the recipes as I suggested in my review comment? Thanks.

ktf:
This:

https://github.com/alisw/alidist/search?utf8=✓&q=BOOST_ROOT&type=

is how we do it elsewhere.

ktf:
Actually, all the DAQ owned packages should probably be fixed

https://github.com/alisw/alidist/search?p=2&q=%22osx+BOOST_ROOT%22&type=&utf8=✓

@Barthelemy

aphecetche:
Hi,

Indeed : 

BOOST_ROOT=${BOOST_ROOT:-$(brew --prefix boost)}

is probably better than total removal like in my original PR ;-) 

Still, brew --prefix boost always return a path, whether or not boost is actually installed or not.

Also, the fact that every and all recipe dealing with boost has to do this trick (correctly or incorrectly) smells a bit, don't you think ? 
Could there be a better way to deal, at alibuild level maybe, with this very well known package in the case the system one is preferred  ?

Barthelemy:
Please see https://github.com/alisw/alidist/pull/956

ktf:
@Barthelemy PR is indeed doing the right thing, as done elsewhere, therefore I am closing this.

PR #954, opened 2017-12-07
Disable HepMC3 due to O2 failed tests

dberzano:


dberzano:
@ktf @preghenella @sawenzel I am testing it to see if it fixes the failing O2 tests.

sawenzel:
Thanks.

dberzano:
By the way @sawenzel is right, we should enable alidist tests to prevent this. Maybe the new PR checker by @brinick addresses this problem too?

dberzano:
Fun fact: this PR does not solve the problem, but I think I have found the culprit. Opening a new one soon.

PR #964, opened 2017-12-12
Fix get-pip.py location on old SLC5

dberzano:
SLC5 curl cannot get it from `bootstrap.pypa.io` due to very old OpenSSL.


dberzano:
Something has been updated on the site and the error from `curl` is:

```
curl: (35) error:1407742E:SSL routines:SSL23_GET_SERVER_HELLO:tlsv1 alert protocol version
```

This has been tested, see [here](https://alijenkins.cern.ch/job/build-any-ib/6570/console).

PR #966, opened 2017-12-12
Add version check to protobuf

dberzano:


dberzano:
@awegrzyn @PascalBoeschoten since Configuration requires at least protobuf 3.0.0, this fix makes aliDoctor behave properly.

PR #974, opened 2017-12-15
Fix Protobuf lookup for Configuration on macOS

lbariogl:
Patch done together with @dberzano


dberzano:
Correctly picks Protobuf from the system on macOS too.

Issue #977, opened 2017-12-30
RapidJSON fails with gcc 7 toolchain


davidrohr:
RapidJSON needs -Wno-error=implicit-fallthrough in CMAKE CXX flags (or a proper fix in the source) in order to run with current GCC

ktf:
@PascalBoeschoten can you please look into this?

PascalBoeschoten:
@davidrohr @ktf made PR: #980

awegrzyn:
The same happens with `grpc`:
```
src/core/lib/support/murmur_hash.c:79:10: error: this statement may fall through [-Werror=implicit-fallthrough=]
src/core/lib/support/murmur_hash.c:81:10: error: this statement may fall through [-Werror=implicit-fallthrough=]
```
What about not treating this warning as error in dependencies that we don't maintain?

ktf:
That would need to be done on a recipe by recipe basis, though. `-Werror=implicit-fallthrough=` is not set by us, but it's a new GCC default. In the end it might actually be easier to move to a newer version of the external which actually fixes the problem.

PascalBoeschoten:
I would agree that in general we should move to newer versions, but sometimes this is non-trivial and a lot harder than adding a compiler flag. At least in the case of 3rd party libraries, I don't think this new GCC warning gives us any benefits.

ktf:
Agreed, but blindly adding a compiler flag everywhere is not something we should do, for the sake of simplicity and maintainability. So if the fastest route is to simply hack one recipe while we prepare for the next version, fine with me, but do not expect global support for this as that is a shortcut, not a proper solution.

In this particular case I am actually surprised that you say that moving to a newer version is complicated, though. Why is that?

PascalBoeschoten:
That was not referring to a particular case, just a general comment.

dberzano:
OK so let's try to bump RapidJSON to see if it fixes the problem upstream.

ktf:
@PascalBoeschoten Ok, agreed then. 

Notice however that the most common case where updating might not be OK is actually ROOT and in that case we do want to see all the possible errors and warnings that a new compiler gives us, because chances are they are genuine issues which we do not want to mask (at least that's my personal experience).

ktf:
closing this as I guess it's not relevant anymore.

PR #980, opened 2018-01-10
Don't treat implicit fallthrough as error with GCC 7+

PascalBoeschoten:
See issue #977

### DIFF
--- a/rapidjson.sh
+++ b/rapidjson.sh
@@ -1,9 +1,9 @@
 package: RapidJSON
 version: 1.1.0
+tag: 4c8fa695f4c1f1c68104ff4c780ab0c6e6ef86ee
 source: https://github.com/alisw/rapidjson
 build_requires:
   - CMake
-tag: 4c8fa695f4c1f1c68104ff4c780ab0c6e6ef86ee
 ---
 #!/bin/sh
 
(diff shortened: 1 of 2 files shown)


rapidjson.sh
```diff
@@ -7,7 +7,13 @@ tag: v1.1.0
...
 #!/bin/sh
 
-cmake $SOURCEDIR                                              \
+GCC_VERSION_MAJOR=$(gcc --version | awk 'NR==1 {print $3}' | cut -d '.' -f1)
+if [ "$GCC_VERSION_MAJOR" -ge "7" ]; then
+  NO_ERROR_FALLTHROUGH="-Wno-error=implicit-fallthrough"
+fi
+
+cmake $SOURCEDIR                                                       \
+      ${NO_ERROR_FALLTHROUGH:+-DCMAKE_CXX_FLAGS=$NO_ERROR_FALLTHROUGH} \
```
ktf: Mmm... Doesn't this override all the CXXFLAGS, including things like -O2 and co?
 
In any case, this was fixed upstream:

https://github.com/Tencent/rapidjson/commit/cba45fe9de6923b858edb0780e257b7257aa4f7b

why don't we simply pick up a newer version? Can you also fix the linter issue, while we are at it? Thanks.

PascalBoeschoten:
I'm not using a newer version because it caused issues with one of the Configuration library's backends. 
After the backend evaluation is done, we'll probably drop that one anyway, so I'd prefer not to spend much time on it.

Thanks for the source-based fix.

PR #973, opened 2017-12-15
Adding gmake dependency for the dim package

matthiasrichter:

PR #984, opened 2018-01-16
Make DIM work without gmake

dberzano:
@matthiasrichter this obsoletes #973 as discussed

### DIFF
--- a/dim.sh
+++ b/dim.sh
@@ -20,6 +20,11 @@ cd $FILE_NAME
 tr -d '\015' < setup.sh > setup2.sh
 mv setup2.sh setup.sh
 
+# DIM installation needs `gmake` (`make` is not sufficient)
+mkdir buildbin
+export PATH="$PWD/buildbin:$PATH"
+ln -nfs $(which make) buildbin/gmake
+
 # Build (needs special environment)
 ( export OS=Linux
   source setup.sh


dberzano:
@Barthelemy any objection?

dim.sh
```diff
@@ -20,6 +20,11 @@ cd $FILE_NAME
 tr -d '\015' < setup.sh > setup2.sh
 mv setup2.sh setup.sh
 
+# DIM installation needs `gmake` (`make` is not sufficient)
```
ktf: Shouldn't we patch DIM and propose the patch upstream?

Barthelemy: We could contact Clara Gaspar indeed. The project is in maintenance but they take patches.

Barthelemy:
It was any ways working on mac and CC7 until now as far as I could test it. The change is ok and does not break anything.

dim.sh
dberzano: That was the original proposal discussed with @matthiasrichter, but we don't know the reaction time and it breaks the recipe for old releases (given that the oldest release probably dates back to years ago...) I actually propose we do both.

dberzano:
@Barthelemy it worked on systems where `gmake -> make`. There are corner cases where the symlink is not available (such as @matthiasrichter's system).

dim.sh
Barthelemy: I think that it makes sense to do both

matthiasrichter:
exactly, I'm working on Ubuntu, `make` is actually `gmake` on such systems because its GNU. And there is no extra gmake package at all.

PR #985, opened 2018-01-16
Set CMAKE_BUILD_TYPE=COVERAGE for lcov dependencies

ktf:
The env stanza is for variables which need to be set in the
build dependencies, not in the recipe itself. Therefore we
make sure that CMAKE_BUILD_TYPE=COVERAGE is done in the lcov
recipe, which will set it to all its dependencies.


ktf:
Again, currently broken. Bypassing.

PR #987, opened 2018-01-18
Bump ROOT to v6-12-04 and remove tests

dberzano:


dberzano:
@MohammadAlTurany fyi
@ihrivnac @sawenzel does this ROOT version contain the fix for the problem you mention in AliceO2Group/AliceO2#703? If so, I suggest we merge.

sawenzel:
If this ROOT version builds correctly on all platforms we should go ahead.

PR #989, opened 2018-01-21
Require Monalisa if AliEn-Runtime is there

ktf:


dberzano:
Do not merge yet, I believe we are missing, in addition, an explicit check on the presence of `monalisa` among the enabled features too. This will avoid similar problems to reoccur... can you add the check? Otherwise I can do it.

dberzano:
OK, as a matter of fact... `monalisa` is checked already. This means that ROOT reports it as enabled, even if it isn't...

dberzano:
I also see from the logfile sent by our user that it exactly breaks there: it surprises me that the build went through on our test machine without this fix. Anyway, this fix seems sufficient, let's approve it.

dberzano:
OK I was able to reproduce it on a fresh installation, and to fix it with this PR. Merging.

PR #990, opened 2018-01-22
Fix Ppconsul modulefile problem on macOS

dberzano:
@mpuccio @Barthelemy @vascobarroso


dberzano:
Ppconsul is disabled on macOS, therefore we should not try to load it.

vascobarroso:
adding @PascalBoeschoten

PR #993, opened 2018-01-23
Update Apache Arrow support

ktf:


ktf:
@pzhristov in case you wanted to experiment with the latest Arrow as well.

PR #997, opened 2018-01-25
Bump ROOT 5 to a GCC7-compat version

dberzano:
Still not to be merged. The aim is to make ROOT 5 builds compatible with GCC v7.


dberzano:
Merging as we use tagged `alidist` for daily builds in any case. This was tested on SLC5, SLC6, CC7, Ubuntu 17.10 and Ubuntu 18.04.

Issue #994, opened 2018-01-24
Problem with ROOT v6-12-04 and AliPhysicsSelection


mfasDa:
In #987 the default ROOT version was moved to v6-12-04. Note that I observed a crash with the AliPhysicsSelection (see below). The crash seems to be related to a new implementation in this ROOT version (RecursiveRemove) and not obviously related to implementation problems on the ALICE side. Affected by this are macOS HighSierra users which try to run local lego train tests for trains including the physics selection. I have reported the problem on the root mailing list, but the reason for the crash is still unexplained. Until solved I recommend to revert the ROOT6 version back to v6-10-08.

```
Stack trace crash:
====================================================
#6  0x0000000000250430 in ?? ()
#7  0x00007f96369b24e5 in TList::FindObject (this=<optimized out>, obj=0x7ffc4505faa0) at /home/markus/alice/sw/SOURCES/ROOT/v6-12-04/v6-12-04/core/cont/src/TList.cxx:614
#8  0x00007f96369a0f15 in THashList::RecursiveRemove (this=0xf1ca70, obj=0x7ffc4505faa0) at /home/markus/alice/sw/SOURCES/ROOT/v6-12-04/v6-12-04/core/cont/src/THashList.cxx:328
#9  0x00007f963688a793 in TROOT::RecursiveRemove (this=0x7f9636d48ec0 <ROOT::Internal::GetROOT1()::alloc>, obj=<optimized out>) at /home/markus/alice/sw/SOURCES/ROOT/v6-12-04/v6-12-04/core/base/src/TROOT.cxx:2440
#10 0x00007f963696a969 in CallRecursiveRemoveIfNeeded (obj=...) at /home/markus/alice/sw/BUILD/215dc709f741d00f751a2ff4247899cc95e9e602/ROOT/include/TROOT.h:387
#11 TObject::~TObject (this=0x7ffc4505faa0, __in_chrg=<optimized out>) at /home/markus/alice/sw/SOURCES/ROOT/v6-12-04/v6-12-04/core/base/src/TObject.cxx:84
#12 0x00007f9636a017a8 in ~TCheckHashRecurveRemoveConsistency (this=0x7ffc4505faa0, __in_chrg=<optimized out>) at /home/markus/alice/sw/SOURCES/ROOT/v6-12-04/v6-12-04/core/meta/src/TCheckHashRecurveRemoveConsistency.h:61
[...]

PR #998, opened 2018-01-26
Roll back ROOT to v6-10-08 for Run2

dberzano:
See #994


dberzano:
Run2 software is currently broken with the latest version of ROOT6. Does not affect Run3 software.

sawenzel:
I am afraid we can't easily. I am already using a couple of the new features.

ktf:
Can we move forward, then? Hoping it fixes the problem?

pzhristov:
Do we have any list of problems with Root v6-12-04?

PR #999, opened 2018-01-26
Improved SWIG support

ktf:
* Move to SWIG `3.0.12`.
* Check for version 3 on system.


ktf:
@sy-c can you check this works for you?

sy-c:
Thanks, that'sfine with me, I did test on a machine with old swig installed, and when no swig installed. alibuild now builds the correct version 3.0.12 in both situations.

(I have a comment about cmake, but I'll keep it in the original discussion as it's not related to this PR)

ktf:
Ok, thanks. I will merge this on monday to avoid the risk of breaking things over the weekend.

sy-c:
thanks!

dberzano:
I am checking it now.

sy-c:
Hello,
after further checks, it seems there's still a problem...

swig works in cmake when invoking alibuild, but not when running afterwards make from the command line (even when having done alienv enter Common-O2/latest...)

with the error
:1: Error: Unable to find 'swig.swg'
:3: Error: Unable to find 'tcl8.swg'

After further investigation, I found out that swig looks for these files in the  wrong directory.
It uses the initial "INSTALL_ROOT" directory with the "big hash"
.../sw/INSTALLROOT/171f2baff60c4f356bfbf18ecf4d15a3030b0894/slc7_x86-64/SWIG/3.0.12-1/share/swig/3.0.12
whereas the base path available at runtime should be .../sw/slc7_x86-64/SWIG/3.0.12-1/share/swig/3.0.12

It works in aliroot because it sets the environment variable SWIG_LIB to the good location.
However, this variable is not available anymore when invoking from the command line.

This should anyway not be done in this way, according to http://www.swig.org/Release/README
...
    4.  The SWIG library can be changed by setting the SWIG_LIB
        environment variable.  However, you really shouldn't
        have to do this.

Could you please provide a fix to this?

I suppose the "configure" command should be given the final directory where it will be executed, and not the intermediate INSTALL_ROOT long-hash build directory path, which can be given in the make install by specifying DESTDIR

When building the swig RPM, I dealt with the relocation using the following:
building:  ./configure --prefix=/usr
installation:  make DESTDIR=${RPM_BUILD_ROOT} install

PR #1000, opened 2018-01-26
Drop gSOAP based monalisa interface

ktf:
Costin says it's not used anywhere, and since it gives
troubles on newer systems we can simply drop it.


dberzano:
Hi, I am not fully convinced we are not using it, I think we still somehow use it on the Grid. @costing why gSOAP (and MonALISA's gSOAP) were there? Why are they no longer needed?

I propose alisw/monalisa-gsoap-client#1 until we are more than 100% sure it's really no longer needed.

Note that since `MonALISA-gSOAP-client` is the only package we use with a branch in `tag:`, we should just not merge that PR until we also sync alidist.

costing:
Ciao Dario,

This is to read data from MonALISA. Last I remember about it from long time ago (the old build system) it was disabled as well. Actually I was surprised to see it showing up again. It was never used in production and can be fully disabled. Let's get rid of it.

Cheers,

.costin

ktf:
Notice also that Redmer seems to imply everything he was doing was ok. Maybe let's mention it at the CB and then remove it after it if no one shouts.

dberzano:
Let me have a quick check at the old build system. When I ported the recipes three years ago, I got it from there. Maybe it was disabled and I just did not notice...

dberzano:
Let me add: we are getting rid of two recipes that "don't build in multicore", this is a great day!

PR #1002, opened 2018-01-27
Provide for protobuf from system in Configuration

dberzano:


dberzano:
@awegrzyn @ktf @shahor02

awegrzyn:
Thanks for improving the recipe.
I think that after tomorrow's WP8 meeting (and TB approval) only selected configuration backend will remain as required. The others can be optional, therefore the recipe most likely will change.

PR #1011, opened 2018-02-04
Explicit disable of unused options, add ALIBUILD_ROOT_CMAKE_ADDOPTS env var for user customisation of ROOT build

adriansev:
To have identical builds it would be best to disable all options that are not used and have overall a tight control on build options. A safeguard is added as an environment variable named ALIBUILD_ROOT_CMAKE_ADDOPTS that, if defined, will add its content to cmake options.
Tested on an "Christmas tree" EL7 installation, no other options beside the intended were picked up.

### DIFF
--- a/root.sh
+++ b/root.sh
@@ -130,6 +130,7 @@ else
         -Dfftw3=OFF                                                                      \
         -Ddcache=OFF                                                                     \
         -Doracle=OFF                                                                     \
+        -Dfitsio=OFF                                                                     \
         ${ALIBUILD_ROOT_CMAKE_ADDOPTS:+${ALIBUILD_ROOT_CMAKE_ADDOPTS}}                   \
         -DCMAKE_PREFIX_PATH="$FREETYPE_ROOT;$SYS_OPENSSL_ROOT;$GSL_ROOT;$ALIEN_RUNTIME_ROOT;$PYTHON_ROOT;$PYTHON_MODULES_ROOT;$LIBPNG_ROOT;$LZMA_ROOT"
   FEATURES="builtin_pcre mathmore xml ssl opengl minuit2 http
--- a/root.sh
+++ b/root.sh
@@ -124,7 +124,6 @@ else
         -Dgfal=OFF                                                                       \
         -Dcastor=OFF                                                                     \
         -Drfio=OFF                                                                       \
-        -Dodbc=OFF                                                                       \
         -Ddcache=OFF                                                                     \
         -Doracle=OFF                                                                     \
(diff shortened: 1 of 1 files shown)


dberzano:
@ktf I insist on surgically fixing _only_ the problem @adriansev had. If the other options aren't causing troubles, let's not change them.

@adriansev the ROOT recipe has ways to check if some features were disabled/enabled _for real_, see the `FEATURES` and `NO_FEATURES` variable. If you disable graphviz, make sure it's also in the `NO_FEATURES` variable 👍

ktf:
@dberzano Fine with me, I did not see your reply and I do not have a particular opinion about this. Notice however that some of them might actually be ticking bombs, e.g. the current hidden dependency on SQLITE.

adriansev:
well, while i will address your requests i think that are a few points worth of debate : 
1. given that i explicit disable things that are not found in cvmfs version i would say that the options are well tested. the point of the patches was to have 1:1 correspondence with cvmfs versions and not have problems like cmake automatically enabling stuff for which it found system installed dependencies 

2. the optimum approach would be actual using -Dminimal and explicit enable what you already have in the cvmfs version. (modulo the dependencies that are picked automatically on build machine and that have to be added explicitly)

So, before i revert the patches could you have time to address the above comments? @dberzano @ktf 
Thank you!

ktf:
My take is that I agree with you @adriansev about the fact we should have explicit dependencies, however I see @dberzano point about doing this somewhat adiabatically. Regarding that I will let you guys figure out the timing on how to proceed.

The only thing I think should really be changed is the fact that you are using an environment variable to inject extra flags. That is a no go for me, at least in its current implementation.

adriansev:
Hi guys! So, in the order : 
1. I reverted the disabled features that might be used (and if dependencies found WILL be compiled) related to some functionality of root
2. I left disabled (but not yet in NO_FEATURES) the grid related features pending your opinion. These features are related to access grid services and if the grid ui is installed on the computer all of those will be enabled and compiled. once again i stress that these are not used at all on cvmfs .. should i still revert them? beside davix (webdav access), bonjour (apple implementation of zeroconf over multicast) and oracle the grid dependencies are : globus, gfal, castor + rfio, dcache .. should i revert them as well?or add them to NO_FEATURES
3. i added gviz to NO_features
What do you think? @dberzano @ktf 
Thank you!

ktf:
This is fine with me. I leave to @dberzano the last word.

adriansev:
hi @dberzano ! any thoughts on this? how should i proceed further?

adriansev:
@dberzano ping? @ktf i have no idea what means "adriansev dismissed ktf’s stale review via e06fb16 8 minutes ago" .. i just tried to rebase the branch on latest master

dberzano:
Hi, I have modified the PR myself by just disabling `gviz`.

PR #1012, opened 2018-02-05
Versions compatible with new FairRoot logger

Barthelemy:


ktf:
Trusting @Barthelemy

Barthelemy:
@ktf thanks :)

PR #1014, opened 2018-02-08
Bump Configuration to v1.3.0

PascalBoeschoten:
This drops the etcd and MySQL back-ends along with their dependencies protobuf & grpc, as well as the dependency on libCommon.

### DIFF
--- a/configuration.sh
+++ b/configuration.sh
@@ -1,16 +1,12 @@
 package: Configuration
 version: "%(tag_basename)s"
-tag:  v1.2.3
+tag:  v1.3.0
 requires:
   - curl
   - boost
   - "GCC-Toolchain:(?!osx)"
-  - protobuf
-  - grpc
-  - Common-O2
   - RapidJSON
   - "Ppconsul:(?!osx)"
-  - "MySQL:slc.*"
 build_requires:
   - CMake
 source: https://github.com/AliceO2Group/Configuration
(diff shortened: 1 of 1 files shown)


PascalBoeschoten:
@awegrzyn

awegrzyn:
This will break Readout, let's wait for https://github.com/AliceO2Group/Readout/pull/2

PascalBoeschoten:
@awegrzyn How come it breaks Readout? It should be backwards compatible, and I don't see CI complaining

awegrzyn:
Because CI does not build `Readout` maybe it should be added.
As `Configuration.h` was delete this will fail:
https://github.com/AliceO2Group/Readout/pull/2/files#diff-440032ce29d313d757b75534e96a33cfR17

awegrzyn:
Also `QualityControl` had to be updated but this was already tested by Barth: https://github.com/AliceO2Group/QualityControl/pull/2

PascalBoeschoten:
Ah OK, I thought we'd already gotten rid of the deps on Configuration.h.

awegrzyn:
Yeah... Ok, so I'd bump `Configuration`, `Monitoring`, `Readout` and `QualityControl` at the same time to avoid breaking anything

awegrzyn:
@ktf This can be merged now. Afterwards QC https://github.com/alisw/alidist/pull/1017 as well.

PR #1021, opened 2018-02-14
Bump GEANT4 to get G4NystromRK4 fix

dberzano:


dberzano:
Patch provided by @ihrivnac and requested by @amorsch and @sawenzel.

dberzano:
Note it will be available along with the next AliRoot tag (v5-09-23).

sawenzel:
@ihrivnac: Is this patch already part of Geant4 v10.3.3 as used by the O2 builds (or do we need to bump this version as well)?

ihrivnac:
The Geant4 v10.3.3 already contains the patch, so there is no need to bump this version.

PR #1029, opened 2018-02-21
O2 PR checks : Fix a problem due to unused but installed files

sawenzel:
Currently, all PR checks of https://github.com/AliceO2Group/AliceO2 fail because the O2 installation directory is not cleaned and it keeps old files from previous installations.

This PR solves the issue by manually making sure the installation directory is emptied before attempting a new incremental build.

Note that this will not affect users who just remake in the build directory without using the incremental build mechanism of AliBuild.

### DIFF
--- a/o2.sh
+++ b/o2.sh
@@ -16,6 +16,11 @@ source: https://github.com/AliceO2Group/AliceO2
 prepend_path:
   ROOT_INCLUDE_PATH: "$O2_ROOT/include"
 incremental_recipe: |
+  # In incremental mode we need to clean the installation area, because we might otherwise
+  # accumulate unused libs/dictionaries which cause 'duplicate symbol' errors in ROOT
+  # leading to failing PR checks
+  find $INSTALLROOT -mindepth 1 -exec rm -r -f {} ';'
+
   make ${JOBS:+-j$JOBS} install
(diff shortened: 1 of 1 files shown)


sawenzel:
The actual build will still be incremental. This PR is not touching the build directories and only makes sure that the installation target is cleared. 

Please go ahead with the alternative solution (which will be slightly more complex to implement), but I think this should be done fast in order to bring back a working PR system for O2 and solving issues

https://github.com/AliceO2Group/AliceO2/issues/870 https://github.com/AliceO2Group/AliceO2/issues/866
https://github.com/AliceO2Group/AliceO2/issues/865
https://github.com/AliceO2Group/AliceO2/issues/863

in the nightlies.

ktf:
For what concerns the errors, I agree they need to be fixed ASAP and I should have just forced a rebuild which should resolve the them.

If that is the case, I would like to wait a bit and discuss a better solution, since I still think it should be a cmake level, not an alidist level, kind of fix. Given we do support people cd-ing in the `BUILD` directory and typing make, that should always work, without having to go back to aliBuild.

To me it should be cmake that notices that dependencies have changed and old files need to be regenerated / removed. @dennisklein will your rewrite take care of issues like this one?

sawenzel:
Closing this PR. Alternative solutions can be discussed elsewhere.

ktf:
Looks like the errors are back. :-/ Can you reopen this please and I will merge it.

PR #1037, opened 2018-03-02
Use own, more stable, libpng mirror

dberzano:
Sourceforce is continuously down.


dberzano:
Travis CI has problems at the moment, but the recipe passes the linter test. Proceeding.

PR #1038, opened 2018-03-02
Get matplotlib from Github and not SourceForge

brinick:


dberzano:
Force-merging as Travis is down...

PR #1046, opened 2018-03-12
Change to flpproto.sh to test codeowners

Barthelemy:


dberzano:
I see something bizarre here. We've required the status:

* **continuous-integration/travis-ci**

to pass, but since this PR comes from another branch on the same repository, Travis CI creates two statuses with a `/pr` or `/push` appended:

* **continuous-integration/travis-ci/pr:** this is because this is a PR
* **continuous-integration/travis-ci/push:** because it's also a push on another repo branch

Can you please try from a fork? Thanks!

Barthelemy:
here is one from a fork : https://github.com/alisw/alidist/pull/1048

dberzano:
OK thanks, as you can see by the way the problem went away... weird.

ktf:
Shouldn't we require only "pr" to pass? push comes from the fact that barth created a branch in the repository (which we should not allow for anyone).

dberzano:
@ktf as said, GitHub gives the ability to select only `continuous-integration/travis-ci` (with no `/pr` or `/push`), and - as a matter of fact - by looking at the other PRs - it does what you want already :-)

Now, I don't know if that means that it only looks for the `/pr` one or for _any_ test starting with `continuous-integration/travis-ci`, or for _all_ of them, but I don't see this granularity in the protected branches configuration page.

ktf:
I think what it does is to require both of them. My point is that in any case for PRs we should not see any ambiguity given that only "/pr" should apply.

dberzano:
OK thanks for clarifying, you are indeed right. In any case, what we could do (probably the only thing we can do) is to disable branch builds in the Travis CI.

(Closing this as we've tested it and it works as expected.)

PR #1042, opened 2018-03-05
Bump DDS version an re-enable building of wn_bin

rbx:
wn_bin was disabled because it did not build on Mac - this is now resolved in the new version.

PR #1054, opened 2018-03-20
Bump DDS version an re-enable building of wn_bin

rbx:
Updated with suggestions from #1042.
Please check if this solution works, as I am not hitting the mentioned issue on any of my test environments..

### DIFF
--- a/dds.sh
+++ b/dds.sh
@@ -1,6 +1,6 @@
 package: DDS
 version: "%(tag_basename)s"
-tag: "1.7"
+tag: "2.0"
 source: https://github.com/FairRootGroup/DDS
 requires:
   - boost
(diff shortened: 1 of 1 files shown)


dds.sh
```diff
@@ -14,6 +14,8 @@ case $ARCHITECTURE in
     [[ ! $BOOST_ROOT ]] && BOOST_ROOT=`brew --prefix boost` ;;
 esac
 
+export DDS_LD_LIBRARY_PATH="${GCC_TOOLCHAIN_ROOT}/lib64"
```
ktf: You should do this only if GCC_TOOLCHAIN_ROOT is defined (e.g. not the case when GCC is not there, for example on mac. Can you please protect it with an if?

rbx: alright, fixed it.

ktf:
@brinick can you verify this builds and merge it? Thanks.

PR #1057, opened 2018-03-21
Fix long shebangs in Python executables

dberzano:
Python installation and setuptools produce by default Python scripts with the
full `python` executable path in the shebang. In some cases, on our installation
directories, that shebang exceeds 128 chars in length (max lenght on most
Linuxes), resulting in a "bad interpreter" error.

Since in our case we always have `python` in our environment, we now replace
every generated shebang with the shorter `/usr/bin/env python`.

We have noticed that this hack is not needed for the Python-modules package.

This complements #907.

### DIFF
--- a/python.sh
+++ b/python.sh
@@ -76,7 +76,13 @@ for U in https://bootstrap.pypa.io/get-pip.py \
   curl -kSsL -o get-pip.py $U || continue && break
 done
 python get-pip.py
-pip install -U pip
+python "$INSTALLROOT/bin/pip" install -U pip
+
+# Remove long shebangs (by default max is 128 chars on Linux)
+pushd "$INSTALLROOT/bin"
+  sed -i.deleteme -e "1 s|^#!${INSTALLROOT}/bin/\(.*\)$|#!/usr/bin/env \1|" * || true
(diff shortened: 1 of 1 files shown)


python.sh
```diff
@@ -76,7 +76,13 @@ for U in https://bootstrap.pypa.io/get-pip.py \
   curl -kSsL -o get-pip.py $U || continue && break
 done
 python get-pip.py
-pip install -U pip
+python "$INSTALLROOT/bin/pip" install -U pip
+
+# Remove long shebangs (by default max is 128 chars on Linux)
+pushd "$INSTALLROOT/bin"
+  sed -i.deleteme -e "1 s|^#!${INSTALLROOT}/bin/\(.*\)$|#!/usr/bin/env \1|" * || true
```
dberzano: Note that this `sed` syntax works both on Linux and macOS.

dberzano:
For the record, all builds (C.I.) succeeded on SLC7, we've observed the problem with [a SLC6 build](https://ali-ci.cern.ch/repo/logs/alisw/AliRoot/557/9ce3f09198f174022b0352b4288e2ed12139f39a/fullLog.txt).

PR #1060, opened 2018-03-22
Disable Python from ROOT 5 builds

dberzano:


dberzano:
ROOT 5 builds require a very old OpenSSL because of AliEn-Runtime. Such OpenSSL version is incompatible with `pip`. This means we cannot install at the same time Python and the ROOT 5 stack. This should fix the PR checks.

PR #1062, opened 2018-03-26
infologger v1.0.5

sy-c:
released infologger v1.0.5


sy-c:
new infologger version released

PR #1063, opened 2018-03-26
new recipe for libinfologger recipe

sy-c:
New recipe for infologger, creating only the client logging library.
It depends on boost and O2-Common.

### DIFF
--- a/libinfologger.sh
+++ b/libinfologger.sh
@@ -1,0 +1,53 @@
+package: libInfoLogger
+version: "%(tag_basename)s"
+tag: v1.0.5
+requires:
+  - boost
+  - "GCC-Toolchain:(?!osx)"
+  - Common-O2
+build_requires:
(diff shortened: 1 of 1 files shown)


awegrzyn:
Either:
- it should renamed to `infologgerlib`
- `libinfologger.sh` added to CODEOWNERS

sy-c:
The name "libinfologger" was proposed by Dario based on conventions already in use. Let's wait his feedback.

dberzano:
Hi @sy-c, @awegrzyn - first off, I am having a look at this now after a workshop week and I'll be on vacation for the rest of the week, so please excuse my slow feedback.

To me `libinfologger.sh` works just fine and it's in line with all the other `lib*.sh` we have - we don't have any `*lib.sh` and I'd keep this convention. But I don't have a strong opinion on that :-)

PR #1068, opened 2018-04-06
Fix Pytha6 recipe

dberzano:


dberzano:
Note: do not merge yet, we need feedback from @sawenzel.

I have removed the double `tar` command (which looked like a way to copy files while preserving permissions, same as `rsync -a`, if I understood correctly). The reason why I have removed it is that it seems that `make install` does the job of copying the files already.

Was there any reason I am overlooking that made the double `tar` necessary?

dberzano:
Note, merging for the moment as this is blocking macOS PR checks.

PR #1069, opened 2018-04-07
Bump FastJet to v3.2.1_1.024-alice3

dberzano:
Fixes an occasional problem with relocation on macOS: alisw/fastjet#2


dberzano:
We have confirmation that this fix works. Linux builds are unaffected.

PR #1080, opened 2018-04-12
Make QualityControl dependent on O2

knopers8:
Needed for https://github.com/AliceO2Group/QualityControl/pull/7

### DIFF
--- a/qualitycontrol.sh
+++ b/qualitycontrol.sh
@@ -10,6 +10,7 @@ requires:
   - DataSampling
   - Monitoring
   - Configuration
+  - O2
 build_requires:
   - CMake
 source: https://github.com/AliceO2Group/QualityControl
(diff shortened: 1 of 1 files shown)


qualitycontrol.sh
```diff
@@ -31,8 +34,10 @@ cmake $SOURCEDIR                                              \
       ${MONITORING_VERSION:+-DMonitoring_ROOT=$MONITORING_ROOT} \
       ${CONFIGURATION_VERSION:+-DConfiguration_ROOT=$CONFIGURATION_ROOT} \
       ${INFOLOGGER_VERSION:+-DInfoLogger_ROOT=$INFOLOGGER_ROOT} \
+      ${O2_VERSION:+-DO2_ROOT=$O2_ROOT}                       \
```
ktf: Why making it optional here and not below?

Barthelemy: Actually most of them should be optional below as well (I guess) or make them non optional here.

ktf: yes, agreed. @knopers8 could you fix that?

knopers8: @ktf Sure. We agreed with @Barthelemy to get rid of all options here. Just give me a while to make sure it doesn't break anything

knopers8: @ktf done

qualitycontrol.sh
```diff
@@ -20,19 +21,23 @@ incremental_recipe: |
 #!/bin/bash -ex
 
 case $ARCHITECTURE in
-    osx*) [[ ! $BOOST_ROOT ]] && BOOST_ROOT=$(brew --prefix boost);;
+  osx*)
+    [[ ! $BOOST_ROOT ]] && BOOST_ROOT=$(brew --prefix boost)
+    [[ ! $GSL_ROOT ]] && GSL_ROOT=`brew --prefix gsl` ;;
```
ktf: `GSL_ROOT` is not `MS_GSL_ROOT`. They are two different libraries (blame Microsoft for the aliasing...).. Is this what you really want????

knopers8: Sorry, I did not know that, I copy-pasted it without enough understanding. Fixed.

ktf:
As a general note you should try to stick to the minimum amount of changes possible when opening the PR and stick to the principle "one PR, one feature / fix". Anyways, now merged.

PR #1081, opened 2018-04-12
Bump GEANT4_VMC to v3-2-p2-alice1

dberzano:
See ALIROOT-7121.


dberzano:
@sawenzel do you agree with this change?

PR #1082, opened 2018-04-12
Remove SWIG from Common deps

awegrzyn:
SWIG was added to `Common` as the TCL wrap file was supposed to be generated via CMake. As it's still not the case I propose to remove it.


awegrzyn:
This is up to @sy-c

sy-c:
The inclusion of swig in cmake was pending the cleanup of cmake to "modern syntax". As this item seems to be postponed to an unknown date, I will investigate if it's possible to do in the current cmake format.

PR #1085, opened 2018-04-13
Fix Protobuf lookup in recent FairRoot

dberzano:
Now FairRoot uses stock FindProtobuf and the lookup mechanism changed slightly.
We explicitly pass a number of variables in case we have our own Protobuf.
Those variables, without any modification required in FairRoot, make sure that
we consistently pick Protobuf from the system only if we don't provide a
Protobuf version ourselves.

### DIFF
--- a/fairroot.sh
+++ b/fairroot.sh
@@ -1,6 +1,6 @@
 package: FairRoot
 version: "%(tag_basename)s"
-tag: "alice-dev-20180215"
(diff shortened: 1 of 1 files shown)


dberzano:
[Better diff (no whitespaces) for clarity](https://github.com/alisw/alidist/pull/1085/files?w=1).

PR #1090, opened 2018-04-23
Bump GEANT4_VMC to tag v3-5-p1 for O2

sawenzel:
Needed for O2 simulation as fixing a problem with empty event handling in Geant4_VMC.


sawenzel:
What is the reason for this revert?

dberzano:
Hi @sawenzel, unfortunately it does not compile, see [here](https://alijenkins.cern.ch/job/DailyBuilds/job/DailyO2/160/console): of course we've merged it before enabling alidist tests...

PR #1091, opened 2018-04-23
Use upstream OpenSSL v1.0.2o

dberzano:


dberzano:
This PR is very important but it needs to be scrutinized carefully. Let's wait for both O2 and AliRoot tests to complete successfully before taking any action.

dberzano:
Closes #826.

PR #1092, opened 2018-04-24
Revert "Bump GEANT4_VMC to tag v3-5-p1 for O2"

dberzano:
Reverts alisw/alidist#1090


dberzano:
Force-merging, this actually fixes a build problem seen [here](https://alijenkins.cern.ch/job/DailyBuilds/job/DailyO2/160/console). With the new alidist checkers, we can catch such problems in advance.

PR #1095, opened 2018-04-24
Add dependency to RapidJSON and googlebenchmark.

aphecetche:
Note that RapidJSON dependency is already there via Configuration btw. I don't know what's the policy wrt dep transitivity though so I restate it here (in case it e.g. disappears from Configuration) 

For info, both dependencies will be needed for O2 PRs to come from MCH for : 
 
- mapping tests and benchmarks (both deps)
- geometry building (RapidJSON)

### DIFF
--- a/o2.sh
+++ b/o2.sh
@@ -11,6 +11,9 @@ requires:
   - Configuration
   - Monitoring
   - ms_gsl
+  - googlebenchmark
+build_requires:
+  - RapidJSON
 source: https://github.com/AliceO2Group/AliceO2
 prepend_path:
   ROOT_INCLUDE_PATH: "$O2_ROOT/include"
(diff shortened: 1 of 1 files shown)


aphecetche:
see also https://github.com/AliceO2Group/AliceO2/pull/1009

awegrzyn:
We foresee dropping RapidJSON in favour of `boost::property_tree::json_parser`

dberzano:
I have nothing against this PR, but could you please @aphecetche tell what you think about using `boost::property_tree::json_parser`?

aphecetche:
@awegrzyn what is the reason for that ? 
@dberzano will first have to see how boost::property_tree::json_parser works and how much it is to convert the RapidJSON-based code to it ...

aphecetche:
@dberzano I was a bit surprised to see a "squash and merge" button available to me, but I hit it nonetheless ;-) as I thought my "powers" were limited to alo stuff...

aphecetche:
@awegrzyn could you please share the rationale behind the idea of moving away from RapidJSON in favour of boost ? thanks.

awegrzyn:
Ah, sorry, I missed your comment. As I took over the project Pascal told me that there was a request to drop `RapidJSON` as there are already other JSON parses available in `alidist`.
So I assume the reasoning was just to get rid of a dependency.

aphecetche:
Ok. I can certainly buy the argument of getting rid of a dependency, as long as the other parser is not (way) slower or more cumbersome to use ;-) Will have a look at some point.

PR #1096, opened 2018-04-24
Bump configuraiton

awegrzyn:


sawenzel:
@awegrzyn: This PR breaks compilation of O2. Could you please revert it for the moment?

awegrzyn:
@sawenzel Actually it is not. The reason O2 is failing is that https://github.com/AliceO2Group/AliceO2/pull/992 is not merged. We've been trying to merge it with Barth since 8am but the build system does not want to pass even though it should. The builds fail more or less randomly. We asked admins to force merge it but no action has been taken.

PR #1099, opened 2018-04-25
Check for the presence of a Modulefile

dberzano:


dberzano:
#1095 was merged without the proper Modulefiles in the recipes, making the package unusable.

PR #1100, opened 2018-04-25
Fix googlebenchmark support in O2

dberzano:
* Make `googlebenchmark` a build requirement: it produces a static library
  therefore it is not needed at runtime
* Add a Modulefile for `googlebenchmark` for completeness

### DIFF
--- a/googlebenchmark.sh
+++ b/googlebenchmark.sh
@@ -6,9 +6,30 @@ build_requires:
  - "GCC-Toolchain:(?!osx)"
  - CMake
 ---
-#!/bin/sh
+#!/bin/bash -e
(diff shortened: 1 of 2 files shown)


dberzano:
@aphecetche can you check?

aphecetche:
@dberzano now (re)compiling... will report asap

dberzano:
So in principle it should work for you if we maintain it as a build requirement. `googlebenchmark` only generates a static library, meaning you won't need the `.a` any longer when running code linked against it (symbols are copied in your executable/library).

Since this works for us (tests are green) I am merging it now (currently, O2 nightlies are broken without it), but feel free to open a new PR in case you see problems!

PR #1103, opened 2018-04-26
Add node recipe

awegrzyn:


awegrzyn:
Failing test is irrelevant to the PR.

node.sh
```diff
@@ -0,0 +1,53 @@
+package: node
+version: "v9.8.0"
+build_requires:
+  - curl
+prefer_system: "(?!slc5)"
+prefer_system_check: |
+  which node || { echo "node is missing"; exit 1; }
```
dberzano: I know it sounds weird but we should not assume `which` exists. Please try to invoke `node` directly, in a way that it returns 0: for instance, `node -h &> /dev/null` instead of `which node` would do (I have checked).

dberzano: I know it sounds weird but we should not assume `which` exists. Please try to invoke `node` directly, in a way that it returns 0: for instance, `node -h &> /dev/null` instead of `which node` would do (I have checked).

node.sh
```diff
@@ -0,0 +1,53 @@
...
+  if [ `$(brew --prefix node)/bin/node --version | tr -d v | awk -F \. {'print substr(0 $1, length($1), length($1) + 1) substr(0 $2, length($2), length($2) + 1)'}` -le 0809 ]; then echo "higher node version required"; exit 1; fi
```
dberzano: We use bash for executing checks, so use `[[` instead of `[`. Alternatively, to the check you have, you could use:

```bash
[[ printf "%04d%04d%04d" $(node --version | sed -e 's/v//g; s/\./ /g') -le 000900090000 ]]
```

but what you've done is OK.

node.sh
```diff
@@ -0,0 +1,53 @@
...
+---
+#!/bin/bash
+
+case $ARCHITECTURE in
```
dberzano: Could you please compact this and use two spaces for indentation?

```bash
case $ARCHITECTURE in
  osx*) NODEOS=darwin ;;
  *)    NODEOS=linux ;;
esac
```

node.sh
```diff
@@ -0,0 +1,53 @@
...
+	;;
+	*)
+		NODEOS=linux
+	;;
+esac
+
+FILE_NAME="$PKGNAME-$PKGVERSION-$NODEOS-x64"
+mkdir ali-node-tmp
+pushd ali-node-tmp
+	TAR_NAME="$FILE_NAME.tar.gz"
```
dberzano: Two spaces for indentation.

node.sh
ktf: How about using the `verge` macro we have in swig.sh for doing the same thing?

node.sh
```diff
@@ -0,0 +1,53 @@
...
+	URL="https://nodejs.org/dist/$PKGVERSION/$TAR_NAME"	
+	curl -O -L $URL
+	tar zxf $TAR_NAME
+popd
+
+/bin/cp -a ./ali-node-tmp/$FILE_NAME/* $INSTALLROOT/.
```
ktf: Why not using rsync?

dberzano: Why `/bin/cp`? Use `cp` and `rm`. Or, actually, here I would use `rsync`.

node.sh
```diff
@@ -0,0 +1,53 @@
...
+		NODEOS=linux
+	;;
+	*)
+		NODEOS=linux
+	;;
+esac
+
+FILE_NAME="$PKGNAME-$PKGVERSION-$NODEOS-x64"
+mkdir ali-node-tmp
+pushd ali-node-tmp
```
ktf: why going through a file, rather than expanding it on the fly?

node.sh
awegrzyn: Sure, is there any documentation how `verge` works?

node.sh
ktf: NODEOS=linux
case $ARCHITECTURE in
  osx*) NODEOS=darwin ;;
esac

node.sh
awegrzyn: Okay, but there are ~20 recipes using `which` including O2 so I assume it must exist.

node.sh
awegrzyn: If I use `[[` the number which start from `0809 ` is treated as octal.

ktf: I would say it's rather obvious:

```
  verge() {
      [  "$1" = "`echo -e "$1\n$2" | sort -V | head -n1`" ]
  }
```

I.e. it checks if the first entry is the same as the one you would get with a natural sort (`sort -V`).

node.sh
dberzano: Sure but don't please :-) The other 20 recipes are "wrong"(-ish). It's in any case better to invoke the command directly if you can.

node.sh
dberzano: That'd be: `curl -L $URL | tar xzf -` (we'd still catch `curl` errors as we have `set -o pipefail`)

node.sh
awegrzyn: ok done

node.sh
awegrzyn: done

awegrzyn: I used what Dario proposed in previous comment

node.sh
awegrzyn: done

awegrzyn:
Is it better now? Shall I delete tar file and unarchived directory after rsync?

node.sh
```diff
@@ -0,0 +1,42 @@
...
+
+NODEOS=linux
+case $ARCHITECTURE in
+  osx*) NODEOS=darwin ;;
+esac
+
+FILE_NAME="$PKGNAME-$PKGVERSION-$NODEOS-x64"
+TAR_NAME="$FILE_NAME.tar.gz"
+URL="https://nodejs.org/dist/$PKGVERSION/$TAR_NAME"
+curl -O -L $URL
```
dberzano: So, indeed, the `curl` + `tar` + `rsync` can become:

```bash
cd "$INSTALLROOT"
curl -L $URL | tar xzf -
```

In practice you unpack directly into the installation dir, and without even needing the intermediate tarball.

dberzano: (So, no cleanup needed.)

awegrzyn: Ok done, but in such case I need to move files after untaring.

node.sh
```diff
@@ -0,0 +1,41 @@
...
+
+NODEOS=linux
+case $ARCHITECTURE in
+  osx*) NODEOS=darwin ;;
+esac
+
+FILE_NAME="$PKGNAME-$PKGVERSION-$NODEOS-x64"
+URL="https://nodejs.org/dist/$PKGVERSION/$FILE_NAME.tar.gz"
+cd "$INSTALLROOT"
+curl -L $URL | tar xzf -
```
ktf: How about:

```bash
curl -L $URL | tar xzf - --strip-components=2 -C $INSTALLROOT
```

ktf: This way you do not need any of the `cd / mv / rm`. On my mac it gives:

```
→ ls -la test-node2
total 288
drwxr-xr-x    9 ktf  staff    288 Apr 27 10:00 .
drwxr-xr-x  168 ktf  staff   5376 Apr 27 10:00 ..
-rw-r--r--    1 ktf  staff  53692 Mar  7 22:31 CHANGELOG.md
-rw-r--r--    1 ktf  staff  60375 Mar  7 22:31 LICENSE
-rw-r--r--    1 ktf  staff  27396 Mar  7 22:31 README.md
drwxr-xr-x    5 ktf  staff    160 Mar  7 22:32 bin
drwxr-xr-x    3 ktf  staff     96 Mar  7 22:31 include
drwxr-xr-x    4 ktf  staff    128 Mar  7 22:31 lib
drwxr-xr-x    5 ktf  staff    160 Mar  7 22:31 share
```

which is the final layout you want, I assume?

awegrzyn: `--strip-components=1` is prefect, is it okay to use it?

awegrzyn: Ok, it works flawlessly on both Linux and Mac

dberzano:
@ktf @awegrzyn I have no strong opinion over `verge` or the implemented solution, maybe using `verge` is more consistent with some other recipes but as said it can be changed in the future! 😄

PR #1110, opened 2018-05-02
Fix check of copyright notice

dstocco:
The sed command on Mac OS cannot properly handle the substitution of newline (\n).
Use tr instead


ktf:
Looks good to me. The failing test is unrelated.

Issue #1112, opened 2018-05-02
Bump FairRoot tag


sawenzel:
We need a new FairRoot tag, containing at least commit 3b01d287cdd.

This is needed in order to merge https://github.com/AliceO2Group/AliceO2/pull/1044.

dennisklein:
Hi,

I suggest to only bump to FairRootGroup/FairRoot@3b01d28, because the next commit splits couple of components into new repos, which will need new recipes. I will propose a PR with the new repos probably beginning of next week.

Cheers

ktf:
Closing in favour of #1114.

sawenzel:
Reopening since this was reverted.

ktf:
Closing in favour of #1150

PR #1114, opened 2018-05-02
Move FairRoot to 3b01d28

ktf:


ktf:
This is to address #1112. I think we should avoid our own tagging of FairRoot, unless we have extra commits on top.

sawenzel:
Can we go on with this?

ktf:
I guess so. I was waiting for @dberzano since I cannot  code review my own PRs, but I will bypass the check now.

PR #1116, opened 2018-05-02
 Avoid spurious warning when not on macOS

ktf:


ktf:
All checks passing for a trivial change. Merging.

PR #1117, opened 2018-05-02
Allow building on top of a generic system view

ktf:
This allows building AliceO2 on top of a system view defined by some
other package manager, e.g. LCGMake, Nix. The variable
ALIBUILD_SYSTEM_VIEW is supposed to point to a directory containing some
incarnation of a system installation which provides externals that
satisfy our prefer_system requirements.


ktf:
@dberzano the check for the modulefiles should recognise recipes that are system requirements and ignore them.

geant4_vmc.sh
```diff
@@ -13,6 +13,8 @@ env:
   G4VMCINSTALL: "$GEANT4_VMC_ROOT"
 ---
 #!/bin/bash -e
+export LD_LIBRARY_PATH=${ALIBUILD_SYSTEM_VIEW:+${ALIBUILD_SYSTEM_VIEW}/lib:}${LD_LIBRARY_PATH}
```
sawenzel: why is this change necessary here (and not in all recipes)? If we have a system view ... I would assume the library environment is handled outside?

vgm.sh
```diff
@@ -13,7 +13,10 @@ cmake "$SOURCEDIR" \
   -DCMAKE_CMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE} \
   -DCMAKE_INSTALL_LIBDIR="lib"                 \
   -DCMAKE_INSTALL_PREFIX="$INSTALLROOT"        \
-  ${CXXSTD:+-DCMAKE_CXX_STANDARD=$CXXSTD}
+  ${CXXSTD:+-DCMAKE_CXX_STANDARD=$CXXSTD}      \
+  -DWITH_EXAMPLES=OFF                          \
+  -DWITH_TESTS=OFF                             \
+  -DCMAKE_INSTALL_PREFIX="$INSTALLROOT"
```
sawenzel: seems unrelated to system views.

geant4_vmc.sh
ktf: The problem comes from the fact that `rootcling`, which comes from our installation, has dependencies on things in a view (e.g. libz), and such a view should not be guaranteed to be in the LD_LIBRARY_PATH at build time. The reason why we do not want the view to be included at build time is that it could provide externals which conflict with the developer environment (typical case is freetype conflicting with various GUIs like emacs) and therefore you want to limit their presence to when you do `alienv enter`.

ktf:
The idea is that this will allow compiling on anything which can provide a "view" like environment where a consistent set of externals are present in a given user specified directory and act as extra "system" externals.

The reason why I think this is better than what you propose are the following:

* So far modulefiles had, by design, only runtime specific information. Going the route you suggest would need to revert that and I do not think is a good idea as it would strongly couple build and runtime environment. This is particularly a bad idea when running on the grid.
* Loading one modulefile at the time will not guarantee that the whole set is consistent, which will eventually cause conflicts and difficult to diagnose errors. Asking the user to use a view like mechanism like `LCG Views` or `nix` will on the other hand result in a always consistent environment, by construction.

sawenzel:
As I said, I think we should discuss this in WP3. ... But views can evidently be implemented using modulfiles plus some soft links and when I say "alienv enter O2" I am already getting a view (or 99% of it) of what is actually needed for O2. In addition,  this is very consistent because one modulefile loads all the others dependent ones.

One more technical question: Would you have to do add "VIEW" changes for any single package that is in a view? (It looks to me that many files need to be touched)
How far away are you to have everything (including DDS, FairROOT, ROOT, Geant4, ... ) in a view so that I actually do not need to build these packages myself?

ktf:
If we can get something like `alienv view O2/latest-build-requires` that would actually be quite cool, I agree.

This development is however somewhat tangential: it's to allow non-aliBuild produced views (e.g. the one provided by SFT via LCG Make, or NixOS / Spack ones) to be integrated as part of the system lookup that aliBuild does.

I will post instructions on how to speed up builds using this as soon as I finish my tests with LCG Views.

ktf:
Closing this, I ended up not liking this approach after all..

PR #1119, opened 2018-05-03
Set O2HLTCATRACKING_ROOT environment variable

sawenzel:
This variable is references in other build recipes. We should set it here for consistency.


sawenzel:
See discussion in PR #1121. Note also that essentially all other modulefiles define this variable (pythia6, pythia8, Vc, ...). [ I would even go as far to say that we should make this a requirement checked by some linting tool. ]

ktf:
See other discussions. This is not how it should be done.

sawenzel:
As a developer, I would like to be able to query this location at runtime. Exactly the same way it is possible for other packages such as arrow.sh maintained by yourself.

ktf:
For the arrow stuff (and other packages we have which can be picked up from the system) we should not expose it to the environment. #1124 should fix it for this particular case, just to show that I think it's wrong and that it should be cleaned up.

PR #1120, opened 2018-05-03
o2.sh: query ARROW_ROOT instead of ARROW_VERSION

sawenzel:
ARROW_ROOT is the actual environment variable used for cmake. It is more consistent and better to query this one instead of ARROW_VERSION (which is not defined by the arrow modulefile).

Same for o2hltcatracking.


dberzano:
Hi @sawenzel, the point is that out CMake should never use environment variables for determining where external software is installed.

Since you've mentioned Arrow, [have a look here](https://github.com/AliceO2Group/AliceO2/blob/dev/cmake/modules/FindArrow.cmake#L25-L27): it actually uses `ARROW_HOME` from the environment, and if not defined, it uses whatever `-DARROW_HOME` you've provided it with.

PR #1121, opened 2018-05-03
Set RAPIDJSON_ROOT in modulefile

sawenzel:


sawenzel:
It is for instance needed when we want to build o2 against a loaded cmvfs enviroment.
Our cmake command line for o2 references this variable. And in any case I think we should put consistent information into modulefiles (it is done for other packages).

F.Y.I. I am currently doing the exercise of compiling o2 against cvmfs nightlies which could accelerate getting started for newcomers.

ktf:
I agree on the feature however I do not think this is not how it should be done. The design decision we had so far is that modulefiles are for runtime only variables and should not be used to expose build time information, which would open a can of worms.

I think the support for build views is a better option for what you want to achieve, because it would ensure consistent environment and keep the above mentioned design decision of not exposing build details in the modulefiles.

ktf:
It is a build detail, given you want to use it for building and it's not needed by the runtime. Giving the false hope that people can rely on `*_ROOT` to point to an external installation will break things whenever the package is picked up from the system and our modulefiles are not loaded.

The design decision so far has been:

* build time requirements and runtime requirements should be separate.
* `modulefiles` only contain runtime requirements coming from alibuild built packages.
* build time dependencies are set by the `env` mechanism of `aliBuild`.
* `*_ROOT` can either come from aliBuild at build time, or be queried from system installation, e.g. via `brew --prefix`.
* `*_VERSION` and `*_REVISION` are only set by aliBuild at build time.

ktf:
I forgot to say that we do have the build environment available, actually, just not in the modulefiles but in the files which are found in `etc/profile.d/init.sh` of each package. The problem of sourcing the correct set of them would still be there, though.

sawenzel:
Exactly. modulefiles and init.sh files should define the same environment. This commit is going one step into this direction. 

Sourcing the correct set is easy: `alienv load everythingneededforo2`

ktf:
No, modulefiles are for runtime dependencies, while init.sh is for build time dependencies. It's important to keep those separated and a mistake to mix the two. If you want to do anything in that direction, you should definitely not go through modulefiles.

What we should probably have is an extension to alienv which sources init.sh, rather the the modulefiles to construct build environments. IMHO that's the correct way to proceed.

sawenzel:
I think this is too much discussion for a simple one line change. Where in the variable name "RAPIDJSON_ROOT" do you see the semantics of a build time dependency? Why is "O2_ROOT", "GEANT4_ROOT" allowed but not "RAPIDJSON_ROOT"? 

I really do not see any danger having a consistent set of variables defined for our packages ... and I think this particular type should be there all the time. At the moment it is not the case consistently.

ktf:
There are two points:

* RapidJSON is one of those things which should have a `prefer_system` option, so we cannot rely on the modulefile (and therefore the variable) being there.
* In general, what you want to achieve using modulefiles is not going to work, because modulefiles will only expose the runtime dependencies. If you really want to do something like that, you should use `etc/profile.d/init.sh` which already defines all those variables and will set them exactly as they are used for a normal `aliBuild build`.
* `GEANT4_ROOT` and `O2_ROOT` are tolerated because in any case we will never pickup those from the system. If a package can be picked up from the system and defines the `_ROOT` in the modulefile, it's a mistake and should be corrected.

sawenzel:
GEANT4 will never be picked from the system? Is that an official decision or is it still discussable?
I was under the impression that we discussed many months ago that alibuild should be able to pickup pre-compiled packages from cvmfs and I believe that that would be a great feature and help to kickstart the work of many people much faster.

Indeed, I can demonstrate that we can pick up GEANT4 from cvmfs and really all I am missing is this little variable here.

ktf:
Well, the decision (which was taken in 2015 when a change of fastjet triggered different results) was that for all the physics related packages we always build them so that we can guarantee what version is being used unless explicit action from the user. Basically we do not silently pick up system versions when we think the package could affect physics results. I do not have a strong preference, but I think it's a more than sensible decision, especially for stuff like ROOT, GEANT4, fastjet.

That said, if we decide to pick GEANT4 from the system, we should drop GEANT4_ROOT as a variable as a system installation in general would not define it, AFAICT.

ktf:
I am also pretty sure that you can pick up GEANT4 from CVMFS by using `etc/profile.d/init.sh`, BTW.

sawenzel:
I don't want to pickup any version of geant4. I would like to be able to say: Load all the environment needed for a compilation of O2 as described by this recipe and pick the stuff from cvmfs.

`alienv enter O2/xxx` almost sets up the environment consistently for this purpose. So you are saying that all we need is another version of this using `init.sh` scripts?

I will close this issue. But I still think there is absolutely nothing wrong with offering this information in the modulefile. I don't think it breaks any system (just because we are in fact no relying on modulefiles).

ktf:
Yes. I agree we need `alienv buildenv O2/latest` and yes, that should use the init scripts, IMHO. In fact, if you sourced the init.sh of O2 you would get exactly that already now, I think, so what you are asking should be just syntactic sugar on top.

sawenzel:
No. The O2 init.sh does not recursively invoke its dependencies. This is why the modulefiles are much better at this stage.

sawenzel:
Another advantage of modulefiles vs init.sh is that I can selectively unload certain modules to actually tell alibuild to rebuild this packages.

ktf:
Sorry, you are wrong. It does contain the whole transitive closure of build time dependencies and the needed build time environment.

Modulefiles on the other hand do not have build time information and will never do because that complicates things on the grid.

```bash
→ cat /cvmfs/alice-nightlies.cern.ch/el7-x86_64/Packages/O2/nightly-20180503-1/etc/profile.d/init.sh
[ "X$ARROW_VERSION" = X ] && source ${WORK_DIR}/slc7_x86-64/arrow/v0.9.0-alice1-1/etc/profile.d/init.sh
[ "X$FAIRROOT_VERSION" = X ] && source ${WORK_DIR}/slc7_x86-64/FairRoot/alice-dev-20180413-3/etc/profile.d/init.sh
[ "X$DDS_VERSION" = X ] && source ${WORK_DIR}/slc7_x86-64/DDS/2.0-5/etc/profile.d/init.sh
[ "X$VC_VERSION" = X ] && source ${WORK_DIR}/slc7_x86-64/Vc/1.3.3-4/etc/profile.d/init.sh
[ "X$HIJING_VERSION" = X ] && source ${WORK_DIR}/slc7_x86-64/hijing/v1.36-4/etc/profile.d/init.sh
[ "X$HEPMC3_VERSION" = X ] && source ${WORK_DIR}/slc7_x86-64/HepMC3/v3.0.0+git_d43693ce0e-1/etc/profile.d/init.sh
[ "X$O2HLTCATRACKING_VERSION" = X ] && source ${WORK_DIR}/slc7_x86-64/O2HLTCATracking/hlt_o2_ca_tracking-v1.7-3/etc/profile.d/init.sh
[ "X$CONFIGURATION_VERSION" = X ] && source ${WORK_DIR}/slc7_x86-64/Configuration/v1.4.1-1/etc/profile.d/init.sh
[ "X$MONITORING_VERSION" = X ] && source ${WORK_DIR}/slc7_x86-64/Monitoring/v1.6.0-1/etc/profile.d/init.sh
[ "X$MS_GSL_VERSION" = X ] && source ${WORK_DIR}/slc7_x86-64/ms_gsl/1-3/etc/profile.d/init.sh
[ "X$RAPIDJSON_VERSION" = X ] && source ${WORK_DIR}/slc7_x86-64/RapidJSON/v1.1.0-alice1-2/etc/profile.d/init.sh
[ "X$GOOGLEBENCHMARK_VERSION" = X ] && source ${WORK_DIR}/slc7_x86-64/googlebenchmark/1.3.0-2/etc/profile.d/init.sh
[ "X$DEFAULTS_RELEASE_VERSION" = X ] && source ${WORK_DIR}/slc7_x86-64/defaults-release/v1-6/etc/profile.d/init.sh
export O2_ROOT=/cvmfs/alice-nightlies.cern.ch/el7-x86_64/Packages/O2/nightly-20180503-1
export O2_VERSION=nightly-20180503
export O2_REVISION=1
export O2_HASH=8bb4e35c540d66deb89ccc80abbae30d00c763cc
export O2_COMMIT=1dd08bbf51
export ROOT_INCLUDE_PATH=$O2_ROOT/include:$ROOT_INCLUDE_PATH
export DYLD_LIBRARY_PATH=$O2_ROOT/lib:$DYLD_LIBRARY_PATH
export PATH=$O2_ROOT/bin:$PATH
export LD_LIBRARY_PATH=$O2_ROOT/lib:$LD_LIBRARY_PATH
```

sawenzel:
The init script is completely wrong and does not work. See https://github.com/alisw/alibuild/issues/539.

ktf:
Yes, you are right. Relocation either failed or is not done for that file. Let's follow that up in alisw/ali-bot#564.

One thing we could do which you might like is to generate automatically a modulefile which loads all the builds dependencies and which you could load to have your build environment. As the automatic generation would use the same mechanism as the one which generated init.sh, the information there would be fully in sync and complete. This can be done.

That said, regarding your point about unloading is unfortunately not so easy.

If you have:

```bash
A->B->C
```

you load B, which loads A and then you unload A and recompile A without recompiling B (because the modulefile is already loaded), your build might end up to be broken in ways which are very difficult to debug, e.g. if you change the ABI of A by introducing a new member in a class in A.

I think we should not generally allow it. I.e. if you loaded a package, regardless of the method, the only way to unload a package is to start a new clean shell and do a different load. This can only be allowed for leaf dependencies. In the case above you should still be allowed to do the trick only for B but not for A.

PR #1123, opened 2018-05-03
[WIP] Add QCG recipe

awegrzyn:
Most of the code could be replaced by `install` but unfortunately required options are not supported on macOS.

### DIFF
--- a/qcg.sh
+++ b/qcg.sh
@@ -2,6 +2,7 @@ package: qcg
 version: "v1.0.4"
 requires:
   - node
+  - QualityControl
 build_requires:
   - node
 ---
@@ -27,7 +28,7 @@ proc ModulesHelp { } {
 set version $PKGVERSION-@@PKGREVISION@$PKGHASH@@
 module-whatis "ALICE Modulefile for $PKGNAME $PKGVERSION-@@PKGREVISION@$PKGHASH@@"
 # Dependencies
-module load BASE/1.0 ${NODE_VERSION:+node/$NODE_VERSION-$NODE_REVISION}
+module load BASE/1.0 ${NODE_VERSION:+node/$NODE_VERSION-$NODE_REVISION} ${QUALITYCONTROL_VERSION:+QualityControl/$QUALITYCONTROL_VERSION-$QUALITYCONTROL_REVISION}
 # Our environment
 setenv QCG_ROOT \$::env(BASEDIR)/$PKGNAME/\$version
 prepend-path PATH \$::env(QCG_ROOT)/bin
--- a/qcg.sh
+++ b/qcg.sh
@@ -1,0 +1,34 @@
+package: qcg
+version: "v1.0.4"
+requires:
+  - node
+build_requires:
+  - node
+---
+#!/bin/bash
+
+cd "$INSTALLROOT"
+npm install @aliceo2/qc@$PKGVERSION --only=production --loglevel error --no-save --no-package-lock
+cp "node_modules/@aliceo2/qc/config-default.js" config.js
+mkdir bin;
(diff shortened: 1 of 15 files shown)


awegrzyn:
@Barthelemy I think it's more clear to run node process and tobject2json separately? What do you think?

qcg.sh
```diff
@@ -0,0 +1,34 @@
...
+  - node
+  - QualityControl
+build_requires:
+  - node
+---
+#!/bin/bash
+
+mkdir -p "$INSTALLROOT/bin"
+npm install @aliceo2/qc@$PKGVERSION --only=production --loglevel error --no-save --no-package-lock --prefix $INSTALLROOT
+rsync -a --ignore-existing "$INSTALLROOT/node_modules/@aliceo2/qc/config-default.js" $INSTALLROOT/config.js
```
ktf: cp -f

qcg.sh
```diff
@@ -0,0 +1,34 @@
...
+proc ModulesHelp { } {
+  global version
+  puts stderr "ALICE Modulefile for $PKGNAME $PKGVERSION-@@PKGREVISION@$PKGHASH@@"
+}
+set version $PKGVERSION-@@PKGREVISION@$PKGHASH@@
+module-whatis "ALICE Modulefile for $PKGNAME $PKGVERSION-@@PKGREVISION@$PKGHASH@@"
+# Dependencies
+module load BASE/1.0 ${NODE_VERSION:+node/$NODE_VERSION-$NODE_REVISION} ${QUALITYCONTROL_VERSION:+QualityControl/$QUALITYCONTROL_VERSION-$QUALITYCONTROL_REVISION}
+# Our environment
+setenv QCG_ROOT \$::env(BASEDIR)/$PKGNAME/\$version
```
ktf: Do you really need `QCG_ROOT` to be exported at runtime? Can't you do without?

sawenzel: It cannot hurt to be able to query/see the actual installation path at runtime.

awegrzyn: This time I actually need it as the users need a way of editing config file: `vi $QCG_ROOT/config.js`. 
Is there alternative way of achieving this?

ktf: `vi $QCC_ROOT/config.js` will actually fail whenever `QCG_ROOT` is not readable (e.g. on an actual production installation), so I would not advertise it. For development, why not using a development package and have a symlink to the actual sources?

Barthelemy: in 99% of the cases people will run, and not develop, the qcg on their dev/test system to be able to see what their QC is producing. They will need a quick way to edit the config. `vi $QCG_ROOT/config.js` allows them to do so. As they use and not develop the qcg I do not think that the development package is a solution.

In production, we usually have a different set of instructions meant for the sysadmins where we would advertise the correct way of editing the file in such an environment. 

I am in favour of keeping the QCG_ROOT variable.

awegrzyn: I am removing `$NODE_ROOT` in https://github.com/alisw/alidist/pull/1126 as it may indeed cause troubles when node is picked up from the system, but this recipe will be always built so I see no reason of dropping `QCG_ROOT`.

ktf: Ok for QCG_ROOT. In the end it's our software. +1 also for dropping `$NODE_ROOT`.

qcg.sh
ktf: I thought you wanted to always overwrite config.js in the installation ROOT, but I now understand what you are doing. You want to allow that user modified configurations get overwritten, right? Then the `rsync` is fine, indeed, for the wanted purpose.

 Iam not sure about it being a good idea in general, as I said elsewhere, but in the end it's your call.

awegrzyn:
@alisw/infrastructure-admins Can you please review?

PR #1125, opened 2018-05-03
Add rationale for variables allowed by modulefiles

ktf:


README.md
```diff
@@ -31,6 +31,7 @@ commodo consequat.
 ```
 rsync -a $SOURCEDIR ./
 ```
+- If a package can / could be picked up from the system, do not provide, in the modulefile, any variable which is not also exposed in general by the system installation. E.g. `ROOTSYS` is fine because that kind of a standard for ROOT installations, `GCC_ROOT` is not because GCC in general does not use `GCC_ROOT`.
```
sawenzel: It do not understand this argument. What can possibly go wrong if we (within our ALICE build infrastructure) expose such additional information in the runtime environment? Can you provide a concrete example (that can happen within our use case) where this will lead to a wrong behaviour?

In any case, alibuild internally depends on a variable GCC_ROOT and I strongly believe that the presence of such variables can even help to reason about the environment more easily. I see that GCC_ROOT is defined, hence I can infer it's installation path and that I am not using the system installation.

Secondly, are you sure that ROOTSYS is defined when you say `apt-get install ROOT`?

ktf: The trivial thing which can happen is that someone hardcodes one of such `Load($SOMETHING_ROOT/lib/libsomething.so)` in a macro, and things get broken when `something` needs to be picked up from (e.g.) homebrew or /usr/lib.

I am not sure I understand your point about GCC_ROOT. Where does aliBuild depend on it being defined in a modulefile? What it's true is that, when building using our own GCC, the variable GCC_VERSION and GCC_REVISION are actually always there (that's why we depend on it to generate modulefiles). When building, GCC_ROOT might or might not be there, depending if GCC comes from our build or from the system, and might be rewritten homebrew in case of a macOS installation. Once again, this build time behaviour has nothing to do with our runtime. 

Regarding `ROOTSYS`: look at <https://root.cern.ch/root/htmldoc/guides/users-guide/ROOTUsersGuide.html#the-organization-of-the-root-framework>. So I think we cannot refrain from actually having it in the environment, same for the various GEANT4 data variables. That does not mean it's actually a good idea to have such a verbosely populated environment.

ktf:
I am merging this as this is the current rationale we have been adopting, to remove this from my "pending items". Of course people can then object to the current rationale and we can amend it.

PR #1131, opened 2018-05-07
New recipes for FairMQ

dennisklein:
FairMQ (and FairLogger) have been separated from the FairRoot repository. The proposed new recipes will be needed once you decide to bump your FairRoot version beyond FairRootGroup/FairRoot@3b01d28.

I have only tested on Linux (Fedora 27) so far:

```
$ aliDoctor --defaults o2 FairMQ
(...)
==> The following packages will be picked up from the system:
    
    - Python-modules
    - GCC-Toolchain
    - autotools
    - CMake
    
    If this is not you want, you have to uninstall / unload them.

==> The following packages will be built by aliBuild because they couldn't be picked up from the system:
    
    - ZeroMQ
    - msgpack
    - boost
    - nanomsg
```

### DIFF
--- a/fairlogger.sh
+++ b/fairlogger.sh
@@ -1,0 +1,43 @@
+package: FairLogger
+version: "%(tag_basename)s"
+tag: v1.1.0
+source: https://github.com/FairRootGroup/FairLogger
+build_requires:
+ - CMake
+ - "GCC-Toolchain:(?!osx)"
+incremental_recipe: |
+  cmake --build . --target install ${JOBS:+-- -j$JOBS}
+  mkdir -p $INSTALLROOT/etc/modulefiles && rsync -a --delete etc/modulefiles/ $INSTALLROOT/etc/modulefiles
+---
+mkdir -p $INSTALLROOT
+
+cmake $SOURCEDIR                                                 \
+      ${CXX_COMPILER:+-DCMAKE_CXX_COMPILER=$CXX_COMPILER}        \
+      ${CMAKE_BUILD_TYPE:+-DCMAKE_BUILD_TYPE=$CMAKE_BUILD_TYPE}  \
+      -DCMAKE_INSTALL_PREFIX=$INSTALLROOT                        \
+      -DDISABLE_COLOR=ON                                         \
+      -DCMAKE_INSTALL_LIBDIR=lib
+
+cmake --build . ${JOBS:+-- -j$JOBS}
+ctest ${JOBS:+-j$JOBS}
+cmake --build . --target install ${JOBS:+-- -j$JOBS}
+
+# ModuleFile
+mkdir -p etc/modulefiles
+cat > etc/modulefiles/$PKGNAME <<EoF
+#%Module1.0
+proc ModulesHelp { } {
+  global version
+  puts stderr "ALICE Modulefile for $PKGNAME $PKGVERSION-@@PKGREVISION@$PKGHASH@@"
+}
+set version $PKGVERSION-@@PKGREVISION@$PKGHASH@@
+module-whatis "ALICE Modulefile for $PKGNAME $PKGVERSION-@@PKGREVISION@$PKGHASH@@"
+# Dependencies
+module load BASE/1.0
+# Our environment
+setenv FAIRLOGGER_ROOT \$::env(BASEDIR)/$PKGNAME/\$version
+prepend-path LD_LIBRARY_PATH \$::env(FAIRLOGGER_ROOT)/lib
+$([[ ${ARCHITECTURE:0:3} == osx ]] && echo "prepend-path DYLD_LIBRARY_PATH \$::env(FAIRLOGGER_ROOT)/lib")
+EoF
+MODULEDIR="$INSTALLROOT/etc/modulefiles"
+mkdir -p $MODULEDIR && rsync -a --delete etc/modulefiles/ $MODULEDIR
--- a/fairmq.sh
+++ b/fairmq.sh
@@ -1,0 +1,79 @@
+package: FairMQ
+version: "%(tag_basename)s"
+tag: v1.2.1
+source: https://github.com/FairRootGroup/FairMQ
+requires:
+ - boost
+ - FairLogger
+ - ZeroMQ
+ - nanomsg
+ - msgpack
+ - DDS
+build_requires:
+ - CMake
+ - "GCC-Toolchain:(?!osx)"
+ - googletest
+incremental_recipe: |
+  cmake --build . --target install ${JOBS:+-- -j$JOBS}
+  mkdir -p $INSTALLROOT/etc/modulefiles && rsync -a --delete etc/modulefiles/ $INSTALLROOT/etc/modulefiles
+---
+mkdir -p $INSTALLROOT
+
+case $ARCHITECTURE in
+  osx*)
+    # If we preferred system tools, we need to make sure we can pick them up.
+    [[ ! $BOOST_ROOT ]] && BOOST_ROOT=`brew --prefix boost`
+    [[ ! $ZEROMQ_ROOT ]] && ZEROMQ_ROOT=`brew --prefix zeromq`
+    [[ ! $NANOMSG_ROOT ]] && NANOMSG_ROOT=`brew --prefix nanomsg`
+    [[ ! $MSGPACK_ROOT ]] && MSGPACK_ROOT=`brew --prefix msgpack`
+  ;;
+esac
+
+cmake $SOURCEDIR                                                 \
+      ${CXX_COMPILER:+-DCMAKE_CXX_COMPILER=$CXX_COMPILER}        \
+      ${CMAKE_BUILD_TYPE:+-DCMAKE_BUILD_TYPE=$CMAKE_BUILD_TYPE}  \
+      -DCMAKE_INSTALL_PREFIX=$INSTALLROOT                        \
+      ${GOOGLETEST_ROOT:+-DGTEST_ROOT=$GOOGLETEST_ROOT}          \
+      ${BOOST_ROOT:+-DBOOST_ROOT=$BOOST_ROOT}                    \
+      ${FAIRLOGGER_ROOT:+-DFAIRLOGGER_ROOT=$FAIRLOGGER_ROOT}     \
+      ${ZEROMQ_ROOT:+-DZEROMQ_ROOT=$ZEROMQ_ROOT}                 \
+      ${NANOMSG_ROOT:+-DNANOMSG_ROOT=$NANOMSG_ROOT}              \
+      ${MSGPACK_ROOT:+-DMSGPACK_ROOT=$MSGPACK_ROOT}              \
+      ${DDS_ROOT:+-DDDS_ROOT=$DDS_ROOT}                          \
+      -DDISABLE_COLOR=ON                                         \
+      -DBUILD_DDS_PLUGIN=ON                                      \
+      -DBUILD_NANOMSG_TRANSPORT=ON                               \
+      -DBUILD_EXAMPLES=OFF                                       \
+      -DCMAKE_INSTALL_LIBDIR=lib                                 \
+      -DCMAKE_INSTALL_BINDIR=bin
+
+cmake --build . ${JOBS:+-- -j$JOBS}
+ctest ${JOBS:+-j$JOBS}
+cmake --build . --target install ${JOBS:+-- -j$JOBS}
+
+# ModuleFile
+mkdir -p etc/modulefiles
+cat > etc/modulefiles/$PKGNAME <<EoF
+#%Module1.0
+proc ModulesHelp { } {
+  global version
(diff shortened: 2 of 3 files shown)


fairlogger.sh
```diff
@@ -0,0 +1,46 @@
...
+      ${CXXSTD:+-DCMAKE_CXX_STANDARD=$CXXSTD}                    \
+      -DCMAKE_INSTALL_PREFIX=$INSTALLROOT                        \
+      -DDISABLE_COLOR=ON
+
+cmake --build . ${JOBS:+-- -j$JOBS}
+ctest ${JOBS:+-j$JOBS}
+cmake --build . --target install
+
+# Detect the lib dir, so we do not have to hardcode it into the modulefile
+LIBDIR=$(cmake -LA .. | grep CMAKE_INSTALL_LIBDIR | cut -d "=" -f2)
```
dberzano: I prefer convention over configuration. Can we have libdir set to `lib` and avoid this detection completely?

fairmq.sh
```diff
@@ -0,0 +1,82 @@
...
+      -DDISABLE_COLOR=ON                                         \
+      -DBUILD_DDS_PLUGIN=ON                                      \
+      -DBUILD_NANOMSG_TRANSPORT=ON                               \
+      -DBUILD_EXAMPLES=OFF
+
+cmake --build . ${JOBS:+-- -j$JOBS}
+ctest ${JOBS:+-j$JOBS}
+cmake --build . --target install
+
+# Detect some dirs, so we do not have to hardcode it into the modulefile
```
dberzano: Same here.

fairmq.sh
```diff
@@ -0,0 +1,82 @@
...
+BINDIR=$(cmake -LA .. | grep CMAKE_INSTALL_BINDIR | cut -d "=" -f2)
+LIBDIR=$(cmake -LA .. | grep CMAKE_INSTALL_LIBDIR | cut -d "=" -f2)
+
+#ModuleFile
```
dberzano: `# Modulefile` (mere aesthetics)

fairlogger.sh
```diff
@@ -0,0 +1,46 @@
...
+
+#ModuleFile
```
dberzano: `# Modulefile` (mere aesthetics)

fairlogger.sh
```diff
@@ -0,0 +1,46 @@
...
+mkdir -p $INSTALLROOT
+
+cmake $SOURCEDIR                                                 \
+      ${CXX_COMPILER:+-DCMAKE_CXX_COMPILER=$CXX_COMPILER}        \
+      ${CMAKE_BUILD_TYPE:+-DCMAKE_BUILD_TYPE=$CMAKE_BUILD_TYPE}  \
+      ${CXXSTD:+-DCMAKE_CXX_STANDARD=$CXXSTD}                    \
+      -DCMAKE_INSTALL_PREFIX=$INSTALLROOT                        \
+      -DDISABLE_COLOR=ON
+
+cmake --build . ${JOBS:+-- -j$JOBS}
```
dberzano: I think `make && make install` are much clearer. We are using CMake to generate Makefiles in any case. I am quite neutral about that but for the moment my feeling is that it makes the recipe less readable.

msgpack.sh
```diff
@@ -0,0 +1,40 @@
...
+cmake $SOURCEDIR                                                 \
+      ${C_COMPILER:+-DCMAKE_C_COMPILER=$C_COMPILER}              \
+      ${CXX_COMPILER:+-DCMAKE_CXX_COMPILER=$CXX_COMPILER}        \
+      -DMSGPACK_CXX11=ON                                         \
+      -DMSGPACK_BUILD_TESTS=OFF                                  \
+      -DCMAKE_INSTALL_PREFIX=$INSTALLROOT
+
+cmake --build . --target install ${JOBS:+-- -j$JOBS}
+
+#ModuleFile
```
dberzano: `# Modulefile` (mere aesthetics)

msgpack.sh
```diff
@@ -0,0 +1,40 @@
...
+  global version
+  puts stderr "ALICE Modulefile for $PKGNAME $PKGVERSION-@@PKGREVISION@$PKGHASH@@"
+}
+set version $PKGVERSION-@@PKGREVISION@$PKGHASH@@
+module-whatis "ALICE Modulefile for $PKGNAME $PKGVERSION-@@PKGREVISION@$PKGHASH@@"
+# Dependencies
+module load BASE/1.0
+# Our environment
+setenv MSGPACK_ROOT \$::env(BASEDIR)/$PKGNAME/\$version
+prepend-path LD_LIBRARY_PATH \$::env(MSGPACK_ROOT)/lib
```
dberzano: No executables in `bin`? Can you confirm?

msgpack.sh
```diff
@@ -0,0 +1,40 @@
+package: msgpack
+version: cpp-2.1.5
```
dberzano: Please use:

```
tag: cpp-2.1.5
version: v2.1.5
```

fairlogger.sh
dennisklein: The project uses the https://cmake.org/cmake/help/v3.0/module/GNUInstallDirs.html module to configure the lib dir. It can be different from platform to platform.

dberzano: In other recipes we use `-DCMAKE_INSTALL_LIBDIR` and set it to `lib`, just grep. Similarly, `-DCMAKE_INSTALL_BINDIR=bin` does the same for `bin`. I feel no need for this generality at directories level, honestly: I would like to keep our recipes stick to simple conventions (`lib` and `bin`). This is what I meant with convention over configuration. I am asking again to please change the recipe accordingly (that's what we do elsewhere).

fairlogger.sh
dberzano: (Actually TBH we are also thinking of using ninja so maybe leave it like this 🙂)

dennisklein: So, do you want it changed or not? ;)

msgpack.sh
dberzano: What we can do is, by default, building it with `-DBUILD_NANOMSG_TRANSPORT=OFF` with a switch that enables it if and only if `msgpack` and `nanomsg` are dependencies. I don't recall the reason.

dennisklein: Ok, since I do not know at the moment how to implement such a switch + condition, I'll leave the dependency list as it is for this PR. When we do the change on FairRoot recipe level, someone on O2 side should anyways decide, which FairMQ components are built by default. Another switch to consider is `-DBUILD_DDS_PLUGIN=ON/OFF`.

dberzano:
@dennisklein do you expect further changes? Can I re-review it? 😄

dennisklein:
Yes pls, I hope I implemented everything we discussed.

awegrzyn:
There are not changes to `fairroot.sh` recipe needed?

dberzano:
Just one more question before proceeding. Is there any reason why the install statements are not provided with a `-j` option? You seem to be using that in the incremental version of the recipes.

dberzano:
For the record: I know that `make install` using GNU Makefiles generated with CMake does not run the `install` target in parallel, but I am wondering if there is any harm in adding the `-j`, maybe using Ninja allows us to do that. Do you know if this is the case?

dberzano:
@awegrzyn this is the first step before modifying the FairRoot recipe. This would also require, in parallel, a version bump. I believe @dennisklein wants to factor the creation of the recipes out of the full migration.

Sorry for the many comments - @dennisklein might be interested in [this discussion](https://alice-talk.web.cern.ch/t/ninja-anyone/66) on Ninja, it seems we're finally trying it 😄

dennisklein:
There is one more point to discuss. Are you fine with overwriting `-DCMAKE_CXX_STANDARD` to `14`? This currently gives you a warning, because in FairMQ we support only C++11. Still, there were a lot of arguments made in our group, that this should be configurable, because in the past there were compatibility problems involved.

If you ask me, I believe it should not be overriden in this case - the language level is IMHO a property of the source code and not so much a build time decision (there might very well be a supported range of language levels for a given source file). Also see the libstdc++ authors statement about it [here](https://stackoverflow.com/questions/46746878/is-it-safe-to-link-c17-c14-and-c11-objects).

dennisklein:
Added the `-j` argument.

dberzano:
> Only downside with interactive use is, that it strips colored compiler output, but that's not an issue here.

Noo that's a big bummer 😢 

> [...] in FairMQ we support only C++11

Then the FairMQ recipe should not use `$CXXSTD` but `11` as value, can you fix that? Thanks!

dennisklein:
> Noo that's a big bummer￼

https://github.com/FairRootGroup/FairRoot/blob/master/CMakeLists.txt#L51-L57 I added this a while back to FairRoot, it gives you color, but also, if you log to files... If you find a better solution, I am happy to learn it. Ideally, I would want to detect interactive use in CMake automatically, but I have not found a way to accomplish that.

> Then the FairMQ recipe should not use $CXXSTD but 11 as value, can you fix that? Thanks!

Done (same for `fairlogger.sh`)

fairmq.sh
```diff
@@ -0,0 +1,79 @@
+package: FairMQ
```
dberzano: FairMQ should be only built using the appropriate defaults, have a look at the `valid_defaults:` section in O2 for instance (FairLogger is a mere dependency so don't add it there).

msgpack.sh
```diff
@@ -0,0 +1,41 @@
+package: msgpack
+version: v2.1.5
+tag: cpp-2.1.5
+source: https://github.com/msgpack/msgpack-c
+build_requires:
+ - CMake
+ - "GCC-Toolchain:(?!osx)"
+prefer_system: "(?!slc5)"
+prefer_system_check: |
+  printf "#include <msgpack/version.hpp>\n#define VERSION (MSGPACK_VERSION_MAJOR * 10000) + (MSGPACK_VERSION_MINOR * 100) + MSGPACK_VERSION_REVISION\n#if(VERSION < 20105)\n#error \"msgpack version >= 2.1.5 needed\"\n#endif\nint main(){}" | c++ -I$(brew --prefix msgpack)/include -xc++ -std=c++11 - -o /dev/null
```
dberzano: Hi @dennisklein, so - I have tested it on my macOS High Sierra with `msgpack` installed from brew (v3.0.0) and without. Everything's fine with "our" msgpack, but with brew's I get:

```
DEBUG:FairMQ:FairMQ:pr1131: -- FairMQ 1.2.1 from Fri May 4 21:57:00 2018 +0200
DEBUG:FairMQ:FairMQ:pr1131: CMake Error at cmake/FairMQLib.cmake:278 (find_package):
DEBUG:FairMQ:FairMQ:pr1131:   Could not find a configuration file for package "msgpack" that is
DEBUG:FairMQ:FairMQ:pr1131:   compatible with requested version "2.1.5".
DEBUG:FairMQ:FairMQ:pr1131:
DEBUG:FairMQ:FairMQ:pr1131:   The following configuration files were considered but not accepted:
DEBUG:FairMQ:FairMQ:pr1131:
DEBUG:FairMQ:FairMQ:pr1131:     /usr/local/opt/msgpack/lib/cmake/msgpack/msgpack-config.cmake, version: 3.0.0
DEBUG:FairMQ:FairMQ:pr1131:     /usr/local/lib/cmake/msgpack/msgpack-config.cmake, version: 3.0.0
DEBUG:FairMQ:FairMQ:pr1131:
DEBUG:FairMQ:FairMQ:pr1131: Call Stack (most recent call first):
DEBUG:FairMQ:FairMQ:pr1131:   CMakeLists.txt:51 (find_package2)
```

Can you check?

dennisklein: So, it looks to me that this is intended behaviour by msgpack. They implemented a version check in their CMake package, that fails on major version change. For now, that means FairMQ `1.2.1` only supports msgpack `2.x`. I will look into it, but this will not be resolved in the context of this PR.

dennisklein: FairRootGroup/FairMQ#32

fairmq.sh
dennisklein: Will look into this later or tomorrow

dberzano:
To sum up, we have recipes for FairLogger and FairMQ _however for the moment they won't be picked if building O2_ because we're using a tagged version of FairRoot still containing both components.

Current FairRoot `dev` branch, more precisely starting from FairRootGroup/FairRoot@3b01d28, has to be built with those two components. We need to add support in O2 for making it work. **This is likely to require some work and will be discussed at the next WP3.**

At the moment, the new recipes only work reliably with our own msgpack due to a bug, see FairRootGroup/FairMQ#32: support for more recent msgpack versions has to be added to FairMQ upstream and won't be fixed in this PR. However, if we will jump to the new FairRoot version before this problem is fixed, **we will work it around by requiring an exact msgpack version for the moment**.

dberzano:
And of course thanks @dennisklein for your PR!

PR #1132, opened 2018-05-07
Run Framework tests serially

dberzano:


o2.sh
```diff
@@ -38,7 +38,8 @@ incremental_recipe: |
     export O2_ROOT=$INSTALLROOT
     # Clean up old coverage data
     find . -name "*.gcov" -o -name "*.gcda" -delete
-    CTEST_OUTPUT_ON_FAILURE=1 ctest ${JOBS+-j $JOBS}
+    ctest -E test_Framework --output-on-failure ${JOBS+-j $JOBS}
+    ctest -R test_Framework --output-on-failure
```
sawenzel: looks good to me. (Is the CTEST_OUTPUT_ON_FAILURE no longer needed?)

dberzano: Replaced by the `--output-on-failure` switch which does the same.

PR #1136, opened 2018-05-09
Add Ninja

aphecetche:
Adding the possibility to using the (faster) Ninja build system instead of make.

Note that for the moment I did not make it the default, so to try it one must do e.g. : 

CMAKE_GENERATOR=Ninja aliBuild --defaults o2 build O2

I only tested on Mac so far.

### DIFF
--- a/ninja.sh
+++ b/ninja.sh
@@ -1,0 +1,36 @@
+package: ninja
+version: "1.8.2"
+tag: "v1.8.2"
+source: https://github.com/ninja-build/ninja
+build_requires:
+ - "GCC-Toolchain:(?!osx)"
+prefer_system: .*
+prefer_system_check: |
+  which ninja && case `ninja --version | sed -e 's/.* //' | cut -d. -f1,2,3 | head -n1` in "^(0|2)\.|^1\.([0-7]|[9-99])\.\d*|^1\.8\.[0-1]" ) exit 1 ;; esac
+---
+#!/bin/bash -e
+
+echo "Building ALICE Ninja. To avoid this install Ninja >= 1.8.2"
+
+$SOURCEDIR/configure.py --bootstrap
+mkdir -p $INSTALLROOT/bin
+cp ./ninja $INSTALLROOT/bin
+
+# Modulefile
+mkdir -p etc/modulefiles
+cat > etc/modulefiles/$PKGNAME <<EoF
+#%Module1.0
+proc ModulesHelp { } {
+  global version
+  puts stderr "ALICE Modulefile for $PKGNAME $PKGVERSION-@@PKGREVISION@$PKGHASH@@"
+}
+set version $PKGVERSION-@@PKGREVISION@$PKGHASH@@
+module-whatis "ALICE Modulefile for $PKGNAME $PKGVERSION-@@PKGREVISION@$PKGHASH@@"
+# Dependencies
+module load BASE/1.0
+# Our environment
+setenv NINJA_ROOT \$::env(BASEDIR)/$PKGNAME/\$version
+prepend-path PATH \$::env(NINJA_ROOT)/bin
+EoF
+
+mkdir -p $INSTALLROOT/etc/modulefiles && rsync -a --delete etc/modulefiles/ $INSTALLROOT/etc/modulefiles
--- a/ninja.sh
+++ b/ninja.sh
@@ -6,11 +6,9 @@ build_requires:
  - "GCC-Toolchain:(?!osx)"
 prefer_system: .*
 prefer_system_check: |
-  which ninja && case `ninja --version | sed -e 's/.* //' | cut -d. -f1,2,3 | head -n1` in "^(0|2)\.|^1\.([0-7]|[9-99])\.\d*|^1\.8\.[0-1]" ) exit 1 ;; esac
+  ninja --version > /dev/null; if test $? = 127; then exit 1; else case `ninja --version | sed -e 's/.* //' | cut -d. -f1,2,3 | head -n1` in "^(0|2)\.|^1\.([0-7]|[9-99])\.\d*|^1\.8\.[0-1]" ) exit 1 ;; esac; fi; exit 0
 ---
 #!/bin/bash -e
-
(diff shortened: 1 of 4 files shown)


ninja.sh
```diff
@@ -0,0 +1,36 @@
+package: ninja
+version: "1.8.2"
+tag: "v1.8.2"
+source: https://github.com/ninja-build/ninja
+build_requires:
+ - "GCC-Toolchain:(?!osx)"
+prefer_system: .*
+prefer_system_check: |
+  which ninja && case `ninja --version | sed -e 's/.* //' | cut -d. -f1,2,3 | head -n1` in "^(0|2)\.|^1\.([0-7]|[9-99])\.\d*|^1\.8\.[0-1]" ) exit 1 ;; esac
```
dberzano: Can you use `ninja --help &> /dev/null` and check if the code is 127? (If it is, `ninja`'s not found.) See `curl` for instance. (I know, I know; I've had this discussion plenty of times, but there are good reasons not to use `which` 😉)

ninja.sh
```diff
@@ -0,0 +1,36 @@
...
+---
+#!/bin/bash -e
+
+echo "Building ALICE Ninja. To avoid this install Ninja >= 1.8.2"
```
dberzano: I'd drop this line. I know it's in other places, however if we override versions using defaults, this line won't change.

ninja.sh
```diff
@@ -0,0 +1,36 @@
...
+  global version
+  puts stderr "ALICE Modulefile for $PKGNAME $PKGVERSION-@@PKGREVISION@$PKGHASH@@"
+}
+set version $PKGVERSION-@@PKGREVISION@$PKGHASH@@
+module-whatis "ALICE Modulefile for $PKGNAME $PKGVERSION-@@PKGREVISION@$PKGHASH@@"
+# Dependencies
+module load BASE/1.0
+# Our environment
+setenv NINJA_ROOT \$::env(BASEDIR)/$PKGNAME/\$version
+prepend-path PATH \$::env(NINJA_ROOT)/bin
```
dberzano: No `lib`, can you confirm? (By the way, if this is the case, it's great. It means we can really construct a system from scratch by having ninja+compiler only, and not also the dependencies to run ninja itself...)

o2.sh
```diff
@@ -89,7 +90,9 @@ esac
 if [[ $ALIBUILD_O2_TESTS ]]; then
   CXXFLAGS="${CXXFLAGS} -Werror"
 fi
+
 cmake $SOURCEDIR -DCMAKE_INSTALL_PREFIX=$INSTALLROOT                                        \
+      ${CMAKE_GENERATOR:+-G "$CMAKE_GENERATOR"}                                             \
```
dberzano: 👍 this allows us to set it with some defaults (your example uses an ext. variable which is frowned upon, but it's the same)

o2.sh
```diff
@@ -124,10 +127,11 @@ cmake $SOURCEDIR -DCMAKE_INSTALL_PREFIX=$INSTALLROOT
       ${ARROW_ROOT:+-DARROW_HOME=$ARROW_ROOT}                                               \
       -Dbenchmark_DIR=${GOOGLEBENCHMARK_ROOT}/lib/cmake/benchmark
 
-if [[ $GIT_TAG == master ]]; then
+if [[ $GIT_TAG == master && "${CMAKE_GENERATOR}" == "Unix Makefiles" ]]; then
```
dberzano: OK so I'd just drop entirely the whole `CONTINUE_ON_ERROR` logic while you're at it. It's nonsense to me (and we don't use it).

dberzano:
You know what, we can make it the default in `dev-fairroot` for starters.

ninja.sh
aphecetche: ok, will do. But for consistency, would be better to change the other recipes as well ;-)

ninja.sh
aphecetche: ok

ninja.sh
aphecetche: No lib besides system ones : 

~/alice/sw/osx_x86-64/ninja/latest/bin
❯ otool -L ninja
ninja:
        /usr/lib/libc++.1.dylib (compatibility version 1.0.0, current version 400.9.0)
        /usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1252.50.4)

o2.sh
aphecetche: my intent was, once tested a bit, to put CMAKE_GENERATOR in the `env` part of the default-o2*  
will that be ok or do I miss something ?

o2.sh
aphecetche: ok

o2.sh
dberzano: It's OK

aphecetche:
Just pushed some changes where I added a `defaults-o2-ninja.sh` (and removed the build_requires from o2.sh). I also changed `fairroot.sh` to use `CMAKE_GENERATOR` (if available). 

For `DDS` it's not that simple. Simply switching to Ninja does not work as there is a custom target  (the main one, wn_bin) that is not digested correctly in the ninja case... Could make it work but requires a couple of changes in DDS CMakeLists.txt... Not sure it is worth it : as we are not DDS developers what is the point of switching to Ninja for it ? (well, unless DDS build time is cut by a large fraction with ninja, but I doubt so...)

dberzano:
Many thanks for this PR! I agree with your stance on DDS.

o2.sh
```diff
@@ -62,6 +62,7 @@ valid_defaults:
   - o2-dev-fairroot
   - alo
   - o2-prod
+  - o2-ninja
```
ktf: Sorry for being late. How about overloading this on the dev-fairroot one? Having defaults to test different tools diverges rather quickly, IMHO.

dberzano: That's precisely what has been discussed if you read the conversation. However, at the moment dev-fairroot does not work for other reasons. I'd rather have it fixed and _then_ I agree to have it on dev-fairroot, that's the plan.

PR #1137, opened 2018-05-11
Fix package: in o2-ninja defaults

dberzano:


dberzano:
@aphecetche small fix 🙂

PR #1144, opened 2018-05-16
Add AliTPCCommon to build_requires

aphecetche:


dberzano:
Resist, if everything's OK next week we'll remove this default and use ninja everywhere 😄

PR #1148, opened 2018-05-17
CMake fix for RHEL6 and removal of obsoleted dependency

davidrohr:


davidrohr:
@dberzano : This fixes the issue that Diego hat, and it removes the now obsolete dependency of O2 on O2HLTTPCCATracking.

PR #1150, opened 2018-05-18
Use a working commit for FairRoot

ktf:
This was already done but reverted citing our convention. Actually the convention states quite clearly:

> Do not create extra branches unless you do need to patch the original sources.

(see https://github.com/alisw/alidist/#guidelines-for-handling-externals-sources). Since this is the case here, I think using the hash is the correct thing to do (and regardless, according to @sawenzel the current tag is not the correct one).

### DIFF
--- a/fairroot.sh
+++ b/fairroot.sh
@@ -1,6 +1,6 @@
 package: FairRoot
-version: "%(tag_basename)s"
-tag: "3b01d28"
+version: "%(short_hash)s"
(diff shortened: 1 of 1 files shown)


dberzano:
As agreed during WP3, we are going to use hashes and `%(short_hash)s` as version number (not the long one).

ktf:
This was checked to be working in previous round of test. The merge did not change the diff.

PR #1153, opened 2018-05-21
Explicit Python3 support in ROOT, add compile_commands to AliRoot and AliPhysics

mpuccio:
This adds the generation of the compile_commands to AliRoot and AliPhysics (just like it is done in O2).
Moreover I added a check on the python3 support in ROOT, to be sure that if PYTHON3 is available the proper option is enabled.

Cheers,
Max


mpuccio:
Closing to split the PR as requested privately by @dberzano

PR #1154, opened 2018-05-22
Fix o2checkcode not guessing O2 source path

dberzano:


dberzano:
@ktf @sawenzel I had to add a modulefile (a dummy one) to keep the linter happy

ktf:
Merging this. In any case the code checker is broken at the moment so this cannot harm more.

ktf:
Yes, as I said "correlation does not mean causation". ;-) I guess AliTPC came at the same time..

PR #1156, opened 2018-05-23
Do not set MANPATH.

aphecetche:
If MANPATH is set to simply o2 dir it will hide regular paths used by
man, so once that module is loaded a simple `man ls` will fail for
instance, which is pretty annoying.

man is smart enough to look in nearby directories of the directories in
PATH, so `man o2` will work just fine.

@matthiasrichter might want to confirm this though, as I guess he was one of the man pages advocates ;-) )

### DIFF
--- a/o2.sh
+++ b/o2.sh
@@ -159,7 +159,6 @@ module load BASE/1.0 FairRoot/$FAIRROOT_VERSION-$FAIRROOT_REVISION ${DDS_VERSION
 setenv O2_ROOT \$::env(BASEDIR)/$PKGNAME/\$version
 setenv VMCWORKDIR \$::env(O2_ROOT)/share
 prepend-path PATH \$::env(O2_ROOT)/bin
-prepend-path MANPATH \$::env(O2_ROOT)/share/man
 prepend-path LD_LIBRARY_PATH \$::env(O2_ROOT)/lib
 prepend-path ROOT_INCLUDE_PATH \$::env(O2_ROOT)/include
 $([[ ${ARCHITECTURE:0:3} == osx ]] && echo "prepend-path DYLD_LIBRARY_PATH \$::env(O2_ROOT)/lib")


dberzano:
So, if you export:

```bash
export MANPATH=/a/random/path
man ls
```

won't work, whereas:

```bash
export MANPATH=:/a/random/path
man ls
```

will (note the colon). This is explained [in the manpage of man](http://man7.org/linux/man-pages/man1/manpath.1.html) (wow so meta!), have a look at the ENVIRONMENT section.

The point is that environment-modules strips colons (I guess), so you cannot just use `prepend-path` there. You probably need to do (already bash-escaped for our recipes):

```
setenv MANPATH :\$::env(O2_ROOT)/share/man
```

Can you give it a try and update the PR please? Thanks!

aphecetche:
Have not seen the colon trick in the Mac man's manpage indeed ;-)
But regardless, my point was : why define the MANPATH at all if it's working just fine without it ?

dberzano:
>  But regardless, my point was : why define the MANPATH at all if it's working just fine without it ?

How's that possible? `man` needs to know where to find the extra manpages. I need to set it in order for it to work.

aphecetche:
well, the man page (on mac at least) explains that man searches "around" the paths in PATH. 

Indeed if you try to put a man page in the expected hierarchy, e.g. $HOME/tmp/share/man/man1/ and put $HOME/tmp/bin in your PATH, then man should be able to find it. Tried on my Mac and on a CentOS7 machine as well.

dberzano:
I see. Then let's drop `MANPATH` if the hierarchy in O2 is correct (is it?)

aphecetche:
It is.

dberzano:
Then I trust your judgement!

matthiasrichter:
I have no strong opinion on that, as long as it works conveniently. Thanks, @aphecetche.
@ktf introduced the man pages, I also like the concept, but we do not use it consistently. On top of it we do not generate the pages correctly. Probably we will have another round soon to improve the situation.

ktf:
Yes, indeed, if it works and I can get the man pages for the o2 devices I've no objections.

PR #1157, opened 2018-05-23
Explicit Python3 support in ROOT

mpuccio:


root.sh
```diff
@@ -60,6 +60,7 @@ fi
 
 # Disable Python for non-ROOT 6 builds
 [[ -d $SOURCEDIR/interpreter/llvm ]] || NO_PYTHON=1
+[[ $(python  'import sys; sys.exit(0 if sys.version_info[0] > 2 else 1)') -eq 0 ]] && [[ ! $NO_PYTHON ]] && IS_PYTHON3=1
```
dberzano: This line is wrong, I believe you meant to write:

```bash
[[ $NO_PYTHON ]] || { [[ $(python 'import sys; print(sys.version_info[0])') < 3 ]] || IS_PYTHON3=1; }
```

Which means:

* Check if we need to _disable_ Python. In case we don't, check Python major version.
* Check if the major version is below 3. In case it isn't (Python 3 case) then set `IS_PYTHON3` to `1`.

dberzano:
While reviewing this PR, an interesting discussion came out about how Bash behaves when using `set -e` and testing for, _i.e._, an empty variable like this:

```bash
set -e
[[ $EMPTY ]]  # where EMPTY is not set
```

* macOS (bash 3.2.57): does not exit
* SLC5 (bash 3.2.25): does not exit
* Ubuntu 16.04 (bash 4.3.48): exits

For starters, macOS has a quite obsolete bash version. In any case, if we want to make our recipes work everywhere, we should do things like:

```bash
[[ $TEST ]] || do_something
```

and not:

```bash
[[ ! $TEST ]] && do_something
```

because the second case would make the recipe exit on some computers depending on the value of `$TEST`. We cannot even do:

```bash
[[ ! $TEST ]] && do_something || true
```

as `|| true` will - correctly - prevent bash from exiting everywhere in case the test returns false, however it also applies if `do_something` fails!

Yes, bash is evil 👿

dberzano:
(Let's merge it after 18h.)

PR #1167, opened 2018-05-28
Add JAliEn

brinick:
First iteration.
I made a separate XRootD for JAlien given that I'm not sure if we can
update and use the existing recipe?


jalien-xrootd.sh
```diff
@@ -0,0 +1,59 @@
+package: JAlien-XRootD
```
dberzano: Do not make a separate package. We have a recipe for that. We can use a different set of defaults for testing and make a single PR out of this one and #1084.

jalien-xrootd.sh
```diff
@@ -0,0 +1,59 @@
...
+ - CMake
+ - "OpenSSL:(?!osx)"
+ - "osx-system-openssl:(osx.*)"
+ - ApMon-CPP
+ - libxml2
+ - "GCC-Toolchain:(?!osx)"
+---
+#! /bin/bash -e
+
+PACKED_NAME="xrootd-$GIT_TAG.tar.gz"
```
dberzano: We build XRootD from Git sources, we don't unpack tarballs. The existing XRootD recipe has proper references already.

jalien.sh
```diff
@@ -0,0 +1,35 @@
...
+version: "%(tag_basename)s"
+tag: master
+source: https://gitlab.cern.ch/jalien/jalien.git
+requires:
+ - Java
+ - JAlien-XRootD
+build_requires:
+ - Java
+---
+#! /bin/bash -e
```
dberzano: Spaces between `#!` and the interpreter [are allowed](https://unix.stackexchange.com/questions/276751/is-space-allowed-between-and-bin-bash-in-shebang) but we don't use them, please remove for consistency.

jalien.sh
```diff
@@ -0,0 +1,35 @@
...
+  global version
+  puts stderr "ALICE Modulefile for $PKGNAME $PKGVERSION-@@PKGREVISION@$PKGHASH@@"
+}
+set version $PKGVERSION-@@PKGREVISION@$PKGHASH@@
+module-whatis "ALICE Modulefile for $PKGNAME $PKGVERSION-@@PKGREVISION@$PKGHASH@@"
+# Dependencies
+module load BASE/1.0 ${JAVA_ROOT:+Java/$JAVA_VERSION-$JAVA_REVISION} ${JALIEN_XROOTD_ROOT:+JAlien-XRootD/$JALIEN_XROOTD_VERSION-$JALIEN_XROOTD_REVISION}
+# Our environment
+setenv JALIEN_ROOT \$::env(BASEDIR)/$PKGNAME/\$version
+prepend-path PATH \$::env(JALIEN_ROOT)
```
dberzano: Hmm this means that JAliEn installs executables _not_ under `bin`? If this is the case, it should be amended (by JAliEn authors). But we can also work it around within the recipe if this is not possible.

java.sh
```diff
@@ -0,0 +1,33 @@
+package: Java
```
dberzano: I'd call it `JDK` not Java for clarity.

java.sh
```diff
@@ -0,0 +1,33 @@
...
+build_requires:
+ - curl
+---
+#! /bin/bash -e
+ZIP_NAME="jdk-10.0.1_linux-x64_bin.tar.gz"
+UNZIP_NAME="jdk-10.0.1"
+URL="http://download.oracle.com/otn-pub/java/jdk/10.0.1+10/fb4372174a714e6b8c52526dc134031e/$ZIP_NAME"
+curl -b "oraclelicense=a" -L -O $URL
+tar xvfz $ZIP_NAME
+rsync -av $UNZIP_NAME/* $INSTALLROOT/
```
dberzano: Do not let the shell expand it:

```
rsync -av $UNZIP_NAME/ $INSTALLROOT/
```

will do.

java.sh
```diff
@@ -0,0 +1,33 @@
+package: Java
+version: "10.0.1"
+build_requires:
+ - curl
+---
+#! /bin/bash -e
+ZIP_NAME="jdk-10.0.1_linux-x64_bin.tar.gz"
```
dberzano: Do not hardcode 10.0.1, you can use `$PKGVERSION`, see [here](http://alisw.github.io/alibuild/reference.html).

java.sh
```diff
@@ -0,0 +1,33 @@
...
+proc ModulesHelp { } {
+  global version
+  puts stderr "ALICE Modulefile for $PKGNAME $PKGVERSION-@@PKGREVISION@$PKGHASH@@"
+}
+set version $PKGVERSION-@@PKGREVISION@$PKGHASH@@
+module-whatis "ALICE Modulefile for $PKGNAME $PKGVERSION-@@PKGREVISION@$PKGHASH@@"
+# Dependencies
+module load BASE/1.0
+# Our environment
+setenv JAVA_ROOT \$::env(BASEDIR)/$PKGNAME/\$version
```
dberzano: I am wondering if we have `CLASSPATH` or something like that to be exported as well?

jalien-xrootd.sh
```diff
@@ -0,0 +1,59 @@
...
+
+# name of the directory once we unzip
+UNPACKED_NAME="xrootd-$GIT_TAG"
+URL="http://xrootd.org/download/v$GIT_TAG/$PACKED_NAME"
+curl -L -O $URL
+tar xfvz $PACKED_NAME
+
+case $ARCHITECTURE in
+   osx*)
+     [ ! "X$OPENSSL_ROOT" = X ] || OPENSSL_ROOT=`brew --prefix openssl`
```
dberzano: Use bash syntax:

```bash
[[ $OPENSSL_ROOT ]] || OPENSSL_ROOT=$(brew --prefix openssl)
```

java.sh
brinick: I was wondering too ;-)

jalien-xrootd.sh
brinick: OK, then this should also be changed in the xrootd.sh recipe as I just ripped it from
there.

dberzano: Just keep the work in a separate PR without merging it so that we don't touch production 🙂

ktf:
@costing @mimarpe can you please have a look at the comments? I think there is a few which are action items on you.

java.sh
costing: No CLASSPATH needed from the JDK. Exporting $PATH and $JAVA_ROOT is enough for this package.

jalien.sh
brinick: Feedback from @costing is that after the compile command,
the only thing that needs to be installed is the alien-users.jar file,
and he seemed to have no preference as to it being put in the
JALIEN_ROOT dir, or a bin subdir

costing: We don't have any executables (yet at least). The command is built by the various services and boils down to something like: 
java -cp alien-users.jar Main

That said, we could organize it in a slightly different way:
- installation target (the only needed file so far): $JALIEN_ROOT/lib/alien-users.jar
- CLASSPATH += $JALIEN_ROOT/lib/alien-users.jar

That would make it even cleaner to run.

costing: JALIEN_ROOT might come handy if we want to build the command line in a different way. Please leave it in for now, we can remove it later if not needed.

Main is just an example / placeholder. The actual entry point will depend on the service, i.e. alien.site.JobAgent, alien.JSh, alien.JCentral ...

ktf: Can't we simply introduce it when actually needed?

costing: Ok, let's try without, CLASSPATH should indeed be enough.

brinick:
@dberzano Can you have another look please?

dberzano:
Let me have a quick look.

jdk.sh
```diff
@@ -0,0 +1,32 @@
...
+ - curl
+---
+#!/bin/bash -e
+ZIP_NAME="jdk-$PKGVERSION_linux-x64_bin.tar.gz"
+UNZIP_NAME="jdk-$PKGVERSION"
+URL="http://download.oracle.com/otn-pub/java/jdk/10.0.1+10/fb4372174a714e6b8c52526dc134031e/$ZIP_NAME"
+curl -b "oraclelicense=a" -L -O $URL
+tar xvfz $ZIP_NAME
+rsync -av $UNZIP_NAME/ $INSTALLROOT/
+export  PATH=$INSTALLROOT/Java/bin:$PATH
```
dberzano: Very good, is it possible to download & unpack directly in the installation directory, without an intermediate tar file and an intermediate unpack directory?

I have tested the following one-liner* (*it's one line plus some boilerplate) and it works:

```bash
mkdir -p "$INSTALLROOT"
curl -b 'oraclelicense=a' -L http://download.oracle.com/otn-pub/java/jdk/10.0.1+10/fb4372174a714e6b8c52526dc134031e/jdk-10.0.1_linux-x64_bin.tar.gz | tar --strip-components 1 -C "$INSTALLROOT" -xvvf -
```

Also, the `export PATH` part is not needed (plus I believe the exported path is wrong).

This recipe is also missing:

* A first line breaking if we are on macOS (`[[ $(uname) != Linux ]] && { echo "Works on Linux only"; false; }`)
* A `prefer_system` to pick the JDK from the system if appropriate (you should check for `javac` I believe, _without using `which`_).

jdk.sh
```diff
@@ -0,0 +1,32 @@
...
+proc ModulesHelp { } {
+  global version
+  puts stderr "ALICE Modulefile for $PKGNAME $PKGVERSION-@@PKGREVISION@$PKGHASH@@"
+}
+set version $PKGVERSION-@@PKGREVISION@$PKGHASH@@
+module-whatis "ALICE Modulefile for $PKGNAME $PKGVERSION-@@PKGREVISION@$PKGHASH@@"
+# Dependencies
+module load BASE/1.0
+# Our environment
+setenv JAVA_ROOT \$::env(BASEDIR)/$PKGNAME/\$version
```
dberzano: As per our (new) policy, use just `set` and not `setenv` (it won't get into the user's environment but it's not needed).

xrootd.sh
```diff
@@ -1,7 +1,7 @@
 package: XRootD
-version: "%(tag_basename)s"
-tag: v3.3.6-alice2
-source: https://github.com/alisw/xrootd.git
+version: "%(short_hash)s"
```
dberzano: You should override `version:`, `tag:`, and `source:` in the `defaults-jalien.sh` file, not here.

defaults-jalien.sh
```diff
@@ -0,0 +1,12 @@
+package: defaults-jalien
+version: v1
+overrides:
+  xrootd:
```
dberzano: Down below you are also setting `tag:` in the `xrootd.sh` recipe, just keep those overrides here and drop the ones in `xrootd.sh`.

jalien.sh
```diff
@@ -0,0 +1,37 @@
...
+ - Java
+valid_defaults:
+ - jalien
+---
+#!/bin/bash -e
+
+rsync -av $SOURCEDIR/ ./
+./compile.sh users
+mkdir -p $INSTALLROOT/lib
+cp alien-users.jar $INSTALLROOT/lib/
```
dberzano: @costing do we still need this distinction between "user" installation and all the rest? Is there a relevant difference in size? Having a single installation would make deployment easier.

@brinick, @costing: how is JAliEn launched? `java -jar alien-users.jar`? If this is the case, I suggest to:

* Keep the `.jar` inside `lib` and the `CLASSPATH` as they are
* Add a dummy `jalien` binary in `bin` (export `PATH` accordingly too) that does:

```bash
#!/bin/bash -e
exec java -jar alien-users.jar "$@"
```

so that one can run it by simply typing `jalien`.

dberzano:
@costing how is JAliEn spelt? Is it jalien, Jalien, JAliEn or JAlien?

mimarpe:
you don't spell it, he spells you :-D

dberzano:
Is the `J` mute?

dberzano:
Oh by the way is it a "he"? This is not very gender inclusive 🏳️‍🌈 😃

mimarpe:
https://ih1.redbubble.net/image.132383933.0910/st%2Csmall%2C215x235-pad%2C210x230%2Cf8f8f8.lite-1u1.jpg

J isn't mute AFAICS...

defaults-jalien.sh
brinick: But the tag setting in xrootd is needed for production isn't it?

xrootd.sh
brinick: I believe I have updated the xrootd recipe to use the official xrootd repo,
hence the new values of tag, version and source. I then override these
in the jalien defaults for @costing particular case.

defaults-jalien.sh
dberzano: So, the one in `xrootd.sh` should not be changed. That one is used for production. You should only override `tag:`, `version:` and `source:` in `defaults-jalien.sh`, that was my point.

brinick:
I believe it/he/she/they is/are spelt JAliEn.
(At least, that's what 👽 told me when they kidnapped me and
took me aboard their space ship, so that's what I'm going with.)

jdk.sh
costing: I'd leave PATH in so we don't have to guess where `java` can be found

defaults-jalien.sh
brinick: I thought we agreed to update to the official repo, and not
use our own fork? Unless you wish to do the xrootd recipe
update in a different PR?

jalien.sh
costing: alien-users.jar is the only needed file. The command line will be composed by whichever service is running, picking one or other entry point in it. Yes, we might have this helper script but not sure it helps much, especially is PATH and CLASSPATH are correctly exposed.

jdk.sh
brinick: The PATH in this case is in the build environment.
I think @dberzano is correct that it is not needed.
You have it set in the runtime environment module:
```
set JAVA_ROOT \$::env(BASEDIR)/$PKGNAME/\$version
prepend-path PATH \$JAVA_ROOT/bin
```

defaults-jalien.sh
brinick: One point however is that @costing requires the xrootd bin and lib dirs
set up at runtime, and I believe that this can only be done by adding
a modulefile to the current xrootd recipe, which has none.

dberzano: So what I am saying is:

* we don't need to set `tag:`, `version:`, `source:` _twice_, we need it once...
* ...and this one place should be _here_, in the `defaults-jalien.sh` configuration
* adding a modulefile to `xrootd.sh` is very fine, I haven't made any "negative" comment about it 🙂

jalien.sh
dberzano: Say, I prefer if we ask our users to run:

```bash
jalien ...
```

rather than:

```bash
java -jar alien-users.jar
```

As a matter of fact: I'd really prefer to run `jalien` like this myself, I find the full command rather obnoxious 🙂

For the record, adding this script is just a matter of doing in the recipe (no changes to JAliEn):

```bash
mkdir $INSTALLROOT/bin
cat > $INSTALLROOT/bin/jalien <<EOF
#!/bin/bash -e
exec java -jar alien-users.jar "$@"
EOF
chmod 0755 $INSTALLROOT/bin/jalien
```

so I really believe we should do it.

dberzano: As discussed yesterday with @costing and @brinick, we drop the creation of the wrapper script and we `rsync` the `bin` directory of the JAliEn repository to `$INSTALLROOT/bin`.

costing: Ok, the "bin/jalien" script is there and a "0.0.0" tag too, to switch to using tags instead of master from now on.

brinick:
@dberzano @costing @mimarpe

dberzano:
Please do not merge yet. Let's wait for the tests.

dberzano:
OK now before approving/merging (I have done some notable modifications, in particular I have completely revamped the JDK version check which was completely wrong), I am running a manual build on Jenkins. If it works, let's deploy (I'll open a separate PR to ali-bot for that) and test.

We're almost there...

dberzano:
For the curious, here's the dependency graph:

schermata 2018-06-19 alle 17 41 57

* **Green** packages are runtime dependencies
* **Purple** packages are build dependencies (not deployed)
* **Gray** arrows indicate a build dependency relation
* **Blue** arrows indicate a runtime dependency relation

dberzano:
Build was successful, now merging alisw/ali-bot#576. When the deployment will report as successful, and all tests are green, we'll merge.

dberzano:
OK, current package does not load properly (says zlib is not found). If you look at the graph above, that's because it's specified as a build dependency and not as a runtime one. This is now fixed, and here's the new plot:

schermata 2018-06-19 alle 18 19 49

It's [rebuilding](https://alijenkins.cern.ch/job/build-any-ib/6774/) and will be redeployed automatically.

dberzano:
@costing @mimarpe enjoy:

```
/cvmfs/alice-nightlies.cern.ch/bin/alienv enter VO_ALICE@JAliEn::0.0.0-2
[JAliEn/0.0.0-2] ~ > jalien
[Grid certificate password: ]
```

Mind the `-2`! When you confirm it works and tests are green (meaning essentially that the changes we've made to the XRootD recipe don't break prod) I'll merge!

Thanks @brinick for this!

mimarpe:
👍😀

costing:
Fantastic! Everything just works! To me it all looks good, can we have the same in the production CVMFS + use tag 1.0.0 for jAliEn?

dberzano:
O2 test failure unrelated. alisw/ali-bot#578 ==> will be published soon. Let's wait for AliRoot, we're almost there.

PR #1168, opened 2018-05-30
Bump to Protobuf 3.5.2 and gRPC 1.9.1

teo:


grpc.sh
```diff
@@ -18,12 +18,16 @@ rsync -a $SOURCEDIR/ ./
...
+env -i HOME="$HOME" LC_CTYPE="${LC_ALL:-${LC_CTYPE:-$LANG}}" PATH="$PATH" USER="$USER" LD_LIBRARY_PATH="$LD_LIBRARY_PATH" make ${JOBS:+-j$JOBS} prefix=$INSTALLROOT
+
+# ldconfig won't work if we're not running make install as root, and that's ok, we don't need it
+sed -i 's/ldconfig/true/' ./Makefile
 make prefix=$INSTALLROOT install
 
 # Add missing symlink. Must be relative, because the directory is moved around after the build
 cd $INSTALLROOT/lib
-ln -s ./libgrpc++.so.1.2.5 ./libgrpc++.so.1
+ln -s ./libgrpc++.so.1.9.1 ./libgrpc++.so.1
```
ktf: How about changing this to be version independent?

teo: It is not clear to me why this is even needed, I only updated it because it was there. The link with the correct soname (`libgrpc++.so.5` as well as `libgrpc++.so`) is already generated.

ktf: Can you check and clean it up?

teo: Yep, I'm on it. I'll update this PR either with a version-independent `ln` or with a removal of that line.

teo:
@ktf I've copied the setting of the `_ROOT` variable from other recipes, in which I've found `setenv` and not `set`. I've tried now with `set` but the module load doesn't seem to work. It works fine with `setenv` though.

dberzano:
@teo `set` sets it for the modulefile only; `setenv` sets it in the environment. If you don't need it in the environment, use `set` only.

teo:
@dberzano I do use it in the Control-OCCPlugin recipe, because I need to tell CMake where to look for gRPC, and it's useful for it to be exported into the environment when the grpc env is loaded, because then the same CMake call that works in the recipe also works when building manually.

teo:
@dberzano @ktf Alright, I've tested with both `set` and `setenv`, and if that's ok with you I would still like to export `GRPC_ROOT` to the environment. I then pass it to Control-OCCPlugin's CMake, to make it easy and smooth to find gRPC and link against it (same goes for Boost, FairRoot and Protobuf).

ktf:
That's unrelated. The modulefile is for the runtime environment, not for the build environment.

teo:
@ktf My point is: once I build grpc with aliBuild, it is useful to me to be able to `alienv load` the grpc module and manually build the OCCPlugin with a CMake call that references `GRPC_ROOT` instead of having to manually provide the path.

ktf:
You do not need that. Any properly written CMake find_module has logic to find libraries from the `LD_LIBRARY_PATH` and that's what you should use. No need to specify any path with environment variables. What you describe will not work if we start picking up things from the system where in general none of those `_ROOT` variable will be defined.

teo:
@ktf The CMake find script for gRPC can indeed find the libraries anywhere without additional variables, and overall it works fine if gRPC is installed in normal filesystem paths, including headers and executables. So picking things up from the system will work either way.
However, if gRPC is installed in a non-system path, the find script can't find the headers or the gRPC protobuf plugin executable without adding the gRPC root dir to the `CMAKE_PREFIX_PATH` list. The `LD_LIBRARY_PATH` does not seem to be enough for this.

ktf:
If you avoid changing the 2.6.1 I think I am fine with this now.

PR #1176, opened 2018-06-04
Add back libgrpc++.so.1 symlink workaround.

teo:
This time with comments describing why it's necessary, so we don't remove it until upstream fixes their build system issues.


grpc.sh
```diff
@@ -21,6 +21,13 @@ git submodule update --init # aie aie aie
 # gRPC has a custom Makefile which readily breaks with some environment vars, better to run it in a clean environment
 env -i HOME="$HOME" LC_CTYPE="${LC_ALL:-${LC_CTYPE:-$LANG}}" PATH="$PATH" USER="$USER" LD_LIBRARY_PATH="$LD_LIBRARY_PATH" make ${JOBS:+-j$JOBS} prefix=$INSTALLROOT
 
+# Add missing symlink. Must be relative, because the directory is moved around after the build.
+# This should normally not be necessary, but it is made necessary by the issues linked in the following upstream pull request: https://github.com/grpc/grpc/pull/13500
+# Once the issue gets fixed, it should be safe to remove this workaround.
+cd $INSTALLROOT/lib
```
ktf: Use
```
pushd $INSTALLROOT/lib
  ...
popd
```

ktf: actually:

```bash
ln -s libgrpc++.so.%(tag_basename)s $INSTALLROOT/lib/libgrpc++.so.1
```

is what you want, no?

teo: It is. Fixing.

PR #1177, opened 2018-06-05
Fix libxml2 git url

teo:


teo:
@ktf doesn't seem to work here https://alijenkins.cern.ch/job/build-any-ib/6736/console

dberzano:
@teo it's a temporary error seen in the past too many times. We don't see it frequently because we don't rebuild libxml2 too much. I'd change the URL in any case to avoid this problem in the future, but after 4/5pm today (this indeed triggers a full rebuild).

ktf:
ahh, gitlab... ;-)

dberzano:
@ktf it looks like the wrong URL is the reason why o2checkcode was failing everywhere... merging with some urgency!

PR #1187, opened 2018-06-11
Bump FairRoot version

sawenzel:
Necessary in order to fix AliceO2Group/AliceO2#1186.


dberzano:
OK that was quick, I think it's fine.

PR #1189, opened 2018-06-12
Added GEANT4_BUILD_MULTITHREADED env variable:

ihrivnac:
- The variable allows different setting of the GEANT4_BUILD_MULTITHREADED CMake option
- The default value "OFF" is overridden ("ON") in defaults-o2.sh and defaults-o2-dev-fairroot.sh

### DIFF
--- a/geant4.sh
+++ b/geant4.sh
@@ -27,10 +27,6 @@ env:
 
 [[ $CXXSTD > 14 ]] && CXXSTD=14 || true  # Only C++14 is supported at the moment
 
-if [ ! $GEANT4_BUILD_MULTITHREADED ]; then
-  GEANT4_BUILD_MULTITHREADED=OFF
-fi
-
 # if this variable is not defined default it to OFF
 : ${GEANT4_BUILD_MULTITHREADED:=OFF}
 
--- a/geant4.sh
+++ b/geant4.sh
@@ -27,24 +27,28 @@ env:
 
 [[ $CXXSTD > 14 ]] && CXXSTD=14 || true  # Only C++14 is supported at the moment
 
-cmake $SOURCEDIR                                    \
-  -DGEANT4_INSTALL_DATA_TIMEOUT=2000                \
-  -DCMAKE_CXX_FLAGS="-fPIC"                         \
-  -DCMAKE_INSTALL_PREFIX:PATH="$INSTALLROOT"        \
-  -DCMAKE_INSTALL_LIBDIR="lib"                      \
-  -DCMAKE_BUILD_TYPE=RelWithDebInfo                 \
-  -DGEANT4_BUILD_TLS_MODEL:STRING="initial-exec"    \
-  -DGEANT4_ENABLE_TESTING=OFF                       \
-  -DBUILD_SHARED_LIBS=ON                            \
-  -DGEANT4_INSTALL_EXAMPLES=OFF                     \
-  -DCLHEP_ROOT_DIR:PATH="$CLHEP_ROOT"               \
-  -DGEANT4_BUILD_MULTITHREADED=OFF                  \
-  -DCMAKE_STATIC_LIBRARY_CXX_FLAGS="-fPIC"          \
-  -DCMAKE_STATIC_LIBRARY_C_FLAGS="-fPIC"            \
-  -DGEANT4_USE_G3TOG4=ON                            \
-  -DGEANT4_INSTALL_DATA=ON                          \
-  -DGEANT4_USE_SYSTEM_EXPAT=OFF                     \
-  ${CXXSTD:+-DCMAKE_CXX_STANDARD=$CXXSTD}           \
+if [ ! $GEANT4_BUILD_MULTITHREADED ]; then
+  GEANT4_BUILD_MULTITHREADED=OFF
+fi
+
+cmake $SOURCEDIR                                             \
+  -DGEANT4_INSTALL_DATA_TIMEOUT=2000                         \
+  -DCMAKE_CXX_FLAGS="-fPIC"                                  \
+  -DCMAKE_INSTALL_PREFIX:PATH="$INSTALLROOT"                 \
+  -DCMAKE_INSTALL_LIBDIR="lib"                               \
+  -DCMAKE_BUILD_TYPE=RelWithDebInfo                          \
+  -DGEANT4_BUILD_TLS_MODEL:STRING="global-dynamic"           \
+  -DGEANT4_ENABLE_TESTING=OFF                                \
+  -DBUILD_SHARED_LIBS=ON                                     \
+  -DGEANT4_INSTALL_EXAMPLES=OFF                              \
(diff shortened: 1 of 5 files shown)


geant4.sh
```diff
@@ -25,6 +25,10 @@ env:
 ---
 #!/bin/bash -e
 
+if [ ! $GEANT4_BUILD_MULTITHREADED ]; then
```
dberzano: Change those three lines to:

```bash
: {GEANT4_BUILD_MULTITHREADED:=OFF}
```

which does exactly the same. (Mind the `:` at the beginning of the line.)

geant4.sh
```diff
@@ -36,7 +40,7 @@ cmake $SOURCEDIR                                    \
   -DBUILD_SHARED_LIBS=ON                            \
   -DGEANT4_INSTALL_EXAMPLES=OFF                     \
   -DCLHEP_ROOT_DIR:PATH="$CLHEP_ROOT"               \
-  -DGEANT4_BUILD_MULTITHREADED=OFF                  \
+  -DGEANT4_BUILD_MULTITHREADED="$GEANT4_BUILD_MULTITHREADED" \
```
dberzano: Please fix the alignment of `\` in the other lines.

sawenzel:
I see not activity here. @ihrivnac : I think this can be merged as soon as the suggestions of @dberzano have been addressed.

sawenzel:
@ihrivnac, @dberzano: I have updated this PR according to the suggestions (since there was no activity for a while). I think we should go on with this.

geant4.sh
```diff
@@ -25,24 +25,27 @@ env:
...
-  -DCLHEP_ROOT_DIR:PATH="$CLHEP_ROOT"               \
-  -DGEANT4_BUILD_MULTITHREADED=OFF                  \
-  -DCMAKE_STATIC_LIBRARY_CXX_FLAGS="-fPIC"          \
-  -DCMAKE_STATIC_LIBRARY_C_FLAGS="-fPIC"            \
-  -DGEANT4_USE_G3TOG4=ON                            \
-  -DGEANT4_INSTALL_DATA=ON                          \
-  -DGEANT4_USE_SYSTEM_EXPAT=OFF                     \
-  ${CXXSTD:+-DCMAKE_CXX_STANDARD=$CXXSTD}           \
+# if this variable is not defined default it to OFF
+: {GEANT4_BUILD_MULTITHREADED:=OFF}
```
dberzano: This should be:

```bash
: ${GEANT4_BUILD_MULTITHREADED:=OFF}
```

(a `$` is missing).

sawenzel:
Actually I can't modify/edit o2-ninja.sh in relation to this PR (on this very branch). I would prefer doing it in a sep PR directly hereafter.

dberzano:
Hm I don't understand why you can't edit it. Let me check.

dberzano:
OK I just did it.

sawenzel:
What is the status here? The CI test seems to be actually failing.

sawenzel:
Can this be merged, please?

ktf:
Aren’t the failures related?

ktf:
Also, could you please rebase it? There are conflicts.

dberzano:
This is ready to be merged at your convenience. I cannot positively review it because I have contributed to it.

@sawenzel, @ktf if you agree can you please review it and merge when green?

@ihrivnac: I had to rebase it once again for the record 🙂

ihrivnac:
There seems to be a problem with switching the GEANT4_BUILD_TLS_MODEL option from "global-dynamic" to "initial-exec"; this option was set to get working the dynamical loading of libraries via ROOT gSystem when running in MT mode, however it brought an overhead (~3%) when running in sequential mode with libraries built with MT enables.
In our local test, it looked that this option is not needed anymore, but from the tests here it looks like it is still needed for certain platforms. I suggest to try everything with the TLS model  "global-dynamic". I will commit the update in my branch.

dberzano:
CC7, GCC v7.3.0

dberzano:
Note that we can selectively enable it on certain platforms if you tell us so.

ihrivnac:
The platforms which we tested locally were:
Ubuntu:16.04 (Sandro), MAC OS HIgh Sierra 10.13.4 (me)
So we may need a selective setting here.

sawenzel:
Fine. But if the production environment has this problem it means that we are back to accepting a 3% penalty just by theoretically enabling this mode (without even using it). At least I have a problem with this. Is there no other way to avoid this? How are other experiments solving this issue?

ktf:
CMS uses `global-dynamic`:

https://github.com/cms-sw/cmsdist/blob/IB/CMSSW_10_3_X/gcc700/geant4.spec

not sure if they have anything extra in the framework itself.

sawenzel:
For the record, the reasoning to put `initial-exec` in commit `86ed6651973a56db` can be found here: https://alice.its.cern.ch/jira/browse/O2-172.

ihrivnac:
In the new log, only the new test o2sim_G4_mt fails, and it happens at exit, so there is some problem with final clean-up, which I will investigate.
On the other hand, according to Geant4 banner, the Geant4 libraries were not rebuilt with changing the build option; also, I don't see in the log if geant4_vmc libraries were re-built. Is it possible to check this on the test machine?
The large number of warning come from FairRoot framework:
    ### [WARN] The method CheckIfSensitive has to be implemented in the detector class which 
   inherits from FairModule
and it is not relevant to this PR. I'd expect that they are present in the Passed tests, too. Could we check the logs of o2sim_G3 and o2sim_G4?

ihrivnac:
I proposed a fixed for the warnings in O2 PR #1382. With introducing the new function InitializeO2Detector() in O2 there was disconnected the initialisation of sensitive volumes, triggered by DefineSensitiveVolumes() in FairRoot.
On the other hand I am not able to reproduce the break at the end of run. Is it possible to produce a back trace from this crash on the test machine?

dberzano:
@ihrivnac there is something systematically wrong in your Git workflow (you keep bringing here commits that don't belong and are already in master).

* If you want to align this pull request to the changes made upstream, simply press the **Update branch** button on the GitHub interface
* If you want to do the same but manually, simply `git pull --rebase origin master && git push -f` when you are on your `geant4_mt` branch

Hope this helps!

ihrivnac:
Thanks. I clicked on "Update branch" in github, and got again the red button:
"Merging is blocked" ...

dberzano:
Yes, merging is blocked because there is a negative review. Were the comments from @ktf addressed?

ktf:
At leasts all the tests are green now, however there is no way to check if the spurious messages. Didn't we copy successful test logs? That said, no more objections as long as things are green.

dberzano:
Successful logs are not retained ATM.

sawenzel:
I would have preferred resolving the performance problem before merging. Will re-open https://alice.its.cern.ch/jira/browse/O2-172 since we are back to `global-dynamic` threading which has  runtime overhead.

PR #1190, opened 2018-06-12
Bump dataflow versions for FairRoot split

Barthelemy:


Barthelemy:
@dberzano thanks, I have removed the WIP now, it should be ok

PR #1192, opened 2018-06-13
Bump FairMQ to fix DPL logging issues

dberzano:
Problem was fixed in FairRootGroup/FairMQ#43

### DIFF
--- a/fairmq.sh
+++ b/fairmq.sh
@@ -1,6 +1,6 @@
 package: FairMQ
-version: "%(short_hash)s"
-tag: 59b04a1a64a8fc6357f5a018dd76a851046591af
+version: "%(tag_basename)s"
+tag: v1.2.5
 source: https://github.com/FairRootGroup/FairMQ
 requires:
  - boost
--- a/fairmq.sh
+++ b/fairmq.sh
@@ -1,6 +1,6 @@
 package: FairMQ
-version: "%(tag_basename)s"
-tag: v1.2.4
+version: "%(short_hash)s"
+tag: 59b04a1a64a8fc6357f5a018dd76a851046591af
 source: https://github.com/FairRootGroup/FairMQ
 requires:
  - boost


dberzano:
So, we have the following warning (which is an error in this context):

```
In file included from /mnt/mesos/sandbox/sandbox/sw/SOURCES/O2/dev/0/Utilities/aliceHLTwrapper/src/WrapperDevice.cxx:33:0:
/mnt/mesos/sandbox/sandbox/sw/slc7_x86-64/FairMQ/v1.2.5-1/include/fairmq/options/FairProgOptions.h:1:2: error: #warning "This header file is deprecated. Use FairMQProgOptions class directly which now contains all FairProgOptions functionality. Note, that FairMQProgOptions is also available if you include FairMQDevice." [-Werror=cpp]
 #warning "This header file is deprecated. Use FairMQProgOptions class directly which now contains all FairProgOptions functionality. Note, that FairMQProgOptions is also available if you include FairMQDevice."
```

This part should be fixed in O2 but if we merge now we'll make all O2 PRs fail.

dennisklein:
Just out of curiosity, why is the warning an error?

dberzano:
We deliberately build with `-Werror` on our continuous integration (it's not the case on user laptops by default). The rationale is to spot problems like these. People would just ignore warnings otherwise 😉

dennisklein:
Should we mark deprecations with something less strong than a warning?

dberzano:
I believe in this context deprecation means that an alternative exists already, right? O2 code should be updated in order to immediately jump to the alternative. Do you know how for the record?

dennisklein:
In the particular case you posted just remove the `#include <options/FairProgOptions.h>` line. If you hit more complex usages, replace `FairProgOptions` class with `FairMQProgOptions`. The latter inherited from the former, but this inheritance was removed.

dennisklein:
By immediately switching to the new API, you effectively drop support for older versions, too. Which would be equivalent to us just breaking the API without deprecation?

dberzano:
> In other words, interpreting deprecation marker as an error, disables the deprecation period, doesn't it?

You have a point. Are there alternatives to using `#warning`? The point is, at the moment, the code is not broken. This failing test is an indication that we should act in order to migrate to the new system.

dennisklein:
Just for our curiosity on how to better deprecate header files: We did some experiments with [clang](https://wandbox.org/permlink/Pm84hV3ipqQ3DQj5) and [gcc](https://wandbox.org/permlink/J0SysiqcNkaN14It). Both do not produce the nicest output, but at least the warning will have a distinct category, which is not caught by your `-Werror=cpp` setting. Maybe we will use such a workaround in the future. Better would be to have a `#warning` which we could supply a category..

PR #1193, opened 2018-06-13
Move RapidJSON as builddep for Configuration

dberzano:
RapidJSON is not needed at runtime as already reflected by the Modulefile as it
is a header-only library


dberzano:
@alisw/dataflow-experts I appreciate your feedback here 🙂

PR #1201, opened 2018-06-15
Use short_hash instead of commit_hash in version numbers

teo:


dberzano:
All those three packages should belong to @alisw/dataflow-experts. Can you amend the `CODEOWNERS` in the same PR?

dberzano:
Also, I'd retain the `-` as separator instead of `.` (matter of taste/consistency mostly eh, dots have special meanings in RPMs and they will be replaced by the published in any case).

PR #1204, opened 2018-06-21
Fix JDK check on multiuser systems

dberzano:
Creating a file under `/tmp` directly was not ideal and could have led to
clashes on multiuser systems


dberzano:
@brinick fyi

PR #1208, opened 2018-06-26
Use OpenJDK on both macOS and Linux

dberzano:


dberzano:
As discussed with @costing, no particular license restrictions apply to OpenJDK for some security features.

cc @brinick

dberzano:
Recipes not tested by the C.I. in any case. Merging (no production recipe changed).

PR #1210, opened 2018-06-27
Add missing build requirements for ROOT

sawenzel:
Building ROOT on a fresh Ubuntu systems reveals
that there is an actual build dependency on libxml2 and ssl.
This dependency is not reported by aliDoctor and the build fails
when these libraries are not installed.

This commit fixes the problem by explicitely declaring
the dependency for o2.

Please correct as necessary and extend to other defaults etc.

### DIFF
--- a/defaults-o2.sh
+++ b/defaults-o2.sh
@@ -40,6 +40,8 @@ overrides:
       - "GCC-Toolchain:(?!osx)"
       - libpng
       - lzma
+      - libxml2
+      - OpenSSL
   AliRoot:
     version: "%(commit_hash)s_O2"
     requires:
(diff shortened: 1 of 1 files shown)


sawenzel:
See also https://alice-talk.web.cern.ch/t/alibuild-fails-while-building-root/73

dberzano:
It makes sense to me. Does it fix the problems described on Discourse?

sawenzel:
At least aliDoctor now correctly checks these packages. I did not check rebuilding from  a bare-bone Ubuntu.

dberzano:
OK @sawenzel thanks for the fix, as you can see I have added one more modification in order to correctly support OpenSSL both on macOS and Linux. Let's wait for the tests, in the meanwhile I am approving it.

sawenzel:
Thanks for the adjustments.

PR #1211, opened 2018-07-02
Version Bump of AliTPCCommon to v1.3

davidrohr:
Added: TPC SCD Common Code, TPC-TRD Tracking, Attachment of adjacent TPC clusters during track fit.

### DIFF
--- a/alitpccommon.sh
+++ b/alitpccommon.sh
@@ -1,6 +1,6 @@
 package: AliTPCCommon
 version: "%(tag_basename)s"
-tag: alitpccommon-v1.2
+tag: alitpccommon-v1.3.1
 source: https://github.com/AliceO2Group/AliTPCCommon
 build_requires:
   - CMake


dberzano:
Hi @davidrohr, before approving I wanted to ask you whether laptop users will be forced to update alidist in case they update AliRoot, so that we plan the merging accordingly. Thanks!

davidrohr:
Hi @dberzano , the current AliRoot master is compatible with AliTPCCommon v1.2 and v1.3.

Old AliRoot versions (before today) are not compatible to new AliTPCCommon v1.3.
So if you update alidist after this PR and use new AliTPCCommon, you'll also have to update AliRoot.

Some AliRoot PR in the future will remove compatibility to AliTPCCommon v1.2.
At that time, people will also have to update alidist.

I tested already with old and new AliRoot and with O2 and with O2 with AliRoot enabled.
No issues so far. Let's see if the central CI build sees something.
But this is a rather large change with quite some new code, so I cannot promise there won't be any side effects.

dberzano:
As a matter of fact, AliRoot tests fail. Can you please have a look?

davidrohr:
It seems to fail due to usage of C++11 feature constexpr.
I thought it would be allowed to use C++11 outside of the modules used for online mode?
I'll have a closer look tomorrow.

davidrohr:
@dberzano : The Problem was ROOT 5 CINT; which is of course not C++11 capable.
I fixed it for now, but it means we have to be careful with C++11 features in AliRoot until we switch to ROOT6 eventually. From my perspective, this can be merged now.

PR #1212, opened 2018-07-02
Bump ROOT to patched v6-14-00 for O2 and AliRoot

dberzano:
* Use ROOT v6-14-00 with root-project/root#2272
* Align AliRoot and O2 ROOT 6 version
* Update Python and Arrow CMake ROOT options
* Generalize CMake generators support (e.g. Ninja)


sawenzel:
Why is this bumping Geant4 in certain cases?

dberzano:
@sawenzel reportedly, the GEANT4_VMC version used previously in our AliRoot ROOT 6 defaults (which are NOT the production ones) does not compile (see #994, I haven't checked it myself). Note that ROOT 6 Run 2 configurations are not tested in the alidist CI.

The reason why we're bumping them is to allow starting an actual Run 2 validation with them, and to align the ROOT 6 versions used in AliRoot and O2.

Since you're the expert here, do you think we can simply bump GEANT4_VMC without touching GEANT4? Or, is there any reason not to bump everything and validate afterwards?

The intention is to initiate the migration to ROOT 6 for Run 2 after CHEP, this bump would allow for a Grid validation.

dberzano:
The scope of this update has been clarified with @sawenzel and @aphecetche so let's proceed with it. Unfortunately we cannot update GEANT4* separately for Run2 as the new ROOT 6 version would break without that bump. The bump will be mentioned in the merged PR message.

sawenzel:
@mfasDa : Thanks for this explanation!

PR #1216, opened 2018-07-04
Downgrade AliTPCCommon to v1.2 for Run2 software

dberzano:
* O2 chains will use v1.3.1
* Run2 software will use v1.2

### DIFF
--- a/alitpccommon.sh
+++ b/alitpccommon.sh
@@ -1,6 +1,6 @@
 package: AliTPCCommon
 version: "%(tag_basename)s"
-tag: alitpccommon-v1.3.1
+tag: alitpccommon-v1.2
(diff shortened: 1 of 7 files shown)


dberzano:
All PR checks on AliPhysics are currently failing because:

* PR tests are run against _tagged_ AliRoot (not master) but alidist master
* _tagged_ AliRoot does not currently support the latest version of AliTPCCommon (v1.3.1)

We are therefore, as discussed with @davidrohr:

* temporarily downgrading (only for AliRoot/AliPhysics) AliTPCCommon to v1.2

and we will bump it up back to v1.3.1 as soon as we prepare the next AliRoot release.

It has to be noted that the error wasn't caught by the CI because the alidist CI checks AliRoot master (not tagged).

PR #1224, opened 2018-07-10
Build ppconsul on mac

awegrzyn:


ppconsul.sh
```diff
@@ -19,8 +19,19 @@ make ${JOBS:+-j$JOBS}
 #make install
 mkdir $INSTALLROOT/lib
 mkdir $INSTALLROOT/include
-cp $BUILDROOT/$PKGNAME/output/libjson11.so $INSTALLROOT/lib/
-cp $BUILDROOT/$PKGNAME/output/libppconsul.so $INSTALLROOT/lib/
+
+LIB_JSON=so
```
ktf: Why do you need two of them? How about using simply `$SONAME` like done elsewhere?

ppconsul.sh
```diff
@@ -19,8 +19,19 @@ make ${JOBS:+-j$JOBS}
...
+case $ARCHITECTURE in
+    osx*)
+      [[ ! $BOOST_ROOT ]] && BOOST_ROOT=$(brew --prefix boost)
+      LIB_JSON=a
+      LIB_CONSUL=dylib
+    ;;
+esac
+
+cp $BUILDROOT/$PKGNAME/output/libjson11.$LIB_JSON $INSTALLROOT/lib/
+cp $BUILDROOT/$PKGNAME/output/libppconsul.$LIB_CONSUL $INSTALLROOT/lib/
```
ktf: While you are at it, we can probably simplify this to:

```bash
cp $BUILDROOT/$PKGNAME/output/*.$SONAME $INSTALLROOT/lib/
```

awegrzyn:
What about now?
The largest problem is that I get following error when I try to run executable:
```
dyld: Library not loaded: @rpath/libppconsul.0.1.dylib
  Referenced from: /Users/awegrzyn/alice/sw/osx_x86-64/Configuration/OCONF-100-1/bin/configuration-put
  Reason: image not found
Abort trap: 6
```

ppconsul.sh
```diff
@@ -1,32 +1,25 @@
...
 ---
-#!/bin/sh
+#!/bin/bash -e
 
-cmake $SOURCEDIR                                              \
-      -DCMAKE_INSTALL_PREFIX=$INSTALLROOT                     \
-      -DBUILD_SHARED_LIBS=ON                                  \
-      ${BOOST_VERSION:+-DBOOST_ROOT=$BOOST_ROOT}              \
+cmake $SOURCEDIR                                 \
+      ${CMAKE_GENERATOR:+-G "$CMAKE_GENERATOR"}  \
```
Barthelemy: what is this ?

dberzano: By default cmake uses Makefiles as generator (meaning it generates stuff that can be compiled through `make`). With this option we open the possibility to use, for instance, `ninja`.

Barthelemy:
@ktf Are there other things to fix on this PR ? 
@awegrzyn if you finished working on this, could you remove the "WIP" from the title ? 

the test `test_DPLUtils_test_RootTreeReader` is failing but I doubt it is due to the change proposed by this PR.

ktf:
I still see that `setenv` is there. `set` should be used.

ktf:
I assume you checked that LD_LIBRARY_PATH is the one you want, right?

awegrzyn:
@ktf What do you mean by that?

awegrzyn:
@ktf Yup, I loaded env successfully on both Mac and CC7

awegrzyn:
Can this be merged?

ktf:
Overriding @alisw/mrrtf signature.

ktf:
Will merge after the final round of tests.

ktf:
Failure unrelated. Merging.

PR #1226, opened 2018-07-18
Use ROOT v6-10-08, restore working userspace

dberzano:
ROOT v6-14-00 causes problems with file compression and merging, all related to
job submission for analysis in Run 2. While the root cause is being
investigated, we restore the userspace by reverting it for Run 2.


dberzano:
@marcovanleeuwen it seems that the problem we're hitting is pretty critical and everybody is quite clueless about that. We are therefore restoring the userspace by reverting to v6-10-08. We have kept an additional set of defaults (accessible via `--defaults next-root6`) for our Grid tests with v6-14-00.

PR #1230, opened 2018-07-31
Enable examples in FairMQ builds

teo:
Useful in this phase to also build FairMQ examples, so I can then use them from RPMs in VMs.

### DIFF
--- a/fairmq.sh
+++ b/fairmq.sh
@@ -43,7 +43,7 @@ cmake $SOURCEDIR                                                 \
       -DDISABLE_COLOR=ON                                         \
       -DBUILD_DDS_PLUGIN=ON                                      \
       -DBUILD_NANOMSG_TRANSPORT=ON                               \
-      -DBUILD_EXAMPLES=OFF                                       \
+      -DBUILD_EXAMPLES=ON                                        \
       -DCMAKE_INSTALL_LIBDIR=lib                                 \
       -DCMAKE_INSTALL_BINDIR=bin
 


dberzano:
I seem to remember there was a reason why they were turned off. @ktf any hint? Otherwise if the test is green it's fine for me.

ktf:
You removed them to save build time, apparently <https://github.com/alisw/alidist/commit/131be1b7ee5ce2b522cbf32f77922d585bb15b47>.

dberzano:
I see. So maybe we should run a benchmark before enabling them back, or enabling them in some cases only. @teo what default are you using for your builds?

sbinet:
alternatively, @teo, what are the examples you need ?
(perhaps I could provide the equivalent Go ones thus building time wouldn't be an issue :P)

dennisklein:
> I see. So maybe we should run a benchmark before enabling them back, or enabling them in some cases only.

Also note, that building the examples in the new FairMQ repo is only a fraction of the build time compared to the FairRoot repo, where the flag `BUILD_EXAMPLES` was building FairMQ examples PLUS all FairRoot examples.

teo:
@dberzano o2-dataflow
@sbinet Thank you for offering! Right now I was looking to run the 1-n-1 example but potentially others as well, with the goal of taking a sure-to-be-working FairMQ configuration and testing O² Control with these generic FairMQ devices. To my great regret a Go version wouldn't fit the requirements in this case because I need to be testing the control&configuration FairMQ-plugin. However (slight off-topic), as of recently the Control executor supports both `FairMQ` and `Direct` control modes, so it can be configured to send either FairMQ-specific state machine events to the OCC plugin of a running FairMQ device, or "uncooked" Control state machine events with no FairMQ-specific handling to any other process with a compatible gRPC server. No such process-side implementation exists yet (we only have `FairMQPlugin_OCC`) so this `Direct` non-FairMQ code path is untested, and fer can be a candidate - depending on the state machine spec you'd prefer. It could also make sense to keep fer's interface completely FairMQ compatible, use the FairMQ control mode and rather reimplement the equivalent to `FairMQPlugin_OCC` in fer.

sbinet:
@teo: feel free to file an issue (either under AliceO2 or `sbinet-alice/fer`) to discuss (and keep of a record of) the things you'd like to have on the `fer` side to perhaps be able to drop the C++ build requirement.

PR #1235, opened 2018-08-03
Remove dependency of thrift on which

ktf:


ktf:
For the moment we allow thrift only from the system. Merging.

PR #1246, opened 2018-08-10
Fix for QA with ROOT6: Will include QA macros and needs INCLUDE path set

davidrohr:


dberzano:
Hi @davidrohr, I cannot have a deeper look as I am on vacation (you are as well if I recall correctly), here's a quick comment. To me, adding the whole `$ALICE_PHYSICS` as ROOT include path seems a little bit too much (but I realize we might not have other ways around). I leave the decision to your best judgement.

p.s. please open pull requests with titles shorter than 50 chars 🙂

davidrohr:
Hi,
I agree that it is perhaps no ideal solution, but I have no other idea.
Alternatively, we could create a macros directory in AliPhysics like the include, and install everything there, but that would require major changes there.

I already tried to keep it short :| And I wasn't sure when you'd leave. Thought perhaps you could start one more validation attempt. Anyway, enjoy your vacation.

davidrohr:
Hi,
I had a closer look, and we could consistently use gROOT->LoadMacro("$ALICE_PHYSICS..... in all AliPhysics macros, then we do not need to set this include path at all. So I would recomment not to merge this pull request.
I am not closing this PR right now, as we might still want to use it for testing the release validation with ROOT 6 temporarily until we have the proper solution implemented in AliPhysics.
(I had a quick look and it seems not too much effort, but I didn't manage to finish it now. I'll be travelling the next days, so probably I'll not be able to finish it soon.)
Cheers
David

PR #1247, opened 2018-08-14
Exclude failing fairmq tests

Barthelemy:


Barthelemy:
On my mac and CC7 , with defaults o2 and o2-dataflow, there are 3 tests failing in FairMQ and they provoke a failure of the build.

dennisklein:
@Barthelemy Hi, if you have time, it would be aweseome, if you could additionally file an issue with FairMQ which describes your environment in more detail under which these tests fail. Thx!

Barthelemy:
@dennisklein with pleasure, will do.

Barthelemy:
@dennisklein The problem was that another program was using port 7777.

PR #1248, opened 2018-08-14
Enable arrow for o2-fairroot-dev default

ktf:


ktf:
Not tested by the CI. Merging as it is the same as we do in `defaults-o2.sh`.

PR #1251, opened 2018-08-16
Add arrow as a dependency to all o2 defaults

ktf:


ktf:
This should also fix the issue with the `o2checkcode` test in `AliceO2` as it seems to use the dataflow default.

ktf:
AliRoot is not affected. Merging.

PR #1252, opened 2018-08-17
added libInfoLogger as dependency

sy-c:
This is to be able to use libInfoLogger library from O2 to inject log messages into infologger system.


o2.sh
```diff
@@ -8,6 +8,7 @@ requires:
   - Vc
   - hijing
   - HepMC3
+  - libInfoLogger
```
ktf: InfoLogger

ktf: sorry, no, you are right.

sy-c: indeed, there are two packages for infologger.
libInfoLogger is the lighter one, to avoid adding the extra dependencies needed for server, DB, etc. It just has the client library instead of the full daemons set.

I have successfully used libInfoLogger in my fork of O2 (and will create the PR when this one is completed): https://github.com/sy-c/AliceO2/commits/dev

ktf:
Will merge once the tests pass.

PR #1254, opened 2018-08-21
Add arrow dependency to QC

Barthelemy:
Needed because arrow is visible in an O2's header.

### DIFF
--- a/qualitycontrol.sh
+++ b/qualitycontrol.sh
@@ -11,6 +11,7 @@ requires:
   - Monitoring
   - Configuration
   - O2
+  - arrow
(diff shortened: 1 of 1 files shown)


awegrzyn:
As I've just verified QC does not depend on `arrow`. Why does it need to be added the dependency list?

Barthelemy:
indeed. This is more like a workaround. We discussed it with Giulio and he mentioned that he plans to refactor this piece of code in O2 to hide the Arrow dependency.

Barthelemy:
but for the time being I would need this workaround to fix the build.

PR #1258, opened 2018-08-22
Use initial-exec as threading model for G4(MT) mode

sawenzel:
There is a performance problem with `global-dynamic` as reported in https://alice.its.cern.ch/jira/browse/O2-172 and
`initial-exec` is the suggested and working solution.

Note that this change does not affect any production as it will only take effect when MT is enabled (which is not the case by default yet).


ktf:
Error unrelated.

PR #1263, opened 2018-08-23
Fix o2checkcode not exiting with nonzero on errors

dberzano:
Fixes O2-283.


sawenzel:
Note that we should now expect some problems in the O2 code base (since a few non-conforming things slipped in meanwhile).

dberzano:
Force-merging as checks do not test it. We expect some O2 PRs to fail (as they should) because of this. Thanks @sawenzel for reporting and for your patience.

PR #1265, opened 2018-08-23
ppconsul: Allow boost to be picked up via brew

ktf:


ppconsul.sh
```diff
@@ -10,12 +10,17 @@ build_requires:
...
+    [[ ! $BOOST_ROOT ]] && BOOST_ROOT=`brew --prefix boost` ;;
+esac
+
 cmake $SOURCEDIR                                 \
       ${CMAKE_GENERATOR:+-G "$CMAKE_GENERATOR"}  \
       -DCMAKE_INSTALL_PREFIX=$INSTALLROOT        \
       -DCMAKE_INSTALL_LIBDIR=lib                 \
       -DBUILD_SHARED_LIBS=ON                     \
-      ${BOOST_VERSION:+-DBOOST_ROOT=$BOOST_ROOT}
+      ${BOOST_ROOT:+-DBOOST_ROOT=$BOOST_ROOT}
```
Barthelemy: At some point I remember being told to use XXX_VERSION instead of XXX_ROOT because the former is guaranteed to be set whereas the latter depends on the other recipe explicit definition. What is the reason for now switching back to XXX_ROOT ?

ktf: In the module file, yes, that's what you want to do. That's because we use modulefiles only for software which we build with alibuild, so you need to use a variable which is only set when the tool is built with albuild (e.g. PACKAGE_VERSION).

When instead looking up for dependencies to be passed to cmake you should use PACKAGE_ROOT whenever you expect it to be picked up from either alibuild or from homebrew on mac, like we should do for the case of boost.

ktf:
I've described the rationale at https://github.com/alisw/alidist/commit/9e4048db326170648b879cf29d3fa7948ade9b61

ktf:
Unless you have further questions / objections, could you sign off, please?

ppconsul.sh
Barthelemy: Ok, I understand, thank you for the clarification.

PR #1266, opened 2018-08-23
common-o2: use BOOST_ROOT for explicit boost lookup

ktf:


ktf:
Same rationale as the one of ppconsul.

PR #1267, opened 2018-08-24
[WIP] Adopt Ninja in O2

dberzano:
* Special Ninja defaults removed; Ninja is used in O2 by default
* Harmonize some differences between `o2-dev-fairroot` and `o2`

See O2-344


ktf:
This should come with a big and bright disclaimer that people should use `cmake --build` (or whatever) to build, no?

dberzano:
Yes, this is one of the reasons why it's a WIP. Some discussion on the methodology is expected at WP3 before wee release it. It's good in any case to see if the chain compiles.

ktf:
How about we have the new root in first? That should be ready to go immediately no?

Barthelemy:
I think that this PR should be split as it does/addresses two (or even 3) different things. I will need to read a bit about `cmake --build` and ninja to give any valid input on the PR content.

dberzano:
@Barthelemy @ktf as explained already:

1. this is a WIP
2. proper migration methodology has to be discussed first
3. yes it's more appropriate to split it however I'd like to see what are the test results with both of them enabled to save my time 😄

Be assured that nothing will be changed unexpectedly and without proper warnings.

dberzano:
With #1270, this PR will be only about Ninja and defaults harmonization.

PR #1275, opened 2018-08-30
Fix o2checkcode when no source files changed

dberzano:


dberzano:
The problem affected, for instance, AliceO2Group/AliceO2#1300. `grep` returned nonzero as no source file changed, causing the recipe to break.

Since this is not tested in the `build/O2/alidist` recipe, I am merging it.

PR #1257, opened 2018-08-22
[WIP] Add explicit Common dependency to O2

sy-c:
libInfoLogger depends on Common-O2
need to define variable for cmake find path

PR #1276, opened 2018-08-30
Add explicit Common dependency to O2

sy-c:
This change is to fix cmake O2 on osx for another PR:
AliceO2Group/AliceO2#1293

Add explicit Common dependency to O2

This is a replacement PR for alisw/alidist#1257


sy-c:
Thanks for spotting this, I did not notice when testing as common-o2 is already in the module load of libinfologger. Fix committed.

PR #1280, opened 2018-09-07
Fix rivet-config relocation

dberzano:


dberzano:
@qgp this should solve your problem, can you please have a look?

rivet-hi.sh
```diff
@@ -8,8 +8,10 @@ requires:
   - fastjet
   - HepMC
   - boost
+build_requires:
+  - GCC-Toolchain:(?!osx)
 prepend_path:
-  PYTHONPATH: $RIVETHI_ROOT/lib64/python2.7/site-packages:$RIVETHI_ROOT/lib/python2.7/site-packages
+  PYTHONPATH: $RIVET_HI_ROOT/lib64/python2.7/site-packages:$RIVET_HI_ROOT/lib/python2.7/site-packages
```
dberzano: @qgp for the record this was wrong in the original recipe

ktf:
Do we really need to export `RIVET_HI_ROOT` at all? Why is this needed?

qgp:
@ktf: It's the prefix path for the installation and is exported as we do for all other packages as well ... Should we also stop exporting e.g. GSL_ROOT then? Maybe I don't understand something here?

ktf:
Yes, correct. We should stop exporting in the **runtime** (build time is fine and done automatically, actually) ALICE specific environment variables (e.g. `RIVET_HI_ROOT`, `GSL_ROOT`, `WHATEVER_ROOT`) which are generally not to be expected (so for example `ROOTSYS` is fine because ROOT expects and handles correctly it). This is to avoid introducing ALICE specific expectations for packages which are not under ALICE control.

qgp:
Where do we take the prefix from then? Do all packages come with a <package>-config that works correctly?

For Rivet it's fine with me if we manage to avoid this. However, I don't think this should block the fix for the relocation now.

ktf:
For example, yes. Or for python packages you rely on the fact that PYTHONPATH will be set correctly. Again, the point is not to eliminate ways of finding the prefix or files inside the folder, the point is to eliminate ALICE specific environment variables.

I agree this should not delay an hot fix, however you still did not reply my question: where at runtime do you need such a prefix?

qgp:
E.g. to find things in $RIVETHI_ROOT/share/Rivet - which admittedly can be taken from rivet-config (if it returns the correctly relocated path). 

To get something like this from all the paths contained in PYTHONPATH seems ugly to me. Anyway, this is getting away from Rivet.

qgp:
Deprecating <package>_ROOT is fine with me if we then do it for all packages. Actually, I agree that this makes sense and is a good idea.

Anyway, for the hot fix now: I would not change the name of this variable, this doesn't solve anything (as far as I can see) and just creates more confusion. Then, in a second go we can drop RIVET(HI)_ROOT.

ktf:
Yes, I agree we should do it everywhere things are not under our control.

qgp:
@dberzano: Can you comment on the RIVETHI_ROOT to RIVET_HI_ROOT renaming?

dberzano:
As discussed offline:

* The `PYTHONPATH` part in the header cannot work in the recipe's header like this (because aliBuild internally uses `RIVET_HI_ROOT`, with the underscore): so if some package has Rivet-hi as a dependency, it will find the `PYTHONPATH` in the _recipe_ environment wrongly set. That fix stays.
* It was agreed to leave the spelling `RIVETHI_ROOT` in the runtime environment for now, but PWGMM is warned of its deprecation in favour of using `rivet-config --prefix` (which is more robust and not aliBuild-dependent)

So, when PWGMM will decide it's time, `RIVETHI_ROOT` in the runtime environment will go 👍

dberzano:
Build tests don't test Rivet or Rivet-hi in any case. Merging.

dberzano:
As agreed, we'll build (with GCC 7):

* A new AliGenerators (which also contains Rivet)
* Rivet-hi (which is an independent package)

PR #1283, opened 2018-09-12
nanomsg: move to 1.1.3

ktf:
Apparently this is required by latest FairMQ.


ktf:
@sawenzel this should fix the error we saw in the other PR.

ktf:
O2 compiles. Not used by AliRoot. Merging.

PR #1284, opened 2018-09-12
msgpack: bump to satisfy latest FairMQ requirement

ktf:


ktf:
Mmm... looks like I need to do it in the defaults... :-(

dberzano:
Are there cases where we need a different version of msgpack? I propose a cleanup if we don't 🙂 

I seem to remember that @teo had a case (but I might be wrong).

dennisklein:
If this fails only because of FairMQ v1.2.6, then merge #1285 first. Since v1.2.7 FairMQ supports msgpack 3.x. (And since v1.3.1 it requires msgpack 3.1+).

PR #1286, opened 2018-09-12
default-o2-dev-fairroot: update msgpack

ktf:


ktf:
Does not get tested by tests. Merging.

PR #1306, opened 2018-09-27
Remove explicit flag for Monitoring

Barthelemy:
It is not needed anymore thanks to the use of modern cmake in monitoring.

### DIFF
--- a/o2.sh
+++ b/o2.sh
@@ -128,7 +128,6 @@ cmake $SOURCEDIR -DCMAKE_INSTALL_PREFIX=$INSTALLROOT                            
       -DCMAKE_EXPORT_COMPILE_COMMANDS=ON                                                    \
       ${CXXSTD:+-DCMAKE_CXX_STANDARD=$CXXSTD}                                               \
       ${ALITPCCOMMON_ROOT:+-DALITPCCOMMON_DIR=$ALITPCCOMMON_ROOT}                           \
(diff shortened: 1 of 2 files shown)


Barthelemy:
@awegrzyn

awegrzyn:
I guess in O2 it needs to stay as they are not using targets.

awegrzyn:
This also needs be merged after https://github.com/AliceO2Group/QualityControl/pull/59

awegrzyn:
Maybe you could also bump QC in the same PR?

dberzano:
@awegrzyn @Barthelemy so just to understand, the corresponding `.cmake` or `.config` module is searched for automatically under `LD_LIBRARY_PATH`?

What happens on macOS?

awegrzyn:
I'm not sure but I guess in docs I found:
> Search the standard system environment variables. Path entries ending in /bin or /sbin are automatically converted to their parent directories ...

awegrzyn:
I did a quick test and indeed after adding `Monitoring/bin` to `PATH` cmake finds the config which is located in `Monitoring/lib/cmake/`

Barthelemy:
@dberzano for me it works on mac  as it is , maybe thanks to the LD_LIBRARY_PATH even though it is on mac

Barthelemy:
@awegrzyn I have bumped QC

PR #1308, opened 2018-09-28
v0.7.1 QC and qcg in flpproto

Barthelemy:


Barthelemy:
related to O2-273 : setup a machine for qcg.

Barthelemy:
@awegrzyn I have added a commit to add lib64 to the LD_LIBRARY_PATH for QualityControl.

ktf:
Out of curiosity, why using `lib64`?

Barthelemy:
@ktf That is a very good question actually :) It is linked to this other PR : https://github.com/AliceO2Group/QualityControl/pull/61

I started using ${CMAKE_INSTALL_LIBDIR} but realized that on my system it points to lib64. Thus libraries were not found. 

As noted by Adam in the other PR, we should review this to make sure we are all on the same line. I have absolutely no opinion on it, I just followed what was done in Monitoring.

awegrzyn:
I've tested how CMake sets `CMAKE_INSTALL_LIBDIR` on different platforms:
- macOS - `lib`
- CC7 - `lib64`
- SLC6 - not set at all (empty)

I'd be strongly in favour of hardcoding this var to a fixed path (most probably `lib`).

ktf:
Yeah, an alternative would be being able to extract it via cmake, but I am not sure if it's actually possible...

Barthelemy:
ok, I can do that. I would prefer than relying on something that changes from platform to platform.

PR #1330, opened 2018-10-12
Use the same variable to enable and check Alien

mpuccio:
If you have both O2 and AliRoot in the same working environment $ALIEN_RUNTIME_ROOT might be set while $ALIEN_RUNTIME_VERSION is not while compiling O2.
With this fix you prevent having alien NOT added to the configuration but checked afterwards (that was breaking ROOT compilation with o2 defaults in some configurations of the working directory/env).


ktf:
Did you give a shot to my suggestion?

ktf:
Did you give a shot to my suggestion?

mpuccio:
Sorry but I had no possibility of trying it, I was having this issue on a workstation of a colleague of mine. Anyway imho this is solving a clear bug in the recipe, can this be merged?

PR #1331, opened 2018-10-14
[WIP] try common_recipe option on two recipes

aphecetche:


aphecetche:
Not a real PR, but more a supporting piece for https://alice.its.cern.ch/jira/browse/O2-414

Barthelemy:
seems more readable, good job.

ktf:
I like the idea. I had something similar in mind at some point, but without using functions. Yours is definitely better. Do you understand the build failure, though?

aphecetche:
concerning the build failure : yes, it's expected as it's not only a change in the recipes. There is an alibuild counterpart : https://github.com/alisw/alibuild/pull/551

dberzano:
This is very nice and simple and reduces so much boilerplate, thanks.

ktf:
closing this for now, as it requires the whole collaboration to move to a new version of aliBuild.

PR #1340, opened 2018-10-16
updated packages in python-modules.sh

MycrofD:
allowed installation of latest versions of all packages. root_numpy works with 4.7.1, not the latest 4.7.3.
matplotlib 1.4.3 throws backend warning. matplotlib 2.0.2 works but matplotlib.pyplot wants pyqt=4. 

abyzovlab/CNVnator#96 (comment)
...........
Hi,
tldr; working version: _python2.7(anaconda), ROOT 6.09.02, root_numpy 4.7.1, rootpy 1.0.1, numpy 1.15.2, matplotlib 2.0.2 (pyplot needs pyqt=4)_

The long version:

so I tried to install latest versions of all packages: _numpy, rootpy, root_numpy_ etc. But there was another error with _root_numpy 4.7.3_:
`ImportError: libCore.so.5: cannot open shared object file: No such file or directory `

It hit me, why don't I try other versions of _root_numpy_? So I did.
With _root_numpy 4.7.2_, similar error:
`ImportError: libCore.so.6.10: cannot open shared object file: No such file or directory `

But then with _root_numpy 4.7.1_, it finally worked. No errors. 

With _matplotlib 1.4.3_ it did work, but threw some backend warning while importing _matplotlib_. 

```
>>> import matplotlib
/user../lib/python2.7/site-packages/matplotlib-1.4.3-py2.7-linux-x86_64.egg/matplotlib/__init__.py:1039: UserWarning: Bad val "pyside" on line #38
	"backend      : pyside
"
	in file "/user/.../lib/python2.7/site-packages/matplotlib-1.4.3-py2.7-linux-x86_64.egg/matplotlib/mpl-data/matplotlibrc"
	Key backend: Unrecognized backend string "pyside": valid strings are [u'pgf', u'cairo', u'MacOSX', u'CocoaAgg', u'gdk', u'ps', u'GTKAgg', u'nbAgg', u'GTK', u'Qt5Agg', u'template', u'emf', u'GTK3Cairo', u'GTK3Agg', u'WX', u'Qt4Agg', u'TkAgg', u'agg', u'svg', u'GTKCairo', u'WXAgg', u'WebAgg', u'pdf']
  (val, error_details, msg))

```
I tried with some newer versions, but there was an error (`ImportError: No module named functools_lru_cache`) and the following answer showed the working version _matplotlib 2.0.2_.
https://stackoverflow.com/a/49335954/3623705

But then again, pyqt=5 was the latest version installed, so I had to downgrade it `conda uninstall pyqt`, and then `conda install pyqt=4`
Other dependencies for _pyqt=4_:     _sip-4.18,    qt-4.8.7,    pyqt-4.11.4_


MycrofD:
@dberzano  After a lot of trials, I got a working environment that includes rootpy and root_numpy (this was the motivation). Although I installed `rootpy` and `root_numpy` separately, as in `pip install <package>`  and not through the script `python-modules.sh`, it worked and so I included the packages in the 'for loop'. Sorry that it doesn't work here. Apparently PyROOT has not been enabled for `rootpy`.

dberzano:
Sorry for being picky, with "motivation" I mean: is this required for achieving something? Do other persons require it?

Users should not require to install and compile Python-modules on their own. By following the [instructions](https://alice-doc.github.io/alice-analysis-tutorial/building/) and adding what's missing for you, you can have your own specific working environment without changing the upstream recipes. We only require Python-modules to be compiled by us for our Grid deployments, to make it clear.

Another note. Conda is a very nice project, but it comes with its own compiler. Mixing-matching Conda-installed packages and aliBuild- or system-provided is a road to disaster, unfortunately.

MycrofD:
I think rootpy needs to be installed separately after installing ROOT by aliroot. That's why I believe it gives the error, as ROOT has not yet been installed and thus PyROOT has not been enabled. True, others might not need it. However, I was not sure how to make this exercise of getting a working environment public. So, I did a PR, although all I wanted was to make a comment: that these are the packages that work.

Thanks for the tip. :)

PR #1359, opened 2018-10-26
bump jalien to 1.0.5

costing:
Main fix is ZIP archive storing of ROOT files (disabling compression to make it fully ROOT-compatible)


Barthelemy:
@costing @mimarpe What about merging ?

PR #1362, opened 2018-10-30
Addition of codingguidelines recipe and use in qualitycontrol (O2-351)

Barthelemy:


Barthelemy:
> , so I trust Barth.

Thanks :) as a matter of fact, I would like some tests to run for QC as well if possible.

ktf:
I think we should define the aforementioned o2 environment project, depending on everything, and test that, actually. Anyways, yes, we need more tests.

Barthelemy:
@ktf the jira ticket is here : https://alice.its.cern.ch/jira/browse/O2-407

PR #1363, opened 2018-10-31
Bump FairMQ to 1.3.6

teo:


ktf:
Actually we should fix FairRoot as well...

teo:
@ktf I know, I submitted this anyway so it's ready to go when FairRoot is sorted out.

ktf:
We also need a new msgpack (and I suspect a new cmake, having seen @mkrzewic PR). I've updated it in #1366 which I will merge as soon as the tests pass.

ktf:
@mkrzewic can you check what is going on? Looks like this breaks the memory resource stuff.

mkrzewic:
the memory resource stuff is now in FairMQ including some deeper level integration - I imagine it might break some more stuff as a number of methods are no longer const... I'll have a look.

ktf:
Thanks.

ktf:
@mkrzewic did you manage to give a look? This is preventing us from upgrading and the new version is needed to start integrating `Control`.

mkrzewic:
@ktf almost there, was tied up in PbPb preparations. Expect a PR today.

mkrzewic:
#1458 should to the trick.

sawenzel:
Shouldn't this be merged now?

dberzano:
Tests are running. I want to be 100% sure they work before doing that. I am actively on it, hold on.

sy-c:
This PR broke flpproto...
curiously (or unsurprisingly) with an error related to boost::container, as in one of the failing test reported in https://github.com/AliceO2Group/AliceO2/pull/1458

the quick fix is to add "container" in the boost components,
the good fix should be to use FairMQ_Boost_COMPONENTS variable, as stated in the FairMQ doc:
(but no one seems to do it in O2 repos, so it all probably works just by chance...)

> In order to succesfully compile and link against the FairMQ::FairMQ target, you need to discover its public package dependencies, too.
> 
> find_package(FairMQ)
> if(FairMQ_FOUND)
>   find_package(FairLogger ${FairMQ_FairLogger_VERSION})
>   find_package(Boost ${FairMQ_Boost_VERSION} COMPONENTS ${FairMQ_Boost_COMPONENTS})
> endif()

PR #1365, opened 2018-10-31
Bump FairRoot

ktf:
This is required to work with the new FairMQ 1.3.5


ktf:
@shahor02 @sawenzel could you green light this? We need it in order to be able to move to FairMQ 1.3.5 (which itself is requested by @teo for Control).

shahor02:
@ktf for me its fine, the sim and rec chains work as before

ktf:
Merging so that we can check if the FairMQ bump becomes green.

PR #1370, opened 2018-11-06
[WIP] Get .clang-format from CodingGuidelines (O2-351)

Barthelemy:


Barthelemy:
@alisw/infrastructure-admins could someone check this PR ?

Barthelemy:
build broken by something else

dberzano:
Hold on. We had to merge a PR in O2 before updating alidist. We are now waiting for the tests to confirm we can bump alidist as well. Some disruption is expected, unfortunately.

dberzano:
Errors will be gradually cleared.

o2.sh
```diff
@@ -91,6 +92,8 @@ case $ARCHITECTURE in
   *) SONAME=so ;;
 esac
 
+rsync -a --ignore-existing $CODINGGUIDELINES_ROOT/.clang-format $SOURCEDIR # Copy .clang-format from CodingGuidelines
```
sawenzel: I still have a doubt here. The file will be copied to the source dir of o2 only at the moment of building. So in the time between a git checkout and the first build the user does not have a valid .clang-format file (and cannot use auto-correction features in eclipse).

ktf: I agree. I though the consensus was to propagate the file in various repositories at commit time, via a travis or similar hook.

dberzano: Yes, @Barthelemy and I discussed this precisely this afternoon offline, and we've agreed that there'll be a proper solution in the next days. Having a hook-or-something would also prevent polluting the source directory (we, in general, reject alidist PRs altering the source directory, and this is the reason why all builds are done out of source if possible, and if not, sources are `rsync`d in the build dir first).

sawenzel: I don't understand arguments about polluting the source directory. As argued in the past, the .clang-format file belongs into the source directory because it is coupled to the code and this is where standard tools (such as git-clang-format, eclipse, etc.) will look. Unfortunately, there is no `git clone` hook.

sawenzel: Also, is there an actual reason/need to go into an intermediate solution?

dberzano: I did not say we should not put it in the source directory of every repo, I said that it's not a recipe's task to do it, citing consistency reasons with other recipes.

We should propagate the `.clang-format` using some kind of hooks as @ktf pointed out (_e.g._ CodingGuidelines master changes --> proper PRs are opened in all the concerned repositories automatically).

This requires shipping credentials with, _e.g._, Travis, but since it should react only upon master branch changes, Travis will allow us to do that quite trivially. @Barthelemy and I can prepare a demonstrator maybe.

sawenzel: Ok. So maybe a compromise would be to ask/force users only to install sources with `aliBuild init` (that can be documented). During that init we could call some sort of hook which does the initial syncing (probably needs some logic in alibuild). Thereafter the procedure described here should be ok to keep the sources up to date (with some delay), but at least I would always have a file present. This would also avoid playing with credentials etc.

sawenzel: In addition, instead of copying the file one might employ soft links to better have the "semantics" of using some resource which is outside of the current repo and which is automatically up-to-date.

sy-c: Hello,
I am surely completely off-topic, but has it been considered to use the cmake ExternalProject feature ? It's usually a convenient way to retrieve files from an outside repo...

sawenzel: @sy-c : I think it was considered. But the problem remains the same since cmake is only executed at build time. (We also considered that cmake could execute a test a check if the file is still up-to-date).

ktf:
https://github.com/AliceO2Group/CodingGuidelines/pull/11 is the correct way of doing this, IMHO.

ktf:
As discussed I think converging on AliceO2Group/CodingGuidelines#11 is better. Closing this.

PR #1377, opened 2018-11-07
Build Python's own libffi

dberzano:
Prevents "libffi.so.5 not found" problem if running on a platform different than
the build one.


dberzano:
This one along with #1376 is needed for the next AliGenerators release (@alisw/pagmc).

PR #1379, opened 2018-11-07
Fix the ROOT include path in qc (QC-112)

Barthelemy:


qualitycontrol.sh
```diff
@@ -71,6 +73,7 @@ setenv QUALITYCONTROL_ROOT \$::env(BASEDIR)/$PKGNAME/\$version
 prepend-path PATH \$::env(QUALITYCONTROL_ROOT)/bin
 prepend-path LD_LIBRARY_PATH \$::env(QUALITYCONTROL_ROOT)/lib
 prepend-path LD_LIBRARY_PATH \$::env(QUALITYCONTROL_ROOT)/lib64
+prepend-path ROOT_INCLUDE_PATH \$::env(QUALITYCONTROL_ROOT)/include # this is for ROOT
```
dberzano: Can you just please get rid of the comment? I am not sure Environment Modules handles them properly in all versions if they are inline. Also, we use `ROOT_INCLUDE_PATH` in many places already so I don't think a comment is necessary 🙂

Barthelemy: sure, done

PR #1380, opened 2018-11-08
Bump to llvm/clang 7.0

sawenzel:
Necessary to resolve https://alice.its.cern.ch/jira/browse/O2-412;
Needs to be merge **after** https://github.com/alisw/clang/pull/2 and a new tag has been created.

### DIFF
--- a/clang.sh
+++ b/clang.sh
@@ -1,7 +1,7 @@
 package: Clang
 version: "7.0"
-source: https://github.com/alisw/clang
 tag: master
+source: https://github.com/alisw/clang
 build_requires:
  - CMake
 ---
--- a/clang.sh
+++ b/clang.sh
@@ -1,5 +1,5 @@
 package: Clang
-version: "3.9"
+version: "7.0"
 source: https://github.com/alisw/clang
 tag: master
 build_requires:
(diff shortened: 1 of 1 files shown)


sawenzel:
Anything holding back this PR? Without it the o2codechecker will stay blind.

ktf:
There is a check that is still red, can you check what is going on there? I am away today.

sawenzel:
I checked and fixed. But the ticket requires intervention from admins also in other places (tag creation for clang) since I don't have the right to do it.

ktf:
I think I've everything in place. Let's check the tests are green and then I will merge. Do we actually build the package though?

sawenzel:
I guess the package is not build in alidist PRs.

ktf:
I do not remember actually. We might actually have some logic in place which checks every package whose recipe has been modified.

dberzano:
Nope, it's not tested (it's just linted). It can be merged without waiting.

sawenzel:
Thanks. I also updated the codechecker: https://github.com/AliceO2Group/O2CodeChecker/pull/27
The system is hopefully consistent now.

PR #1381, opened 2018-11-08
Add libtirpc

adriansev:
starting with glibc 2.28 rpc.h is no longer part of glibc
and instead of /usr/include/rpc/rpc.h it is found in libtirpc
package in /usr/include/tirpc/rpc/rpc.h

### DIFF
--- a/libtirpc.sh
+++ b/libtirpc.sh
@@ -3,7 +3,7 @@ version: "%(tag_basename)s"
 tag: "libtirpc-1-1-4"
 source: git://git.linux-nfs.org/projects/steved/libtirpc.git
 build_requires:
-  - CMake
+  - autotools
   - "GCC-Toolchain:(?!osx)"
 ---
 #!/bin/bash -e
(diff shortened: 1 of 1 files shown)


adriansev:
@dberzano Ping? :)

adriansev:
@dberzano done!

libtirpc.sh
```diff
@@ -0,0 +1,33 @@
+package: libtirpc
+version: "%(tag_basename)s"
+tag: "libtirpc-1-1-4"
+source: git://git.linux-nfs.org/projects/steved/libtirpc.git
+build_requires:
+  - CMake
```
dberzano: A small change: it does not seem the module requires CMake (I see `./configure` below) so you can drop this dependency. Just check whether it depends on stuff like `autotools` and add it properly (we have a recipe for that already).

adriansev: right! copy-paste artifacts, done

adriansev:
@dberzano done! thx!!

adriansev:
@dberzano all green, thanks!!

PR #1386, opened 2018-11-13
Fix loading JAliEn-ROOT as an external ROOT plugin

Atlantic777:
This commit reverses order of loading external and internal plugins.
First, the external plugins are loaded, then internal. Order of loading
the plugins defines also the lookup order. The plugin priorities are
accross group, not accross all plugins available.

NOTE: If both external and internal plugin have the same base and
have the same regex, then the external plugin will be removed from
the plugin manager! However, the internal plugin will not replace it,
it will be registered at the end of list as any other internal plugin.

NOTE2: This modification should not affect behavior of any existing
ROOT plugin, is relevant only for JAliEn-ROOT and requires no further
modifications in ROOT.


Atlantic777:
One example is accessing the remote XML files (TFile::Cp("something.xml", "alien:///target.xml"). The problem is that plugin priorities are bound to the same plugin directory, they are not global. This means that even if external plugin has higher precedence over an internal plugin, the internal one would be loaded.

To be precise, here is what is going on:
- (internal plugin) TAlienFile (^alien:)
- (internal plugin) TXMLFile (".+[.]xml$)
- (external plugin) TJAlienFile ("^alien:)

If we want to load a remote XML file, TJAlienFile will kick out TAlienFile, but in the end the TXMLFile will be used. This is a regression and unexpected behaviour.

The ROOT's plugin manager has a replacement mechanism (based on the plugin type and regex), it will kick out the existing matching plugin, but it will not have priority over any other internal plugin that has matching regex.

Another solution would be to improve ROOT's plugin manager and make priorities work across the directories. We discussed this solution with the ROOT team but concluded that we should find a way to bypass this on our side. Here's an example that fails: https://gitlab.cern.ch/nhardi/jalien-ci/blob/test-xml-files/tests/007_test_cp_XML_to_grid/test_XML_files.C

dberzano:
Ohh thanks for clarifying. Indeed, the ROOT plugin loading mechanism sucks badly. I just want to understand better how it works.

You load a file like:

```c++
TFile::Open("alien://blah/file.xml")
```

* You probably expect that both the JAliEn and XML plugins are loaded.
* What happens when you don't have JAliEn is that both AliEn and XML are loaded, correctly.
* What happens now with JAliEn is that XML only is loaded.

Are my statements above correct?

If you reverse the order, however, what I see happening is that the old AliEn will **always** kick out XML. But maybe I am not fully understanding what is going on here 😕

Atlantic777:
To be included in this pull request:
- [ ] change JAliEn-ROOT repository to gitlab.cern.ch/jalien/jalien-root <- I'm working on that

Atlantic777:
Alright, we can merge this as far as I am concerned.

PR #1388, opened 2018-11-13
Pass Ppconsul_DIR to cmake

awegrzyn:


ktf:
mmmm... I think this is breaking my O2 build on mac:

```
DEBUG:O2:O2:0: [ 64%] Linking CXX shared library ../../lib/libFramework.dylib
DEBUG:O2:O2:0: ld: file not found: @rpath/libppconsul.0.1.dylib for architecture x86_64
DEBUG:O2:O2:0: clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

I suspect the problem is the rpath...

Barthelemy:
@awegrzyn Did you develop on Mac ? I tested on CC7.

Barthelemy:
@ktf Don't we test on Mac during the PR ?

awegrzyn:
Well, I didn't verify I'm Mac as build fails due to FairMQ `CreateMessage`. Let's revert it then.

Barthelemy:
I confirm the problem on Mac

dberzano:
They are not tested and never were. As discussed, it looks like the persons responsible are not really willing to buy a new Mac. ATM we don't have enough resources for doing that.

PR #1390, opened 2018-11-14
Update nanomsg

ktf:


sawenzel:
Shouldn't this be merged? Was approved few months ago.

PR #1391, opened 2018-11-14
Revert "Pass Ppconsul_DIR to cmake"

awegrzyn:
Reverts alisw/alidist#1388


ktf:
I can confirm reverting fixes the problem. Not sure what is going on though, I suspect something with the way we build ppconsul.

PR #1392, opened 2018-11-15
bump readout v0.14.1

sy-c:
fix for https://github.com/alisw/alidist/pull/1363

### DIFF
--- a/readout.sh
+++ b/readout.sh
@@ -1,6 +1,6 @@
 package: Readout
 version: "%(tag_basename)s"
-tag: v0.14.0
+tag: v0.14.1
 requires:
   - boost
   - "GCC-Toolchain:(?!osx)"


awegrzyn:
Hi,
This fails on both Mac and CC7 due to:
```
/home/awegrzyn/alice/sw/SOURCES/Readout/v0.14.1/v0.14.1/src/testROC.cxx:15:10: fatal error: ReadoutCard/Driver.h: No such file or directory
 #include <ReadoutCard/Driver.h>
```

sy-c:
I confirm that ReadoutCard v0.9.1 broke the interface,
it works well with v0.9.0

kostorr:
@sy-c , I removed the Driver class and the header file. Corresponding initialisations are now done with the construction of the DMA channel. I believe it's enough to remove the include and the driver initialisation part from Readout/src/ReadoutEquipmentRORC.cxx and Readout/src/testROC.cxx

sy-c:
sure, the fix is easy, but it should be advertised beforehand when there is an incompatible interface change (and this usually implies a major version change, as recommended in https://semver.org/)

sy-c:
readout v0.14.2 released... compilation now ok, but will have to check with a CRU

PR #1395, opened 2018-11-15
bump readout v0.14.2

sy-c:
to comply with interface change in ReadoutCard v0.9.1


Barthelemy:
this is outdated right ? there is another PR for 0.15

PR #1397, opened 2018-11-19
Update clang.sh to use v7.0.0-alice1

ktf:


ktf:
Nothing will test this. Merging.

PR #1400, opened 2018-11-19
WIP Using temporary tag for DDS

AndreyLebedev:
Only required for Grid tests.

### DIFF
--- a/dds.sh
+++ b/dds.sh
@@ -1,6 +1,6 @@
 package: DDS
 version: "%(tag_basename)s"
-tag: "2.0"
+tag: "GridTest"
 source: https://github.com/FairRootGroup/DDS
 requires:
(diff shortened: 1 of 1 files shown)


dberzano:
@AndreyLebedev can you explain the rationale? Do you need this to be merged? Also: the error says:

```
  CMake 3.11.1 or higher is required.  You are running version 3.11.0
```

Can't you do something to make it work with CMake 3.11.0? Otherwise, this update will be extremely disruptive (everything needs to be rebuilt upon CMake updates).

AndreyLebedev:
We have been asked to perform massive Grid tests before 2.2 release. We, therefore, need this version available on cvmfs.
We'll fix cmake version and will send another pull request.

PR #1403, opened 2018-11-21
Add O2Suite and tag O2

dberzano:
* The O2Suite metapackage builds the whole O2 software components. It will be
  also used for testing the Pull Requests more accurately
* O2 has been bumped to 1.0.0. All software built with the `o2` default will
  use O2 1.0.0 as a reference. `o2-dataflow`, `o2-ninja`, `o2-dev-fairroot`
  will instead use `dev`

See O2-426.

### DIFF
--- a/defaults-o2-dataflow.sh
+++ b/defaults-o2-dataflow.sh
@@ -64,6 +64,7 @@ overrides:
       which cmake && case `cmake --version | sed -e 's/.* //' | cut -d. -f1,2,3 | head -n1` in [0-2]*|3.[0-9].*|3.10.*) exit 1 ;; esac
   O2:
     version: "%(short_hash)s%(defaults_upper)s"
+    tag: dev
 ---
 # This file is included in any build recipe and it's only used to set
 # environment variables. Which file to actually include can be defined by the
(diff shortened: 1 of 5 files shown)


dberzano:
@ktf @Barthelemy Any comments?

ktf:
We should actually change the CI to build O2Suite, rather than AliceO2.

Barthelemy:
Yes, it looks good. Thank you

dberzano:
> We should actually change the CI to build O2Suite, rather than AliceO2.
 Yes, that's exactly the idea; we need to merge this first. Will do right away.

awegrzyn:
Do we still need `flpproto`? It looks to me that `O2Suite` does more or less the same.

dberzano:
I have agreed with @Barthelemy a transitional period "just in case", but yes, it definitely can be removed IMHO. Oh, for the record: all O2 alidist PR tests are using O2Suite now (testing also, for instance, QualityControl and Readout) 🎉

Barthelemy:
Thanks a lot Dario. 
@awegrzyn I asked for the grace period because of our documentation that might still refer to "flpproto". Also our test machines need to be updated.

PR #1404, opened 2018-11-22
Align ROOT options of *next-root6

dberzano:
There is no reason to have different options. By using the same, most of the
cached builds for `next-root6` will also work for `user-next-root6`.


dberzano:
Not tested by the CI and does not affect production: merging it!

PR #1406, opened 2018-11-22
Add dependencies required by the ITS GPU builds

mpuccio:
Hi @dberzano ,

this header only package will be required as a dependency of O2 for building the ITS CUDA tracker.
No other PR depends on this so... low priority.

Cheers!
Max

### DIFF
--- a/cub.sh
+++ b/cub.sh
@@ -1,0 +1,24 @@
+package: cub
+version: "%(tag_basename)s"
+tag: v1.8.0
+source: https://github.com/NVlabs/cub.git
+---
+#!/bin/bash -e
+rsync -a --delete --exclude='**/.git' --delete-excluded "$SOURCEDIR/" "$INSTALLROOT/"
+
+# Modulefile
+MODULEDIR="$INSTALLROOT/etc/modulefiles"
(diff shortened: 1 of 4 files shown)


cub.sh
```diff
@@ -0,0 +1,25 @@
+package: cub
+version: "%(tag_basename)s"
+tag: v1.8.0
+source: https://github.com/NVlabs/cub.git
+
+---
+#!/bin/bash -e
+rsync -avc ${SOURCEDIR}/ ${INSTALLROOT}/
```
dberzano: Use literally:

```bash
rsync -a --delete --exclude='**/.git' --delete-excluded "${SOURCEDIR}/" .
```

Drop `-v` and `-c` (exp. the latter).

cub.sh
```diff
@@ -0,0 +1,25 @@
+package: cub
+version: "%(tag_basename)s"
+tag: v1.8.0
+source: https://github.com/NVlabs/cub.git
+
```
dberzano: Remove the empty line

mpuccio:
All comments addressed!

PR #1409, opened 2018-11-27
Fix JAliEn wrong GCC dependencies

dberzano:
GCC-Toolchain is a runtime dependency, not just a build dependency


dberzano:
@costing this fixes your XRootD problem, I am building a new JAliEn right away.

dberzano:
@costing says he wants to bump JAliEn as well, let's hold on (in any case, this set of defaults is not tested by the PR checker)

PR #1412, opened 2018-11-29
Introduce treelite

mpuccio:
Hi @dberzano ,
this is the changes I was discussing with you... now.
It is necessary to apply the BDT (from xgboost and lightGBM) inference on the GRID.

Cheers,
Max

### DIFF
--- a/treelite.sh
+++ b/treelite.sh
@@ -1,0 +1,44 @@
+package: treelite
+version: "%(tag_basename)s"
+tag: "2a12742269c1d3de553d9a12ff36bc2a5d874239"
+source: https://github.com/dmlc/treelite
+requires:
+  - "GCC-Toolchain:(?!osx)"
+build_requires:
+  - CMake
+  - "Xcode:(osx.*)"
+---
+#!/bin/bash -e
+
+rsync -a $SOURCEDIR/ src/
+pushd src
+  git submodule update --init --recursive
+popd
+
+cmake src                                   \
+  ${CMAKE_GENERATOR:+-G "$CMAKE_GENERATOR"} \
+  -DCMAKE_INSTALL_PREFIX="$INSTALLROOT"
+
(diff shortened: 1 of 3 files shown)


treelite.sh
```diff
@@ -0,0 +1,45 @@
...
+  - CMake
+  - "Xcode:(osx.*)"
+incremental_recipe: |
+  make ${JOBS:+-j$JOBS} install
+  mkdir -p $INSTALLROOT/etc/modulefiles && rsync -a --delete etc/modulefiles/ $INSTALLROOT/etc/modulefiles
+---
+#!/bin/bash -e
+
+pushd $SOURCEDIR
+git submodule update --init --recursive
```
dberzano: Please `rsync` to the current dir, and then run the submodule command. Please also indent the `pushd...popd` "block" - yes, it's not a block, and Bash's not a proper language, but still :-)

treelite.sh
```diff
@@ -0,0 +1,45 @@
...
+popd
+
+cmake $SOURCEDIR -DCMAKE_INSTALL_PREFIX="$INSTALLROOT"
+
+make ${JOBS+-j $JOBS}
```
dberzano: Use the generic `cmake --build` command instead of `make`

treelite.sh
```diff
@@ -0,0 +1,45 @@
...
+  make ${JOBS:+-j$JOBS} install
+  mkdir -p $INSTALLROOT/etc/modulefiles && rsync -a --delete etc/modulefiles/ $INSTALLROOT/etc/modulefiles
+---
+#!/bin/bash -e
+
+pushd $SOURCEDIR
+git submodule update --init --recursive
+popd
+
+cmake $SOURCEDIR -DCMAKE_INSTALL_PREFIX="$INSTALLROOT"
```
dberzano: Pass the generator if available (see the ninja recipes)

PR #1415, opened 2018-12-03
Align Run2/Run3 GCC and GEANT4s

dberzano:
As requested by @pzhristov


pzhristov:
Hi @dberzano 
Thank you very much! Would you add also CMake and vgm to the list of aligned packages?
Cheers,
Peter

geant4.sh
```diff
@@ -1,7 +1,7 @@
 package: GEANT4
 version: "%(tag_basename)s"
-tag: v4.10.01.p03-alice1
-source: https://github.com/alisw/geant4
+tag: v10.3.3
```
sawenzel: this is quite a big version change for Run2. Is this supported by the DPG? Will they have to rerun all the Geant4 validation productions? Do we have a validation procedure when bumping Geant4?
@alisw/dpg

dberzano: IIRC it was discussed a couple of weeks ago at the CB, @pzhristov can maybe confirm that we've even taken a decision to this respect? I let him comment on this change.

pzhristov: Yes, since Geant4 is not in production anyway, we better align all the versions. This will help to make also the comparisons in a meaningful way avoiding differences from the compiler.

dberzano:
@ktf I am doing that as a matter of fact. Hold on.

dberzano:
@pzhristov what is the GEANT4 version you suggest we use?

pzhristov:
Hi @dberzano,

I think we should test the build with the last tagged version that Ivana proposed, 10.4.2.

https://github.com/Geant4/geant4/releases/tag/v10.4.2

Cheers,
Peter

dberzano:
OK I have bumped GEANT4 (but I am not sure whether we should bump the VMC as well).

pzhristov:
@dberzano, We need geant4_vmc version v3-6-p1, the current one does not compile. There is also a problem with ninja generator in AliRoot, I see 

CMake Error at CMakeLists.txt:318 (enable_language):
  The Ninja generator does not support Fortran using Ninja version

    1.5.1

  due to lack of required features.  Kitware has implemented the required
  features but as of this version of CMake they have not been integrated to
  upstream ninja.  Pending integration, Kitware maintains a branch at:

    https://github.com/Kitware/ninja/tree/features-for-fortran#readme

  with the required features.  One may build ninja from that branch to get
  support for Fortran.


-- Configuring incomplete, errors occurred!

dberzano:
@pzhristov I do know about the ninja stuff. The ninja recipe takes exactly the steps you illustrate. However, I did not deliberately enable it, there must be a problem with my recipe that I am going to fix. Thanks for the heads-up on g4vmc!

pzhristov:
@dberzano The patch excluding ninja solves the compilation issues, now everything is fine on Ubuntu 16.04. The gun test is OK with the patched master of Root6.

PR #1417, opened 2018-12-04
Bump QC 

Barthelemy:


readoutcard.sh
```diff
@@ -18,7 +18,7 @@ incremental_recipe: |
   mkdir -p $INSTALLROOT/etc/modulefiles && rsync -a --delete etc/modulefiles/ $INSTALLROOT/etc/modulefiles
 ---
 #!/bin/bash -ex
-
+echo ""
```
awegrzyn: I think it's better to remove this line

Barthelemy: indeed :)

Barthelemy:
fixed

PR #1431, opened 2018-12-11
Fix gRPC recipe on Mac

teo:


grpc.sh
```diff
@@ -40,7 +40,7 @@ env -i HOME="$HOME"                             \
 
 # ldconfig won't work if we're not running make install as root, and that's ok,
 # we don't need it
-sed -i 's/ldconfig/true/' ./Makefile
+sed -i -e 's/ldconfig/true/' Makefile
```
dberzano: Unfortunately this won't work on macOS, where you need to provide a zero-length extension to `-i`. The way you wrote it creates `Makefile-e` as backup, since `-e` is seen as the extension.

One universal (BSD/GNU) way of fixing the issue is:

```bash
sed -i.deleteme -e 's/ldconfig/true/' Makefile
rm -f *.deleteme
```

Can you fix it please? Thanks!

teo: Sure, fixing immediately, thank you for the pointer :)

sawenzel:
This was approved. In reason for not merging it?

ktf:
Can't this be fixed by moving to a more recent version? Doesn't  gRPC support mac?

teo:
@ktf I don't think so, this issue has been going on for a long time and affects both Linux and Mac. I hope they'll pick a single good build tool eventually, but in the meantime we're stuck with the crazy broken mess that is the gRPC Makefile.

PR #1432, opened 2018-12-11
Coconut requires gRPC

teo:


Barthelemy:
Maybe there is something missing in the modulefile part but that should eventually be generated automatically.

PR #1439, opened 2018-12-18
Correctly pick up CXX standard in roounfold.sh

sawenzel:
Fixing compile error on Ubuntu 16.04 with `aliBuild build AliPhysics defaults=o2`. In this case, the compilation did not use (or pickup) the correct C++ standard and failed.
I believe that similar things should be done for every recipe using cmake.


sawenzel:
build error (in O2) unrelated

sawenzel:
Ping.

PR #1443, opened 2018-12-20
o2: require MCStepLogger

sawenzel:
Since the code was externalized, we need to mention this
as a dependency.


sawenzel:
@dberzano : Could you please review and merge?

ktf:
I am looking into the compilation error of O2.

sawenzel:
The O2 build seems to be fine.

PR #1447, opened 2018-12-20
Use libInfoLogger instead of InfoLogger in QC

awegrzyn:


Barthelemy:
ok qc has been bumped

awegrzyn:
Well not really, you bumped QC to `v0.9.4` and the build is still using `v0.9.3`

ktf:
Did you really bump qc? i see `v0.9.3` in `qualitycontrol.sh`.

Barthelemy:
Indeed, it seems I did not ? let me check and redo it if needed.

PR #1448, opened 2018-12-20
Adding RooUnfold to AliPhysics in build and module

mfasDa:
Enabling builds of AliPhysics with RooUnfold and
include it also in the modulefile. This is necessary
in order to have access to RooUnfold classes in
AliPhysics, needed in order to fill ND response
matrices on the grid.

### DIFF
--- a/aliphysics.sh
+++ b/aliphysics.sh
@@ -43,6 +43,7 @@ cmake "$SOURCEDIR"                                                 \
       ${CMAKE_BUILD_TYPE:+-DCMAKE_BUILD_TYPE="$CMAKE_BUILD_TYPE"}  \
       ${ALIEN_RUNTIME_ROOT:+-DALIEN="$ALIEN_RUNTIME_ROOT"}         \
       ${FASTJET_ROOT:+-DFASTJET="$FASTJET_ROOT"}                   \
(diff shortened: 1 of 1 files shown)


dberzano:
Hi @mfasDa, (cc'ing @pzhristov and @chiarazampolli), this is perfectly fine for me, given that it impacts only the ROOT 6 builds.

However, no new Grid tag will be created with it _unless there is an explicit urgency from the DPG_, as Christmas is coming 🙂

I just wanted to double check this with you.

mfasDa:
Hi @dberzano,

ROOT6 builds are absolutely fine for us. As RooUnfold is disabled for the release builds it shouldn't affect ROOT5 builds. For us having normal daily ROOT6 AN-tags with RooUnfold dependency would be sufficient.

Cheers

Markus

dberzano:
@mfasDa my point is that we are building AliRoot/AliPhysics using a _tagged_ version of alidist. The same tag, as per policy, is used for ROOT 5 and 6 builds, no exception is made.

I can merge, however this will not - on purpose - change the production versions we have on the Grid. This policy has been adopted for a couple of years now and we have it for keeping things simple and under control.

Out of consistency, if you want this and you need it urgently, a porting request should be opened and the DPG should agree both on the pull request and the urgency. I am pretty sure a new tag will be requested in any case, so just piggyback on that.

PR #1452, opened 2019-01-07
Bump quality_control

Barthelemy:


Barthelemy:
#1447

ktf:
Looks like the compilation error is actually compiler / runtime issue. I found a similar one:

https://bugs.llvm.org/show_bug.cgi?id=33222

I am trying to find a workaround on my setup.

ktf:
I am merging this since in any case we are not yet testing O2Suite.

PR #1454, opened 2019-01-08
Fix cgal and fastjet build with c++17 on MacOS

davidrohr:
@sawenzel : I had the same issue now, this PR fixes the build at least for cgal.
I thought it is simpler this way than forking cgal in github alisw.

Btw: There are some more c++17 problems in fastjet, which uses random_shuffle, which has been removed. Besides, AliRoot builds on macos with defaults o2 for me.

### DIFF
--- a/cgal.sh
+++ b/cgal.sh
@@ -1,5 +1,5 @@
 package: cgal
-version: "v4.6.3"
+version: "4.6.3"
 requires:
   - boost
 build_requires:
@@ -15,17 +15,15 @@ case $ARCHITECTURE in
     [[ ! $BOOST_ROOT ]] && BOOST_ROOT=`brew --prefix boost`
   ;;
 esac
-PKGID=35136
-URL="https://gforge.inria.fr/frs/download.php/file/${PKGID}/"
+URL="https://github.com/CGAL/cgal/releases/download/releases%2FCGAL-$PKGVERSION/CGAL-$PKGVERSION.tar.xz"
 
-curl -kLo cgal.tar.bz2 "$URL"
-tar xjf cgal.tar.bz2
+curl -kLo cgal.tar.xz "$URL"
+tar xJf cgal.tar.xz
 cd CGAL-*
 
 if [[ "$BOOST_ROOT" != '' ]]; then
   export LDFLAGS="-L$BOOST_ROOT/lib"
   export LD_LIBRARY_PATH="$BOOST_ROOT/lib:$LD_LIBRARY_PATH"
-  export DYLD_LIBRARY_PATH="$BOOST_ROOT/lib:$DYLD_LIBRARY_PATH"
 fi
 
 export MPFR_LIB_DIR="${MPFR_ROOT}/lib"
@@ -87,5 +85,4 @@ module load BASE/1.0 ${BOOST_VERSION:+boost/$BOOST_VERSION-$BOOST_REVISION}
 setenv CGAL_ROOT \$::env(BASEDIR)/$PKGNAME/\$version
 prepend-path PATH \$::env(CGAL_ROOT)/bin
 prepend-path LD_LIBRARY_PATH \$::env(CGAL_ROOT)/lib
-$([[ ${ARCHITECTURE:0:3} == osx ]] && echo "prepend-path DYLD_LIBRARY_PATH \$::env(CGAL_ROOT)/lib")
 EoF
--- a/defaults-o2.sh
+++ b/defaults-o2.sh
@@ -29,6 +29,10 @@ overrides:
     version: "%(commit_hash)s_O2"
   ROOT:
     tag: "v6-18-04" 
+  cgal:
+    version: "4.11"
+  fastjet:
+    tag: "v3.3.2_1.041-alice1"
 ---
 # This file is included in any build recipe and it's only used to set
 # environment variables. Which file to actually include can be defined by the
--- a/fastjet.sh
+++ b/fastjet.sh
@@ -24,7 +24,7 @@ printf "void main() {}" | c++ -xc ${BOOST_ROOT:+-L$BOOST_ROOT/lib} -lboost_threa
   || BOOST_LIBS="${BOOST_ROOT+-L$BOOST_ROOT/lib} -lboost_thread-mt"
 BOOST_LIBS="$BOOST_LIBS -lboost_system"
 
-rsync -a --delete --cvs-exclude $SOURCEDIR/ ./
+rsync -a --delete --cvs-exclude --exclude .git $SOURCEDIR/ ./
 
 # FastJet
 pushd fastjet
@@ -79,5 +79,4 @@ setenv FASTJET_ROOT \$::env(BASEDIR)/$PKGNAME/\$version
 prepend-path PATH \$::env(FASTJET_ROOT)/bin
 prepend-path LD_LIBRARY_PATH \$::env(FASTJET_ROOT)/lib
 prepend-path ROOT_INCLUDE_PATH \$::env(FASTJET_ROOT)/include
-$([[ ${ARCHITECTURE:0:3} == osx ]] && echo "prepend-path DYLD_LIBRARY_PATH \$::env(FASTJET_ROOT)/lib")
 EoF


cgal.sh
```diff
@@ -66,6 +66,14 @@ cmake . \
       -DCGAL_IGNORE_PRECONFIGURED_MPFR:BOOL=YES \
       ${BOOST_ROOT:+-DBoost_NO_SYSTEM_PATHS:BOOL=TRUE -DBOOST_ROOT:PATH="$BOOST_ROOT"}
 
+sed -ibackup "s/register int/int/g" src/CGAL/Turkowski_numeric_solver.cpp
```
dberzano: Use: `sed -ideleteme -e 's/one/two/g' filename` and then `rm -f *.deleteme`. It's the only way to make it work reliably both with GNU sed and BSD's (e.g. on macOS).

sawenzel:
Very neat idea.

dberzano:
I have noticed that the WIP check is going a bit bananas lately. I am merging this one.

dberzano:
(OK, WIP got unstuck. Let's wait for the greens to come again then.)

davidrohr:
@dberzano @sawenzel : Should we fix fastjet as well, it is the only package failing to compile on MacOS with C++17? (For the moment, I am just using --disable fastjet)

dberzano:
Yes, in the same PR (this one) possibly. Is the change similar? Fastjet has an external repository where we import the code:

https://github.com/alisw/fastjet

so we need to patch it there and bump the version number in the recipe rather than monkeypatching in the recipe.

dberzano:
OK in the meanwhile I've put this one on hold (it's WIP again). Maybe Fastjet v3.2.2 (see [here](http://fastjet.fr/)) fixes the issue?

davidrohr:
@dberzano : It needs both the new version of CGAL and the new version of fastjet.
In addition, fastjet needs a patch to fix a misdetection during configure on MacOS.
I just created a PR to alisw/fastjet#3, which bumps the version and removes the deprecated interface.

I changed this PR to bump the cgal and fastjet versions, and in addition changed the cgal recipe to fetch the files from the github git repo instead of downloading a tarball.
Hope this is how it is supposed to be.

You'll have to create a tag in the fastjet repo first, and then change this PR accordingly to use the correct fastjet tag, I have enabled edits for maintainers.
If you are not satisfied with the new recipe, please just change it as well.

Cheers
David

davidrohr:
Note: Do not merge yet, the new fastjet version seems to break AliPhysics

davidrohr:
@dberzano @pzhristov @sawenzel : Hi guys,
as mentioned here: alisw/fastjet#3 , there is a problem with the AliPhysics EMCAL Jet analysis that is not compatible to the current fastjet. This stalls an update of fastjet, and therefore also this PR here, which would make AliRoot compatible to C++17. As far as I have seen, this is the last showstopper.
Whom should we ask to have a look, or do we forget about C++17 for AliRoot?

sawenzel:
I think this should be fixed. @mfasDa: Could someone in EMCAL take a look how to update the jet-code to newer version of fastjet?

mfasDa:
Hi @sawenzel,

we can certainly take care about the compatibility but before moving to a new fastjet version for the release and the ROOT6 build we would first need to validate that results are in agreement with the current fastjet version, so for the grid tags it would be nice if you could stay with the current fastjet for a bit of time.

Thanks!

Cheers

Markus

davidrohr:
Hi @mfasDa : Sure, we cannot switch yet anyway, as it does not compile with AliPhysics. Can you check this PR: alisw/fastjet#3 , which explains the problems we are seeing?
Once you have updated AliPhysics to be compatible, and once it is verified that the new fastjet works correctly, we can do the switch.
Cheers
David

lcunquei:
Hi,  I think that the best solution is to keep the current version of fastjet in AliPhysics and disable fastjet in O2 (is fastjet used in O2 at the moment? i don't think so. Updating fastjet in AliPhysics requires validating  and it is not the right moment now from the physics point of view. cheers.

davidrohr:
That is basically what we currently have. The problem is that the current version of fastjet uses deprecated features which breaks compilation of AliRoot / AliPhysics with C++17.
So either we fix these deprecations manually, or we switch to the current fastjet version, or we are stuck with C++11.

lcunquei:
Hi and why do we need to compile AliPhysics against C++17. Can't we hve C++17 for O2 and C++11 for AliPhysics?

davidrohr:
This is not urgently needed, but in the long run this might be very convenient. And as fastjet seems to be the only showstopper, it might be worth to invest some time.

mfasDa:
Fixes in AliPhysics handled here: https://github.com/alisw/AliPhysics/pull/8368

sawenzel:
It is definitely needed to compile AliRoot and O2 using the same defaults for the purpose of simulation comparison and reusing some Monte Carlo generators available in AliRoot. (But this does not actually need AliPhysics)

sawenzel:
@mfasDa : Thanks. So if we keep the old AliPhysics in production and use the patch with `--defaults o2`  we should be completely ok.

mfasDa:
@sawenzel Yes, however in this case the build recipes need to be changed:

- fasjet.sh: go revert to current fastjet version
- defaults_o2.sh (and other o2-based defaults): specify new fasjet version

This is also what we would prefer. Would it be possible to modify the pull request accordingly?

Cheers

Markus

davidrohr:
For fastjet we don't need to change the recipe, we can simply specify the fastjet version in the defaults_o2.sh, that should be easy. But what would we do with cgal. For cgal, my PR changes the recipe to use their new git repo, instead of downloading a tarball. I guess we can also fetch the old version from git, but this should be tested.

davidrohr:
And we can btw still apply the AliPhysics patch since it is backward-compatible?

mfasDa:
Yes, the AliPhysics patch supports both versions (based on the FASTJET_VERSION_NUMBER). It actually passed a the build test using the existing fastjet.

davidrohr:
OK, I updated this PR: Fastjet builds the old version by default, but takes the sources from their new git repo. This should keep AliRoot/AliPhysics as they are.

With O2 defaults, we should have the new fastjet/cgal versions.
This should compile with AliRoot/AliPhysics and O2 defaults once alisw/AliPhysics#8368 and alisw/fastjet#3 are merged, and someone needs to create a new tag in alisw/fastjet for "v3.3.2_1.041"

Anyway, in the long rung we could really update the fastjet version, such that we can use C++17 consistently.

davidrohr:
@dberzano : I have updated this PR to use the fastjet tag you created (v3.3.2_1.041-alice1) for the O2 defaults. This leaves AliRoot / AliPhysics untouched, but will enable us to build AliRoot with O2 defaults and C++17.
So basically, we can merge this right now instead of waiting for Physics Coordination Approval for alisw/alidist#1494
When the new fastjet version is approved, we can bump fastjet in general also for AliRoot/AliPhysics without O2 defaults.

ktf:
Is this still relevant?

pzhristov:
I think we should merge this pull request. Now I am trying to have C++17 for AliRoot/AliPhysics, otherwise I do not have enough space on my laptop to have both.

davidrohr:
@ktf @pzhristov : This PR would bump fastjet and cgal only for O2 defaults, so it would leave the current AliRoot/AliPhysics setup unchanged.
@dberzano said he'd like to instead first check the physics of the new version, and then bump the version in general, not only for O2 defaults.
I do not really care. I use AliRoot with O2 defaults, so I use my PR locally anyway.

mfasDa:
@ktf @pzhristov The issue is still relevant, fastjet doesn't compile under macOS under the c++17 standard. As mentioned before by @lcunquei and me before making the new fastjet the default in the release and root6 builds it would first require a set of tests that results are consistent, while for the o2 defaults it is not really relevant as they are (normally) not used for user analyses.

pzhristov:
@mfasDa @lcunquei Is it difficult to check the new versions of these packages? I agree it is a good idea to have single version both for AliRoot/AliPhysics and O2.

lcunquei:
Hello,there are several papers/analysis undergoing final checks before publication and a change of fastjet would be disruptive since validation tests would be needed. Validation tests are not difficult but are not really convenient at this moment, since people are swamped by other physics problems. The best is to wait a bit to have the single version.

davidrohr:
@sawenzel @pzhristov @ihrivnac : This has been open since almost a year now, and nothing has happened in the fastjet validation.

This PR as is does not change fastjet for AliPhysics, but it just bumps it for O2 defaults, to become compatible to C++ 17.

I have just rebased it and fixed the collisions. I propose we merge it as is for now. Once it is validated, we can remove the new versions for the O2 defaults and use them generally.

davidrohr:
@pzhristov : Can we merge this, as we discussed in the WP3 meeting 2 weeks ago?

ktf:
Is the github repository still huge as it was? If yes, we should think about using a stripped off mirror.

davidrohr:
@ktf : CGAL seems indeed rather large. If that is a problem, we can switch back to fetching a tarball. But then, I't fetch a release tarball from github, instead from this weird URL:
```
PKGID=35136
URL="https://gforge.inria.fr/frs/download.php/file/${PKGID}/"
```

ktf:
Yes, agreed. We had https://github.com/alisw/CGAL at some point, I do not remember why it was changed to use the tarball. Maybe because the tarball and the git differ in the way they need to be built?

davidrohr:
github only has tar.xz tarballs. I suppose we can assume xz being available?

jolopezl:
This PR actually keeps fastjet broken (https://github.com/alisw/alidist/issues/1866).

PR #1464, opened 2019-01-15
Add support for arrow 0.12.0

ktf:
* Also enable support for parquet, which is now shipped with arrow
* Build our own lz4 and thrift packages, rather than consider them
  system dependencies.


ktf:
Needs https://github.com/AliceO2Group/AliceO2/pull/1595 to be merged (done now).

PR #1467, opened 2019-01-16
DDS: workaround issue with wn_bin on macOS

ktf:
See also discussion at https://alice-talk.web.cern.ch/t/o2-build-failing-on-mojave-in-dds/198/4

### DIFF
--- a/dds.sh
+++ b/dds.sh
@@ -25,7 +25,14 @@ cmake $SOURCEDIR                                                         \
 JOBS=$((${JOBS:-1}*2/5))
 [[ $JOBS -gt 0 ]] || JOBS=1
 
+# This is needed because https://github.com/Homebrew/homebrew-core/pull/35735
+# seems to break the creation of the tarball.
+case $ARCHITECTURE in
+  osx*)
 make -j$JOBS wn_bin
+  ;;
+esac
+
 make -j$JOBS install
 
 MODULEDIR="$INSTALLROOT/etc/modulefiles"


ktf:
@pzhristov This is the issue you were also facing. It's indeed some change in homebrew which broke the `wn_bin` creation.

@AnarManafov @dennisklein @AndreyLebedev is this known?

AnarManafov:
@ktf, I would recomend to build BOOST without icu support. I am not sure you use it at all and it also makes boost libs bigger.
The problem is that it is not exactly DDS, but cmake's prerequisites functions can't find this library.

If you still want to use icu for some reason, then you can try to define DDS_LD_LIBRARY_PATH with the path to the icu library, before calling cmake on DDS.

ktf:
I do not want to particularly use ICU, but it comes out of the box with homebrew. I will think about `DDS_LD_LIBRARY_PATH`.

dennisklein:
Apparently, it is all in the name of a better UX, see

https://github.com/Homebrew/homebrew-core/issues/31510
https://github.com/Homebrew/homebrew-core/pull/35735

tldr; Homebrew changed the default boost configuration and removed the option to control whether one wants icu support or not. Judging from all the sad smileys and thumbs down votes, these decisions by homebrew are not that famous. I had similar issues, when I tried to contribute an option for the zeromq formula, to build with zmq's cmake build system instead of autotools/make. But it got rejected (with the same argument, UX).

AnarManafov:
Another solution is to get a habit to compile boost on your own. I personally don't use precompiled versions. BOOST is too important for us to let others compile it with their weird option :D

PR #1474, opened 2019-01-18
pythia: version bump

ktf:
Pick up correct initialisation fix.


ktf:
This is currently broken, so I merge it.

PR #1476, opened 2019-01-22
Thrift: fix issue when OpenSSL comes from the system

ktf:


ktf:
@awegrzyn can you check if this fixes your issue?

awegrzyn:
Yup it does, thanks

PR #1478, opened 2019-01-22
Add aliECS GUI recipe

awegrzyn:


ecsg.sh
```diff
@@ -0,0 +1,40 @@
...
+proc ModulesHelp { } {
+  global version
+  puts stderr "ALICE Modulefile for $PKGNAME $PKGVERSION-@@PKGREVISION@$PKGHASH@@"
+}
+set version $PKGVERSION-@@PKGREVISION@$PKGHASH@@
+module-whatis "ALICE Modulefile for $PKGNAME $PKGVERSION-@@PKGREVISION@$PKGHASH@@"
+# Dependencies
+module load BASE/1.0 ${NODE_VERSION:+node/$NODE_VERSION-$NODE_REVISION}
+# Our environment
+setenv ECG_ROOT \$::env(BASEDIR)/$PKGNAME/\$version
```
ktf: Do you really need this? If yes, shouldn't it be `ECS**G**_ROOT`

awegrzyn: Not really, I actually plan to drop it (in `qcg` as well), for the time being just trying to make RPM generation work

awegrzyn:
@ktf

PR #1480, opened 2019-01-23
arrow: disable parquet support for now

ktf:
Thrift dependency requires a newer version of bison which is not
available on the Mac build machine.


ktf:
@sawenzel this should bring the O2 mac builds back in business.

PR #1491, opened 2019-01-31
Fix typo cmake -> CMake in build requirements

mfasDa:


dennisklein:
Not really related to this PR, but dont know any better place to put:

I have seen couple of recipes with

```
cmake --build . -- ${JOBS+-j $JOBS} install
```

or similar, shouldn't it rather be

```
cmake --build . --target install -- ${JOBS+-j $JOBS}
```

?

PR #1494, opened 2019-02-05
WIP Bump FastJet + FJContrib to v3.3.2 + 1.041

dberzano:


dberzano:
@davidrohr @sawenzel this must not be merged until we have the approval from our Physics Coordination.

@davidrohr I had to redo your work on alisw/fastjet: upstream, untouched versions are imported on master, and tagged; then, from each version, we spring a new `alice/` branch with our own patches. For reference: alisw/fastjet#3. In any case bumping FJ versions as said is not a solution to the problem. The Physics changes, therefore proper testing and approval has to be performed by the PhysCoord.

mfasDa:
@dberzano as discussed in #1454 by @lcunquei and me for the builds on the grid it would first need a set of tests to validate that the results stay consistent. This could be done if we would get a test build on the grid available for the lego train. Otherwise keeping the fastjet version would be preferred as it is used in a couple of analyses that are on track for publication.

ktf:
Is this still relevant?

davidrohr:
We have merged #1454 instead. For the future, we still want to bump fastjet generally, when our physics group finally validates it...
But I think we can close this.

mfasDa:
For what concerns the general bump of the fastjet version the fastjet authors currently work on implementing a patch fixing the issue with the static linking of libgmp and fastjet-config --libs, responsible for the issues we have with the MC generators relying on fastjet. This fix should be part in the newest release of fastjet - one should probably wait for this for a general validation and bumping of the fastjet version.

ktf:
Ok, I am closing this, we can reopen once that is done.

PR #1498, opened 2019-02-07
Add librdkafka recipe

awegrzyn:


dberzano:
The error is unrelated (and this test is not yet mandatory as per [O2-436](https://alice.its.cern.ch/jira/browse/O2-436)) so we can merge.

PR #1499, opened 2019-02-08
Bump AliTPCCommon to v2.3.2.1 - Update flatOject with verison from Ruben

davidrohr:
Minor change in AliTPCCommon containing new version of FlatObject header, should be used by the LUT class AliceO2Group/AliceO2#1263


davidrohr:
@dberzano : The O2 ctest failure is unrelated, can we merge this?

PR #1503, opened 2019-02-18
Introduce AliEn-ROOT-Legacy as external TAlien ROOT plugin

Atlantic777:
This pull request introduces the Old-AliEn-ROOT, TAlien as an external ROOT plugin. The Old-AliEn-ROOT was created by extracting the `net/alien` directory from the ROOT repository. This modification (still) does not affect any code in upstream ROOT repository and is done just on our side.

This modification affects default ROOT5 and ROOT6 builds but is functionally equivalent to previous builds. This reorganization brings the TAlien to the same version for both ROOT5 and ROOT6. The commit history is preserved and includes the recent patch in TAlienFile by Miguel (@mimarpe).

This reorganization gives us more control over TAlien code and lets us make further improvements in TAlien required for future JAliEn-ROOT (TJAlien) integration.

The Old-AliEn-ROOT repository can be found here: https://gitlab.cern.ch/jalien/old-alien-root

The pull request is in WIP state until Volodymyr and I finish with our tests. We expect to finish testing by the end of week. Meanwhile, we invite everyone to review and test this setup. Thanks!


dberzano:
@Atlantic777 any chance we can call it AliEn-ROOT-Legacy instead of "Old"? p.s. the O2 test timed out at some point, it will clear out automatically I hope

Atlantic777:
@dberzano ok, no problem, I have just renamed it. By the way, we should not forget that we need to distribute AliEn-ROOT-Legacy on CernVM-FS. I guess this needs to be enabled somewhere.

dberzano:
@Atlantic777 we deploy AliPhysics. All the dependencies are automatically included. So, no need to enable it anywhere :-)

Atlantic777:
The build log for the failed Ci task is missing (build/AliRoot/alidist). Can we get more info what happened? It builds fine on my machine.

Atlantic777:
So, I see that AliRoot and O2 CI tasks are now green, except for O2Suite on macOS. I see in the logs that macOS CI task failed due to some O2 simulation tests. This doesn't seem to be related to this PR. I hope I didn't miss any other problems in the log.

We finished testing this setup from our side and everything looks good. However, I'm keeping the WIP status until we have a new AliRoot tag. Once we have commits from https://github.com/alisw/AliRoot/pull/889 included in an AliRoot tag, we can proceed with merging this PR.

Atlantic777:
This pull request can be merged now.

alien-root-legacy.sh
```diff
@@ -0,0 +1,59 @@
...
+  global version
+  puts stderr "ALICE Modulefile for $PKGNAME $PKGVERSION-@@PKGREVISION@$PKGHASH@@"
+}
+set version $PKGVERSION-@@PKGREVISION@$PKGHASH@@
+module-whatis "ALICE Modulefile for $PKGNAME $PKGVERSION-@@PKGREVISION@$PKGHASH@@"
+# Dependencies
+module load BASE/1.0 ${GCC_TOOLCHAIN_VERSION:+GCC-Toolchain/$GCC_TOOLCHAIN_VERSION-$GCC_TOOLCHAIN_REVISION} ROOT/${ROOT_VERSION}-${ROOT_REVISION}
+
+# Our environment
+setenv ALIEN_ROOT_LEGACY_ROOT \$::env(BASEDIR)/$PKGNAME/\$version
```
ktf: Do you really need this to be an environment variable? How about just using `set`?

Atlantic777: As far as I can tell it doesn't matter for us. I just copy-pasted from the JAliEn-ROOT recipe that was already there. I updated this and dropped it from an environment variable to a regular variable, both in AliEn-ROOT-Legacy and JAliEn-ROOT.

dberzano:
OK give me a minute to have a quick look!

root.sh
```diff
@@ -39,6 +39,11 @@ incremental_recipe: |
 
   cmake --build . -- ${JOBS:+-j$JOBS} install
   mkdir -p $INSTALLROOT/etc/modulefiles && rsync -a --delete etc/modulefiles/ $INSTALLROOT/etc/modulefiles
+
+  echo "Building with external Alien plugin, removing interal plugin handlers."
+  rm -v "$INSTALLROOT/etc/plugins/TGrid/P010_TAlien.C"
```
dberzano: This will break the day they remove AliEn from ROOT. Changing it to `-vf` for forward compatibility. I am taking care of it.

jalien-root.sh
```diff
@@ -1,9 +1,13 @@
 package: JAliEn-ROOT
 version: "%(tag_basename)s"
-tag: "0.1.0"
-source: https://gitlab.cern.ch/jalien/jalien-root
+tag: "0.2.0"
+source: https://gitlab.cern.ch/jalien/jalien-root.git
 requires:
   - ROOT
+env:
```
dberzano: This is not effective: `env:` does not make envvars available in this recipe at build time. It makes, however, those variables available when building all the recipes that depend on it, incorrectly overriding the main defaults. Removing this section. I am taking care of it.

alien-root-legacy.sh
```diff
@@ -0,0 +1,59 @@
+package: AliEn-ROOT-Legacy
+version: "%(tag_basename)s"
+tag: "0.0.4"
```
Atlantic777: AliEn-ROOT-Legacy version bump needed to support new location for user tokens.

dberzano:
For the record I have tested it locally and AliEn works, thanks. Let's wait for the bump.

alien-root-legacy.sh
Atlantic777: AliEn-ROOT-Legacy actually doesn't need to be updated.

shahor02:
Hi @dberzano @Atlantic777 , not sure if you follow the alice-project-analysis-task-force@cern.ch : 
@akubera wrote that now the alien is not installed at all for default aliroot installation. BTW, the same is true for o2-defaults, where AliEn-Runtime and ApMon-CPP are disabled, w/o obvious reason.

akubera:
I think the issue was just `defaults-user-root6` overrides the AliRoot requires, and AliEn-ROOT-Legacy was missing from the list. I'm rebuilding with [the missing line](https://github.com/alisw/alidist/compare/master...akubera:master) added. I will edit this comment after I verify I can do alien-token-init again successfully.

PR #1507, opened 2019-02-20
Data Distribution repository for AliceO2

ironMann:
Adding:
- DataDistribution recipe (`datadistribution.sh`)
- ~~Defaults with shorter dependency chain for the DataDistribution~~

### DIFF
--- a/datadistribution.sh
+++ b/datadistribution.sh
@@ -1,0 +1,68 @@
+package: DataDistribution
+version: "%(tag_basename)s"
+tag: v0.5.1
+requires:
+  - boost
+  - FairLogger
+  - FairMQ
+  - Monitoring
+  - O2
+build_requires:
+  - CMake
+  - "GCC-Toolchain:(?!osx)"
+  - ms_gsl
+source: https://github.com/AliceO2Group/DataDistribution
+incremental_recipe: |
+  make ${JOBS:+-j$JOBS} install
+  mkdir -p $INSTALLROOT/etc/modulefiles && rsync -a --delete etc/modulefiles/ $INSTALLROOT/etc/modulefiles
+---
+#!/bin/bash -ex
+
+case $ARCHITECTURE in
+    osx*) [[ ! $BOOST_ROOT ]] && BOOST_ROOT=$(brew --prefix boost);;
+esac
+
+cmake $SOURCEDIR                                              \
+      ${CMAKE_GENERATOR:+-G "$CMAKE_GENERATOR"}               \
+      -DCMAKE_INSTALL_PREFIX=$INSTALLROOT                     \
+      ${BOOST_ROOT:+-DBOOST_ROOT=$BOOST_ROOT}                 \
+      ${FAIRLOGGER_ROOT:+-DFairLogger_ROOT=$FAIRLOGGER_ROOT}  \
+      ${FAIRMQ_ROOT:+-DFairMQ_ROOT=$FAIRMQ_ROOT}              \
+      ${O2_ROOT:+-DO2_ROOT=$O2_ROOT}                          \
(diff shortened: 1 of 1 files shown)


ironMann:
No objections. Updated

datadistribution.sh
```diff
@@ -0,0 +1,67 @@
...
+module load BASE/1.0                                                                              \\
+            ${BOOST_VERSION:+boost/$BOOST_VERSION-$BOOST_REVISION}                                \\
+            ${GCC_TOOLCHAIN_VERSION:+GCC-Toolchain/$GCC_TOOLCHAIN_VERSION-$GCC_TOOLCHAIN_REVISION}\\
+            ${MONITORING_VERSION:+Monitoring/$MONITORING_VERSION-$MONITORING_REVISION}            \\
+            ${FAIRLOGGER_VERSION:+FairLogger/$FAIRLOGGER_VERSION-$FAIRLOGGER_REVISION}            \\
+            ${FAIRMQ_VERSION:+FairMQ/$FAIRMQ_VERSION-$FAIRMQ_REVISION}                            \\
+            ${O2_VERSION:+O2/$O2_VERSION-$O2_REVISION}
+
+# Our environment
+setenv DATADISTRIBUTION_ROOT \$::env(BASEDIR)/$PKGNAME/\$version
```
ktf: Do you really need this to be in the environment?

datadistribution.sh
```diff
@@ -0,0 +1,67 @@
...
+  - boost
+  - FairLogger
+  - FairMQ
+  - Monitoring
+  - O2
+build_requires:
+  - CMake
+  - "GCC-Toolchain:(?!osx)"
+  - ms_gsl
+  - CodingGuidelines
```
ktf: Why do you need the CodingGuidelines?

ironMann: leftover from template copy/paste...
Removed

datadistribution.sh
ironMann: For now I only _need_ to fudge the PATH.
updated

ktf: can you simply use set? That way it does not pollute the environment. I know the variable is under our control, but I would still prefer we minimise exposing `_ROOT` everywhere.

ironMann: Sure, I already changed it as you described

datadistribution.sh
```diff
@@ -0,0 +1,67 @@
...
+      -DCMAKE_INSTALL_PREFIX=$INSTALLROOT                     \
+      ${BOOST_ROOT:+-DBOOST_ROOT=$BOOST_ROOT}                 \
+      ${FAIRLOGGER_ROOT:+-DFairLogger_ROOT=$FAIRLOGGER_ROOT}  \
+      ${FAIRMQ_ROOT:+-DFairMQ_ROOT=$FAIRMQ_ROOT}              \
+      ${O2_ROOT:+-DO2_ROOT=$O2_ROOT}                          \
+      ${MONITORING_ROOT:+-DMonitoring_ROOT=$MONITORING_ROOT}  \
+      -DCMAKE_EXPORT_COMPILE_COMMANDS=ON
+
+cp ${BUILDDIR}/compile_commands.json ${INSTALLROOT}
+make ${JOBS+-j $JOBS} install
```
dberzano: While we are at it: can you use a generic `cmake` command in case we use a different generator (like Ninja)?

This would mean to add:

```
cmake ... ${CMAKE_GENERATOR:+-G "$CMAKE_GENERATOR"} ...
```

to the cmake list of options, and then replacing the `make install` line with:

```
cmake --build . -- ${JOBS+-j $JOBS} install
```

ironMann: OK. updated

Issue #1510, opened 2019-02-25
Clang recipe : clang-format load fails


Barthelemy:
On CC7 : 
1. `aliBuild build --defaults o2 Clang`
2. `alienv load Clang/latest  # or enter`
3. `clang-format -style=file -i *.cpp`
3. Get : 
```
[Clang/latest] ~/dev/alice/QualityControl $> clang-format -style=file -i *.cpp
clang-format: /lib64/libstdc++.so.6: version `GLIBCXX_3.4.21' not found (required by clang-format)
clang-format: /lib64/libstdc++.so.6: version `GLIBCXX_3.4.20' not found (required by /local/alice/sw/slc7_x86-64/Clang/7.0.0-1/bin/../lib/libLLVMSupport.so.7)
clang-format: /lib64/libstdc++.so.6: version `GLIBCXX_3.4.22' not found (required by /local/alice/sw/slc7_x86-64/Clang/7.0.0-1/bin/../lib/libLLVMSupport.so.7)
clang-format: /lib64/libstdc++.so.6: version `GLIBCXX_3.4.21' not found (required by /local/alice/sw/slc7_x86-64/Clang/7.0.0-1/bin/../lib/libLLVMSupport.so.7)
clang-format: /lib64/libstdc++.so.6: version `GLIBCXX_3.4.20' not found (required by /local/alice/sw/slc7_x86-64/Clang/7.0.0-1/bin/../lib/libclangBasic.so.7)
clang-format: /lib64/libstdc++.so.6: version `GLIBCXX_3.4.21' not found (required by /local/alice/sw/slc7_x86-64/Clang/7.0.0-1/bin/../lib/libclangBasic.so.7)
clang-format: /lib64/libstdc++.so.6: version `GLIBCXX_3.4.21' not found (required by /local/alice/sw/slc7_x86-64/Clang/7.0.0-1/bin/../lib/libclangFormat.so.7)
clang-format: /lib64/libstdc++.so.6: version `GLIBCXX_3.4.21' not found (required by /local/alice/sw/slc7_x86-64/Clang/7.0.0-1/b
```

bazinski:
you need the compiler as well to get the corresponding glibc.

for me:
alienv enter Clang/latest,GCC-Toolchain/latest

Barthelemy:
Thanks for the prompt reply. I would say that Clang should therefore depend on GCC-Toolchain in the recipe. If you agree with this, I can prepare the fix.

bazinski:
[...]

PR #1511, opened 2019-02-25
Add GCC as dependency of clang

Barthelemy:
#1510


Barthelemy:
@ktf Yes, this is a good idea. I just don't know how to do that right away. This being said, even if it is cleaner, it would not change much for the end users : they build/download any ways GCC.

ktf:
agreed.

PR #1512, opened 2019-02-25
Bump Control to v0.5

teo:


teo:
> -1. We do not want to have go as a mandatory dependency for InfoLogger, IMHO. Could you make that optional somehow, e.g. by moving the go bindings of infologger to a separate package?

v0.6 already depends on libInfoLogger instead of InfoLogger, but since v0.5 is already tagged I'd like to keep the InfoLogger dependency only for this version.
Perhaps we could build InfoLogger without Go+SWIG and libInfoLogger with it?

ktf:
Nope, the opposite I would not object (InfoLogger in Go, libInfoLogger without), but asking people in O2 to forcefully depend on Go is not an option at the moment, IMHO.

teo:
> Nope, the opposite I would not object (InfoLogger in Go, libInfoLogger without), but asking people in O2 to forcefully depend on Go is not an option at the moment, IMHO.

Fine with me, I'll revert the change to libInfoLogger and keep depending on InfoLogger.

ktf:
@dberzano logs are missing...

teo:
@dberzano @ktf Is someone looking into this?

ktf:
I've approved the changes. Could you trigger a rebuild?

ktf:
Ping @alisw/dataflow-experts

PR #1513, opened 2019-02-26
Python3 environment for MachineLearning Framework

svallero:
Python3 is supposed to work both on SLC6 and CentOS7. The ML modules work only on CentOS7, due to too old system libraries (i.e. glibc) on SLC6.


ktf:
Closing this as most / all of the needs are addressed (?) by the python3 migration.

PR #1516, opened 2019-02-27
Do not require the full InfoLogger for readout

ktf:


ktf:
This requires a minor modification also in Readout and ReadoutCard.

readout.sh
```diff
@@ -23,16 +23,16 @@ case $ARCHITECTURE in
...
+      ${BOOST_VERSION:+-DBOOST_ROOT=$BOOST_ROOT}                         \
+      ${COMMON_O2_VERSION:+-DCommon_ROOT=$COMMON_O2_ROOT}                \
+      ${MONITORING_VERSION:+-DMonitoring_ROOT=$MONITORING_ROOT}          \
       ${CONFIGURATION_VERSION:+-DConfiguration_ROOT=$CONFIGURATION_ROOT} \
-      ${READOUTCARD_VERSION:+-DReadoutCard_ROOT=$READOUTCARD_ROOT} \
-      ${INFOLOGGER_VERSION:+-DInfoLogger_ROOT=$INFOLOGGER_ROOT} \
-      ${FAIRROOT_VERSION:+-DFAIRROOTPATH=$FAIRROOT_ROOT}      \
-      ${FAIRROOT_VERSION:+-DFairRoot_DIR=$FAIRROOT_ROOT}      \
+      ${READOUTCARD_VERSION:+-DReadoutCard_ROOT=$READOUTCARD_ROOT}       \
+      ${LIBINFOLOGGER_VERSION:+-DInfoLogger_ROOT=$LIBINFOLOGGER_ROOT}    \
```
teo: Is this correct? Looking at the libInfoLogger recipe, it looks like it's still exporting `INFOLOGGER_ROOT` instead of `LIBINFOLOGGER_ROOT`.

ktf: Yes. `LIBINFOLOGGER_ROOT` is what alibuild generates automatically at build time and it's not available at runtime.

Barthelemy:
@sy-c 

To me this seems ok.

ktf:
Notice the fix in the header detection business. Do we need new tags for this? Could you please take care for them?

ktf:
An alternative solution would be to also copy InfoLogger.h in `libInfoLogger`, as you prefer.

sy-c:
Readout v0.18 (#1517) has the proper cmake fix using infologger target.

PR #1520, opened 2019-03-02
Move ROOT to v6-14-08

ktf:


aphecetche:
@ktf, out of curiosity, why not the  prod version  6-16-00 ?

ktf:
Nothing really... I was trying to solve an issue which is apparently fixed in v6-14-08. We can also move to 6-16-00 if @dberzano deems it's ok.

aphecetche:
In fact I pushed the curiosity a bit further and it might not be completely straightforward.

First, I had to add the `-DPYTHON_EXECUTABLE` to the root recipe (otherwise CMake got confused with my python installation).

And then : 

```
/Users/laurent/alice/sw/SOURCES/O2/1.0.0/0/Framework/Core/src/RCombinedDS.cxx:39:10: fatal error: 'ROOT/RDFUtils.hxx' file not found
#include <ROOT/RDFUtils.hxx>
```

So, staying with 6.14 was maybe a smart move after all ;-)

ktf:
Could you provide a PR of how you actually fixed the `-DPYTHON_EXECUTABLE` business? The other one is a typo on my side and AliceO2Group/AliceO2#1752 should fix it.

pzhristov:
I think we have to move to v6-16-00 and use it both for O2 and AliRoot/AliPhysics.

aphecetche:
@ktf well, what I've done on my Mac was a quick hack : specify -DPYTHON_EXECUTABLE=/usr/local/bin/python3 . Don't think that's general enough ;-) 

Also, on a CentOS7 machine, compilation of Root 6.16.00 works without any modification of the recipe, as far as I can tell (tried on a dev FLP)

ktf:
I am merging this since it's a trivial fix and it is being applied manually by a user. We can move to `v6-16-00` once @dberzano is back.

PR #1529, opened 2019-03-07
bump readout v0.19.1

sy-c:


sy-c:
The osx issue in the failing test has been fixed. How to trigger again a test ?

ktf:
I retriggered it by changing some cosmetics..

sy-c:
I don't have the impression the check was relaunched, is there a way to see the corresponding timestamp?
After the fix, it was tested to build ok on macosx without libnuma.

ktf:
The tests were relaunched for sure, as I saw them yellow. Maybe it's yet another thing which does not get picked up by incremental builds (because the configuration already happened). I will try locally on my Mac and then merge.

awegrzyn:
I re-triggered the build and the output is the same.

PR #1530, opened 2019-03-08
Use native ppconsul target

awegrzyn:


awegrzyn:
It follows https://github.com/AliceO2Group/Configuration/pull/42

PR #1533, opened 2019-03-08
Add DIM to @alisw/dataflow-experts for reviews

dberzano:
Via CODEOWNERS


awegrzyn:
It's fine but there's no FLP package that requires DIM.

dberzano:
OK, to my understanding DIM is more dataflow-related than offline, for sure 😄Proceeding. Thanks @awegrzyn for the feedback.

PR #1535, opened 2019-03-11
Bump ROOT to v6-16-00 (O2-538)

dberzano:


dberzano:
Beware, we must not merge at the moment. QC seems not to like the new ROOT:

```
make[2]: *** No rule to make target `/build/ci_checks/alidist_checker_macos_o2_ci_0/sw/BUILD/aded4058143ef5fd5789564e519555dee89e2c19/ROOT/lib/libtbb.dylib', needed by `lib/libQualityControl.dylib'.  Stop.
```

dberzano:
@Barthelemy is it possible that one of [the `ROOT::` libraries](https://github.com/AliceO2Group/QualityControl/blob/master/Framework/CMakeLists.txt#L99-L119) from here now carries `libtbb` with it?

dberzano:
I have found the problem. ROOT's generated CMake files (used correctly by QualityControl) refer to full paths in build directories. I have worked around the issue to make them relocatable by aliBuild. I have successfully tested the PR locally (CC7 only, full stack), let's see if tests turn green here before merging.

Barthelemy:
Thank you Dario, nice work and nice catch ! It seems that all checks have passed.

PR #1536, opened 2019-03-11
o2checkcode: Explicitely list enabled checks

sawenzel:
Rather then disabling certain checks, we now explicitely enumerate the
checks which are enabled. We will no longer be (implicitely) affected due to a
change in clang-tidy version (which might add more checks).

The list of enabled checks are the ones previously used.


dberzano:
Can be merged as far as I am concerned. I see no WIP so I am proceeding.

PR #1537, opened 2019-03-11
Fix InfoLogger dependency in Control

teo:


control-occplugin.sh
```diff
@@ -8,10 +8,10 @@ requires:
   - grpc
   - protobuf
   - "GCC-Toolchain:(?!osx)"
+  - InfoLogger
```
Barthelemy: couldn't you use libinfologger instead of infologger ?

teo: No, because the libInfoLogger recipe is built without Go bindings, and alisw/infrastructure-admins prefer to keep it that way.

teo: Actually you're right, I might be able to use libInfoLogger instead of InfoLogger at least in the OCCplugin.

PR #1539, opened 2019-03-12
Vc: build vc in any case

ktf:
Our version of Vc contains a fix for GCC which is not in any of the official v1.3.x releases, making using Vc from the system impossible.


ktf:
@jgrosseo

ktf:
This currently prevents building on Ubuntu if Vc is already installed. Merging.

PR #1546, opened 2019-03-18
[WIP] Add CUDA default (O2-545)

dberzano:
Enable building O2 with `--defaults o2-cuda`. All O2 dependencies built with
`--defaults o2` can be reused without being rebuilt

### DIFF
--- a/defaults-o2-cuda.sh
+++ b/defaults-o2-cuda.sh
@@ -1,0 +1,1 @@
+defaults-o2.sh
--- a/o2.sh
+++ b/o2.sh
@@ -84,8 +84,9 @@ valid_defaults:
   - alo
   - o2-prod
   - o2-ninja
+  - o2-cuda
 ---
-#!/bin/sh
+#!/bin/bash -ex
 export ROOTSYS=$ROOT_ROOT
 
 # Making sure people do not have SIMPATH set when they build fairroot.
@@ -120,6 +121,10 @@ if [[ ! $CMAKE_GENERATOR && $DISABLE_NINJA != 1 && $DEVEL_SOURCES != $SOURCEDIR 
   type "$NINJA_BIN" &> /dev/null || NINJA_BIN=
   [[ $NINJA_BIN ]] && CMAKE_GENERATOR=Ninja || true
   unset NINJA_BIN
+fi
+
+if [[ $BUILD_FAMILY == *o2-cuda ]]; then
+  HAS_CUDA=1
 fi
 
 unset DYLD_LIBRARY_PATH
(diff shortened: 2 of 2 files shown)


dberzano:
@mconcas @davidrohr @mpuccio what do you think?

ktf:
Why do we need another default?

ktf:
My point being: given if CUDA is there you most likely want to have CUDA support in O2, why not simply enabling it if nvcc  is found (or whatever way to detect it makes sense)?

mconcas:
Hi @ktf. 

I see your point. However, when we build with CUDA, we do not have both CPU and GPU version of the code compiled. 
This for the moment is made by design, to keep interfaces transparent (GPU methods override the CPU ones).

Regardless of what method (`defaults` or something else) you would use to enable this, I would, therefore, like to have the possibility to decide whether to enable CUDA or not in my build.

davidrohr:
I am principally against introducing new defaults. In principle, we could just auto-detect. I just have some comments:
- I want to switch to moat up to date cuda versions once they are released. No idea whether some distros might lack behind.
- With new gcc versions, cuda is often incompatible when released (i.e. gcc 8 is not supported right now). This can be hackes by compiling the cuda part with a second old gcc but then linking with the new one. (this is what i do right now).
- The tracking will not run with old gpus, but this is a runtime thing and should not affect compilation. As long as the gpu parameter is not set for the tracking, the gpu won't be used. We have to think what we do with gpu ctests though.

Another point is what we do with hip/opencl. Do we want autodetection here as well? At least for opencl we have a very particular build flow that will not work with every opencl sdk.

Am 18. März 2019 11:41:33 MEZ schrieb Giulio Eulisse <notifications@github.com>:
>My point being: given if CUDA is there you most likely want to have
>CUDA support in O2, why not simply enabling it if nvcc  is found (or
>whatever way to detect it makes sense)?
>
>-- 
>You are receiving this because you were mentioned.
>Reply to this email directly or view it on GitHub:
>https://github.com/alisw/alidist/pull/1546#issuecomment-473858308

-- 
Diese Nachricht wurde von meinem Android-Gerät mit K-9 Mail gesendet.

mconcas:
[EDIT] I was wrong

davidrohr:
In any case i'd leave the -denable_cuda switch. we could just make this 'auto' by default.

Am 18. März 2019 12:17:05 MEZ schrieb Matteo Concas <notifications@github.com>:
>Hi @ktf. 
>
>I see your point. However, when we build with CUDA, we do not have both
>CPU and GPU version of the code compiled. 
>This for the moment is made by design, to keep interfaces transparent
>(GPU methods override the CPU ones).
>
>Regardless of what method (`defaults` or something else) you would use
>to enable this, I would, therefore, like to have the possibility to
>decide whether to enable CUDA or not in my build. 
>
>-- 
>You are receiving this because you were mentioned.
>Reply to this email directly or view it on GitHub:
>https://github.com/alisw/alidist/pull/1546#issuecomment-473869424

-- 
Diese Nachricht wurde von meinem Android-Gerät mit K-9 Mail gesendet.

mconcas:
Indeed, putting -denable_cuda switch as a default and have a correctly working auto-detection would work for me.

dberzano:
OK then we don't need this PR any longer.

davidrohr:
I think we should discuss how we deploy cuda. And for the autodetection, i strongly suggesr to require a minimum version, which should always be the latest one.

Am 18. März 2019 12:31:19 MEZ schrieb Dario Berzano <notifications@github.com>:
>OK then we don't need this PR any longer.
>
>-- 
>You are receiving this because you were mentioned.
>Reply to this email directly or view it on GitHub:
>https://github.com/alisw/alidist/pull/1546#issuecomment-473873490

-- 
Diese Nachricht wurde von meinem Android-Gerät mit K-9 Mail gesendet.

PR #1547, opened 2019-03-18
Add missing ROOT_INCLUDE_PATH

dberzano:
See AliceO2Group/AliceO2#1800. Affected:

* Vc
* ms_gsl
* GEANT4
* GEANT4_VMC
* AliTPCCommon


ktf:
I guess the multiple include paths will hit us at some point, but for now, good enough since it fixes an actual problem.

PR #1549, opened 2019-03-19
Bump Rivet to 2.7.0

qgp:
requires YODA 1.7.4


pzhristov:
I tested the compilation on Mac an it is OK. The problem on Mac is in the package Readout and has nothing to do with Rivet. I will approve this pull request.

pzhristov:
The tests on MacOS are successful.

PR #1550, opened 2019-03-20
Points lhapdf pdfset retrieval script to the proper archive repo on hepforge

aalkin:


lhapdf5.sh
```diff
@@ -19,7 +19,7 @@ make install
 
 PDFSETS="cteq6l cteq6ll CT10 CT10nlo MSTW2008nnlo EPS09LOR_208 EPS09NLOR_208"
 pushd $INSTALLROOT/share/lhapdf
-  $INSTALLROOT/bin/lhapdf-getdata $PDFSETS
+  $INSTALLROOT/bin/lhapdf-getdata --repo="https://www.hepforge.org/archive/lhapdf/pdfsets/5.9.1" $PDFSETS
```
qgp: I don't think we should hard code the version here

aalkin: I agree. Will "%(tag_basename)s" work here in the same way it is used at the start of the recipe?

aalkin: I made the repo link use the version number from the recipe tag.

qgp: I think you'll need to strip it, e.g. ```${${version#v}%-*}```.

Maybe @dberzano has a better idea?

qgp:
Fine with me, but we still have to update the versions to the new tags:
v5.9.1-alice2
v6.2.1-alice3

Cheers,
   Jochen

aalkin:
@qgp  I don't see tag v5.9.1-alice2 in alisw/lhapdf yet, but I've changed the tag in recipe to point at it.

ktf:
Is this still relevant? If yes can you update it and resolve the conflicts? Thanks.

ktf:
Ping @alisw/pagmc for approval.

PR #1552, opened 2019-03-22
Enable cassertions and macro tests in O2 build in o2.sh depending on ALIBUILD_O2_TESTS variable

davidrohr:
@dberzano : Add cmake variable to enable the assertions, and also move the setting of CHECK_ROOTMACRO_COMPILE when ALIBUILD_O2_TESTS is set to o2.sh.

I'll update AliceO2Group/AliceO2#1799 accordingly.


dberzano:
Hi, as said in AliceO2Group/AliceO2#1799, this is not needed for checking the macros (as it's handled properly in AliceO2Group/AliceO2#1800).

I'd simply do the `-DENABLE_CASSERT` part.

PR #1553, opened 2019-03-22
defaults o2-dataflow should belong to dataflow-experts

Barthelemy:


dberzano:
Numa error aside, the point here is that what we test and deploy is `O2Suite` against `o2`, not against `o2-dataflow`. I agree in transferring the ownership as long as the two defaults don't end up diverging much (it's fine if some packages/dependencies are disabled, it's fine if some temporary bump of deps is done, as long as it's quickly ported everywhere).

Barthelemy:
I see your point. I don't think that it shall diverge much.

sawenzel:
This PR was approved 10 days ago but not yet merged. I am wondering what is the workflow about approving and merging? I guess if an admin approves, it should be merged immediately (and maybe we could have this done automatically).

Barthelemy:
Good point. I presume that it was on me to merge it after approval ? It is also true that I expected the thing to be in as soon as it was approved by an admin. I like your idea of automatic merging.

ktf:
I would start to assign PRs to usual suspects for merging. I would not replace lack of communication with tools. There are plenty of reasons why one could approve a PR (e.g. because it looks fine) and delay the merging (for example to wait for something to happen on the server side, like a build).

sawenzel:
Ok. But in order to avoid waiting time or having to ping back, I would suggest to merge immediately when an admin approves or to leave a message communicating why it should wait.

PR #1554, opened 2019-03-25
Monitoring: Enforce no warnings and run tests in CI

awegrzyn:


awegrzyn:
@sy-c Could you take a look what's happening in the Readout? macOS build fails with `ld: library not found for -lnuma`

sy-c:
I'm pretty sure this is fixed with the latest readout version. Is there a way to check the readout git commit hash used for this build? I suspect it's not using the good one.

dberzano:
From the fifth line of the log, the version is Readout v0.19.1. The hash is what corresponds to that git tag, meaning [this](https://github.com/AliceO2Group/Readout/tree/v0.19.1).

sy-c:
There was a hickup in the procedure to release this readout version, it was at some point re-tagged. That's why I want to check the hash used for the build and not the tag version.

dberzano:
That's what I suspected. Unfortunately we don't support retag (the issue popped up already a couple of times in the past), the only course of action is to produce a new tag instead and update the recipe properly.

sy-c:
ok, I have pushed a fresh v0.19.2 tag for readout https://github.com/alisw/alidist/pull/1555
(code is unchanged)

PR #1556, opened 2019-03-25
o2: clean (ROOT) test artefacts

sawenzel:
Test artefacts may not be compatible between builds
when reusing the build area. It is better to clean them
up after building. In particular done for ROOT files.


sawenzel:
This should fix problems as seen in the interference of https://github.com/AliceO2Group/AliceO2/pull/1810 and other PRs of O2.

ktf:
Can you do it before the tests are run? That way this will cleanup a broken area.

sawenzel:
Ok. Now deleting before running ctest. Not using `-delete` as this does not remove files with '/' (recursively) in the name.

ktf:
What I meant is:

```bash
find $PWD -name "*.root" -delete
```

That should work correctly, no?

sawenzel:
Fine. Misunderstood the man-page.

shahor02:
I would delete also ``*.ini`` and ``*.txt`` files.

sawenzel:
@shahor02 : Not sure. This would delete things like CMakeCache.txt and real config files...

shahor02:
@sawenzel more reasons to run the tests in a dedicated directory. Soon or late the behaviour of the tests will be affected by the ini files (not sure about the text or json files).

sawenzel:
Agreed but this requires more changes. Can we do this later?

sawenzel:
@dberzano : This is solving a real problem but I have no spare time currently to improve it further. There are no root files other than those produced by tests.

dberzano:
@sawenzel @shahor02 As WP3 convener I think such problems should be solved properly. This is going to create more problems in the future.

I understand your urgency, but I don't want to lose track of the problem either. Here's a compromise. Open a Jira where you explain:

* what is the problem you are trying to solve,
* why this PR is a hotfix working for now,
* what would be the more appropriate way of dealing with it in the future.

As long as the problem is known and tracked, I am fine with it and you can go on with merging. Agreed?

sawenzel:
@ktf: Modified according to your suggestion. I think we can go on now.

PR #1557, opened 2019-03-26
Bump FairMQ to v1.4.2 and require libfabric in the system

dennisklein:
This adds the necessary recipes to be able to `-DBUILD_OFI_TRANSPORT=ON` with FairMQ `v1.4.2+`. This build option existed even before, but needed more dependencies.

@dberzano has already [expressed concerns](https://alice-talk.web.cern.ch/t/alibuild-version-dependent-cmake-option/244/6) about the condition in the `fairmq.sh` recipe. I am happy to move to a different solution, e.g. just a manual switch. In any case, I would need some advice on how to implement this. Also on how to make the package dependency in the yaml section conditional on that same condition. Thx!

We should also decide, if this transport should really be built on any desktop/laptop, since it is only meaningful for developers (on non-Infiniband machines) at the moment.


awegrzyn:
Hi,
I would disable build on macOS, and build on Linux only when `libfabric` and  `libfabric-devel` are present.

dennisklein:
Can someone post the error log from the failed checks? (there is no link for me)

> I would disable build on macOS

Probably a good idea, have never tried, if `libfabric` even builds on macOS. Will add an OS guard.

> , and build on Linux only when libfabric and libfabric-devel are present.

We needed the `ofi.sh` recipe on the Alice FLP/EPN prototype machines, because the present ofi installation did not work (not sure, if it was from the official package or from some manual installation/OFED distribution). At least good to have it around, if there are problems with the system packages.

ofi.sh
```diff
@@ -0,0 +1,38 @@
+package: ofi
+version: "%(tag_basename)s"
+tag: v1.7.0
+source: https://github.com/ofiwg/libfabric
+requires:
```
awegrzyn: There are no logs because aliBuild fails on this line, I guess `requires` can't be empty.

dennisklein: removed it

dennisklein:
**`build/O2Suite/alidist_macOS`**: This is fixed in `v1.4.0`. Not sure, why nobody has complained yet for `v1.3.8`.

**`build/O2/alidist`**: I don't understand.

awegrzyn:
@dberzano Could you check why CC7 build fails?

awegrzyn:
@dennisklein CC7 fails on ambiguous `byte` as well

fairmq.sh
```diff
@@ -30,9 +31,22 @@ case $ARCHITECTURE in
     [[ ! $NANOMSG_ROOT ]] && NANOMSG_ROOT=`brew --prefix nanomsg`
     [[ ! $MSGPACK_ROOT ]] && MSGPACK_ROOT=`brew --prefix msgpack`
   ;;
+  *)
+    # Check, if we want to build the ofi transport, which is for all versions v1.4.2+
+    function fairmq_build_ofi() {
+      pushd $SOURCEDIR
+      git --no-pager tag -l --merged HEAD --contains "v1.4.2" >/dev/null 2>&1
```
dberzano: As said separately, and as you say yourself, you really cannot rely on this to work at all.

dberzano: For the record I am looking for a way to detect it properly. In any case, better to use the version variable than the `git` approach.

dennisklein: I am fine to make it a manual switch. How should I implement it?
* Can you point to a recipe with an example?
* Can I just look for an environment variable?
* How can I make a `require:` item depend on the presence of an environment variable?

dberzano: Switching to the comment section because it's clearer 😄

dberzano:
I agree with @awegrzyn. I don't like having dangling recipes that may end up being rarely used. I would remove the recipe and turn it into a _system_ recipe only that merely checks whether the package is available.

@dennisklein: `requires:` that are version-dependent are not possible. You can check the version of your dependency by looking for `$FAIRMQ_VERSION` (in all packages depending on FairMQ) or `$PKGVERSION` (in the FairMQ recipe itself).

For checking version numbers, you can find inspiration on `sort -V`. Have a look at how the version number is parsed [here](https://github.com/alisw/alidist/blob/01a3e0031ff4a798abdd1166db4d1802310e7dca/cmake.sh#L8-L10) for instance.

For a more phenomenological approach, in places when we need to tell ROOT 5 from ROOT 6, we simply check for the presence of certain files in the ROOT installation directory, [see here](https://github.com/alisw/alidist/blob/01a3e0031ff4a798abdd1166db4d1802310e7dca/alien-root-legacy.sh#L20-L21).

dennisklein:
> I agree with @awegrzyn. I don't like having dangling recipes that may end up being rarely used. I would remove the recipe and turn it into a _system_ recipe only that merely checks whether the package is available.
> 
> @dennisklein: `requires:` that are version-dependent are not possible. You can check the version of your dependency by looking for `$FAIRMQ_VERSION` (in all packages depending on FairMQ) or `$PKGVERSION` (in the FairMQ recipe itself).
> 
> For checking version numbers, you can find inspiration on `sort -V`. Have a look at how the version number is parsed [here](https://github.com/alisw/alidist/blob/01a3e0031ff4a798abdd1166db4d1802310e7dca/cmake.sh#L8-L10) for instance.
> 
> For a more phenomenological approach, in places when we need to tell ROOT 5 from ROOT 6, we simply check for the presence of certain files in the ROOT installation directory, [see here](https://github.com/alisw/alidist/blob/01a3e0031ff4a798abdd1166db4d1802310e7dca/alien-root-legacy.sh#L20-L21).

Why are we still talking about this? I agreed with you already in the discourse thread and asked specific questions for an alternative solution which I repeated here and they get ignored. Anyways, you have basically all the code needed (I leave the branch public), you are welcome to use it, change it, not use it.

awegrzyn:
I'm reopening as we are interested in the `v1.4.0` features. Let's try to summarise what's missing:
- [ ] Remove `ofi` recipe, `asofi` transport is built only when  `libfabric`, `libfabric-devel` found in the system
- [ ] In the `FairMQ` recipe, version check, that enables `asofi` should be done using `$PKGVERSION` and `sort -V` (if not look for presence of specific files)
- [ ] FairMQ version bump to `v1.4.2` (otherwise it does not build)

Am I missing anything?

awegrzyn:
@dberzano Is the following check fine with you?
```
if [ "$(printf '%s\n' "1.4.2" "${PKGVERSION#?}" | sort -V | head -n1)" = "1.4.2" ]; then
  BUILD_OFI=ON
fi  
```

dberzano:
@awegrzyn I'd do it like this (note the bashisms: `[[`, lack of `"` when not needed, etc.: this is fine as we ensure we always use bash as shell):

```bash
BUILD_OFI=ON
if [[ $(printf '%s\n' "1.4.2" "${PKGVERSION:1}" | sort -V | head -n1) != "1.4.2" ]]; then
  # Legacy behaviour
  BUILD_OFI=OFF  # or whatever
fi  
```

In practice, bashisms aside, I'd reverse the check in order to make the recipe more forward-compatible, and to make it easier to clean up old code in the future.

In other words, the default behaviour (outside the `if` clause) should be the one valid from now on, and the alternative behaviour (in case the `if` branch is true) is the legacy one.

awegrzyn:
@dberzano Ok thanks. Could you give me a hand with making `ofi.sh` a _system recipe_? What happens when package is not found?

awegrzyn:
Ok I guess this is what you meant: https://github.com/alisw/alidist/blob/31ba0256d65fca999ebd7eb4c196a511866ec0af/opengl.sh

dberzano:
Yes. curl too. aliBuild will just fail if not found.

awegrzyn:
Okay, so these are last changes necessary before merge:
- [ ] https://github.com/AliceO2Group/AliceO2/pull/1874
- [ ] https://github.com/AliceO2Group/Readout/pull/38

awegrzyn:
@dberzano Could you confirm that `build/O2/alidist` fails because of missing `libfabric` and `libfabric-devel`. Would it be possible to add it?

awegrzyn:
Ok macOS build is green it means that as soon as `libfabric-devel` is present at build machines we can merge it.

dberzano:
@awegrzyn we also need to add another library for the Python 3 migration.

awegrzyn:
@dberzano I'm still not able to merge it due to Travis claiming that ofi recipe does not have modulefile. Could you force-merge?

Barthelemy:
@dberzano @ktf @pzhristov Could someone either merge it or clarify what remains to be done ? it has been days since the last update.

pzhristov:
If the Python 3 question is solved, I propose to merge this pull request.

dberzano:
Looks good to me and the error is unrelated.

ktf:
This now breaks the PR checker:

https://ali-ci.cern.ch/repo/logs/AliceO2Group/AliceO2/1913/6d12edf82b4b1c7ee274a17531c99bd8c266c489/build_o2_macos/fullLog.txt

How come this was not spotted earlier?

PR #1559, opened 2019-03-27
Add GPU copyright notice exception

dberzano:


dberzano:
As discussed with @davidrohr, we are adding an exception for the `GPU` directory. Its code is GPLv3 compatible, and it can be added to O2. More discussions on the copyright owners fall under the competences of the Steering Board.

PR #1563, opened 2019-03-28
Disable Fortran in ROOT for user builds

dberzano:


dberzano:
@mpuccio this is your patch, updated (slightly)

pzhristov:
Hi @dberzano ,

The problem is that we link Fortran-compiled code from Hbook and h2root using clang++. It doesn’t work anymore without adding -lgfrotran, so in turn gfrotran has to be added to the list of libraries in the corresponding CMakeLists of Root. Of course we can stop compiling Hbook related code since probably nobody uses it, as proposed in this patch.

Cheers,
Peter

dberzano:
@pzhristov yes, the idea is to disable Fortran when not needed (for analyzers for instance), this is not to be considered a fix to the actual problem.

dstocco:
Hi @dberzano ,
thanks a lot for the patch.
However, I have this issue with fortran also when compiling O2 (not sure why: the issue popped up out of the blue, a couple of days before upgrading clang).
Will your patch solve also the compilation with --defaults o2?

Thanks in advance,
Cheers,
Diego

dberzano:
I don't think so because there's no no-Fortran defaults in O2 ATM.

PR #1566, opened 2019-03-29
Bump ReadoutCard to v0.9.18

kostorr:


awegrzyn:
@dberzano Would it be possible to build `Readout` instead of `O2` when changes concern `ReadoutCard`?
Otherwise PRs to `ReadoutCard` are untested.

dberzano:
@awegrzyn we could add an extra check that triggers only in that case, but be sure you remind us today at WP3.

dberzano:
@awegrzyn let me correct. What gets built by the `build/O2/alidist` test is the O2Suite package, which contains both `Readout` and `ReadoutCard` as dependencies (have a look at the [attached graph](https://github.com/alisw/alidist/files/3067563/o2suite.pdf)). The same applies to the `build/O2Suite/alidist_macOS` test. So, it's already the case: when bumping `Readout*` and the PR is green, you can merge with a certain degree of confidence. Am I missing something?

awegrzyn:
@dberzano Okay thanks. You are right, everything is correct.

PR #1570, opened 2019-04-02
Fix xalienfs scripts relocation and dylibs on MacOS

Atlantic777:
The xalienfs becomes dependency of AliEn-ROOT-Legacy and because of that the $GSHELL_ROOT environment variable should point to it as well.

### DIFF
--- a/alien-root-legacy.sh
+++ b/alien-root-legacy.sh
@@ -1,6 +1,6 @@
 package: AliEn-ROOT-Legacy
 version: "%(tag_basename)s"
-tag: "0.0.5"
+tag: "0.0.6"
 source: https://gitlab.cern.ch/jalien/alien-root-legacy.git
 requires:
   - CMake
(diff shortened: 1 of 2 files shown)


alien-root-legacy.sh
```diff
@@ -55,6 +58,7 @@ module load BASE/1.0 ${GCC_TOOLCHAIN_VERSION:+GCC-Toolchain/$GCC_TOOLCHAIN_VERSI
 
 # Our environment
 setenv ALIEN_ROOT_LEGACY_ROOT \$::env(BASEDIR)/$PKGNAME/\$version
+setenv GSHELL_ROOT \$::env(ALIEN_ROOT_LEGACY_ROOT)
```
dberzano: Please add GSHELL_NO_GCC=1 in the Modulefile as well, thanks!

Atlantic777: Done.

Atlantic777:
Back to WIP, the fix for .dylibs on MacOS will be added to this pull request.

alien-root-legacy.sh
```diff
@@ -40,6 +43,11 @@ for RPKG in $BUILD_REQUIRES; do
   rm -f $INSTALLROOT/etc/modulefiles/{$RPKG,$RPKG.unrelocated} || true
 done
 
+if [[ $ARCHITECTURE == osx* ]]; then
+  ln -nfs $INSTALLROOT/lib/libAliEnROOTLegacy.dylib $INSTALLROOT/lib/libAliEnROOTLegacy.so
```
dberzano: The two symlinks here will not be relocated at install time, please use relative paths:

```bash
ln -nfs libBlahBlah.dylib $INSTALLROOT/lib/libBlahBlah.so
```

PR #1571, opened 2019-04-03
Fix for lhapdf pdfsets needed for generator packages

mfasDa:
Issues in pdfsets:
- LHAPDF5 was downloading pdfsets from LHAPDF6, which are incompatible. The LHAPDF version 
   needs to be specified in the address for the download
- Avoid using lhapdf / lhapdf-getdata script for downloading as it runs into errors, rather download explicitly via curl.
- Issue with PYTHONPATH for LHAPDF6: On some platforms the site-package is installed in lib64 
  instead of lib, which results in lhapdf script failing due to missing python bindings. Unless the 
  directory for the site packages can be handled automatically both options need to be added to the 
  PYTHONPATH.

Without the fixes the installation of generators (i.e. POWHEG/HERWIG) run into build errors when getting the pdfsets

### DIFF
--- a/lhapdf-pdfsets.sh
+++ b/lhapdf-pdfsets.sh
@@ -5,7 +5,7 @@ build_requires:
 ---
 #!/bin/bash -ex
 
-PDFSETS="cteq6l1 MMHT2014lo68cl MMHT2014nlo68cl cteq66"
+PDFSETS="cteq6l1 MMHT2014lo68cl MMHT2014nlo68cl cteq66 CT10nlo"
 if [ ! -d $INSTALLROOT/share/LHAPDF ]; then mkdir -p $INSTALLROOT/share/LHAPDF; fi
 pushd $INSTALLROOT/share/LHAPDF
   REPO=https://lhapdf.hepforge.org/downloads?f=pdfsets/6.2.1
--- a/lhapdf-pdfsets.sh
+++ b/lhapdf-pdfsets.sh
@@ -6,9 +6,17 @@ build_requires:
 #!/bin/bash -ex
 
 PDFSETS="cteq6l1 MMHT2014lo68cl MMHT2014nlo68cl cteq66"
-lhapdf --pdfdir=$INSTALLROOT/share/LHAPDF  \
-       --listdir=$LHAPDF_ROOT/share/LHAPDF \
-       install $PDFSETS
+if [ ! -d $INSTALLROOT/share/LHAPDF ]; then mkdir -p $INSTALLROOT/share/LHAPDF; fi
+pushd $INSTALLROOT/share/LHAPDF
+  REPO=https://lhapdf.hepforge.org/downloads?f=pdfsets/6.2.1
+  for P in $PDFSETS; do
+    PDFPACK=$(printf "%s.tar.gz" $P)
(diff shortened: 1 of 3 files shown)


qgp:
@pzhristov, @ktf, @dberzano: why was this merged without alerting PAG-MC/PWG-MM?

pzhristov:
@qgp You got the alert on 03/04/2019. After 9 days I merged the pull request since, as stated in the description, "Without the fixes the installation of generators (i.e. POWHEG/HERWIG) run into build errors when getting the pdfsets".

qgp:
@pzhristov: So, I must have missed it. It would be nice if people making changes to MC packages would get in touch with us on problems and possibly ping us in case of the lack of a response.

My concern about this "fix" is that the lhapdf-getdata script remains broken and we should rather fix the upstream package instead of implementing workarounds in the build recipes.

mfasDa:
PWGMM was also informed about this, on the mailing list, but there was no reaction. Therefore I asked offline responsibles directly for having a look.

qgp:
> PWGMM was also informed about this, on the mailing list, but there was no reaction. Therefore I asked offline responsibles directly for having a look.

My bad, it slipped and I apologise for that. In the future, please ping us in the case of no answer.

pzhristov:
@qgp Is there anything we can do to fix the remaining problems? I was not aware of any broken script...

qgp:
@pzhristov: I would suggest to fix this in LHAPDF directly to choose the correct download location in the lhapdf-getdata script according to the version of the package (as is already done for LHAPDF 6). I can look into that but right now I have rather limited Internet access (at LHCP in Mexico), which makes testing something rather difficult. Until fixed there, I would suggest to keep Markus' fix in the recipe so that the builds work.

Sorry again for not having raised this earlier, blame is on me.

mfasDa:
@qgp I recall again what the problem was: Internally lhapdf-getdata is reading a webpage with all the datasets and is parsing its content in order to determine which pdfsets are available. So it tries to parse a table via regular expressions, but this table is different for lhapdf5 and lhapdf6 (i.e. because the format changes). So either it downloaded nothing, because it claimed that the pdfset didn't exist (which is not true), or it downloaded something but the content was an http 404, because it was redirecting to a page that didn't exist. Eventually the parsing can be fixed, but I am not sure it can be done generally for LHAPDF5 and LHAPDF6. At that point I gave up as I hadn't more time to spend here (also as this page is quite complex and has a lot of formatting in), and the curl-based solution worked. Eventually it is better to write the script from scratch, eventually using a real html engine.

qgp:
@mfasDa: Thanks for the recapitulation of the issues at hand. As far as I can see there are two parts of the problem:
- parsing of the PDF list from the HTML output
  there was a fix by PR alisw/LHAPDF#5, did you observe failures before or after that fix?
- download URL for the PDF sets
  ```current``` now always points to LHAPDF6 sets, which is not yet fixed in our LHAPDF5 package

mfasDa:
@qgp I was working at this pretty much at the same time. I used fresh alidist master for my tests, but I am not  sure they were updated before or after March 28th.

Indeed there are two separate problems.

PR #1574, opened 2019-04-09
Bump geant4_vmc to v3-6-p3

sawenzel:
This is needed in order to use Geant4 "as-a-service" in O2 simulation.


sawenzel:
log files are not available (404)

PR #1578, opened 2019-04-10
Drop support for nanomsg

ktf:
Turns out ALICE will not use nanomsg in production, so there is
little point in having it available.


ktf:
@MohammadAlTurany @ironMann I guess this is a reasonable assumption, right? AFAIU, we will not need `nanomsg` for either in-node or intra-node transport, correct?

ironMann:
I don't have plans to use ZeroMQ or nanomsg directly. Currently, I'm tracking FairMQ version from alidist (v1.3.8).

MohammadAlTurany:
@ktf Yes it is a reasonable suggestion 
@ironMann You should test and use v1.4.2!!

ktf:
Ok, great. I will merge this as soon as the tests are green. I will also check if we have any problems in bumping zeromq to 1.4.2.

rbx:
Nanomsg transport in FairMQ uses msgpack to implement multipart. If you drop nanomsg then you can also drop msgpack, unless you use it for something else outside FairMQ.

dennisklein:
> I will also check if we have any problems in bumping zeromq to 1.4.2.

If you encounter any problems, this may help in resolving them:
* [Slide 5](https://indico.cern.ch/event/806602/contributions/3377237/attachments/1824622/2985652/FairMQ_State_Machine_Update.pdf) on how to deal with the API changes from `v1.3` to v`1.4`.
* In AliceO2Group/Readout#38 and AliceO2Group/DataDistribution#3 I have hinted the minimal changes needed - the respective authors may want to use a more sophisticated adaptation.

awegrzyn:
@ktf Shall we drop `msgpack` as @rbx mentioned? I've verified it works without it.

ktf:
I still liked the idea of having msgpack around for some idea I am tinkering with, although I will probably go with flatbuffers in the end, since arrow in any case requires it.. If it's not an issue, I would keep it around a bit longer.

PR #1580, opened 2019-04-11
WIP Switch off Fortran: h2root and hbook are not needed and their compila…

pzhristov:
…tion causes problems on MacOS

### DIFF
--- a/root.sh
+++ b/root.sh
@@ -111,13 +111,6 @@ if [[ $ALICE_DAQ ]]; then
             soversion ${CXX11:+cxx11} ${CXX14:+cxx14} ${CXX17:+cxx17} mysql xml"
   NO_FEATURES="ssl alien"
 else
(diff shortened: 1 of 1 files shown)


pzhristov:
The error we see in the linking of AliRoot/MICROCERN library on SLC7 is similar to what we had in Root, and has to be fixed separately.

dberzano:
There's also GEANT3. GEANT3 takes the Fortran config from ROOT. We should apply `-DCMAKE_Fortran_compiler=gfortran` to the `geant3.sh` recipe.

pzhristov:
In theory if we add Fortran to project(Geant3) (together with CXX and C), this should be enough, CMake should be able to find it automatically.

dberzano:
In principle yes, but we must really specify `-DCMAKE_Fortran_COMPILER` to avoid ambiguities (it's a small change).

PR #1581, opened 2019-04-12
Overhaul of Protobuf&gRPC recipes to build them with CMake

teo:
* Bumped Protobuf to 3.7.1
* Added recipe for c-ares which needs to be built as a dependency of gRPC (note that this is not a new dependency, it is simply built out of gRPC's tree)
* Bumped gRPC to 1.19.1, removed lots of build hacks and switched to upstream (alisw fork not needed any more)
* Bumped AliECS to 0.6.3 which was [released](https://github.com/AliceO2Group/Control/releases/tag/v0.6.3) simultaneously to adapt to Protobuf&gRPC changes.

Also see https://github.com/AliceO2Group/AliceO2/pull/1880

### DIFF
--- a/grpc.sh
+++ b/grpc.sh
@@ -1,11 +1,13 @@
 package: grpc
 version: "%(tag_basename)s"
-tag:  v1.15.1-alice1
+tag:  v1.19.1
 requires:
   - protobuf
+  - c-ares
 build_requires:
   - "GCC-Toolchain:(?!osx)"
-source: https://github.com/alisw/grpc
+  - CMake
+source: https://github.com/grpc/grpc
 prefer_system: "(?!slc5)"
 prefer_system_check: which grpc_cpp_plugin
 incremental_recipe: |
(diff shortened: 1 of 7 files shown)


grpc.sh
```diff
@@ -14,44 +16,26 @@ incremental_recipe: |
...
-       LD_LIBRARY_PATH="$LD_LIBRARY_PATH"       \
-       make ${JOBS:+-j$JOBS} prefix=$INSTALLROOT
-
-# ldconfig won't work if we're not running make install as root, and that's ok,
-# we don't need it
-sed -i.deleteme -e 's/ldconfig/true/' Makefile
-rm -f *.deleteme
-
-make prefix=$INSTALLROOT install
+opts=(
```
ktf: What's the point of this? Can't you simply put it inline?

teo: No point except having a list of things where you don't have to remember to add a `\` at the end of each line.

teo: Do you want me to change it to inline?

ktf: Yes, please, let’s stay consistent and keep it simple.

grpc.sh
```diff
@@ -21,20 +21,18 @@ git checkout "$GIT_TAG"
...
-    -DgRPC_BUILD_TESTS=OFF
-    -DBUILD_SHARED_LIBS=ON
-    -DgRPC_SSL_PROVIDER=package
-    -DgRPC_ZLIB_PROVIDER=package
-    -DgRPC_GFLAGS_PROVIDER=packet
-    -DgRPC_PROTOBUF_PROVIDER=package
-    -DgRPC_BENCHMARK_PROVIDER=packet
-    -DgRPC_CARES_PROVIDER=package
-)
+cmake $SOURCEDIR \
```
ktf: Please align the \... also you probably want to drop the last one.

protobuf.sh
```diff
@@ -1,17 +1,19 @@
...
  - "GCC-Toolchain:(?!osx)"
 prefer_system: "(?!slc5)"
 prefer_system_check: |
   printf "#include \"google/protobuf/stubs/common.h\"\n#if (GOOGLE_PROTOBUF_VERSION < 3000000)\n#error \"At least protobuf 3.0.0 is required.\"\n#endif\nint main(){}" | c++ -I$(brew --prefix protobuf)/include -Wno-deprecated-declarations -xc++ - -o /dev/null && protoc -h &> /dev/null
 ---
 
-rsync -av --delete --exclude="**/.git" $SOURCEDIR/ .
-autoreconf -ivf
-./configure --prefix="$INSTALLROOT"
+cmake $SOURCEDIR/cmake \
```
ktf: Align the `\`.

ktf:
Thanks. It looks like it breaks the build of O2, though. Could you take a look?

teo:
@ktf I already did, I believe this https://github.com/AliceO2Group/AliceO2/pull/1880 should fix it.

ktf:
Ok. I will let @dberzano take it on from here.

ktf:
Merging to get things going.

ktf:
There is actually a problem with this. You have:

```
find_package(protobuf CONFIG REQUIRED)
```

rather than the old:

```
find_package(Protobuf REQUIRED)
```

The former does not work if the system protobuf is picked up and it was not built using cmake (like in my case). Do you really need to using the CONFIG or can you get away with the usual setup?

PR #1585, opened 2019-04-13
Fix for: undefined reference to `_gfortran_stop_numeric_f08'

marcovanleeuwen:
Without this fix, on Ubuntu, the aliroot compilation uses /usr/bin/f95 which leads to trouble with libgfortran when linking.

This seems to be a side-effect/result of the decision to compile root without fortran, and this commit https://github.com/alisw/alidist/pull/1563 

NB: not tested on Max OSX; probably should be checked there with the new xcode, since there was apparently some issue with gfortran there.


pzhristov:
Hi Marco,
We have to revise AliRoot/CMakeLists.txt too. The way FORTRAN is treated there is tot optimal.

Cheers,
Peter

PR #1587, opened 2019-04-15
Bump AliRoot to v5-09-48

dberzano:


dberzano:
@pzhristov @ktf documentation is coming to avoid confusion in the future, don't worry (give me 10 more minutes) 😄

ktf:
Thank you. Can we also streamline the 9 defaults we have at the moment?

ktf:
Notice also that AliPhysics 48 was not created (can you actually take care of it)? I think we need a simpler workflow for the case in which we have a change in AliPhysics which requires one in AliRoot.

dberzano:
@ktf That is a separate issue and concerns macOS unfortunatey as @pzhristov knows. I did not have confirmation that macOS users can now launch ROOT 5 jobs from their ROOT 6 installation.

For the documentation: alisw/alisw.github.io#69. And for the future (this is in the doc):

* an unvalidated release should be called `v5-09-47a` (not `48`, this is how the DPG wants it)
* all old release references in alidist should be changed (this PR does it)
* alidist must be tagged with `AliPhysics-v5-09-48-01` (doing it now after merging)

PR #1589, opened 2019-04-15
Move to Python 3

dberzano:


dberzano:
Still a WIP because I want to test it on an environment with Python coming from the system.

dberzano:
OK so I have pushed for the moment:

* an update that does not check for the presence of Modulefiles in system dependencies
* a fix that should make boost work both with and without Python, according to the configuration

From what I see locally (but let's wait for the tests) I have a problem building ReadoutCard with the o2 defaults. ReadoutCard depends on Python but the dependency is not expressed: it finds at the same time a boost version built with our Python, and system's Python, which are incompatible.

I am trying to address it.

dberzano:
The Readout problem is being addressed. The [macOS problem](https://ali-ci.cern.ch/repo/logs/alisw/alidist/1589/ed581b4eeeb13a94f0f2892e63647743b807fcc4/build_O2Suite_alidist_macOS/fullLog.txt) requires a bit more attention.

dberzano:
@sy-c it does not, but it links to ReadoutCard which does. Since it uses the ReadoutCardConfig.cmake generated by ReadoutCard, one must pass to it the Python variable all the same. Another option would be to hardcode the Python path into the ReadoutCardConfig.cmake, but this would, unfortunately, make it unrelocatable and break our deployment. This is the neater fix to the problem I am afraid 😢

dberzano:
I think I fixed the macOS problem with Python-modules.

dberzano:
From my local tests (but I want confirmation from the C.I.):

* macOS should now be fine
* qcg is problematic because of an interplay between ZeroMQ/Python 3 so we expect O2Suite to fail there, but @awegrzyn wanted to move ZeroMQ away from qcg (this would unblock the situation)

I'd say we're almost there...

dberzano:
As expected, the error was in ZeroMQ on QCG, but it has been fixed upstream. Also, Travis has moved to Python 3 by default yesterday, so the linter if ran now would fail.

Everything has been fixed by myself and @awegrzyn in #1592. Rebasing on top of master and waiting for the tests.

dberzano:
Another problem with `node-gyp` which impressively does not support Python 3. I have a "workaround" to make tests green, hold on.

dberzano:
Building O2Suite is still in progress on Linux, but I have tested it on the same C.I. setup successfully. Let's wait for the green tests.

**Let's not merge it before 4pm!**

For the record, I made a local C.I.-like test with merging both this one and #1591 at the same time, with success. From my side, no need to recheck that one when this one's merged.

The plan is: if I see a clear green light, I'll merge this one and #1591 at the same time later today.

dberzano:
Before merging I have tested the SLC6 builds (production ones for Run2). They appear to succeed.

python.sh
```diff
@@ -1,30 +1,24 @@
...
  - libpng
  - sqlite
+ - "GCC-Toolchain:(?!osx)"
 build_requires:
  - curl
 env:
-  SSL_CERT_FILE: "$(export PATH=$PYTHON_ROOT/bin:$PATH; export LD_LIBRARY_PATH=$PYTHON_ROOT/lib:$LD_LIBRARY_PATH; python -c \"import certifi; print certifi.where()\")"
+  SSL_CERT_FILE: "$(export PATH=$PYTHON_ROOT/bin:$PATH; export LD_LIBRARY_PATH=$PYTHON_ROOT/lib:$LD_LIBRARY_PATH; python -c \"import certifi; print(certifi.where())\")"
   PYTHONHOME: "$PYTHON_ROOT"
-prefer_system: (?!slc5)
```
ktf: This effectively removes the `prefer_system`, I guess you meant `prefer_system: .*`

boost.sh
```diff
@@ -17,20 +17,43 @@ prepend_path:
...
+  else
+    # Using system's Python. We want to make sure `python-config` is available in $PATH and points
+    # to the Python 3 version. Note that a symlink will not work due to the automatic prefix
+    # calculation of the python-config script. Our own Python does not require tricks
+    if ! type python3-config &> /dev/null; then
+      echo "FATAL: cannot find python3-config in your \$PATH. Cannot enable boost_python"
+      exit 1
+    fi
+    mkdir fake_bin
+    cat > fake_bin/python-config <<\EOF
```
ktf: I think we need the same trick for `python3m-config`...

PR #1594, opened 2019-04-18
Fix Python-modules shebangs

dberzano:
Binaries are created with the original Python path hardcoded. The path is not
found when deployed. Use Python from the environment instead, without requiring
relocation.

Also change requirements (build only on SLC*, get from system elsewhere).


dberzano:
As discussed in WP3.

PR #1597, opened 2019-04-24
fairmq: disable OFI transport if OFI not available

ktf:


ktf:
@shahor02 can you check if this solves your issue?

PR #1598, opened 2019-04-24
Avoid explicit dependency on gcc

ktf:
Fixes boost, yaml-cpp, lz4 detection.


ktf:
Errors unrelated. Merging.

PR #1599, opened 2019-04-24
freetype - Add pkg-config check for freetype cflags

akubera:
The Ubuntu 19.04 libfreetype6-dev package removes `freetype-config` and only uses the `freetype2.pc` file for use with pkg-config. This patch adds a check in freetype, xdevel, and python-modules for the `freetype-config` executable, with fallback to pkg-config if not found.

Alternatively, the check could be the shorter:
 `$((pkg-config --cflags freetype2 || freetype-config --cflags) 2>/dev/null)`

### DIFF
--- a/freetype.sh
+++ b/freetype.sh
@@ -7,7 +7,7 @@ build_requires:
   - curl
 prefer_system: (?!slc5)
 prefer_system_check: |
-  printf "#include <ft2build.h>\n" | c++ -xc++ - `freetype-config --cflags` -c -M 2>&1;
+  printf "#include <ft2build.h>\n" | c++ -xc++ - `if [ -x "$(which freetype-config)" ]; then freetype-config --cflags; else pkg-config freetype2 --cflags; fi` -c -M 2>&1;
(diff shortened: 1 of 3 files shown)


ktf:
I think this will not work for the case in which we provide our own autotools (which includes pkg-config).

akubera:
Ok, I just prepended `/usr/lib/x86_64-linux-gnu/pkgconfig` to PKG_CONFIG_PATH before calling

ktf:
Thanks, but don't you have a better way to detect the system `PKG_CONFIG_PATH`?

akubera:
Are there other supported environments where freetype-config is not installed? I thought "the latest and greatest" Ubuntu was targeted, so I was submitting my change that fixed it. Those who use aliBuild autotools should submit their own patch; I can't build it without errors on 19.04 *or* 18.04 (`aliBuild GCC-Toolchain --no-system --defaults=user`) so I can't test it.

pzhristov:
Fixed in PR https://github.com/alisw/alidist/pull/2211

PR #1603, opened 2019-04-29
Build own version of OFI if not found on the system

ktf:
Build OFI on mac as well, for consistency.

### DIFF
--- a/fairmq.sh
+++ b/fairmq.sh
@@ -8,7 +8,7 @@ requires:
  - ZeroMQ
  - msgpack
  - DDS
- - "asiofi:(?!osx)"
+ - asiofi
 build_requires:
  - CMake
  - "GCC-Toolchain:(?!osx)"
--- a/ofi.sh
+++ b/ofi.sh
@@ -10,7 +10,7 @@ prefer_system_check: |
 ---
 rsync -a --exclude='**/.git' --delete --delete-excluded "$SOURCEDIR/" ./
 autoreconf -ivf
-./configure --prefix="$INSTALLROOT"
+./configure --prefix="$INSTALLROOT" --enable-mlx=no 
 make ${JOBS:+-j $JOBS} install
 
 # Modulefile
--- a/ofi.sh
+++ b/ofi.sh
@@ -1,10 +1,34 @@
 package: ofi
-version: "1.6.0"
-system_requirement_missing: |
-  libfabric and its development package are missing from your system.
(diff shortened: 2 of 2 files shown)


ktf:
@shahor02 could you try this?

ktf:
@dennisklein is that really true you need 1.6.0+? That prevents using system packages for a bunch of platforms.

dennisklein:
I do not know, never tried older versions. There is some piece of initialization which requests a minimal ofi api version currently, this would need to be relaxed to allow older versions.

That said, I would say `-DBUILD_OFI_TRANSPORT=ON` is a linux only feature, even though it is not enforced on FairMQ side. AFAIK, there is libfabric for macos (which I have never tried), but not libibverbs. (?)

May I ask, what is the problem?

dennisklein:
Here I request ofi api version `1.5` during initialization, see https://github.com/FairRootGroup/asiofi/blob/ce3a5125fba61ac10cff53aa5229186719b2693c/include/asiofi/info.hpp#L129. Not sure to which range of package versions this corresponds and if I use any api that is only available since `1.5`. One would need to try.

chiarazampolli:
Ciao,

In case you are interested, and it is useful, for me this PR worked for me with this issue. 

Chiara

ktf:
Multiple people have issues because a recent enough ofi was not available for their setup (ubuntu, IIUC). `FI_VERSION` seems to be exactly the OFI version, no?

shahor02:
@ktf  Thanks, worked also on Ubuntu 18.04!
@dennisklein : [here](https://alice-talk.web.cern.ch/t/alibuild-o2-fails-on-ubuntu-18-04-after-python3-ofi-updated-in-alidist/253) is the description of the problem.

dennisklein:
> Multiple people have issues because a recent enough ofi was not available for their setup (ubuntu, IIUC). `FI_VERSION` seems to be exactly the OFI version, no?

Until `1.6` there was [this comment](https://github.com/ofiwg/libfabric/blob/v1.6.x/include/rdma/fabric.h#L78-L80) stating that it is not the case. In `master` this comment is gone, so maybe they match the package version now?

dennisklein:
@shahor02 Thx and sry I missed your question on alice talk. I wrote a reply there now.

dennisklein:
@ktf On many hosts I have installed ofi myself so far where an OFED installation was present, I had problems building without the `--enable-mlx=no` configure option. Just FYI.

ktf:
I've added your suggestion..

PR #1609, opened 2019-04-30
Build GCC without Python (O2-628)

dberzano:


dberzano:
@sy-c @costing this should solve the problem of `gdb` logs mixed up with Python errors. Let's wait for the tests and, if green, let's merge it > 4pm.

PR #1610, opened 2019-04-30
Workaround missing ABI suffix in python include path

ktf:


Barthelemy:
Could someone review and merge ?

PR #1613, opened 2019-05-02
Remove AliTPCCommon

davidrohr:
@dberzano : Should be merged after https://github.com/alisw/AliRoot/pull/914 and probably in conjunction with creating a new AliRoot / AliPhysics tag, in order not to break AliPhysics PRs.


davidrohr:
rebased and all green now, so whenever the new AliRoot tag is ready

ktf:
@davidrohr can't we actually close this now?

davidrohr:
@ktf : No, this should be merged. Right now, alitpccommon i still in alidist.
Unfortunately, we can only merge this once there is a new AliRoot tag. Otherwise, AliPhysics CI will fail, since it builds with alidist master versus the last AliRoot tag, which still requires AliTPCCommon.

There is already a JIRA for the next tag: https://alice.its.cern.ch/jira/browse/ALIROOT-8256

Issue #1628, opened 2019-05-09
python-modules build failure on centos7 (matplotlib)


ironMann:
Hi, I have a build failure in python-modules while trying to build o2 with o2-dataflow defaults on a centos7 server. (manual alibuild without docker)

Logs [here](https://cernbox.cern.ch/index.php/s/6Hyx3F0stEUFreV)

This diff fixes it for me:
```diff
diff --git a/python-modules.sh b/python-modules.sh
index 13caefd..94262c5 100644
--- a/python-modules.sh
+++ b/python-modules.sh
@@ -59,6 +59,8 @@ env PYTHONUSERBASE="$INSTALLROOT" pip3 install --user -IU -r requirements.txt
 pushd "$INSTALLROOT"
   if [[ -d lib64 ]]; then
     ln -nfs lib64 lib  # creates lib pointing to lib64
+  else
+    ln -nfs lib lib64
   fi
   pushd lib
     ln -nfs python$PYVER python
```

ktf:
I guess it makes sense. Can you open a PR with it? Could you also add a test making sure that lib itself exists?

PR #1630, opened 2019-05-09
python-modules: create a lib64 symlink pointing to lib in the custom …

ironMann:
…python module install path

Needed for centos7 matplotlib installation

Closes: #1628


ktf:
Thanks. I will merge this once the daily tags for AliPhysics have started.

python-modules.sh
```diff
@@ -59,6 +59,10 @@ env PYTHONUSERBASE="$INSTALLROOT" pip3 install --user -IU -r requirements.txt
 pushd "$INSTALLROOT"
   if [[ -d lib64 ]]; then
     ln -nfs lib64 lib  # creates lib pointing to lib64
+  else
+    if [[ -d lib ]]; then
```
ktf: `elif` ...

PR #1631, opened 2019-05-09
bump readout v0.25

sy-c:
see release notes: https://github.com/sy-c/Readout/blob/master/doc/releaseNotes.md

### DIFF
--- a/readout.sh
+++ b/readout.sh
@@ -1,6 +1,6 @@
 package: Readout
 version: "%(tag_basename)s"
-tag: v0.25
+tag: v0.25.1
 requires:
   - boost
   - "GCC-Toolchain:(?!osx)"
@@ -11,6 +11,8 @@ requires:
   - Monitoring
   - Configuration
   - ReadoutCard
+  - lz4
+#  - Control-OCCPlugin
 build_requires:
   - CMake
 source: https://github.com/AliceO2Group/Readout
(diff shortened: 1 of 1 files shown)


sy-c:
@alisw/dataflow-experts 
anybody to push this PR ?

build/O2/alidist  fails with unrelated Error in <ACLiC>: Compilation failed!

ktf:
On mac it actually fails linking with:

```
ld: unknown option: -soname=liblz4.so.1
make[3]: *** [liblz4.so.1.9.1] Error 1
make[2]: *** [lz4-prefix/src/lz4-stamp/lz4-build] Error 2
make[1]: *** [CMakeFiles/lz4.dir/all] Error 2
```

Do you actually build lz4 yourself???

sy-c:
yes, readout builds lz4
unfortunately I can't check what happens, alibuild does not work at the moment on the only mac server I have access to (it brakes in boost).

ktf:
We do ship with a lzma package, any reason for not use that?

sy-c:
This for real-time compression of the data. According to ITS, LZ4 is the fastest algorithm for their use case.

ktf:
sorry, I meant lz4... My point is that we do have lz4 packaged up, IIRC....

ktf:
https://github.com/alisw/alidist/blob/master/lz4.sh

sy-c:
Indeed, very good point, I had not noticed lz4 was available.
recipe now make use of it with readout v0.25.1.
it should also fix the osx build.

ktf:
Thanks! I will merge this once all tests are green.

sy-c:
There's one failure but it does not seem related. macos build is ok.

PR #1642, opened 2019-05-16
Addition of two python packages for the QC (QC-61)

Barthelemy:
responses is to mock requests in tests. 
dryable is to add a --dry-run option.


Barthelemy:
Error seems unrelated. @ktf @dberzano could you reapprove the PR ? thank you

ktf:
After the mac tests, sure.

PR #1645, opened 2019-05-18
Update the detection of the python3 executable location.

aphecetche:
Needed for example on systems using "shims-based" strategies to switch
between different versions of Python, like pyenv or asdf.

In those systems which python gives something like :

~/.asdf/shims/python

which is a script giving access to the right python executable, but not
directly the python executable.

Without this fix the Root cmake configuration can possibly be confused
and fails :

-- Found PythonInterp: /Users/laurent/.asdf/shims/python3 (found version
"3.6.6")
-- Found PythonLibs:
/usr/local/Frameworks/Python.framework/Versions/3.7/lib/libpython3.7m.dylib
(found version "3.7.3")
-- Looking for numpy (python package)
-- Found NUMPY:
/Users/laurent/alice/qc/sw/osx_x86-64/Python-modules/1.0-3/lib/python/site-packages/numpy/core/include
(found version "1.16.2")
CMake Error at cmake/modules/SearchInstalledSoftware.cmake:480
(message):
  Version mismatch between Python interpreter (3.6.6) and libraries
  (3.7.3).

    ROOT cannot work with this configuration.  Please specify only
      PYTHON_EXECUTABLE to CMake with an absolute path to ensure
      matching
        versions are found.

### DIFF
--- a/root.sh
+++ b/root.sh
@@ -71,6 +71,7 @@ if [[ -d $SOURCEDIR/interpreter/llvm ]]; then
   ROOT_PYTHON_FEATURES="python"
   ROOT_HAS_PYTHON=1
   # One can explicitly pick a Python version with -DPYTHON_EXECUTABLE=... -DPYTHON_INCLUDE_DIR=<path_to_Python.h>
+  PYTHON_EXECUTABLE=$( $(which python3) -c 'import sys; print(sys.executable)')
 else
   # Non-ROOT 6 builds: disable Python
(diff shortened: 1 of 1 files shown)


aphecetche:
Actually my solution is probably not enough. Might want to use `realpath` instead. Will test more.

aphecetche:
Removing WIP as it seems to be working fine.

aphecetche:
@ktf @dberzano Any comment on this one ?
I've been using it with my asdf-based setup for the last couple of weeks without any issue so far.

PR #1659, opened 2019-05-31
Bump QC to v0.13.1

knopers8:


ktf:
For some reason this does not seem to fix the macOS builds. I will merge it, but are you sure this is the correct version?

knopers8:
Indeed, the fix wasn't actually in that version, my mistake...

PR #1662, opened 2019-06-04
Use Azul's JDK

costing:
This build is optimised for memory usage. Further on, on Linux we will use the 32bit JDK while only on Mac we will keep using the 64bit one (as it's the only one available).


ktf:
Why don't you want to use the 64 bit one also on linux?

ktf:
Waiting for the mac builds before merging.

PR #1666, opened 2019-06-05
root: mitigate memory issues when linking

ktf:


root.sh
```diff
@@ -37,13 +37,20 @@ incremental_recipe: |
...
 
+
 COMPILER_CC=cc
 COMPILER_CXX=c++
 COMPILER_LD=c++
 [[ "$CXXFLAGS" == *'-std=c++11'* ]] && CXX11=1 || true
 [[ "$CXXFLAGS" == *'-std=c++14'* ]] && CXX14=1 || true
 [[ "$CXXFLAGS" == *'-std=c++17'* ]] && CXX17=1 || true
 
+# We do not use global options for ROOT, otherwise the -g will kill compilation
```
sawenzel: What about finally get rid of the global `-g` ?

ktf: We could do that as well, however it will kill the stack traces, so we need to enable it on a per project basis before we remove it globally.

Issue #1656, opened 2019-05-29
Boost recipe using headers from brew boost on macOS


mfasDa:
Dear experts,

It turned out that

export CPLUS_INCLUDE_PATH="$CPLUS_INCLUDE_PATH:$(python3 -c 'import sysconfig; print(sysconfig.get_path("include"))')"

makes the boost build picking up boost headers from brew in case the python3 installation comes from brew as well. As on mac boost is anyhow not built with boost support can we put this line within a protection ("if [ $ARCHITECTURE == osx ] ...")?

Remove boost from brew is not a solution as users might have installed programs that rely on boost from brew.

Cheers

Markus

mfasDa:
@dberzano @ktf Would you agree with this fix?

ktf:
I agree this is an issue, I am not sure what you want to do, though. Could you open a PR?

mfasDa:
Thanks for the reply! I will open a PR.

What I want to do is the following: The reason why this happens is that both the boost and the python headers are taken from /usr/local/include on macOS when python3 is installed via homebrew (which I assume is the standard case for mac), and by setting CPLUS_INCLUDE_PATH consequently the boost headers from homebrew are propagated. As the boost versions in brew and alidist are different this creates the compiler errors. As currently boost is anyhow configured on mac without python support (see a couple of lines above) setting the include path for python on mac is not really needed and this bypasses the problem for the moment. It is not really a solution as maybe at some point you want to add python support for boost also on mac. However I also think you cannot force the users to remove boost from homebrew as this might raise other issues (boost is pretty fundamental). So maybe for mac you will have to install a custom python3 ...

ktf:
I understand the issue and I agree. Let's move the discussion to the PR.

PR #1671, opened 2019-06-10
Prevent setting CPLUS_INCLUDE_PATH on mac

mfasDa:
Fixing alisw/alidist#1656: In case boost and
python3 are taken from homebrew a collision
in the boost headers might happen as brew
installs a more recent boost version. As python
support is in any case currently disabled on
macOS dropping propagating the include path
for the python headers is safe on this platform.

### DIFF
--- a/boost.sh
+++ b/boost.sh
@@ -67,7 +67,10 @@ cd $BUILDDIR/tools/build
 # the ABI suffix. E.g. ../include/python3 rather than ../include/python3m.
 # This is causing havok on different combinations of Ubuntu / Anaconda
 # installations.
-export CPLUS_INCLUDE_PATH="$CPLUS_INCLUDE_PATH:$(python3 -c 'import sysconfig; print(sysconfig.get_path("include"))')"
+case $ARCHITECTURE in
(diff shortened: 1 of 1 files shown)


ktf:
Looks like you have other changes in this PR, can you please remove them and only address the problem you report in the title?

That said, I was thinking that maybe a better way to detect the python installation folder is to use `brew --prefix python3`, check if the reported directory actually exists and if not simply leave `CPLUS_INCLUDE_PATH` empty.

ktf:
Ah, no, leaving it empty would not work...

ktf:
Ok, I think you convinced me your fix is what we want, at least for now... Could you however use:

```
case $ARCHITECTURE in
  osx*)  ;;
esac

i.e. avoid bash-isms? I've read reports that the new macOS will come with ZSH and will not ship bash by default, so I think we should start avoiding them.

mfasDa:
@ktf Thanks a lot for your review! I adapted the PR accordingly. Please have a look.

ktf:
LGTM. I will wait for the tests and merge it once today daily is done.

PR #1683, opened 2019-06-14
Bump QCG to v1.5.0

awegrzyn:


awegrzyn:
@alisw/alien-experts Why tag can't start with a `@`? What shall I do to escape it?

Barthelemy:
@awegrzyn I guess you wanted to call for @alisw/infrastructure-admins

ktf:
No idea why, I suspect some yaml-ism, but it looks like a good idea to avoid it... ;-) Why do you need it?

The culprit seems to be the fact that yaml scanner does not support it:

```
yaml.scanner.ScannerError: while scanning for the next token
found character '@' that cannot start any token
  in "<unicode string>", line 3, column 6:
```

can you try to put it in quotes.

ktf:
From https://yaml.org/spec/1.2/spec.html

```
The “@” (#x40, at) and “`” (#x60, grave accent) are reserved for future use
```

awegrzyn:
Thanks for debugging, we need this as we have multiple Node packages within the same repository (https://github.com/AliceO2Group/WebUi), we are using full Node names to tag our releases: `@org_name/pkg_name@version`.

ktf:
Issue with Readout being followed up at https://alice.its.cern.ch/jira/browse/O2-756. @sy-c

Issue #1680, opened 2019-06-13
Recipe for glfw


aphecetche:
(talking about dependencies on Mac...)

I'd like to be able to compile w/ or w/o glfw on my Mac, without having to brew install/remove it. 
If there was a glfw recipe that would be kind of easy, right ?

Or is there a reason of _not_ having such a recipe ?

(and yes, if we agree that'd be good to have, I'll make it ;-) )

ktf:
The idea was that the GUI should not impose a requirement, especially in the beginning when I was not sure how it would have been welcome. Now that we give it a bit for granted, I guess we can have a proper recipe as well. That's my 2 cents, at least.

Barthelemy:
The only risk I see is with alidock. If we install glfw automatically, the GUIs will be built and I am not sure what will happen when users will run a DPL workflow. Will it crash ?

aphecetche:
Hi Barth,

Please see #1684 : I'm actually not proposing to install glfw automatically, just to have a recipe that makes it easy to install it. 

Use case here is mainly for Macs I would say :  on those, if you install glfw with brew, then you are somehow forced to use it in O2 builds. With a recipe, you have the option (emphasis on option here) to uninstall glfw from brew, install it with the recipe and then, depending on whether or not you use the resulting module in the terminal where you do alibuild, you'll get a build with or without glfw. In that respect, it's not change, but an optional addition only.

Barthelemy:
Hi, 
Thank you for the clarification. It makes sense. Good.
See you

ktf:
For the record. DPL should work just fine with the GUI compiled in but non-functional, e.g. because of missing display. Also notice that modules are not for compile time dependencies. At this point, where the GUI is largely accepted and actually missed if not there, why making it optional?

--
Ciao,
Giulio

aphecetche:
The initial idea was to be able to easily test the cmake build with (what we usually have on Macs) and without glfw (what we usually(?) have on CC7 servers) on the same machine (either Mac or CC7). And as stated above, the mere existence of a glfw recipe makes that possible without changing anything to the current situation, which I guess is just a win-win deal.

[...]

PR #1684, opened 2019-06-14
Add recipe for GLFW

aphecetche:
Here's an attempt at #1680 

For the record, so far I'm happy with just having this recipe available, no need to propagate the dependency explicitly anywhere, as far as I'm concerned.

The way I use it : 

- I don't have glfw installed via brew
- if I do nothing special then alibuild builds o2 without glfw
- if I do `module load GLFW` (after a proper alibuild build GLFW) and then alibuild o2, as GLFW_ROOT is set, the O2 CMakeLists.txt picks it and I get an O2 build w/ GLFW. (at least it's working now on my cmake-migration-step-1 branch where I use `find_package(GLFW NAMES glfw3 CONFIG)`

### DIFF
--- a/glfw.sh
+++ b/glfw.sh
@@ -1,6 +1,9 @@
 package: GLFW
 version: "3.3"
 source: https://github.com/glfw/glfw.git
+build_requires:
+  - CMake
+  - "GCC-Toolchain:(?!osx)"
 ---
 #!/bin/sh
 
@@ -28,4 +31,6 @@ module-whatis "ALICE Modulefile for $PKGNAME $PKGVERSION-@@PKGREVISION@$PKGHASH@
 # Dependencies
 module load BASE/1.0
 setenv GLFW_ROOT \$::env(BASEDIR)/$PKGNAME/\$version
+prepend-path LD_LIBRARY_PATH \$::env(BASEDIR)/$PKGNAME/\$version/lib
+$([[ ${ARCHITECTURE:0:3} == osx ]] && echo "prepend-path DYLD_LIBRARY_PATH \$::env(BASEDIR)/$PKGNAME/\$version/lib")
 EoF
--- a/glfw.sh
+++ b/glfw.sh
@@ -1,0 +1,31 @@
+package: GLFW
+version: "3.3"
+source: https://github.com/glfw/glfw.git
+---
+#!/bin/sh
+
+cmake $SOURCEDIR -DCMAKE_INSTALL_PREFIX=$INSTALLROOT \
+  ${CMAKE_GENERATOR:+-G "$CMAKE_GENERATOR"} \
+  -DBUILD_SHARED_LIBS=ON \
+  -DGLFW_BUILD_EXAMPLES=OFF \
+  -DGLFW_BUILD_TESTS=OFF \
+  -DGLFW_BUILD_DOCS=OFF
+
+cmake --build . -- ${JOBS+-j $JOBS} install
+
+# Modulefile
+MODULEDIR="$INSTALLROOT/etc/modulefiles"
+MODULEFILE="$MODULEDIR/$PKGNAME"
+mkdir -p "$MODULEDIR"
+cat >"$MODULEFILE" <<EoF
+#%Module1.0
+proc ModulesHelp { } {
+  global version
+  puts stderr "ALICE Modulefile for $PKGNAME $PKGVERSION-@@PKGREVISION@$PKGHASH@@"
+}
+set version $PKGVERSION-@@PKGREVISION@$PKGHASH@@
+module-whatis "ALICE Modulefile for $PKGNAME $PKGVERSION-@@PKGREVISION@$PKGHASH@@"
+# Dependencies
+module load BASE/1.0
+setenv GLFW_ROOT \$::env(BASEDIR)/$PKGNAME/\$version
+EoF
--- a/glfw.sh
+++ b/glfw.sh
@@ -5,13 +5,11 @@ build_requires:
   - CMake
   - "GCC-Toolchain:(?!osx)"
 ---
-#!/bin/sh
-
 cmake $SOURCEDIR -DCMAKE_INSTALL_PREFIX=$INSTALLROOT \
-  ${CMAKE_GENERATOR:+-G "$CMAKE_GENERATOR"} \
-  -DBUILD_SHARED_LIBS=ON \
-  -DGLFW_BUILD_EXAMPLES=OFF \
-  -DGLFW_BUILD_TESTS=OFF \
+  ${CMAKE_GENERATOR:+-G "$CMAKE_GENERATOR"}          \
+  -DBUILD_SHARED_LIBS=ON                             \
+  -DGLFW_BUILD_EXAMPLES=OFF                          \
+  -DGLFW_BUILD_TESTS=OFF                             \
   -DGLFW_BUILD_DOCS=OFF
 
 cmake --build . -- ${JOBS+-j $JOBS} install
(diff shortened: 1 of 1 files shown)


ktf:
Sorry for taking long. I am back from holidays. Will review it now.

glfw.sh
```diff
@@ -0,0 +1,31 @@
+package: GLFW
+version: "3.3"
+source: https://github.com/glfw/glfw.git
+---
+#!/bin/sh
+
+cmake $SOURCEDIR -DCMAKE_INSTALL_PREFIX=$INSTALLROOT \
```
ktf: Please align the `\`.

glfw.sh
```diff
@@ -0,0 +1,31 @@
+package: GLFW
+version: "3.3"
+source: https://github.com/glfw/glfw.git
```
ktf: I guess you need at least a `build_requires` on `CMake` and `GCC-ToolChain`?

glfw.sh
```diff
@@ -0,0 +1,31 @@
...
+#%Module1.0
+proc ModulesHelp { } {
+  global version
+  puts stderr "ALICE Modulefile for $PKGNAME $PKGVERSION-@@PKGREVISION@$PKGHASH@@"
+}
+set version $PKGVERSION-@@PKGREVISION@$PKGHASH@@
+module-whatis "ALICE Modulefile for $PKGNAME $PKGVERSION-@@PKGREVISION@$PKGHASH@@"
+# Dependencies
+module load BASE/1.0
+setenv GLFW_ROOT \$::env(BASEDIR)/$PKGNAME/\$version
```
ktf: The module file looks a bit too basic. Don't you need at least `LD_LIBRARY_PATH`? Also please do not use `setenv` but `set`.

aphecetche: Indeed. I'm missing (DY)LD_LIBRARY_PATH. Will add it.
But why set instead of setenv ?

glfw.sh
aphecetche: yep. 

Did not notice the GCC-ToolChain is missing as my zshrc on my CentOS dev. machine has : 

[[ -a /opt/rh/devtoolset-7/enable ]] && source /opt/rh/devtoolset-7/enable

in it...

glfw.sh
ktf: `DYLD_LIBRARY_PATH` is not actually needed anymore, since we do use macOS rpath equivalent on mac.

I would always avoid doing a `setenv` for something which is not in general expected by a package to avoid polluting the environment and to make sure that if we start to use things from outside `aliBuild`, we do not rely on `aliBuild-isms`. All of the `_ROOT` for external packages fall in this category.

aphecetche: Hum, but then if I change the setenv to a set then I indeed can no longer use that recipe for setting correctly my build with/without GLFW, right ?

For the DYLD_LIBRARY_PATH I believe it is still needed in cases where the library is loaded at runtime only (e.g. as a plugin), right ?

aphecetche: For GLFW_ROOT, I would say it's not completely specific to aliBuild : CMake uses also -D<PackageName>_ROOT variables in its (rather complex and/or complete, however you want to put it) search procedure for <PackageName>Config.cmake file. So having a GLFW_ROOT makes the cmake configure stage (a lot) easier.

ktf: Agreed, but that's a compile time option and indeed if you use the per package generated `etc/profile.d/init.sh` you get all the `_ROOT` variables automagically. Environment modules are for runtime dependencies, and if we start relying on `SOMETHING_ROOT` but then the, e.g., home-brew version does not export it, we are in trouble.

If you want to optionally enable glfw (but why, at this point?) how about you simply source the relevant `init.sh`?

aphecetche: 1. sad realization : I did not know (or I forgot) about the generated etc/ structure of the packages. But that _is_ an alibuild-ism, right ? How are they supposed to be used ?
2. I still don't get why you keep saying that module files are only for runtime dependencies. You mean O2 decided to use them only like that ? Because on e.g. HPC systems, module files are used to select compilers for instance.
3. optional GLFW : see https://github.com/alisw/alidist/issues/1680#issuecomment-502656689

ktf: 1. That is a `alibuild-ism`, but given you are using it to then rebuild a package built with aliBuild, I think it's the correct thing to do. `init.sh` is exactly the environment aliBuild itself uses for compilation. At the moment to use it you need to do `WORK_DIR=<path-to-sw> source sw/package/version/etc/profile.d/init.sh`, IIRC. This will give you the same environment as the one the build recipe gets when being built.
2. Yes, I mean that in O2 the information provided in module files is in general not sufficient to actually be able to use it for compilation, albeit it might work for many packages. Notice also that while it's true that on HPC systems module allow you to select the compiler executables, most often they simply modify `$PATH` and then you then rely on something like `pkg-config` / `cmake` / `<something>-config` to select the dependencies. This has the advantage that you do not need to rely on a particular convention of your build environment to do your builds. Maybe what we should really do is to make sure pkg-config environment is provided by the module file and then in the recipe we use that for the detection? This would have the advantage that it will work for any `pkg-config` enabled system....
3. Would't proper defaults solve the optional dependency issue?

aphecetche: Concerning point 2 : given that many of our dependencies are building with cmake anyway, I have the feeling that things would be simpler if our install zone would ressemble more closely what cmake [find_package](https://cmake.org/cmake/help/v3.14/command/find_package.html#search-procedure) expects (I guess that would help pkg-config systems also, but here I'm talking without knowing ;-) )
As far as I can tell the "only" thing that gets in the way of this are the "latest" subdirectories, i.e. we have : 

prefix/package1/latest/lib ...
prefix/package2/latest/lib ...

while if we had 

prefix/latest/package1
prefix/latest/package2

and used prefix/latest as CMAKE_INSTALL_PREFIX I believe cmake configure would just find the dependencies we are building without us having to specify a ton of -DXXX_ROOT variables.

I've not actually tested what I'm saying here though.

ktf: Mmm.. I am not sure I understand. If you set CMAKE_INSTALL_PREFIX=prefix/latest, how do you  specify where to install the separate package? In principle we could easily create symlinks for `
prefix/latest/package1` and leave the current real installation as is, but it's not clear to me how you would be able to do that.

aphecetche: We might imagine to reverse the logic and install in CMAKE_INSTALL_PREFIX=prefix/latest and then move to prefix/package/latest (and link back from there to prefix/latest). A bit convoluted but I guess that _should_ work.

Other, less changing, option that I have to investigate is e.g. the CMAKE_SYSROOT variable. 

btw, I guess we should move this discussion to a Jira ticket, as it's clearly not specific to the GLFW recipe ;-) which I'll do once I have a clearer picture of the CMAKE_SYSROOT for instance.

ktf: The problem I see with using `prefix/package` as main source is that we actually build in `sw/BUILD/<SOME-HASH>/package`. The `<SOME-HASH>` bit is critical because that's what we use to make sure we can relocate the package (basically we replace `.../<SOME-HASH>/` with `/prefix`).

aphecetche: I've yet to make a proper JIRA ticket for this, but a small follow-up : I guess the CMake variable I was after is CMAKE_PREFIX_PATH (not CMAKE_SYSROOT which is targeted at cross-compiling). 

Using the quick and dirty script below to set up links in an existing install zone, a `cmake -DCMAKE_PREFIX_PATH=$ALIBUILD_WORK_DIR/slc7_x86-64/zoby -DCMAKE_INSTALL_PREFIX=$ALIBUILD_WORK_DIR/slc7_x86-64/` goes a long way towards the goal I would say (with one of my cmake-migration branches, at least, as it requires proper usage of the find stuff on our side). Boost still requires the setting of BOOST_ROOT, but most other deps were found just fine.

```
#!/bin/sh

prefix=${1:-$ALIBUILD_WORK_DIR/slc7_x86-64}
label=${2:-latest}

if ! test -d $prefix; then
  echo "directory $prefix does not exist"
  exit 65
fi

targetdir=${3:-zoby}

targetdir=$prefix/$targetdir

echo "prefix=$prefix"
echo "targetdir=$targetdir"
echo "label=$label"

if ! test -d $targetdir; then
  mkdir $targetdir
fi

find $prefix -maxdepth 2 -type l -name $label | while read dir; do
  ln -sf $dir $targetdir/$(basename $(dirname $dir))
done
```

ktf:
I took the liberty to rework this a bit, so that we can go forward with the fully agreed part. Let's discuss the rest of the things where we do not agree in a separate PR.

ktf:
Error on macOS unrelated, tested locally. Merging.

PR #1685, opened 2019-06-17
WIP: Fix JAliEn-ROOT build on MacOS

Atlantic777:
This PR introduces modifications needed to build JAliEn-ROOT on MacOS. The XRootD 4.9.x seems to have troubles building on some MacOS machines so we will stick to 4.8.x until we understand better what is the issue.

After this PR is merged we can trigger another special AliPhysics build with JAliEn support and publish it in CVMFS.

### DIFF
--- a/defaults-jalien.sh
+++ b/defaults-jalien.sh
@@ -10,7 +10,7 @@ env:
 overrides:
   XRootD:
     version: "%(tag_basename)s_JALIEN"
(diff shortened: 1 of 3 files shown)


pzhristov:
The build on MacOS fails because of OpenSSL and is not related to this pull request.

dberzano:
Building now.

PR #1694, opened 2019-06-19
Prep work for cmake migration

aphecetche:
I've tried hard to ensure that my cmake migration requires the minimum amount of changes compared to the existing so it's easy to test, but I'm afraid I'd need one change in this recipe though. 

In the new scheme the tests are meant be ran from the build directory directly (i.e. before install) (but can still be ran after install, though I generally don't see the point). This is possible because there's now a "stage" area in the build tree which mimics the install area somehow (not a lot different from the current situation, just that build/bin build/lib are now build/stage/bin and build/stage/lib). But that stage area also contains important data files (e.g. the magnetic field) that are inputs for some tests, and which are unfortunately wiped out by the o2.sh recipe. This PR fixes this.

(this is one of the reasons for the failure of https://github.com/AliceO2Group/AliceO2/pull/2100 - not the only one unfortunately ;-) )

### DIFF
--- a/o2.sh
+++ b/o2.sh
@@ -53,7 +53,7 @@ incremental_recipe: |
     # Clean up old coverage data and tests logs
     find . -name "*.gcov" -o -name "*.gcda" -delete
     # cleanup ROOT files created by tests in build area
-    find $PWD -path $PWD/stage -prune -o -name "*.root" -delete
+    find $PWD -not \( -path $PWD/stage -prune \) -name \*.root -exec rm {} \;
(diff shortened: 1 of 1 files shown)


aphecetche:
Hold on. Something's still not correct. Investigating...

aphecetche:
Sorry for the multiple messages.
I think I can cook something in my new CMakeLists.txt to avoid patching the o2.sh recipe right now (will be needed only later, in a "cleaning phase"), so I'm closing this one for now.

PR #1695, opened 2019-06-19
Add treelite among the modules loaded

mpuccio:


mpuccio:
Hi, it looks like the Mac test has some issues. Can someone have a look to both that and this PR?

ktf:
Also, could you please provide sensible / short names for the version? VO_ALICE@treelite::2a12742269c1d3de553d9a12ff36bc2a5d874239-2 is a bit of pain. Notice you can separate tag and version...

mpuccio:
Ciao Giulio,
we would need the TREELITE_ROOT variable for setting the include paths in the rootlogon.C macro, would it be a problem to keep it?
On the second issue: unfortunately the treelite people are not tagging the versions... Is there a way to have our own label for that? (apart forking the repo and tagging ourself)

ktf:
Can't you simply add the required entry to ROOT_INCLUDE_PATH?

mpuccio:
OK! As discussed offline I improved the PR according to your comments

treelite.sh
```diff
@@ -37,8 +37,8 @@ module-whatis "ALICE Modulefile for $PKGNAME $PKGVERSION-@@PKGREVISION@$PKGHASH@
...
 # Our environment
 set osname [uname sysname]
-setenv TREELITE_ROOT \$::env(BASEDIR)/$PKGNAME/\$version
-prepend-path PATH \$::env(TREELITE_ROOT)/bin
-prepend-path LD_LIBRARY_PATH \$::env(TREELITE_ROOT)/lib
-$([[ ${ARCHITECTURE:0:3} == osx ]] && echo "prepend-path DYLD_LIBRARY_PATH \$::env(TREELITE_ROOT)/lib")
+prepend-path PATH \$::env(BASEDIR)/$PKGNAME/\$version/bin
+prepend-path ROOT_INCLUDE_PATH \$::env(BASEDIR)/$PKGNAME/\$version/include
+prepend-path LD_LIBRARY_PATH \$::env(BASEDIR)/$PKGNAME/\$version/lib
+$([[ ${ARCHITECTURE:0:3} == osx ]] && echo "prepend-path DYLD_LIBRARY_PATH \$::env(BASEDIR)/$PKGNAME/\$version/lib")
```
ktf: While you are at it, can you drop `DYLD_LIBRARY_PATH`? I think we do not need it anymore with the new way relocation is done.

ktf:
Thanks. I will merge it after today daily to avoid issues on friday...

treelite.sh
mpuccio: OK! I push it here, but maybe could you wait for merging that I test locally in my mac? I feel a bit uneasy to do this change with the macos tests down.

PR #1697, opened 2019-06-20
Exclusion no longer needed, GPU code has adopted O2 license

davidrohr:


davidrohr:
@dberzano : Done
no idea why it fails in the MacOS CI, but I cannot imagine it is related to this PR.

PR #1699, opened 2019-06-21
Make sure we can pick openssl from the system on mac

Barthelemy:
It is not able to pick up openssl from the system because the `-I xxx` is missing in the second compilation test.


ktf:
This seems to fail actually...

Barthelemy:
@ktf I guess it is then a subsequent problem as describe here : https://alice.its.cern.ch/jira/browse/QC-200

ktf:
Errors unrelated.

PR #1708, opened 2019-07-01
Run tests during PRs in QC (QC-181)

Barthelemy:


qualitycontrol.sh
```diff
@@ -57,6 +57,10 @@ JOBS=$((${JOBS:-1}*3/5))
 [[ $JOBS -gt 0 ]] || JOBS=1
 cmake --build . -- ${JOBS:+-j$JOBS} install
 
+if [[ $ALIBUILD_O2_TESTS ]]; then
+  make test
```
aphecetche: Why not calling ctest directly ? Otherwise, if we use ninja for building, that line will fail...

Barthelemy: sure, I will do it.

qualitycontrol.sh
```diff
@@ -58,7 +58,18 @@ JOBS=$((${JOBS:-1}*3/5))
 cmake --build . -- ${JOBS:+-j$JOBS} install
 
 if [[ $ALIBUILD_O2_TESTS ]]; then
-  make test
+  TESTERR=
+  ctest --output-on-failure ${JOBS+-j $JOBS} || TESTERR=$?
+  # Display additional logs for tests that timed out in a non-fatal way
+  set +x
+  for LOG in test_logs/*.nonfatal; do
```
aphecetche: hum, as far as I can tell, this will be ineffective (at best), as the `test_logs` structure is not a ctest feature, but one the O2 `tests-wrapper.sh` script.
Besides, that's not need in QC a there's no such thing as a non-fatal test in QC, right ?

Barthelemy:
I WIP it because the test fails due to OpenSSL and I would like to see if the tests pass.

qualitycontrol.sh
Barthelemy: correct, I went too fast on the copy-paste road.

Barthelemy:
I had to add two environment variables 
1. QUALITYCONTROL_ROOT to let the framework find something in etc (could be avoided somehow)
2. Addition to LD_LIBRARY_PATH: library directory to find the modules libraries. 

I am not really happy about it. Not sure how to do though.

aphecetche:
hum, indeed the QUALITYCONTROL_ROOT should no longer be needed at all for the tests (replaced by the getTestDataDirectory function) ... unless I forgot some usages, of course ;-)

Barthelemy:
QUALITYCONTROL_ROOT is used by the framework to access old streamerinfos in the form of a root file in `etc`. What I could do is make it optional and get rid of this line. The LD_LIBRARY_PATH seems needed though.

ktf:
Why don't you have `QUALITYCONTROL_ROOT` provided by some `quality-control-config` executable? This way you can simply put it in `$PATH` without polluting the environment...

Barthelemy:
I have modified QC to survive not having this variable set

Barthelemy:
@ktf mid-term I'll do what you propose

Barthelemy:
There are failing tests in O2, many failing tests, and it prevents me from evaluating this PR.

ktf:
Is this ready to be merged?

Barthelemy:
@ktf I don't see in the logs the QC being built and tested. I thought that a PR in alidist triggered a build downstream (and actually it was doing it earlier in this same PR). Is it just missing from the logs or is it not executed ?

Barthelemy:
@alisw/dataflow-experts could someone review ?

ktf:
IIRC, PRs of alidist do not set the variable. Only checking of the PRs for actual package does.

Barthelemy:
@ktf ok thank you

PR #1711, opened 2019-07-01
Add RAPID_JSON_HOME variable to module file

MichaelLettrich:
exports the base directory of rapidJSON as environment variable RAPID_JSON_HOME, additions to PATH and LD_LIBRARY_PATH are derived from it. This allows easier development of applications inside the alienv environment.


MichaelLettrich:
Modules are for runtime only.

PR #1725, opened 2019-07-08
Set LD_LIBRARY_PATH as a CMake variable so ctest (of macros) works al…

aphecetche:
…so with SIP enabled

This is not critical nor urgent (as the recommended configuration is to disable SIP) but might be considered at some point (once https://github.com/AliceO2Group/AliceO2/pull/2172 is in, anyway). Without this a ctest fails on most of the macros because an empty LD_LIBRARY_PATH is passed to the root process launching the macro.

### DIFF
--- a/o2.sh
+++ b/o2.sh
@@ -168,8 +168,10 @@ cmake $SOURCEDIR -DCMAKE_INSTALL_PREFIX=$INSTALLROOT                            
       ${ARROW_ROOT:+-DARROW_HOME=$ARROW_ROOT}                                               \
       -Dbenchmark_DIR=${GOOGLEBENCHMARK_ROOT}/lib/cmake/benchmark                           \
       ${GLFW_ROOT:+-DGLFW_LOCATION=$GLFW_ROOT}                                              \
-      ${FMT_ROOT:+-DFMT_ROOT=${FMT_ROOT}}                                                    \
-      ${CUB_ROOT:+-DCUB_ROOT=$CUB_ROOT}
+      ${FMT_ROOT:+-DFMT_ROOT=${FMT_ROOT}}                                                   \
+      ${CUB_ROOT:+-DCUB_ROOT=$CUB_ROOT}                                                     \
+      ${LD_LIBRARY_PATH:+-DLD_LIBRARY_PATH=${LD_LIBRARY_PATH}}
+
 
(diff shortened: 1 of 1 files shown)


ktf:
You mean `DYLD_LIBRARY_PATH`? We concluded in some unrelated thread `LD_LIBRARY_PATH` does not do anything anymore.

aphecetche:
Nope. I mean LD_LIBRARY_PATH. 

A kind of lame/convoluted reproducer of the issue (launching root from a subprocess that no longer have LD_LIBRARY_PATH set) could be : 
(FairRoot used for no good reason) 

```
#!/bin/sh

INSTDIR=$ALIBUILD_WORK_DIR/osx_x86-64

cat << EOF > test.C
#include "FairGenerator.h"
#include "FairPrimaryGenerator.h"

class Test : public FairGenerator {

public:
  Bool_t ReadEvent(FairPrimaryGenerator *) override { return kTRUE; }
};

void test() { Test a; }
EOF


export ROOT_INCLUDE_PATH=$INSTDIR/FairRoot/latest/include

LD_LIBRARY_PATH=$INSTDIR/FairRoot/latest/lib:$INSTDIR/pythia/latest/lib:$INSTDIR/lhapdf/latest/lib:$INSTDIR/pythia6/latest/lib:$INSTDIR/protobuf/latest/lib:$INSTDIR/GEANT4_VMC/latest/lib:$INSTDIR/vgm/latest/lib:$INSTDIR/GEANT4/latest/lib:$INSTDIR/GEANT3/latest/lib64:$INSTDIR/FairMQ/latest/lib:$INSTDIR/DDS/latest/lib:$INSTDIR/asiofi/latest/lib:$INSTDIR/ofi/latest/lib:$INSTDIR/ZeroMQ/latest/lib:$INSTDIR/FairLogger/latest/lib

echo "This should work"

LD_LIBRARY_PATH=$LD_LIBRARY_PATH $INSTDIR/ROOT/latest/bin/root -b << EOF
.L test.C++
.q
EOF

echo "This should fail is SIP is enabled"

$INSTDIR/ROOT/latest/bin/root -b << EOF
.L test.C++
.q
EOF
```

ktf:
In principle we now do the correct `@rpath` tricks so tests should run even if `(DY)LD_LIBRARY_PATH` is not set or if you have SIP enabled.

PR #1733, opened 2019-07-15
Reduce ROOT CMake CXX INTERFACE COMPILE FEATURE from 17 to 14 to be compatible with CUDA

davidrohr:
This is an alternative to AliceO2Group/AliceO2#2186
It just replaces all ROOT CXX requirements from 17 to 14 in our alidist recipe.

I have to say I do not like this hack either, but it is much better than adding tens of manual include dependencies in O2. And after spending quite some time searching for the solution, I think there are only the following possibilities (out of which I like this PR best):
- The much more intrusive PR AliceO2Group/AliceO2#2186 (which handles all includes for CUDA compilation manually in O2).
- This PR, which hacks the ROOT C++17 requirements out (This does not change anything, as we anyway set the CXX standard manually).
- Go back from modern CMake for ROOT to the old style, using ROOT_INCLUDE_DIRS and ROOT_LIBRARIES instead of the targets.
- Go back from modern CMake with CUDA as language to FindCUDA (which is deprecated), since this does not do the language feature check (since it does not consider CUDA a language).
- Implement a new feature in CMake which allows to manually specify compiler and compile flags for language dialects unknown to CMake (rather long term, since it would need to be developed, and we'd need to bump the CMake version eventually).

Finally, I'd like to mention that the same problem will hit us with any dependency that will impose a CXX standard on CMake library interface level. So far, only ROOT does this. If this happens in the future, we'd have to apply the same hack as well.

@ktf @ihrivnac @aphecetche @dberzano : Please let me know what you think. If there are no objections, I'd suggest we go ahead with this solution, since it is not a dramatic hack. It will be mandatory for ROOT 6.18, and also for ROOT 6.16 it will allow me to remove some workarounds.

### DIFF
--- a/root.sh
+++ b/root.sh
@@ -32,6 +32,10 @@ incremental_recipe: |
          "$INSTALLROOT/etc/plugins/TSystem/P030_TAlienSystem.C" \
          "$INSTALLROOT/etc/plugins/TFile/P070_TAlienFile.C"
 
+  # Reduce ROOT C++ Standard Interface Requirement to C++14 for CUDA compatibility
+  sed -i.deleteme -e "s/cxx_std_17/cxx_std_14/" "$INSTALLROOT/cmake/ROOTConfig-targets.cmake" || true
+  find . -name '*.deleteme' -exec rm -f '{}' \; || true
+
   mkdir -p $INSTALLROOT/etc/modulefiles && rsync -a --delete etc/modulefiles/ $INSTALLROOT/etc/modulefiles
 ---
 #!/bin/bash -e
@@ -176,6 +180,8 @@ fi
 
 # Make some CMake files used by other projects relocatable
 sed -i.deleteme -e "s!$BUILDDIR!$INSTALLROOT!g" $(find "$INSTALLROOT" -name '*.cmake') || true
+# Reduce ROOT C++ Standard Interface Requirement to C++14 for CUDA compatibility
+sed -i.deleteme -e "s/cxx_std_17/cxx_std_14/" "$INSTALLROOT/cmake/ROOTConfig-targets.cmake" || true
 find . -name '*.deleteme' -exec rm -f '{}' \; || true
 
 rm -vf "$INSTALLROOT/etc/plugins/TGrid/P010_TAlien.C"         \


ktf:
LGTM; I assume all the interfaces root exposes are C++14 compatible, correct?

davidrohr:
Well, one will actually need c++17 for some root headers (i.e. i must not include certain root headers inside .ci files), but we have this flag anyway since we set in the o2-defaults and also in the O2 CMake the c++ level to 17. So for the actual compile commands, nothing changes at all. It is always -std=c++17.

 Only if we'd totally mess up something and compile within o2-defaults ROOT with c++17 but O2 with c++14, this would not be noticed at CMake level and we would get compile errors in case an incompatible interface is used.

ihrivnac:
Hi @davidrohr ,
I tried to understand what you propose and from the file changes this looks like you let Root build with cxx17 but you change afterwards the ROOTConfig-targets.cmake created by ROOT and where you replace all cxx_std_17 with cxx_std_14, what I find quite dangerous.

 I wanted to test on my side but after updating of everything (including aliBuild, and alidist from your branch) my build is stuck at:
```
==> Building Python-modules@1.0
==> Python-modules is being built (use --debug for full output)...  
```
and does not progress.
From the log of Python-modules in BUILD it looks like the build is finished, but aliBuild does not move on the next package.

davidrohr:
What this changes is the minimum c++ standard needed by a compiler to link to the ROOT library. This is purely a CMake requirement. Alibuild will build O2 (basically everything with o2-defaults) with C++17. So basically nothing changes. The command lines created by CMake are identical.

The only thing is that now it is possible to have a chain of dependencies from something that is c++14 down to a ROOT library, which is unavoidable for our GPU code.

It is quite ugly but I don't think it is dangerous. As long as we do not reduce the c++ standard in O2 / o2-defaults to c++14 while at the same time still compiling ROOT with c++17, nothing can happen.

ihrivnac:
Thanks @davidrohr for your explanations.  At the end my local build passed the Python-modules, it just require more time, and I was able to build everything. I'd just suggest to add a clear comment for the added lines so that we can understand later why they were added.

davidrohr:
@ihrivnac : comments have been added.

ihrivnac:
Ok, thanks. There are no objections from my side now.

ktf:
@ihrivnac can you please move to alibuild 1.5.5.rc1 (`pip install alibuild=1.5.5.rc1`) which should speedup the building of `Python-modules`?

PR #1734, opened 2019-07-16
Add git-clang-format to the binary installation.

ktf:


ktf:
Not merging this yet, because I guess we could simply add git-clang-format to our mirror of the clang repository.

PR #1749, opened 2019-07-24
O2: add GLFW as an explicit dependency

ktf:
Given not having this as a dependency is creating more troubles it solves, I propose to add the dependency and leave to different defaults / `--disable` option the choice of whether to build it or not.


ktf:
@ihrivnac to be discussed at next WP3, I guess.

ihrivnac:
Hi @ktf,
Could you be, please, more explicit?  Which code/libraru in O2 needs this dependency?
And could you add this item into the WP3 meeting document linked at https://indico.cern.ch/event/836954/?

Thanks,

ktf:
We discussed this a bit in WP3. We should make sure that if OpenGL is not present on the machine, we do not build it.

PR #1751, opened 2019-07-25
Bump ReadoutCard to v0.11.8

kostorr:


ktf:
@MichaelLettrich could you check what is going on with the macOS builds?

MichaelLettrich:
On it.

MichaelLettrich:
macOS machines should be back on track. We were running out of disk on all macOS machines for alidist. Problem should be solved.

PR #1754, opened 2019-07-29
Enforce no warning code in Readout

awegrzyn:


awegrzyn:
When it can be done?

sy-c:
end August

ktf:
Again, I think this is an actual error?

ktf:
Can it be merged now?

PR #1764, opened 2019-08-06
O2-809 add recipe for python modules that use ROOT

MichaelLettrich:
Several Python packages of scikit-hep compile and link against the
installed version of ROOT. This receipe installs them, linking against
the ROOT version installed by alidist.

### DIFF
--- a/python-modules-root.sh
+++ b/python-modules-root.sh
@@ -1,0 +1,89 @@
+package: Python-modules-ROOT
+version: "1.0"
+requires:
+  - "Python:slc.*"
+  - "Python-system:(?!slc.*)"
+  - FreeType
+  - libpng
+  - Python-modules
+  - ROOT
+build_requires:
+  - curl
+prepend_path:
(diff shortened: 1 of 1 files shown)


ktf:
This looks ok, however I suspect you will duplicate some pip dependencies between this and Python-modules. Maybe you should post process `$INSTALLROOT/lib/python...` and drop the duplicaticates?

MichaelLettrich:
True. What I am thinking of now: how do we proceed if versions of the dependencies  don't match?

MichaelLettrich:
As discussed with @ktf and @Atlantic777 there currently is a blocker for merging:
alidist packages are always installed as an immutable entity. As a consequence each alidist python recipe will install a full chain of dependencies.

Possible solutions around it:
- find a way to make pip aware of the packages installed in site-packages of the recipes it depends on
- Installation and partial deletion: Prepend all packages that are going to be installed by the recipes we depend on in our recipe, install our own packages. Finally delete all duplicates.

ktf:
@MichaelLettrich is this ready to go? Was the blocker fixed?

PR #1765, opened 2019-08-06
O2-810 Cleanup alidist Python-modules recipe

MichaelLettrich:
There are some workarounds in the python-modules recipe that are no longer needed. remove them.


ktf:
Will merge after the daily.

PR #1768, opened 2019-08-07
Update arrow to 0.14.1

ktf:
* Pick up double-conversion from our own installation.
* Move to use Release builds
* double-conversion: init at v3.1.5
* Force usage of our own tools

### DIFF
--- a/arrow.sh
+++ b/arrow.sh
@@ -1,7 +1,7 @@
 package: arrow
-version: v0.12.0-alice6
-tag: apache-arrow-0.12.0-alice6
-source: https://github.com/alisw/arrow
+version: v0.14.1
+tag: apache-arrow-0.14.1
+source: https://github.com/apache/arrow
 requires:
   - boost
   - lz4
@@ -10,6 +10,7 @@ build_requires:
   - zlib
   - flatbuffers
   - CMake
+  - double-conversion
 env:
   ARROW_HOME: "$ARROW_ROOT"
 ---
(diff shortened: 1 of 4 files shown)


ktf:
@dennisklein do you understand why on mac liblz4 is required by MpsAPI? I cannot find any reference to it.

dennisklein:
> @dennisklein do you understand why on mac liblz4 is required by MpsAPI? I cannot find any reference to it.

I suspect this https://github.com/FairRootGroup/FairRoot/blob/master/MbsAPI/CMakeLists.txt#L73. It is probably overlinking unused Root libraries that need liblz4 ?! I can check later or tomorrow.

dennisklein:
On linux there is a tool called `lddtree` as part of the `pax-utils` package which should reveal that `liblz4` is not a direct dependency of MbsAPI. If it is, I am puzzled.

dennisklein:
https://github.com/gentoo/pax-utils. Appears to work on Mac too.

ktf:
Monday is fine.

ktf:
I've updated the PR to pick your commit in FairRoot. @davidrohr @sawenzel Any objections to pick that up?

davidrohr:
Hi Giulio,

Sure, you can update the Fairroot commit. The commit I have still lacks peter's fix for MacOS as pointed out by Adam.

I would prefer to stick to the official fairroot repo instead of denis' fork. Can't they just merge the fixes into master?

ktf:
Yes, sure, Dennis branch is just for testing

davidrohr:
Ok, makes sense. Are you including Peter's fix for MacOS?

ktf:
@dennisklein your commit does not seem to be enough. I suspect the extra dependency comes from the fact that libTree -> arrow -> lz4.

ktf:
So apparently the problem is that lz4 embeds some bogus id by default. Not sure why. I am replacing it by hand now.

PR #1771, opened 2019-08-08
QualityControl: do not reduce number of builders

ktf:


ktf:
In an effort to improve the build times of `O2Suite`, this seems to be a low hanging fruit. What was the reason for reducing the number of workers? Does it still matter?

Barthelemy:
It was done by Dario in this commit : https://github.com/alisw/alidist/commit/6e315811ee28de3c6f7b34c1f5aa7e6d084f7e7f

I don't know why it was needed or whether it was needed.

Barthelemy:
I can't see the details of the build failure.

ktf:
Yeah, I saw that, but there is no mention of it the commit message. @dberzano do you recall?

PR #1776, opened 2019-08-13
Bump boost / DDS / FairMQ / FairROOT

davidrohr:
@sawenzel : Here is the PR. On my laptop, I used system boost 1.70, so I didn't test this with alidist. Let's see if the alidist recipe compiles with 1.70.

### DIFF
--- a/boost.sh
+++ b/boost.sh
@@ -1,6 +1,6 @@
 package: boost
 version: "%(tag_basename)s"
-tag: v1.68.0
+tag: v1.70.0
 source: https://github.com/alisw/boost.git
 requires:
   - "GCC-Toolchain:(?!osx)"
@@ -11,7 +11,7 @@ build_requires:
   - "bz2"
 prefer_system: (?!slc5)
 prefer_system_check: |
-  printf "#include \"boost/version.hpp\"\n# if (BOOST_VERSION < 106800 || BOOST_VERSION > 106899)\n#error \"Cannot use system's boost: boost 1.68 required.\"\n#endif\nint main(){}" | c++ -I$(brew --prefix boost)/include -xc++ - -o /dev/null
+  printf "#include \"boost/version.hpp\"\n# if (BOOST_VERSION < 107000 || BOOST_VERSION > 107099)\n#error \"Cannot use system's boost: boost 1.70 required.\"\n#endif\nint main(){}" | c++ -I$(brew --prefix boost)/include -xc++ - -o /dev/null
 prepend_path:
   ROOT_INCLUDE_PATH: "$BOOST_ROOT/include"
 ---
--- a/dds.sh
+++ b/dds.sh
@@ -1,6 +1,6 @@
 package: DDS
 version: "%(tag_basename)s"
-tag: "2.2"
+tag: "2.4"
 source: https://github.com/FairRootGroup/DDS
 requires:
   - boost
--- a/fairmq.sh
+++ b/fairmq.sh
@@ -1,6 +1,6 @@
 package: FairMQ
 version: "%(tag_basename)s"
-tag: v1.4.2
+tag: v1.4.7
 source: https://github.com/FairRootGroup/FairMQ
 requires:
  - boost
--- a/fairroot.sh
+++ b/fairroot.sh
@@ -1,6 +1,6 @@
 package: FairRoot
 version: "%(short_hash)s"
-tag: "2b44237b19dc05078c974597f4508632e9b95049"
+tag: "6af7e848585ce720e96a01d5565b1bd5e3884430"
 source: https://github.com/FairRootGroup/FairRoot
 requires:
   - generators
@@ -11,6 +11,7 @@ requires:
   - DDS
   - FairLogger
   - FairMQ
+  - yaml-cpp
   - "GCC-Toolchain:(?!osx)"
 env:
   VMCWORKDIR: "$FAIRROOT_ROOT/share/fairbase/examples"
--- a/readout.sh
+++ b/readout.sh
@@ -1,6 +1,6 @@
 package: Readout
 version: "%(tag_basename)s"
-tag: v1.0.5
+tag: v1.0.6
 requires:
   - boost
   - "GCC-Toolchain:(?!osx)"
--- a/yaml-cpp.sh
+++ b/yaml-cpp.sh
@@ -8,7 +8,7 @@ build_requires:
   - CMake
 prefer_system: (?!slc5)
 prefer_system_check: |
-  printf "#include \"yaml-cpp/yaml.h\"\n" | c++ -std=c++17 -I`brew --prefix yaml-cpp`/include -I`brew --prefix boost`/include -xc++ - -c -o /dev/null
+  pkg-config --atleast-version=0.6.2 yaml-cpp && printf "#include \"yaml-cpp/yaml.h\"\n" | c++ -std=c++17 -I`brew --prefix yaml-cpp`/include -I`brew --prefix boost`/include -xc++ - -c -o /dev/null
 ---
 #!/bin/sh
 case $ARCHITECTURE in


sawenzel:
On Ubuntu, not seeing a compiler crash but the following problem during compile of FairRoot:
```
In file included from /usr/include/yaml-cpp/node/iterator.h:13:                                                                                                              
/usr/include/yaml-cpp/node/detail/iterator.h:48:54: error: no member named 'next' in namespace 'boost'                                                                       
  void increment() { this->base_reference() = boost::next(this->base()); }                                                                                                   
                                              ~~~~~~~^                                                                                                                       
Error: /home/swenzel/alisw_new/sw/ubuntu1804_x86-64/ROOT/v6-18-00-2/bin/rootcint: compilation failure (/home/swenzel/alisw_new/sw/BUILD/f96661bec0ecb283be3e495266d786c89457b
179/FairRoot/fairtools/MCConfigurator/G__MCConfiguratorDict69942550eb_dictUmbrella.h)                                                                                        
fairtools/MCConfigurator/CMakeFiles/MCConfigurator.dir/build.make:62: recipe for target 'fairtools/MCConfigurator/G__MCConfiguratorDict.cxx' failed                          
make[2]: *** [fairtools/MCConfigurator/G__MCConfiguratorDict.cxx] Error 1           
```

davidrohr:
@sawenzel : Can you confirm you have yaml-cpp 0.6.2?
I think it is picked up from the system, and the system-check doesn't enforce the new version.
@pzhristov mentioned some issue with older yaml versions, which I didn't see as my system is at 0.6.2.

ktf:
@MichaelLettrich I think you still need to update ali-bot on the mac machines? Could you confirm?

ktf:
The issue in O2/alidist is most likely a memory one. I've increased the memory available to the builder container to 8GB.

sawenzel:
If we require a specific yaml, then we should put the corresponding check into the recipe.

davidrohr:
Indeed, but I don't which version won't work. Anyway, 0.6.2 is from May 2018 already.

MichaelLettrich:
> @MichaelLettrich I think you still need to update ali-bot on the mac machines? Could you confirm?

fixed.

sawenzel:
I found 
```
Package: libyaml-cpp-dev
Architecture: amd64
Version: 0.5.2-4ubuntu1
```

davidrohr:
ok, can you try to add a version check to the yaml-cpp.sh alidist recipe for 0.62? Then it should bump to 0.62 and hopefully compile

sawenzel:
This works for me if building own yaml-cpp enforced.

davidrohr:
@sawenzel : Shouldn't we also add a version check for 0.6.2 in the prefer_system_check of yaml-cpp.sh?

awegrzyn:
With `o2-dataflow` default FairRoot fails with:
```
[ 33%] Generating G__BaseDict.cxx, G__BaseDict_rdict.pcm, ../lib/libBase.rootmap
input_line_12:19:10: fatal error: 'TG4VUserFastSimulation.h' file not found
```

sawenzel:
@awegrzyn : It seems FairRoot that now depends explicitly on Geant4.

sawenzel:
Seems like this will be fixed with https://github.com/FairRootGroup/FairRoot/pull/912

davidrohr:
Updated to bump to FairRoot with the fix and fixed version check of yaml-cpp thx to @ktf

awegrzyn:
Ok, `o2-dataflow` compiles fine atm.

Btw. do we plan to disable building boost on macOS or pointing to our fork patched by Peter? Otherwise:
```
/Users/awegrzyn/alice/sw/osx_x86-64/boost/v1.70.0-1/include/boost/process/detail/posix/wait_for_exit.hpp:60:7: error: expected unqualified-id
     ::sigemptyset(&sigset);
```

davidrohr:
Also control fails right now with new FairMQ.

@awegrzyn : I actually built this PR with CMake 3.15.1 and didn't encounter any issue?
So what patch do we need, or perhaps they fixed it already?

@sawenzel : I'll be in vacation till end of next week, perhaps you could follow this up in the meantime.

awegrzyn:
@davidrohr Indeed, I've just checked once again and there are no issues with cmake 3.15.*

Suddenly I got such error when compiling O2:
```
[ 62%] Generating G__O2Generators.cxx, ../stage/lib/G__O2Generators_rdict.pcm, ../stage/lib/libO2Generators.rootmap
ERROR: please fix the warnings below about unused class rule
[ 63%] Built target O2lib-TRDBase
Warning: Unused class rule: o2::eventgen::GeneratorFactory
Warning: Unused class rule: o2::eventgen::Pythia8Generator
```

sawenzel:
@awegrzyn : For info,  this error has nothing to do with this PR (see https://alice.its.cern.ch/jira/browse/O2-823). As a side-thought, I am starting to think that maintaining the separate dataflow defaults is causing more pain than good (especially now that we have pre-build binaries). Is there any chance we could simplify things and just use the same software stack (defaults-o2) for everyone? (Discussion probably to be continued in WP3)

sawenzel:
I just checked once more: This compiles fine (linux) with defaults-o2.sh but for some reason for "defaults-o2-dataflow" I am getting an error: cannot find -lG4clhep. Also to me it appears that for some reason cmake is picking up an existing G4 installation although this shouldn't be the case with the dataflow default.

ktf:
It also looks like bumping the Xcode on the server introduces new errors in `Control-OCCPlugin`:

```
In file included from /build/ci_checks/alidist_checker_macos_o2_ci_0/sw/SOURCES/Control-OCCPlugin/v0.10.0/v0.10.0/occ/occlib/OccInstance.cxx:29:
/build/ci_checks/alidist_checker_macos_o2_ci_0/sw/SOURCES/Control-OCCPlugin/v0.10.0/v0.10.0/occ/occlib/RuntimeControlledObject.h:111:43: warning: '/*' within block comment [-Wcomment]
     *                   "address": "tcp://*:5555",
                                          ^
/build/ci_checks/alidist_checker_macos_o2_ci_0/sw/SOURCES/Control-OCCPlugin/v0.10.0/v0.10.0/occ/occlib/RuntimeControlledObject.h:151:130: warning: '/*' within block comment [-Wcomment]
     *                   "address": "tcp://target.hostname.cern.ch:5555",  // if "method" is "bind", "address" can be e.g. "tcp://*:5555"
                                                                                                                                 ^
In file included from /build/ci_checks/alidist_checker_macos_o2_ci_0/sw/SOURCES/Control-OCCPlugin/v0.10.0/v0.10.0/occ/occlib/OccServer.cxx:36:
/build/ci_checks/alidist_checker_macos_o2_ci_0/sw/SOURCES/Control-OCCPlugin/v0.10.0/v0.10.0/occ/occlib/RuntimeControlledObject.h:111:43: warning: '/*' within block comment [-Wcomment]
     *                   "address": "tcp://*:5555",
                                          ^
/build/ci_checks/alidist_checker_macos_o2_ci_0/sw/SOURCES/Control-OCCPlugin/v0.10.0/v0.10.0/occ/occlib/RuntimeControlledObject.h:151:130: warning: '/*' within block comment [-Wcomment]
     *                   "address": "tcp://target.hostname.cern.ch:5555",  // if "method" is "bind", "address" can be e.g. "tcp://*:5555"
                                                                                                                                 ^
/build/ci_checks/alidist_checker_macos_o2_ci_0/sw/SOURCES/Control-OCCPlugin/v0.10.0/v0.10.0/occ/occlib/OccServer.cxx:234:9: warning: format specifies type 'unsigned long' but the argument has type 'RunNumber' (aka 'unsigned long long') [-Wformat]
        newRunNumber);
        ^~~~~~~~~~~~
[ 63%] Building CXX object CMakeFiles/Occ.dir/occlib/RuntimeControlledObject.cxx.o
/build/ci_checks/alidist_checker_macos_o2_ci_0/sw/SOURCES/Control-OCCPlugin/v0.10.0/v0.10.0/occ/plugin/OccPluginServer.cxx:361:46: error: no type named 'InvalidStateError' in 'fair::mq::PluginServices'
            catch (fair::mq::PluginServices::InvalidStateError &e) {
                   ~~~~~~~~~~~~~~~~~~~~~~~~~~^
/build/ci_checks/alidist_checker_macos_o2_ci_0/sw/SOURCES/Control-OCCPlugin/v0.10.0/v0.10.0/occ/plugin/OccPluginServer.cxx:384:32: warning: '&&' within '||' [-Wlogical-op-parentheses]
            !newStates.empty() && isIntermediateState(newStates.back())) {
            ~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/build/ci_checks/alidist_checker_macos_o2_ci_0/sw/SOURCES/Control-OCCPlugin/v0.10.0/v0.10.0/occ/plugin/OccPluginServer.cxx:384:32: note: place parentheses around the '&&' expression to silence this warning
            !newStates.empty() && isIntermediateState(newStates.back())) {
                               ^
            (                                                          )
```

@teo

ktf:
Since @teo is away, @awegrzyn could you take a look?

ktf:
Both @teo and @awegrzyn are away. AliceO2Group/Control#61 should fix the warnings.

sawenzel:
For me this now builds fine also with dataflow.

sawenzel:
@ktf: The warnings in Control-Plugin have nothing to do with this PR, don't they?

ktf:
Well, they break the compilation so we do not know if we actually get to the end. Looking at the errors (some of which are actual errors, not warnings turned into errors), I am not completely sure how it's possible that the new `Control-OCCPlugin` passed the tests.

ktf:
#1808 to fix the warnings / erorrs in Control.

ktf:
For the record, there was no need to force push. Fixing the conflict with the GUI would have been enough.

davidrohr:
Sorry, race conflict.
Btw, didn't know this can be fixed in the gui

ktf:
Don't you get a "resolve" button when you have a conflict? In this case it's even your PR, so it should have been there regardless of your permissions.

davidrohr:
Actually, no idea... Usually just fix it in a console.
Next time I'll have a look. Thx for the hint!

davidrohr:
fix for readout in AliceO2Group/Readout#67
Built O2Suite locally with that fix and it works, need to wait till readout is bumped, then I'll rebase this PR.

Barthelemy:
Merged in Readout

awegrzyn:
> Merged in Readout

I guess we need to bump it in alidist

davidrohr:
> I guess we need to bump it in alidist

@awegrzyn : Yes, we need to bump. You can either bump yourself, or just create a new tag for readout, and I'll include the readout bump in here. That way, we run the CI only once.

Barthelemy:
Ok so I have released Readout v1.0.6.

davidrohr:
@ktf : Now FairRoot fails again, I didn't change anything with it. Any idea perhaps?

ktf:
No... :-/

davidrohr:
ok, I can reproduce it locally.
Very funny. So why did it work before?

Investigating again....

davidrohr:
@ktf : How can travis be stuck? I never saw this before. Usually it takes only 1 minute for alidist.
Anyway, the rest is finally green!

ktf:
It actually finished successfully but apparently did not update the status correctly. I retriggered a build and it finished. Merging this before something else fails... ;-)

davidrohr:
> Merging this before something else fails... ;-)

Thx, that was my point! :|

sawenzel:
This PR is referencing v1.70.0 of https://github.com/alisw/boost.git which apparently does not exist. I am wondering how the build of this actually passed ... probably because boost was installed as system package on the CI machine?

davidrohr:
Puh, that is pretty weird. It did show up all the boost 1.70 issues. Does the CI machine have boost 1.70 as system package?
So should we switch to https://github.com/boostorg/boost/tree/boost-1.70.0 or add 1.70 to our fork? Personally, I prefer to just use the upstream versions.

ktf:
Also, the repo is there, AFAICT. Will explain more later.

--
Ciao,
Giulio

ktf:
We cannot use the official one.

--
Ciao,
Giulio

davidrohr:
https://github.com/alisw/boost is there, but does not have the 1.70 tag as far as I see.

davidrohr:
I created a PR to bump boost in our fork here: boost#3

sawenzel:
There are potentially more problems to come up. Currently I can't build on MacOS after this PR was merged. I tested the following combinations
a) boost 1.70 from brew ---> fails in FairRoot cmake stage
b) no system boost but boost 1.70 from https://github.com/alisw/boost/pull/2 or https://github.com/alisw/boost/pull/3 ---> fails in compiling Ppconsul (Ppconsul does no longer find the headers).

davidrohr:
Damn, I only tried the new boost on Linux.
But how did the MacOS CI passed in the first place?

What is the failure in FairRoot cmake stage?

pzhristov:
I have brew boost 1.70.0 and FairRoot dev e6cd7da7d60d0983cd0c1014a611376a1e8fe0c4 фром 06/07, it compiles for me.

ktf:
I think the macOS CIs test with some of the externals from home-brew, because that's what we tell people to use (reasonably so, I would say).

davidrohr:
But then why does it fail with brew-boost for @sawenzel ?

ktf:
No idea, the build nodes use:

```
→ mpssh -l alibuildmac "/usr/local/bin/brew info boost | grep boost:"
MPSSH - Mass Parallel Ssh Ver.1.3.3
(c)2005-2013 Nikolay Denev <ndenev@gmail.com>

  [*] read (4) hosts from the list
  [*] executing "/usr/local/bin/brew info boost | grep boost:" as user "ktf"
  [*] only on hosts labeled "alibuildmac"
  [*] spawning 4 parallel ssh sessions

alibuildmac01.cern.ch -> boost: stable 1.70.0 (bottled), HEAD
alibuildmac02.cern.ch -> boost: stable 1.70.0 (bottled), HEAD
alibuildmac04.cern.ch -> boost: stable 1.70.0 (bottled), HEAD
alibuildmac03.cern.ch -> boost: stable 1.70.0 (bottled), HEAD

  Done. 4 hosts processed.
```

@sawenzel what is the error you get?

sawenzel:
Sorry. I had FairRoot checked out as dev package and forgot to adjust the commit. It builds now with brew boost.

That still leaves us to fix the build with non-system boost or to make boost a necessary system  dependency on MacOS.

ktf:
Using the system boost is actually the correct thing to do for macs, since that's what people following the instructions will end-up with. Also I would keep it an optional behaviour, since forcing it from the system will complicate the update process (suddenly you need to do brew update every now and then, with all that implies on your other system tools).

I think the only improvement possible is to have yet another build which checks with `--no-system`, at least for alidist. @MichaelLettrich do we have cores for that?

aphecetche:
+1 for not making the system boost a requirement (i.e. keeping both options available and maintained). In the past we've also been in the reverse situation (where boost from the system was _not_ working for our builds while "our boost" was), so I guess that might happen again in the future ?

davidrohr:
I would not add yet another CI. Let's just fix the boost build for MacOS anyway and the CI can stay as is, as that is what most people use? I wouldn't enforce system boost.

sawenzel:
Slightly disagree here: I think it is imperative to test compilation of the software stack using no system packages (for those where it is optional). This would for sure make the system more stable.

davidrohr:
ok, point taken. When we adapt the CI, we could make the minimal config run with no system packages to test everything. The question is: do we want this both for Linux and for MaxOS?

ktf:
@sawenzel we do that on linux. On macOS most of the users have system boost, and until recently we had only one underpowered builder, now we have two. We need to pick our fights.

@davidrohr on linux most users build without any system packages because we now have precompiled tarballs by default.

Maybe actually the best thing is to make sure we provide precompiled tarballs to everyone, macOS and Linux. Now that we have an S3 endpoint available that will most likely scale to our needs.

ktf:
Also, being able to have our own boost has proven to be handy in the past, so I would not force system one either. This is especially true on mac, because homebrew is very aggressive in changing to a new boost whenever available and we are not as fast adapting.

davidrohr:
@ktf : OK, then I'd suggest when we adapt the CI, we make the maximum configuration built without any system packages. In that case, we make sure that all packages are tested.
For me, testing MacOS without system packages would not be a priority right now, could be added later if needed.

Btw: How did the Linux CI succeed for this PR. The MaxOS CI user boost 1.70 from brew, but where did the Linux CI get it from? Does the builder have boost 1.70 as system package?

sawenzel:
Looking at the CI logs of one of the PRs today, https://github.com/AliceO2Group/AliceO2/pull/2380, it seems that the linux CI is still building against boost 1.69.0:
```
  dependencies/O2Dependencies.cmake:45 (find_package)
  dependencies/CMakeLists.txt:11 (include)
  CMakeLists.txt:45 (include)


-- Boost version: 1.69.0
-- Found the following Boost libraries:
--   container
--   thread
--   system
--   timer
--   program_options
--   random
--   filesystem
```

So probably it is building against an already installed version from previous builds.

davidrohr:
I am not sure if this might be misleading. I remember this could be something like >= 1.69?

Also, I though the CI ran with boost 1.70, since I had to fix 1.70 related issues.
But perhaps that was the MacOS CI.

@ktf @MichaelLettrich : Could you just check on the builder what is the actual boost version, which is used during the CI?

sawenzel:
On MacOS, I am getting the expected:
 ```
 dependencies/CMakeLists.txt:11 (include)
  CMakeLists.txt:45 (include)


-- Boost version: 1.70.0
-- Found the following Boost libraries:
--   container
```
so I strongly believe that indeed 1.69.0 is used on Linux.

**Comment by davidrohr** [comment]
Actually, do you find where this message is coming from:
```
-- Boost version: 1.69.0
-- Found the following Boost libraries:
```
I don't find that in any CMake file on my system

If it is indeed 1.69, then to me this is a severe bug in our build system

**Comment by sawenzel** [comment]
This is in the log files of alibuild for O2. Should be reported by find_package(boost).
To me, this also looks like a bug. Likely alibuild cannot check out the required tag v1.70.0 because it does not exist. Then instead of failing, it probably uses the last successful build of boost. 

On the other hand, starting from a completely fresh build, the compilation fails.

**Comment by davidrohr** [comment]
I meant I did not find from where inside `find_package(boost)` these messages come.

OK, I'll try locally to build boost 1.69 with alibuild and then 1.70. If you are right, it should build 1.69 correctly, and then do not fail but assume 1.69 as 1.70. That would be a bug in alibuild and should be fixed.

**Comment by davidrohr** [comment]
Aehm, actually, to make it even more weird. We bumped from 1.68 to 1.70, never had 1.69? So how can the build report 1.69?

**Comment by davidrohr** [comment]
@ktf @sawenzel @MichaelLettrich : So what happens is:
If alibuild tries to build a package with a version whose tag does not exist in the git repository, it clones the repository, tries to check out the tag, and then it fails.
If you try to build the same package/version again, it finds the package already checked out in the SOURCES directory in the master version (which is the wrong version because checking out the tag failed). However, alibuild, it does not double-check that the tag is correct, and just starts building.

I.e., the first time you build with an empty build dir, it fails. If you just retry the build, it suceeds with the current master version, which currently is 1.69.

In alicache, we have a boost 1.70 package which is in fact 1.69. Probably, this was created just in the way described above. So we should first fix this bug in alibuild, and then remove the bogus package in the store. I am having a look at alibuild, but if I don't find it fast, I'll leave it to @ktf.

In parallel, we should try building the stack with the correct boost 1.70, and fix potential issues. I am right now doing this in the alidock container for linux.

**Comment by davidrohr** [comment]
Fix for alibuild: alisw/alibuild#578

**Comment by davidrohr** [comment]
Also for me, both boost PRs from @dennisklein and from myself break Ppconsol in the Linux alidock container. I am wondering how it built with my system boost 1.70.0 in the first place.

**Comment by sawenzel** [comment]
Thanks @davidrohr !!

So a quick fix is potentially to use 1.69.00 as the official version (allowing the brew version 1.70.00 on MacOS). And meanwhile we can fix the 1.70.0 problems.

**Comment by davidrohr** [comment]
Fix in #1817 (see there for the full story...)
I am just building the full O2 setup locally, looks good so far.
@sawenzel : Could you check for MacOS?

In order to fix it properly, we have to merge:
#1817 
alisw/alibuild#578
alisw/boost#2
I can't do any of this, so it will be up to @ktf I guess.

For the next boost bump, we should clean up the mess with the boost cmake config files as described in the other ticked.

**Comment by ktf** [comment]
#1817 and alisw/boost#2 are now in. I assume alisw/alibuild#578 is not really required though? It seems to have some legit issues in one of the tests.

**Comment by ktf** [comment]
I also added the boost v1.70.0 tag.

**Comment by davidrohr** [comment]
Ok, thx, I'll double check the alibuild pr tomorrow.
Kind Regards
David Rohr

**Comment by davidrohr** [comment]
@ktf : alisw/alibuild#578 is not really required ti fix boost for now, but in general it is required since it is the source of this mess we had with the CI not working and alibuild building wrong package version without us knowing.
I think the travis errors for alibuild are unrelated, and I double checked that the PR works for me.
Could you have a closer look.

btw: i checked the newest build in alistore, and it indeed now contains the correct boost 1.70.

**Comment by ktf** [comment]
@davidrohr: ok, but today it's a bit complicated for me. I cannot promise. Do I understand correctly that the issue is only if there is a missing tag?

**Comment by davidrohr** [comment]
@ktf : Yes, only for missing tag, so for now it is OK.
No worry, I don't think this needs fixing on the weekend.

**Comment by dstocco** [comment]
Hi all,
sorry but I was a little bit lost here. What is the current method to install o2 on Mac?
I tried to compile after this PR was merged: boost compiles but DDS does not with the usual error message:
```
DEBUG:alo-o2:DDS:o2-master: In file included from /Users/stocco/Documents/projects/alice/software/sw/SOURCES/DDS/2.4/2.4/MiscCommon/dds_ncf/src/ncf.cpp:12:
DEBUG:alo-o2:DDS:o2-master: In file included from /Users/stocco/Documents/projects/alice/software/sw/SOURCES/DDS/2.4/2.4/MiscCommon/MiscUtils.h:14:
DEBUG:alo-o2:DDS:o2-master: In file included from /Users/stocco/Documents/projects/alice/software/sw/osx_x86-64/boost/v1.70.0-1/include/boost/process.hpp:24:
DEBUG:alo-o2:DDS:o2-master: In file included from /Users/stocco/Documents/projects/alice/software/sw/osx_x86-64/boost/v1.70.0-1/include/boost/process/async_system.hpp:22:
DEBUG:alo-o2:DDS:o2-master: In file included from /Users/stocco/Documents/projects/alice/software/sw/osx_x86-64/boost/v1.70.0-1/include/boost/process/child.hpp:21:
DEBUG:alo-o2:DDS:o2-master: In file included from /Users/stocco/Documents/projects/alice/software/sw/osx_x86-64/boost/v1.70.0-1/include/boost/process/detail/child_decl.hpp:30:
DEBUG:alo-o2:DDS:o2-master: /Users/stocco/Documents/projects/alice/software/sw/osx_x86-64/boost/v1.70.0-1/include/boost/process/detail/posix/wait_for_exit.hpp:60:7: error: expected unqualified-id
DEBUG:alo-o2:DDS:o2-master:     ::sigemptyset(&sigset);
DEBUG:alo-o2:DDS:o2-master:       ^
DEBUG:alo-o2:DDS:o2-master: /Library/Developer/CommandLineTools/SDKs/MacOSX10.14.sdk/usr/include/signal.h:125:26: note: expanded from macro 'sigemptyset'
DEBUG:alo-o2:DDS:o2-master: #define sigemptyset(set)        (*(set) = 0, 0)
```
I also tried to install the system boost with brew (that by the way is version 1.71)...but for some reasons aliBuild still tries to install boost 1.70...and then I think it mixes up the 1.70 and 1.71 versions, resulting in another error.

Thanks in advance,
best regards,
Diego

**Comment by davidrohr** [comment]
@dstocco : Could you try to delete $HOME/alice/sw/SOURCES/boost* before building?

**Comment by dstocco** [comment]
Hi @davidrohr ,
I did it. Actually, not only I removed SOURCES, but also:
```
osx_x86-64/boost
SOURCES/boost
MIRROR/boost
SPECS/osx_x86-64/boost
TARS/osx_x86-64/boost
```
And then recompiled after installing system boost.
The first thing that aliBuild did was to download again boost in MIRROR and then put it back in SOURCES....

davidrohr:
@dstocco : Did you perhaps update to CMake 3.15? There is a problem with that on MacOS.

dstocco:
@davidrohr : yes, unfortunately I did. I can try to downgrade and see if the issue is gone.
Thanks!

dstocco:
Hi @davidrohr ,
I uninstalled make from brew, so that alibuild uses its own (3.13.1)...but I get exactly the same error as before.
Anyways, according to a [comment in this same thread:](https://github.com/alisw/alidist/pull/1776#issuecomment-524196002), it looks like DDS was compiling with make 3.15.1....

davidrohr:
Yes, that was another problem on Mac but it should be fixed.
Are you sure you have current alidist, and no development packages for boost / FairRoot / DDS?

dstocco:
Yes, I'm sure.
I'm at alidist master (`66794448384583cac35401863deabde3439123aa`)
and my only development packages are O2 and alo-o2....
Anyways, if I'm the only one having this issue among Mac user, I can try to scratch the sw and start from the beginning....
The only question is: should I use system boost or boost coming from alibuild?

davidrohr:
I am afraid I am not a mac user myself and cannot tell you much more.
I'd rather use the boost from alidist if in doubt.
If it doesn't work after a rebuild from scratch with empty sw, please report again, and provide the build logs for boost and DDS.

dstocco:
Hi @davidrohr ,
I started from scratch and the compilation is ok now. Sorry for the noise and thanks again for the suggestions.

PR #1782, opened 2019-08-16
 Adding KFParticle as required package

pzhristov:


ktf:
Merging this now so that we get it in the nightly before the weekend.

PR #1785, opened 2019-08-19
Allow building with Ninja

aphecetche:


Barthelemy:
Nice !
Do I understand correctly that if I install ninja then it should automatically use it ?

aphecetche:
Yes. At least that's how I'm using it for AliceO2 (this PR is basically a cut and paste from the o2.sh recipe, btw).

PR #1786, opened 2019-08-19
Bump XRootD version to v4.10.0 for JALiEn and disable xrdpy bindings

Atlantic777:
The XRootD can be upgraded to v4.10.0 on all platforms when we disable Python bindings. The bindings will be re-enabled after we merge the fix into XRootD upstream. This commit affects JAliEn builds only.


Atlantic777:
@pzhristov this is stuck. We force merge it?

PR #1788, opened 2019-08-19
Simplify o2.sh

aphecetche:
Following the work (initiated in https://github.com/AliceO2Group/AliceO2/pull/2313)  to solve https://alice.its.cern.ch/jira/browse/ALIROOT-8326, this PR simplifies the o2.sh recipe.
 
A test on my Mac looks fine but using the CI to confirm that on a wider scope...
I'm meanwhile checking the original problem is indeed gone (it should, as with this change o2 will no longer use the O2RecipeAdapter, but to be confirmed...)

### DIFF
--- a/o2.sh
+++ b/o2.sh
@@ -134,44 +134,10 @@ fi
 unset DYLD_LIBRARY_PATH
 cmake $SOURCEDIR -DCMAKE_INSTALL_PREFIX=$INSTALLROOT                                        \
       ${CMAKE_GENERATOR:+-G "$CMAKE_GENERATOR"}                                             \
-      -DCMAKE_MODULE_PATH="$SOURCEDIR/cmake/modules;$FAIRROOT_ROOT/share/fairbase/cmake/modules;$FAIRROOT_ROOT/share/fairbase/cmake/modules_old"  \
-      -DFairRoot_DIR=$FAIRROOT_ROOT                                                         \
-      -DALICEO2_MODULAR_BUILD=ON                                                            \
-      -DROOTSYS=$ROOTSYS                                                                    \
-      ${PYTHIA6_ROOT:+-DPythia6_LIBRARY_DIR=$PYTHIA6_ROOT/lib}                              \
-      ${GEANT3_ROOT:+-DGeant3_DIR=$GEANT3_ROOT}                                             \
-      ${GEANT4_ROOT:+-DGeant4_DIR=$GEANT4_ROOT}                                             \
-      ${VGM_ROOT:+-DVGM_DIR=$VGM_ROOT}                                                      \
-      ${GEANT4_VMC_ROOT:+-DGEANT4_VMC_DIR=$GEANT4_VMC_ROOT}                                 \
-      -DFAIRROOTPATH=$FAIRROOT_ROOT                                                         \
-      ${BOOST_ROOT:+-DBOOST_ROOT=$BOOST_ROOT}                                               \
-      ${DDS_ROOT:+-DDDS_PATH=$DDS_ROOT}                                                     \
-      -DZMQ_DIR=$ZEROMQ_ROOT                                                                \
-      -DZMQ_INCLUDE_DIR=$ZEROMQ_ROOT/include                                                \
-      ${ALIROOT_VERSION:+-DALIROOT=$ALIROOT_ROOT}                                           \
-      ${PROTOBUF_ROOT:+-DProtobuf_LIBRARY=$PROTOBUF_ROOT/lib/libprotobuf.$SONAME}           \
-      ${PROTOBUF_ROOT:+-DProtobuf_LITE_LIBRARY=$PROTOBUF_ROOT/lib/libprotobuf-lite.$SONAME} \
-      ${PROTOBUF_ROOT:+-DProtobuf_PROTOC_LIBRARY=$PROTOBUF_ROOT/lib/libprotoc.$SONAME}      \
-      ${PROTOBUF_ROOT:+-DProtobuf_INCLUDE_DIR=$PROTOBUF_ROOT/include}                       \
-      ${PROTOBUF_ROOT:+-DProtobuf_PROTOC_EXECUTABLE=$PROTOBUF_ROOT/bin/protoc}              \
-      ${GSL_ROOT:+-DGSL_DIR=$GSL_ROOT}                                                      \
-      ${PYTHIA_ROOT:+-DPYTHIA8_INCLUDE_DIR=$PYTHIA_ROOT/include}                            \
-      ${HEPMC3_ROOT:+-DHEPMC3_DIR=$HEPMC3_ROOT}                                             \
       ${CMAKE_BUILD_TYPE:+-DCMAKE_BUILD_TYPE=$CMAKE_BUILD_TYPE}                             \
       ${ALIBUILD_O2_TESTS:+-DENABLE_CASSERT=ON}                                             \
-      -DMS_GSL_INCLUDE_DIR=$MS_GSL_ROOT/include                                             \
       -DCMAKE_EXPORT_COMPILE_COMMANDS=ON                                                    \
-      ${CXXSTD:+-DCMAKE_CXX_STANDARD=$CXXSTD}                                               \
-      ${MONITORING_VERSION:+-DMonitoring_ROOT=$MONITORING_ROOT}                             \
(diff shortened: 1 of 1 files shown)


ktf:
Anyone can think of something which might become painful with this change?

o2.sh
```diff
@@ -134,44 +134,10 @@ fi
...
-      ${PROTOBUF_ROOT:+-DProtobuf_LITE_LIBRARY=$PROTOBUF_ROOT/lib/libprotobuf-lite.$SONAME} \
-      ${PROTOBUF_ROOT:+-DProtobuf_PROTOC_LIBRARY=$PROTOBUF_ROOT/lib/libprotoc.$SONAME}      \
-      ${PROTOBUF_ROOT:+-DProtobuf_INCLUDE_DIR=$PROTOBUF_ROOT/include}                       \
-      ${PROTOBUF_ROOT:+-DProtobuf_PROTOC_EXECUTABLE=$PROTOBUF_ROOT/bin/protoc}              \
-      ${GSL_ROOT:+-DGSL_DIR=$GSL_ROOT}                                                      \
-      ${PYTHIA_ROOT:+-DPYTHIA8_INCLUDE_DIR=$PYTHIA_ROOT/include}                            \
-      ${HEPMC3_ROOT:+-DHEPMC3_DIR=$HEPMC3_ROOT}                                             \
       ${CMAKE_BUILD_TYPE:+-DCMAKE_BUILD_TYPE=$CMAKE_BUILD_TYPE}                             \
       ${ALIBUILD_O2_TESTS:+-DENABLE_CASSERT=ON}                                             \
-      -DMS_GSL_INCLUDE_DIR=$MS_GSL_ROOT/include                                             \
```
ktf: we made this name up, IIRC, does it really find it?

aphecetche: yes, using the right combination of find_package call `(find_package(ms_gsl MODULE))` and corresponding [find module](https://github.com/AliceO2Group/AliceO2/blob/dev/dependencies/Findms_gsl.cmake)

(but now that we're using MS_GSL_ROOT, we might as well consider ditching it...)

ktf:
One of the advantages of the current setup is that you see immediately in the logs what and where we pick up various externals and we can cut & paste the string found in the logs, even without having the environment set, and it will work. I am therefore still in favour of having:

${SOMETHING_ROOT:+-D$SOMETHING_ROOT}

in the command line. This would already be an improvement since right now for many thing we have a few `-D` while this would result into one.

sawenzel:
I have some doubts and thoughts as well, as much as this is appealing:

Firs of all, I didn't mind the long recipes since they tell me exactly where things come from, understand what we are using, and allow me to possibly take some package from some other place.
I don't know the new system but I guess that now this magic happens behind the scenes and needs the presence of some (**which**?) environment variables.

In the past we were using the recipe + log to generate the CLion configuration. Is this still possible or got even easier?? (Do I now need to be in some environment when I launch CLion??)

My biggest question is if I can still optionally say `-DPackage_DIR=...` to specify an alternative path where to take a package from?

If everything is still possible, I have nothing against going this step.

aphecetche:
OK, two objections are more than enough, so I'll take a step back and go with the ${SOMETHING_ROOT:+-D$SOMETHING_ROOT} proposal (will have to rename the PR then ;-) )

Now, just to clear a few things.

1. @sawenzel the changes here were only meant to clean a bit the recipe, as within the recipe we are sure to be within alibuild realm, i.e. with a bunch of XXX_ROOT environment variables defined. Those XXX_ROOOT environment variables (or their -D counterpart, the so called cmake variables) are the first place where find_package(XXX ...) looks for (number 1 in the https://cmake.org/cmake/help/v3.13/command/find_package.html#search-procedure), but that search procedure goes further if need be. For instance XXX_DIR is number 3. But I must be honest here and say I've not thoroughly check that the few custom FindXXX.cmake we have left actually work with all the steps in that procedure. I would guess they are (as they are themselves using other find methods which follow that procedure), but I've not 100% checked. So _in principle_, yes you can still use -DXXX_DIR if you'd like to. 
In any case if the new setup by chance kills your current (CLion) workflow, please tell me and I'll try to fix that asap.

2. There is no (longer) any magic in the search procedure (unless you use -DALIBUILD_BASEDIR which I'd like to keep for a bit for my own consumption) (or unless you count find_package search procedure as magic, but that's not on me ;-) )

3. Personally my dream setup would be to be able to do `cmake location_of_source_dir -DCMAKE_PREFIX_PATH=someinstalldir` without _any_ prior environment and get a working build out of that configuration stage. But that's currently out of reach for a number of reasons. So the next best thing is something like : 

```
source somewhere/sw/platform/O2/latest/etc/profile.d/init.sh
cmake location_of_source_dir -DCMAKE_GENERATOR=Ninja -DCMAKE_BUILD_TYPE=Debug
```

where the first line sets all the XXX_ROOT environment variables.

sawenzel:
Fine. I wouldn't count my comment as a real objection. As far as I understand previous features are still available.

Barthelemy:
@aphecetche I concur with Sandro about Clion. How do you handle that ? Till now we could copy paste the resulting cmake config line.

aphecetche:
@Barthelemy in CLion (which I no longer use, so kind of forgot how the setup was) how do you handle the environment ?

sawenzel:
@Barthelemy : As far as I understand, you need to build once with aliBuild. Then you source  `somewhere/sw/platform/O2/latest/etc/profile.d/init.sh` and then you launch CLion within that environment (if that is possible). The CMake of O2 should then automatically be usable without further configuration. It could actually be a simplification.

ihrivnac:
@Barthelemy , @sawenzel ,
Could you try it out with your Clion? When the environment variables are set, the CMake does not need these extra CMake variables.

Barthelemy:
@aphecetche @sawenzel  The environment can be set via variables in the parameters of the build. However, I am not 100% sure how to run `somewhere/sw/platform/O2/latest/etc/profile.d/init.sh` before launching clion on Mac, i.e. should I launch the application from the command line ? 

@ihrivnac I will try

ktf:
If your editor supports it, now you can always use `direnv` to load the environment, but still, the logs would show a line which cannot be simply cut & pasted anymore.

Barthelemy:
Unfortunately direnv is not usable in Clion and it is not possible to call a script before running cmake (only before building). 

There is a question [here](https://intellij-support.jetbrains.com/hc/en-us/community/posts/360003487420-Source-script-before-building?input_string=Setup%20environment%20with%20a%20script%20before%20build) from someone at CERN. It points to using "external tools" which I have tried. Attempts and problems so far : 
- (...)/alice/sw/osx_x86-64/O2/latest/etc/profile.d/init.sh is not runnable
- sourcing init.sh fails because WORK_DIR needs to be set
- creating a wrapper script to be called from clion fails as well as the variables are not available in the cmake environment (apparently). 

I am going to try something different. Sorry for the noise.

dennisklein:
I have no preference whether to specify the dependency search hints explicitly or not in the recipe. But I support the general goal of @aphecetche to support the [CMake standard search procedure](https://cmake.org/cmake/help/v3.13/command/find_package.html#search-procedure) by default.

*FYI*: On top of the standard CMake search procedure, in the FairLogger, FairMQ, and asiofi repos we support `<UPPERCASE_PKGNAME>_ROOT` env (read-only) and CMake variables for convenient aliBuild integration (if specified they get prepended to the `CMAKE_PREFIX_PATH` temporarily for a particular `find_package` invocation, so they are effective at search procedure precedence level 2 (or 1, if `<PkgName>` equal to `<UPPERCASE_PKGNAME>` in newer CMake versions)). FairRoot does not yet fully support this convention for all dependencies, but it is on the roadmap.

aphecetche:
@Barthelemy as I said, I'm no longer a CLion user so cannot test it myself, but at least I've found back the [small script](https://github.com/aphecetche/scripts/blob/master/clion/alien2clion.sh) I was using at the time to get the env. from alienv in a form suitable for cut and paste in CLion : 

```shell
alienv printenv O2/latest | tr ";" "\n" | grep -v export | sed -e 's/[[:space:]]*$//'
```

Have no idea if that's still usable though...

Barthelemy:
@aphecetche very nice, a lot better than what I was actually trying to do. It works. I have documented it here : https://github.com/AliceO2Group/AliceO2Group.github.io/pull/32

ktf:
Since everyone is pretty much ok with this let's go ahead and in case we want the explicit setup, we add it back when it makes sense.

PR #1793, opened 2019-08-22
Add recipe for ALF

kostorr:


Barthelemy:
Nothing depends on ALF, correct ? 
The recipe seems ok to me.

kostorr:
@Barthelemy yes that is correct. Thanks

ktf:
Next time please keep the bump and the the introduction of a new recipe separate, please.

Also, could you add alf to the CODEOWNERS file in the FLP section?

PR #1796, opened 2019-08-23
Add a dependency on glfw for the RPM of QC

Barthelemy:
Needed in case the build machine has glfw. The library of QC will then expect to link against glfw and
if it is not present in the machine where we install nothing works.


Barthelemy:
@teo

ktf:
Ping.

PR #1798, opened 2019-08-26
o2: Use DebugGUI external package

ktf:
* Add DebugGUI as an O2 dependency
* Move to latest version of DebugGUI
* Allow ninja usage for DebugGUI


ktf:
@davidrohr this will actually enable the new DebugGUI external package.

davidrohr:
This PR is OK as is for me, need some changes on the O2 side, and in the external DebugGUI itself, as commented in AliceO2#2340.

ktf:
@davidrohr I think I managed to make the compilation optional in case OpenGL is not there on linux. Could you check in your environment? In the builder it works for me.

davidrohr:
Sorry, missed that. I'll rebuild with this PR tomorrow.

davidrohr:
@ktf : I rebuilt with this patch.
- It auto-detects system GLFW and uses that (not rebuilding it).
- It builds DebugGUI and O2 correctly.
- The event display in O2 is working.
So all fine from my side.

ktf:
Ok, perfect. I just need to merge https://github.com/AliceO2Group/AliceO2/pull/2511 to cleanup the headless macOS issue and then I can merge this.

davidrohr:
@ktf : thx, finally :| Running a test now to see if it still works for me.

ktf:
yeah, sorry. it turns out the issue was really somewhere completely different. anyways let me know if you have any regression.

PR #1801, opened 2019-08-26
Export the relevant variables

mpuccio:
Ciao @ktf ,
thanks to @fcatalan92 I realised that there is a bug in the interplay between the AliPhysics recipe and the treelite one, this should be fixing the issue.
An alternative fix would be removing the check on the `TREELITE_VERSION` env variable in the aliphysics recipe, tell me if you prefer that solution I don't have a strong opinion.

Cheers,
Max

### DIFF
--- a/treelite.sh
+++ b/treelite.sh
@@ -37,6 +37,9 @@ module-whatis "ALICE Modulefile for $PKGNAME $PKGVERSION-@@PKGREVISION@$PKGHASH@
 module load BASE/1.0
 # Our environment
 set osname [uname sysname]
+setenv TREELITE_RELEASE \$version
+setenv TREELITE_VERSION $PKGVERSION
+setenv TREELITE_ROOT \$::env(BASEDIR)/$PKGNAME/\$::env(TREELITE_RELEASE)
 prepend-path PATH \$::env(BASEDIR)/$PKGNAME/\$version/bin
 prepend-path ROOT_INCLUDE_PATH \$::env(BASEDIR)/$PKGNAME/\$version/include
 prepend-path LD_LIBRARY_PATH \$::env(BASEDIR)/$PKGNAME/\$version/lib


ktf:
@MichaelLettrich same here.

ktf:
I prefer if you remove the check for `TREELITE_VERSION`. In general we should not rely (in the runtime) on variables which are ALICE / aliBuild only, rather than expected by the external tool itself. E.g. `ROOTSYS` is fine, because root expects it, `$ROOT_ROOT` is not, because it's an aliBuild-ism.

MichaelLettrich:
@ktf should be solved.

ktf:
Yes, I would remove anything which is not expected by upstream.

mpuccio:
Ciao @ktf ,

you were right. Actually this PR is not needed. If I look at the modulefile of v5-09-50-01_ROOT6-1 on cvmfs I see

`module load BASE/1.0 AliRoot/v5-09-50_ROOT6-1 RooUnfold/V02-00-01-alice4-15 treelite/2a12742269-1`
(that is what I see locally, so treelite it is correctly loaded with AliPhysics)

however if I look at the latest tag (vAN-20190829_ROOT6-1) I just see:

`module load BASE/1.0 AliRoot/v5-09-49_ROOT6-1 RooUnfold/V02-00-01-alice4-11`

is it expected that the version of AliRoot is different? It looks like that for the daily the old version of alidist is used...

Cheers,
Max

ktf:
Looks like the dailies are not updated to use v5-09-50. I will fix this tomorrow.

mpuccio:
Close enough! You fixed for the ROOT5 builds, but for ROOT6 we still use the old version :P
Screenshot 2019-09-02 at 18 33 14

ktf:
This is now the case for both ROOT5 and ROOT6. Closing this.

PR #1803, opened 2019-08-27
Bump QCG version to v1.6.0

graduta:


awegrzyn:
@ktf

ktf:
macOS builders are out of disk space at the moment. @MichaelLettrich is looking into it. I will merge this since they are not likely to come back soonish.

PR #1804, opened 2019-08-27
Bump Rivet

pkarczma:


ktf:
Merging this. In any case we do not build in production with macOS for this.

PR #1814, opened 2019-09-05
Bump to CMake 3.16

davidrohr:
First test, should hopefully work now that FairRoot / boost / everything is bumped.

### DIFF
--- a/cmake.sh
+++ b/cmake.sh
@@ -1,14 +1,14 @@
 package: CMake
 version: "%(tag_basename)s"
-tag: "v3.13.1"
-source: https://github.com/Kitware/CMake
+tag: "v3.16.3-alice1"
+source: https://github.com/davidrohr/CMake
 build_requires:
  - "GCC-Toolchain:(?!osx)"
  - make
 prefer_system: .*
 prefer_system_check: |
   verge() { [[  "$1" = "`echo -e "$1\n$2" | sort -V | head -n1`" ]]; }
-  type cmake && verge 3.13.1 `cmake --version | sed -e 's/.* //' | cut -d. -f1,2,3`
+  type cmake && verge 3.16.3 `cmake --version | sed -e 's/.* //' | cut -d. -f1,2,3`
 ---
 #!/bin/bash -e
 


ktf:
I think we will need at least 6d8c35996c5ab2dc27bfda24c02b0624a10fa248 for FairRoot.

davidrohr:
Not sure, I built it locally with our current tag and it worked.
But fine, we can also boost FairRoot further by these 4 commits.

MichaelLettrich:
The problem with macOS is a known issue in the macOS build environment. We will fix it asap.

davidrohr:
@ktf @MichaelLettrich : Is there any progress with the MacOS issue with CMake 3.15?

ktf:
In principle the issues should be solved now. I triggered a rebuild by bumping the minor version.

davidrohr:
No, still same issue as before

davidrohr:
@ktf @MichaelLettrich : What is the plan for getting the new CMake run on MacOS? Now that CHEP is over, we might retry this.

davidrohr:
@MichaelLettrich : The new cmake is still failing on MacOS with the float header problems. Can this eventually be fixed?

MichaelLettrich:
There seems to be some bug either in a CMake File or in CMake itself because the Readout package fails to build with `-Werror`. We have the same Errors with the system CMake in macOS Catalina. See [O2-1074](https://alice.its.cern.ch/jira/browse/O2-1074)

MichaelLettrich:
#2031 should bring a fix and we should finally be able to merge this.

MichaelLettrich:
#2031 was merged. @davidrohr  can you bump this for retesting?

davidrohr:
@MichaelLettrich : It failed again with the same problem.
This error is in O2, so the workaround for readout cannot help.
We could implement the same workaround for O2.
But actually, I would prefer to try to find a proper fix for CMake instead of disabling CI features.
Is there already a bug report to CMake and are people working on this?

davidrohr:
I just updated this and added the workaround also to o2, just for the time being.
Still, I'd prefer a proper solution.

MichaelLettrich:
Can you please revert the workaround? There are 7 receipes which need that treatment. I will address them in its own PR. 

Also I have been able to find a [CMake bugreport](https://gitlab.kitware.com/cmake/cmake/issues/19845#note_689595) and use it to create a small reproducer that fails outside of our environment. So this is definitely a bug in CMake which we need to work around for now.

davidrohr:
ok, I'll wait for your fix and then rebase. I'll keep the CI running here, just to see whether it passes O2 now.

MichaelLettrich:
Hopefully after PR #2037 has been merged, we will finally be able to merge this.

davidrohr:
@ktf : The CI has finally passed. Can we merge this? Then we can close #2037 and revert #2031 (which will be no longer needed), and then I can also open a PR to revert the ROOT CMake workaround we have in place.
@MichaelLettrich : ping

ktf:
We should move to the alisw organisation, but apart from that this is fine for me.

davidrohr:
You'll have to do this ... :)
Anyhow, trying to get this merged into CMake upstream. It's vlearly a bug on their end. Don't want to keep many custom patches.

Kind Regards
David Rohr

davidrohr:
FYI @ktf @MichaelLettrich : The fix should make it into the CMake 3.17 release.

PR #1815, opened 2019-09-05
[WIP] Fix boost git url

davidrohr:
@sawenzel : Does this fix the issue (of #1776).
Not sure if we can trust the CI in this respect, should be investigated independently.

### DIFF
--- a/boost.sh
+++ b/boost.sh
@@ -1,7 +1,7 @@
 package: boost
 version: "%(tag_basename)s"
-tag: v1.70.0
-source: https://github.com/alisw/boost.git
+tag: boost-1.70.0
+source: https://github.com/boostorg/boost.git
 requires:
   - "GCC-Toolchain:(?!osx)"
   - "Python-modules:(?!osx)"


ktf:
Do not use the official one.

--
Ciao,
Giulio

davidrohr:
OK, I see, the problem with building from upstream is anyway that it uses git submodules and alibuild does not fetch the full build tree. I'll close this one, and try locally with a full build tree, and open a PR to our boost fork.

ktf:
Yes, submodules are one of the problems (which could be solved, but would require everyone move to a new version of aliBuild).

The other is that when we have huge git history we decided to either download the tarballs (like for CGAL, which has other "peculiarities") or to import the tarballs in our own github repository. 

This is not of so easy solution, because alibuild uses git history to facilitate development with features like development packages and incremental builds. We could use the GitHub API, but that complicates things (because users would need to get their own access token) and slow them down (because using a local git repo is 1000x faster (if not more) than using the REST API for most of the repositories).

ktf:
BTW, the instructions are here:

https://github.com/alisw/alidist#guidelines-for-handling-externals-sources

Issue #1614, opened 2019-05-02
Boost 1.70.0 and CMake lessons


dennisklein:
Dear alidist maintainers,

I would like to share some lessons I learned with Boost `1.70` to save some time when these issues pop up. Since version `1.70` Boost installs its own CMake package by default. While I applaud this effort (no need to update CMake any more for newest `FindBoost.cmake`), it comes with some issues:
* The `QUIET` argument passed to `find_package()` is ignored ([this is fixed](https://github.com/boostorg/boost_install/issues/3) in the next release on Boost side, CMake side [will be fixed](https://gitlab.kitware.com/cmake/cmake/issues/19186) in `3.15`)
* When building with `link=shared` (as we currently do), a simple `find_package(Boost)` may fail by complaining about an unsuitable build variant.
  * The error string which hints the solution (`... BUILD_SHARED_LIBS not ON, set Boost_USE_STATIC_LIBS=OFF to override ...`) does not work ([this is fixed](https://github.com/boostorg/boost_install/commit/160c7cb2b2c720e74463865ef0454d4c4cd9ae7c) in the next Boost release).
  * (The problem of inspecting the `BUILD_SHARED_LIBS` variable at all [is reported](https://github.com/boostorg/boost_install/issues/5), too.)
* If the [basic function signature (or module mode) of `find_package()`](https://cmake.org/cmake/help/latest/command/find_package.html) is used - should be most of our uses - CMake will first call the `FindBoost.cmake` module which will try to find a CMake package for Boost. If it succeeds, it may print wrong status messages. This affects CMake up to `3.14` and [will be fixed](https://gitlab.kitware.com/cmake/cmake/issues/19186) in `3.15` (hopefully, merge pending).

Considering the above, I recommend to disable the CMake package temporarily (pass `--no-cmake-config` to bjam) for `1.70` and enable once we go for `1.71`. Next thing to look out for is how homebrew will decide here for `1.70` ...

Barthelemy:
thanks a lot

Barthelemy:
By the way, I have just tried cmake 3.15 on mac and fairroot does not build : 
```
DEBUG:QualityControl:FairRoot:0: -- Build CbmModule with Gdml support
DEBUG:QualityControl:FairRoot:0: CMake Error at MbsAPI/CMakeLists.txt:19 (CHECK_SYMBOL_EXISTS):
[...]

PR #1817, opened 2019-09-06
Do not install boost's CMake config files, since they break some stupid dependent packages

davidrohr:
As of boost 1.70.0, boost automatically installs CMake Config files.
Dependent packages pick them up, instead of the CMake Module file that comes with CMake.
This leads to 2 problems:
1. Packages e.g. Ppconsul are unable to work with the CMake Config, and thus fail to set library and include paths. By chance, they still work if boost 1.70 is a system installation, since it automatically finds headers and libraries.
2. In the way we build boost, the CMake Config files seem to be broken. At least, requiring some of the boost libraries in `find_package(Boost)` fails, despite the library being present. I couldn't figure out why. With my (and MacOS brew) system boost, this problem does not appear, since the Boost Config files are correct.

The solution for now: we just delete the config files before we install them. Dependent packages will pick up CMakes boost module as it was before. This should be fixed in the future but not on this Friday evening.

### DIFF
--- a/boost.sh
+++ b/boost.sh
@@ -100,6 +100,10 @@ b2 -q                                            \
    ${CXXSTD:+cxxstd=$CXXSTD}                     \
    install
 
+# Remove CMake Config files, some of our dependent packages pick them up, but fail to use them
+# So for now we rely on the boost module FindBoost which comes with CMake
+rm -Rf "$INSTALLROOT"/lib/cmake
+
 # If boost_python is enabled, check if it was really compiled
 [[ $BOOST_PYTHON ]] && ls -1 "$INSTALLROOT"/lib/*boost_python* > /dev/null
 


dennisklein:
There is a flag to disable the CMake config, see my comments in #1614.

davidrohr:
Great, so you encountered the same issues. We should stop repeating work...

dennisklein:
There is also the counterpart switch on the consuming project side: if they have troubles with the boost 1.70 cmake package, you could configure them with `-DBoost_NO_BOOST_CMAKE=ON` to skip the cmake package and load via find module.

In Cmake 3.15 we added a compatibility layer in the upstream FindBoost.cmake *config mode* code path, so that resulting variables are closer to what the pure find module code path did in the past. Unfortunately, this may not be enough to catch all the legacy use cases in people's projects.

Since they are released, a Cmake 3.15.3+ and Boost 1.71.0 combo could be worth a try rather than fighting the broken boost 1.70.

davidrohr:
#1776 with boost 1.70 should work now with these fixes.
As I wrote above, we should do it properly for the next boost update.

PR #1824, opened 2019-09-13
bump qc to v0.18.3

Barthelemy:


Barthelemy:
why was this warning not detected in the QC PR ?

Barthelemy:
@ktf @MichaelLettrich Is `$ALIBUILD_O2_TESTS` not set in the builds in QC ? because the `-WError` should have been enabled there.

Barthelemy:
ok, I think that I know what happened. The build was broken by something else and it must have been in the shadow.

ktf:
There is actually an error in QC:

```
In file included from /build/ci_checks/alidist_checker_macos_o2_ci_0/sw/SOURCES/QualityControl/v0.18.2/v0.18.2/Modules/TPC/src/PID.cxx:25:
/build/ci_checks/alidist_checker_macos_o2_ci_0/sw/SOURCES/QualityControl/v0.18.2/v0.18.2/Modules/TPC/include/TPC/PID.h:55:9: error: private field 'mH' is not used [-Werror,-Wunused-private-field]
  TH1F* mH{ nullptr };
        ^
1 error generated.
```

Barthelemy:
@ktf Yes, this is what I meant. There was a genuine problem (a warning turned error) but we could not see it during the QC build because it was in the shadow of an unrelated problem. It is fixed in https://github.com/AliceO2Group/QualityControl/pull/244 but the CI has failed and I can't see the logs.

davidrohr:
@ktf : Opening the fulllog.txt for the alidist macOS CI still triggers a download for me. Could you check whether it has the correct Content-Type?

davidrohr:
@Barthelemy : Also here, the mac OS CI failed with a random QC ctest failure (different one than before). I am attaching the log in case the CI rerun and the log becomes unaccessible.
[fullLog.txt.tar.gz](https://github.com/alisw/alidist/files/3627306/fullLog.txt.tar.gz)

ktf:
@MichaelLettrich which hash of ali–bot is used by the macOS builders? Could you update to the tip of master?

Barthelemy:
@davidrohr Indeed. It confirms that the tests are not actually run in the QC PRs. @ktf @MichaelLettrich any idea why the tests are not executed in the QC ?

ktf:
Yes, as I said the problem is that your recipe has an incremental_recipe which does not run the tests.

PR #1827, opened 2019-09-16
Force enable GPU when O2_ALIBUILD_FORCE_CPU is set

ktf:


ktf:
@davidrohr I think I forgot this.

davidrohr:
@ktf : Yes, we need this as well in any case. But as far as I have seen, the env variables ALIBUILD_O2_FORCE_GPU and AMDAPPSDKROOT are not passed through to aliBuild in the CI.
If what I say is correct, this PR will not do anything, since the environment is missing.
(If the environment is there, I am clueless why it didn't detect OpenCL and ROCm before.)

ktf:
I had to force pull the images.. Let's see if it gets better...

PR #1831, opened 2019-09-20
Enable treelite in AliPhysics

ktf:


ktf:
@pzhristov Apparently now we need treelite for AliPhysics. This was changed in the CI defaults but not in the release building defaults. Was there any reason to keep it disabled?

ktf:
O2 tests not affected by this change. Actually we do not seem to test aliphysics at all. I will add a builder for that as well.

mpuccio:
This actually breaks the compilation of AliPhysics... Was the issue only related to treelite not being used with ROOT5? Currently people are anyway using it with ROOT6...

ktf:
Yeah, but we still build with ROOT5 for the AliPhysics releases, so we need to disable whatever uses treelite in aliphysics, I guess...

mpuccio:
Yes, this is already the case. The module ML is disabled if it does not find treelite and the code depending on that is disabled too

ktf:
Ok, @mpuccio and myself discussed privately and the issue is actually that the `root6-next` default should be used, not root6, so I reverted this.

PR #1834, opened 2019-09-20
bump qc v0.18.3

Barthelemy:


Barthelemy:
@ktf There is again a test without logs. Does it show it failing because there are no logs ?

PR #1836, opened 2019-09-21
boost: attempt at 1.72.0 as discussed in WP3

ktf:


davidrohr:
NIce, so boost 1.71 already builds with the recipe as is.
@dennisklein : Last time you mentioned with boost 1.71, we could simplify some things.
Would you have time to adapt the recipe accordingly, or could you let some of us know what you meant and what would need to be done?

dennisklein:
Which recipe?

With Boost 1.71 and CMake 3.15.3 I expect we do not need to change our CMakeLists.txt from what we have done pre Boost 1.70 (assuming we already used `Boost::*` imported targets).

davidrohr:
@dennisklein : ok, right now we have this workaround to not install the boost cmake config files. YOu are saying, once we bump cmake and boost, we can just disable this workaround? Anything else?
Thx, David

dennisklein:
https://gitlab.kitware.com/cmake/cmake/issues/19714. This affects boost 1.71 from brew at the moment. Let's see how brew resolves it.

dennisklein:
> YOu are saying, once we bump cmake and boost, we can just disable this workaround?

Yes, but we still need to test. Some switches/defaults may still have changed as you can see in the issue above. The upstream maintainer of the boost cmake package still sees no problem in breaking everyone... even after all the issues with the Boost 1.70 release.. there is a limit to what we could do on the CMake side of things.

davidrohr:
@ktf : Boost 1.72 is broken, it doesn't work with DDS and some other packages.
There is a fix here https://github.com/OpenMandrivaAssociation/boost/commit/167e6d8982319d16d846f8d337b93723b807de9a (found by @dennisklein ).
I have already tried locally with this patch applied, and then 1.72 will work.
Do you have this patch in the alisw fork?

Otherwise, we have to wait for 1.73, where it is fixed.

ktf:
Ok. This was done to followup a request from @chiarazampolli and @shahor02. I will try to understand the state of our fork...

davidrohr:
@ktf : I think I set up our fork, and it will not work. I already tried.
[boost 167e6d8982319d16d846f8d337b93723b807de9a.patch.gz](https://github.com/alisw/alidist/files/4409105/boost.167e6d8982319d16d846f8d337b93723b807de9a.patch.gz)

Please apply the attached patch to it, then it should do.

ktf:
I guess you meant "and it will work". I updated our fork and this PR. Let's see. I think it will allow us to cleanup some cmake stuff as well.

chiarazampolli:
It seems the macOS part fails, but I could not see the log. The dataflow part also fails, but is not required, correct?

davidrohr:
@chiarazampolli : dataflow is new and not yet required.
macOS is failing for many PRs, it seems there is an unrelated general problem on the mac CI.

chiarazampolli:
Ok, so this can be merged?

davidrohr:
no, we should still check whether it works with macos, since it already has some patches in it. I am just saying the fact that the mac CI fails at the moment doesn't tell yes or no, the other problem must be fixed first.

MichaelLettrich:
macOS passed after I cleaned up the mirror this morning.

awegrzyn:
Hi @ktf @MichaelLettrich,
I guess on macOS it will pass as long as there is NO boost 1.72 from brew. Otherwise, it will pick up boost 1.72 from brew and fail on executor_type @ FairMQ, am I correct? I'm testing this now.

davidrohr:
Yes, we'll have to disallow system boost in any case for 1.72, since the official version is broken. We need the alidist-compiled one with the patch. For 1.73, we can revert to using it vanilla.

Kind Regards
David Rohr

dennisklein:
> fail on executor_type @ FairMQ, am I correct?

For reference, see https://github.com/FairRootGroup/FairMQ/issues/237 (still looking into replacing Boost.Process, but no fix is yet finished) and https://github.com/FairRootGroup/DDS/issues/286 (wontfix).

chiarazampolli:
FYI, I had to apply either of the following fixes to compile on Ubuntu:
From @preghenella 
{code}
diff --git a/boost.sh b/boost.sh
index e071283..6153c45 100644
--- a/boost.sh
+++ b/boost.sh
@@ -69,7 +69,7 @@ cd $BUILDDIR/tools/build
 # installations.
 case $ARCHITECTURE in
   osx*)  ;;
-  *) export CPLUS_INCLUDE_PATH="$CPLUS_INCLUDE_PATH:$(python3 -c 'import sysconfig; print(sysconfig.get_path("include"))')" ;;
+  *)     ;; ### export CPLUS_INCLUDE_PATH="$CPLUS_INCLUDE_PATH:$(python3 -c 'import sysconfig; print(sysconfig.get_path("include"))')" ;;
 esac
 bash bootstrap.sh $TOOLSET
 mkdir -p $TMPB2
{code}

or from @aalkin :
{code}
diff --git a/boost.sh b/boost.sh
index e071283..3f4fd75 100644
--- a/boost.sh
+++ b/boost.sh
@@ -67,11 +67,11 @@ cd $BUILDDIR/tools/build
 # the ABI suffix. E.g. ../include/python3 rather than ../include/python3m.
 # This is causing havok on different combinations of Ubuntu / Anaconda
 # installations.
+bash bootstrap.sh $TOOLSET
 case $ARCHITECTURE in
   osx*)  ;;
 esac
-bash bootstrap.sh $TOOLSET
 mkdir -p $TMPB2
 ./b2 install --prefix=$TMPB2
 export PATH=$TMPB2/bin:$PATH 
{code}

Both worked equally well.

PR #1838, opened 2019-09-24
Bump HepMC to v2.06.10

qgp:


ktf:
Looks like you need to point to the official sources?

jgrosseo:
I think in that case, we again go to our repository. But as long as we don't need, do you prefer to clone it? Actually, that version is not yet in our repo, that is why Giulio changed the source...

ktf:
Yes, exactly. I think we should only use our own repository only if we have an immanent need to patch, not "just in case".

qgp:
I am not following here. Giulio, why was it that we "need" to point to the official sources? At least so far, I had understood that the policy was to take the software from the alisw repositories, which are fully under our control. That is why I important the new version there and also added the new tag there. What is missing?

If you are saying that the policy is to use upstream repos, fine with me. We should just try to have some consistency and make sure that we don't break our builds by potential upstream changes (as happened for LHAPDF for example).

For PAG-MC approval, what do I have to do? Isn't it enough that I am saying it's ok?

davidrohr:
Just to give my opinion:
I am in favor of using upstream repos, if we use a tagged version, and if we do not need to apply special patches, and if the upstream repo is not unbearably large, and if the upstream repo does not need git submodules.
If this is fulfilled, there should be no way how changes in upstream break our build (assuming they do not overwrite tags).
In any other case, I'd clone it to alisw.

ktf:
@qgp I thought the tag was not there in our own repo, because the build was silently failing, which is usually due to a missing tag. That's why I moved it.

Since I just noticed the tag is there, then it's clearly up to you to decide where to point, but I would recommend against having our own tag unless we encounter one of the issues listed by David

qgp:
As said, fine with me. I was just following the policy as explained some time ago by Dario. If we use the upstream repo here, we should also do this in other cases (from here on).

Remains the question about the PAG-MC approval, maybe @maireiphc can give his blessing. It seems I can't ...

maireiphc:
If I get you correctly, it means that we will need to clean many of the alisw repositories for MC and start from the official repos wherever we can.
Fine for me but to be done little by little.
As for HEPmc, let's start with this one.
Approved for merging with the official repo from HEPmc.

maireiphc:
Sorry maybe to quick in the merging stage. Only 5 out of 6 checks have been successful.
We still do not get why the build silently fails, do we ?
Based on what @ktf said, it could be due to a missing tag. Could it be due to hepmc.sh which calls for "**v**2.06.10" while the existing tag under the official repo (https://gitlab.cern.ch/hepmc/HepMC) seems to be : "2.06.10" or "HEPMC_02_06_10" ?
I don't know which one we intend to pick (they looks similar based on the date of the latest commits...)

PR #1839, opened 2019-09-24
readoutcard: depend on Python rather than Python-modules (O2-853)

ktf:


ktf:
@teo If this works, this should solve most of the issues with the updatable RPMs, actually. Is there any particular reason why `Python-modules` is used, rather than just `Python`?

teo:
> @teo If this works, it should solve most of the issues with the updatable RPMs, actually. Is there any particular reason why `Python-modules` is used, rather than just `Python`?

@kostorr question for you ↑

I can only test this once RPMs are buildable again, issue: https://alice.its.cern.ch/jira/browse/O2-852

ktf:
@teo, could you at least fix the FLP packages?

teo:
> @teo, could you at least fix the FLP packages?

Of course, we'll fix the recipes we maintain as soon as we can build&test. Thank you for your effort @ktf

kostorr:
@ktf I have no idea... I see no problems with the change on my builds, so it's fine by me.

PR #1844, opened 2019-09-28
Fix kfparticle CXX std

davidrohr:
KFParticle unconditionally overrides the CXX std to C++11, which fails to build with ROOT, if ROOT was built with a higher standard, which breaks our O2 defaults.
I asked Maksim and Ivan and apparently they do not want to fix this upstream for the time being.
So I propose we apply this fix.

### DIFF
--- a/kfparticle.sh
+++ b/kfparticle.sh
@@ -11,6 +11,8 @@ build_requires:
 ---
 #!/bin/bash -e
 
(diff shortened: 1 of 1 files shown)


davidrohr:
Fine with me, but this some with write access to alisw has to do.

ktf:
I've created the mirror and updated this PR to use the new url / tag.

davidrohr:
thx, looks good

PR #1862, opened 2019-10-14
Drop xrootd from alien-runtime dependencies

ktf:
As discussed with @Atlantic777, we need to drop those in order to be able to break the circular dependency when building xrootd python bindings.

### DIFF
--- a/alien-root-legacy.sh
+++ b/alien-root-legacy.sh
@@ -14,9 +14,6 @@ append_path:
   ROOT_PLUGIN_PATH: "$ALIEN_ROOT_LEGACY_ROOT/etc/plugins"
 prepend_path:
   ROOT_INCLUDE_PATH: "$ALIEN_ROOT_LEGACY_ROOT/include"
-env:
-  GSHELL_ROOT: "$ALIEN_ROOT_LEGACY_ROOT"
-  GSHELL_NO_GCC: "1"
 ---
 #!/bin/bash -e
 
@@ -39,16 +36,6 @@ cmake $BUILDDIR                                          \
       -DROOT_VERSION="$ROOT_MAJOR"
 
 cmake --build . --target install ${JOBS:+-- -j$JOBS}
-
-for RPKG in $BUILD_REQUIRES; do
-  RPKG_UP=$(echo $RPKG|tr '[:lower:]' '[:upper:]'|tr '-' '_')
-  RPKG_ROOT=$(eval echo "\$${RPKG_UP}_ROOT")
-  rsync -a $RPKG_ROOT/ $INSTALLROOT/
-  pushd $INSTALLROOT/../../..
-    env WORK_DIR=$PWD sh -e $INSTALLROOT/relocate-me.sh
-  popd
-  rm -f $INSTALLROOT/etc/modulefiles/{$RPKG,$RPKG.unrelocated} || true
-done
 
 if [[ $ARCHITECTURE == osx* ]]; then
   # Due to some ROOT quirks, we create .so symlinks pointing to the real .dylib libs on macOS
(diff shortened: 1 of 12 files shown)


Atlantic777:
@ktf it looks like we actually want to keep ApMon-CPP in AliEn-Runtime if possible.

root.sh
```diff
@@ -225,10 +225,10 @@ module load BASE/1.0 ${ALIEN_RUNTIME_VERSION:+AliEn-Runtime/$ALIEN_RUNTIME_VERSI
                      ${LZMA_VERSION:+lzma/$LZMA_VERSION-$LZMA_REVISION}
 # Our environment
 setenv ROOT_RELEASE \$version
-setenv ROOT_BASEDIR \$::env(BASEDIR)/$PKGNAME
-setenv ROOTSYS \$::env(ROOT_BASEDIR)/\$::env(ROOT_RELEASE)
-prepend-path PYTHONPATH \$::env(ROOTSYS)/lib
-prepend-path PATH \$::env(ROOTSYS)/bin
-prepend-path LD_LIBRARY_PATH \$::env(ROOTSYS)/lib
+setenv ROOT_BASEDIR \$BASEDIR/$PKGNAME
```
ktf: I think here you do need $::env

ktf: It's only prepend-path which cannot have `:`

xalienfs.sh
```diff
@@ -61,7 +64,9 @@ module load BASE/1.0 XRootD/${XROOTD_VERSION}-${XROOTD_REVISION}             \\
             AliEn-Runtime/${ALIEN_RUNTIME_VERSION}-${ALIEN_RUNTIME_REVISION}
 
 # Our environment
-setenv XALIENFS_ROOT \$::env(BASEDIR)/$PKGNAME/\$version
-prepend-path LD_LIBRARY_PATH \$::env(XALIENFS_ROOT)/lib
-prepend-path PATH \$::env(XALIENFS_ROOT)/bin
+set XALIENFS_ROOT \$BASEDIR/$PKGNAME/\$version
```
ktf: You need `::env`

alien-root-legacy.sh
```diff
@@ -77,12 +64,10 @@ module load BASE/1.0 ${GCC_TOOLCHAIN_VERSION:+GCC-Toolchain/$GCC_TOOLCHAIN_VERSI
...
 
 # Our environment
-setenv ALIEN_ROOT_LEGACY_ROOT \$::env(BASEDIR)/$PKGNAME/\$version
-setenv GSHELL_ROOT \$::env(ALIEN_ROOT_LEGACY_ROOT)
-setenv GSHELL_NO_GCC 1
-prepend-path PATH \$::env(ALIEN_ROOT_LEGACY_ROOT)/bin
-prepend-path LD_LIBRARY_PATH \$::env(ALIEN_ROOT_LEGACY_ROOT)/lib
-append-path ROOT_PLUGIN_PATH \$::env(ALIEN_ROOT_LEGACY_ROOT)/etc/plugins
-prepend-path ROOT_INCLUDE_PATH \$::env(ALIEN_ROOT_LEGACY_ROOT)/include
+set ALIEN_ROOT_LEGACY_ROOT \$BASEDIR/$PKGNAME/\$version
```
ktf: You need `::env`

Atlantic777:
@ktf so, this PR is good to go as far as I am concerned.

root.sh
```diff
@@ -227,8 +227,10 @@ module load BASE/1.0 ${ALIEN_RUNTIME_VERSION:+AliEn-Runtime/$ALIEN_RUNTIME_VERSI
 setenv ROOT_RELEASE \$version
 setenv ROOT_BASEDIR \$::env(BASEDIR)/$PKGNAME
 setenv ROOTSYS \$::env(ROOT_BASEDIR)/\$::env(ROOT_RELEASE)
-prepend-path PYTHONPATH \$::env(ROOTSYS)/lib
-prepend-path PATH \$::env(ROOTSYS)/bin
-prepend-path LD_LIBRARY_PATH \$::env(ROOTSYS)/lib
+
+set ROOTSYS  \$::env(ROOTSYS)
+prepend-path PYTHONPATH \$ROOTSYS/lib
```
ktf: Please let's try to keep it uniform with the rest, why not using `\$ROOT_ROOT`?

Atlantic777: Well, as ROOTSYS was an env variable I just created a copy of it that is a normal variable. I could also rename this "normal ROOTSYS" variable to ROOT_ROOT. I'm on it.

PR #1865, opened 2019-10-16
libpng: cleanup for modules 3 -> 4 migration

ktf:


ktf:
@teo, try to adapt your modulefiles as in this PR to see if modules4 digests it.

libpng.sh
```diff
@@ -37,8 +37,7 @@ module-whatis "ALICE Modulefile for $PKGNAME $PKGVERSION-@@PKGREVISION@$PKGHASH@
 # Dependencies
 module load BASE/1.0 $([[ $ALIEN_RUNTIME_VERSION ]] && echo "AliEn-Runtime/$ALIEN_RUNTIME_VERSION-$ALIEN_RUNTIME_REVISION" || echo "${ZLIB_VERSION:+zlib/$ZLIB_VERSION-$ZLIB_REVISION}")
 # Our environment
-setenv LIBPNG_ROOT \$::env(BASEDIR)/$PKGNAME/\$version
-prepend-path PATH $::env(LIBPNG_ROOT)/bin
-prepend-path LD_LIBRARY_PATH $::env(LIBPNG_ROOT)/lib
-$([[ ${ARCHITECTURE:0:3} == osx ]] && echo "prepend-path DYLD_LIBRARY_PATH $::env(LIBPNG_ROOT)/lib")
+set LIBPNG_ROOT \$::env(BASEDIR)/$PKGNAME/\$version
```
teo: Just to confirm, we can keep using `::env` but only with `set` and never `setenv`?

ktf: I don't know. :) I have two ideas on what might be wrong:

1) `append-path` does not allow for ":" anymore in 4 (at least as far as I can tell) and `$::env` should really become `$env` or something equivalent.
2) if you setenv, the variable is not available anymore within the same module file. 

I suspect the problem is 1.

teo: I confirm that your change fixes this.

ktf: Ok, then we have quite some cleanup to do... ;-)

PR #1867, opened 2019-10-17
linter: do not allow ':' in prepend path anymore

ktf:


ktf:
@alisw/o2-wp3 @teo we need to cleanup this everywhere if we want to have some hope to move to modules4. See also O2-877 and O2-876.

PR #1868, opened 2019-10-18
Bump cgal, fixes MacOS issue with C++17

davidrohr:
@jolopezl : Seems to be a problem that affected only MacOS with C++17, it should be fixed after this version bump. Could you try if it works for you?

### DIFF
--- a/defaults-o2.sh
+++ b/defaults-o2.sh
@@ -30,7 +30,7 @@ overrides:
   ROOT:
     tag: "v6-18-04" 
   cgal:
-    version: "4.11"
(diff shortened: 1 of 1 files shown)


jolopezl:
Hi @davidrohr: I confirm that `$> aliBuild build AliRoot --defaults o2` builds AliRoot with o2 defaults successfully. Thanks!

davidrohr:
ok, good, @ktf : Can we merge this? It affects only O2 defaults, and fixes a MacOS issue.

PR #1870, opened 2019-10-21
Add Configuration as a dependecy for ReadoutCard

kostorr:


Barthelemy:
@ktf Could you refresh my memory about the failing travis test ?

ktf:
you should remove the line with DYLD... which is not needed anymore.

readoutcard.sh
```diff
@@ -32,6 +33,7 @@ cmake $SOURCEDIR                                                      \
       -DCMAKE_INSTALL_PREFIX=$INSTALLROOT                             \
       ${BOOST_VERSION:+-DBOOST_ROOT=$BOOST_ROOT}                      \
       ${COMMON_O2_VERSION:+-DCommon_ROOT=$COMMON_O2_ROOT}             \
+      ${CONFIGURATION_VERSION:+-DConfiguration_ROOT=$CONFIGURATION_ROOT} \
```
ktf: Can configuration be optional or coming from the system? If not I would simply drop the `${CONFIGURATION_VERSION:+` part.

ktf:
Merging this. Consider making Configuration mandatory and remove the support for optional presence, I guess.

PR #1872, opened 2019-10-28
FairMQ: bump to 1.4.9

ktf:


ktf:
This should address a regression when using shmem transport. See also AliceO2Group/AliceO2#2547.

ktf:
@dennisklein @rbx any ETA for 1.4.9? IIRC we need that for the shared memory issue.

dennisklein:
> @dennisklein @rbx any ETA for 1.4.9? IIRC we need that for the shared memory issue.

We were [waiting for your feedback](https://github.com/AliceO2Group/AliceO2/pull/2547#issuecomment-547591245). We are ready to move forward. Will do the release right now

dennisklein:
https://github.com/FairRootGroup/FairMQ/tree/v1.4.9

ktf:
@rbx @dennisklein what happened to Uuid()?

rbx:
Nothing, `std::string Uuid();` is still in `Tools.h` and/or `tools/Unique.h`, why?

rbx:
Ah, there is a new target `FairMQ::Tools` that you need to depend on.

rbx:
Although it should have been picked up from FairMQ target, since it is a public dependency of it...

ktf:
@rbx Can you check why that is not the case?

rbx:
The file that has problem with it is https://github.com/AliceO2Group/Readout/blob/v1.2.0/src/ConsumerFMQchannel.cxx. Looking at it more closely I see that there is no tools included anywhere in the entire Readout repository (neither `<fairmq/Tools.h>` nor `<fairmq/tools/Unique.h>`).
In the 1.4.9 we did a bit of cleanup of the include statements. Looks like ConsumerFMQchannel depended on one of the FairMQ headers that it does include having a tools include that is now gone.

Anyway, the correct solution would be for ConsumerFMQchannel.cxx to include the tools that it uses, either `<fairmq/Tools.h>` or `<fairmq/tools/Unique.h>`.

ktf:
Ah, ok, understood.

ktf:
See AliceO2Group/Readout#84

ktf:
@MichaelLettrich are the mac builds just slow?

PR #1876, opened 2019-11-08
Bump ROOT for AliRoot/AliPhysics to fix compilation on Catalina

mpuccio:
Hey @ktf, @pzhristov,
apparently this was already known to ROOT developers (see [v6-18-00 release notes](https://root.cern.ch/content/release-61800)), but v6-16-00 does not compile correctly on Catalina.
Since we moved to v6-18-04 in O2 we might do the same for AliRoot/AliPhysics.
Additionally, since this is another ROOT6 bump, I think it is the right timing to remove the old/unmaintained root6 defaults.

Let me know what you think!


ktf:
Thanks for this. I've fixed the linter issue and will merge this if everything goes fine.

PR #1878, opened 2019-11-12
RapidJSON: do not expose bogus runtime paths

ktf:
RapidJSON is actually headers only. Proper fix would actually be to use RapidJSON as `build_requires` only.


Barthelemy:
macos build logs are missing

ktf:
Looks like the mirror for ReadoutCard got corrupted on one of the machines. I removed it and it will hopefully work at the next iteration.

awegrzyn:
@ktf FLP packages do not use RapidJSON anymore, what about modifying `CODEOWNERS` files?

ktf:
macOS issue unrelated. Merging.

PR #1880, opened 2019-11-12
arrow: enable Gandiva and Parquet support

ktf:
* Add new tool LLVM which provides llvm and clang 7.1.0. This
  is a required dependency for Gandiva. In principle, since
  now LLVM provides its own full fledged repository, including
  clang, we should probably drop using our own mirror elsewhere.
* Include re2 compile time dependency, needed by Gandiva.
* Make sure thrift can compile correctly with modern compilers.
* Enable Gandiva and Parquet in Arrow.
* Improve detection of Arrow in ROOT.

### DIFF
--- a/arrow.sh
+++ b/arrow.sh
@@ -6,11 +6,15 @@ requires:
   - boost
   - lz4
   - RapidJSON
+  - thrift
+  - LLVM
+  - protobuf
 build_requires:
   - zlib
   - flatbuffers
   - CMake
   - double-conversion
+  - re2
 env:
   ARROW_HOME: "$ARROW_ROOT"
 ---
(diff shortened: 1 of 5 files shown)


ktf:
Checks now pass. Merging this so that on monday people will be able to download a precompiled copy.

sawenzel:
This PR brings in a new LLVM dependency. In principle we already had llvm in the clang.sh recipe (used for the code checks). I am wondering if we could potentially marry the two, or does Gandiva need a very particular version of LLVM?

ktf:
Gandiva requires 7.0 or 7.1, so in principle we could indeed merge the two. I did not reuse the `clang.sh` because I was not sure if there was any specific requirement for it. On my side the only weird thing is that I need clang to be out of the `$PATH` to avoid interference with GCC.

PR #1890, opened 2019-11-18
Yacc-like: have a way to build our own bison / flex

ktf:
This is required because thrift requires bison 2.5 which
is not default on some platforms.


ktf:
@Barthelemy this should fix O2-918

Barthelemy:
thank you !

ktf:
`brew install flex bison` should also fix it.

ktf:
Will merge this after today daily.

PR #1891, opened 2019-11-18
thrift: build without java

sawenzel:


ktf:
Will merge after the daily.

PR #1893, opened 2019-11-19
yacc-like: avoid cyclic dependency.

ktf:


ktf:
@awegrzyn this should fix your problem.

ktf:
Merging since the builds are currently broken due to this.

PR #1894, opened 2019-11-19
Detect circular dependencies in recipes

ktf:


ktf:
This should prevent the issue.

PR #1898, opened 2019-11-19
yacc-like: do not compile docs

ktf:


ktf:
This should remove the need for help2man.

PR #1899, opened 2019-11-19
[WIP]: More circular dependencies checks

ktf:


ktf:
Works correctly, at least to a first loop. The message is slightly erroneous because the circular dependency might be elsewhere.

PR #1903, opened 2019-11-22
Enable treelite for jalien defaults

Atlantic777:
This should fix the AliPhysics + JAliEn daily builds.


ktf:
Only affects JAlien. Merging.

PR #1904, opened 2019-11-24
LLVM: fix for symbols exposed to Cling

aalkin:
- Suppress building of unnecessary tools and tests
- Use -fvisibility=hidden


aalkin:
@ktf looks like the crucial part was `CLANG_TOOL_LIBCLANG_BUILD=OFF`

PR #1905, opened 2019-11-25
bump readout v1.2.0

sy-c:


sy-c:
@alisw/infrastructure-admins 
I opened this PR 2 hours ago, and the macOS test is still in state "Expected — Waiting for status to be reported". There is no other recent PR. I would expect tests to complete within minutes.
Can you please check what is going on?

PR #1906, opened 2019-11-25
Use own OpenSSL also on mac

ktf:
Now that homebrew ships with OpenSSL 1.1.x only, we cannot rely on
it for old alien installations, so we bring our own and remove
the old osx-system-openssl hack.


openssl.sh
```diff
@@ -4,19 +4,32 @@ tag: OpenSSL_1_0_2o
...
+    EXTRA_OPTS="-Wno-nonportable-include-path"
+  ;;
+  *) 
+    EXTRA_DEPS=zlib
+    EXTRA_OPTS="-Wa,--noexecstack" 
+  :;
+esac
 
-./config --prefix="$INSTALLROOT"                   \
+export KERNEL_BITS=64
```
ktf: @mfasDa allegedly this is the trick to make it build 64 bit on macOS.

mfasDa: Works, but it results in an error in the install part:
```
DEBUG:Readout:OpenSSL:aliceo2: + install_name_tool -id /Users/markusfasel/alice/sw/osx_x86-64/OpenSSL/v1.0.2o-1/lib/libcrypto.1.0.0.dylib osx_x86-64/OpenSSL/v1.0.2o-1/lib/libcrypto.1.0.0.dylib
DEBUG:Readout:OpenSSL:aliceo2: error: /Library/Developer/CommandLineTools/usr/bin/install_name_tool: can't open input file: osx_x86-64/OpenSSL/v1.0.2o-1/lib/libcrypto.1.0.0.dylib for writing (Permission denied)
DEBUG:Readout:OpenSSL:aliceo2: error: /Library/Developer/CommandLineTools/usr/bin/install_name_tool: can't lseek to offset: 0 in file: osx_x86-64/OpenSSL/v1.0.2o-1/lib/libcrypto.1.0.0.dylib for writing (Bad file descriptor)
DEBUG:Readout:OpenSSL:aliceo2: error: /Library/Developer/CommandLineTools/usr/bin/install_name_tool: can't write new headers in file: osx_x86-64/OpenSSL/v1.0.2o-1/lib/libcrypto.1.0.0.dylib (Bad file descriptor)
DEBUG:Readout:OpenSSL:aliceo2: error: /Library/Developer/CommandLineTools/usr/bin/install_name_tool: can't close written on input file: osx_x86-64/OpenSSL/v1.0.2o-1/lib/libcrypto.1.0.0.dylib (Bad file descriptor)
```

ktf: Do you also have:

https://github.com/alisw/alidist/pull/1906/files#diff-bc6a953101d106eb0b4914f811bcbb84R55-R57

?

mfasDa: Works for OpenSSL! Thanks! Now of course the packages depending on OpenSLL run into errors and need to be adapted until the PR is merged.

mfasDa:
With the changes in the recipes I am able to build AliPhysics including legacy AliEn, but it screws up the module files:

```
alienv enter AliEn-ROOT-Legacy/latest
OpenSSL(13):ERROR:105: Unable to locate a modulefile for 'OpenSSL/-4'
XRootD(13):ERROR:105: Unable to locate a modulefile for 'XRootD/-2'
Currently Loaded Modulefiles:
  1) BASE/1.0                   2) AliEn-Runtime/v2-19-le-2   3) Python-modules/1.0-1       4) ROOT/v6-16-00-1            5) AliEn-ROOT-Legacy/latest
Use alienv list to list loaded modules. Use exit to exit this environment.
```

Connection with AliEn is therefore not possible due to missing libraries:

```
[AliEn-ROOT-Legacy/latest] ~ $> alien-token-init mfasel
---------------------------------------------------------------
Setting central config:
===============================================================
export alien_API_SERVER_LIST="pcapiserv03.cern.ch:10000|pcapiserv08.cern.ch:10000|"
export TERMINFO=/usr/share/terminfo
===============================================================
---------------------------------------------------------------
Setting closest site to: CERN
===============================================================
Using X509_CERT_DIR=/Users/markusfasel/alice/sw/osx_x86-64/AliEn-Runtime/v2-19-le-2/globus/share/certificates
Error: not all shared libraries are in the DYLD_LIBRARY_PATH!   Check with otool -L
```

mpuccio:
@ktf @mfasDa did you find a solution for this issue? I cannot compile on my mac...

mfasDa:
@mpuccio I got finally a working environment again merging this pull request with https://github.com/alisw/alidist/pull/1862, commenting in root.sh lines
```
#if [[ $ALIEN_RUNTIME_VERSION ]]; then
#  # Get them from AliEn-Runtime in the Modulefile
#  unset OPENSSL_VERSION XROOTD_VERSION LIBXML2_VERSION
#endif
```

and after loading the modules doing

```
export DYLD_LIBRARY_PATH=$LD_LIBRARY_PATH
```

Now also the old alien works again. However grpc is still not compiling in the o2 suite, which is needed for readout. I think it would be helpful to post a warning not to upgrade homebrew until this issue is fixed not to have more people running into the issue.

Atlantic777:
@mfasDa do alien-token-init and aliensh also work fine? I still see the problem with segfault in xrdgsiproxy when running alien-token-init.

mfasDa:
@Atlantic777 Yes, after manually setting the DYLD_LIBRARY_PATH alien-token-init, aliensh and the alien command line tools work for me. I didn't observe a segfault.

openssl.sh
```diff
@@ -25,18 +38,23 @@ rsync -av --delete --exclude="**/.git" $SOURCEDIR/ .
          no-ecdsa                                  \
```
mfasDa: It seems like one of the three lines is responsible for the failing build of grpc in the o2 builds. If I remove these lines grpc builds.

mfasDa:
In at least two build recipes `OPENSSL_ROOT_DIR` must be replaced with `OPENSSL_ROOT`:

- control-occplugin.sh
- readout.sh

After the changes and removing the -no-ec... in the openssl build recipe I am finally able again to also build readout from the O2 stack.

mpuccio:
Hi, it looks to me that this might be a good occasion to stop trying to workaround the things and move to jalien (since anyway it should go in production sooner or later).
I am trying it locally and I think it would be good if we start having large scale testing of it (and Catalina users might be the perfect test case).

mfasDa:
@mpuccio As many users use alien command line tools in their scripts for practical use having xjalienfs is necessary before going to jalien. I also tried jalien locally and it works fine for me, and I would also be in favor of stopping try workarounds for legacy code, but having alien command line tools is a must for me before switching.

mpuccio:
Completely agreed. I am trying it locally and indeed without alien command line tools it looks quite cumbersome to me. However they should not be too much complicated to be implemented and this is the kind of feedback that @Atlantic777 and others needs to make jalien ready for the users.

Atlantic777:
The xjalienfs is ready to be merged since last week. I believe it will be included in the todays AliPhysics daily build.

adriansev:
@mfasDa @mpuccio @ktf @Atlantic777 please take a look here https://gitlab.cern.ch/jalien/xjalienfs/tree/master/bin and tell me if there any other alien_ style commands that should be added (or other shortcuts that could be useful even if there were not present in legacy tools). Thank you!

mfasDa:
@adriansev In my opinion for productive work at least the following commands from xalienfs which are not yet implemented would be needed:

- alien_erase
- alien_mkdir
- alien_rm
- alien_rmdir
- alien_mv

To have at least the most important file system operations implemented (mkdir, rmdir, ls, mv, cp, rm, find). In some cases also queue operations are performed in scripts, which would require

- alien_submit
- (alien_queue)
- (alien_ps)

Not sure if others are using other commands from the command line tools.

Cheers

Markus

adriansev:
@mfasDa added the missing shortcuts. note that erase and queue verbs does not exists on jalien server (at least yet :) ). still for scripts i would recommend alien_cmd "${@}" or better just alien.py "${@}" to skip some sub-processes. alien.py can be used standalone so you can give it a drive test and tell me what's missing, what can be better. thanks!

mfasDa:
@adriansev alien_ls shortcut would also be helpful. I see alien.py already supports ls, so the shortcut is probably straight forward to add.

adriansev:
@mfasDa done, thanks a lot for feedback!

Atlantic777:
After rebuilding with https://github.com/alisw/alidist/pull/1906/commits/4d1a3b3569de074cbf3fb2197bd845fd9b8d4d34 the alien-token-init segfault went away on my machine.

pzhristov:
What is the situation with this PR?

ktf:
I suspect it entered in a different form. Closing it.

PR #1907, opened 2019-11-26
arrow: do not expose LLVM outside gandiva

ktf:
This is to prevent it conflicting with the LLVM coming from ROOT.


ktf:
@Atlantic777 could you please sign off?

Atlantic777:
+1

PR #1910, opened 2019-11-27
[WIP] (O2-932) Use PACKAGE_REVISION rather than PACKAGE_VERSION as a guard

Barthelemy:


Barthelemy:
@ktf @MichaelLettrich There seems to be a problem with the build machines, or at least the logs.

PR #1912, opened 2019-11-28
bump readout v1.2.3

sy-c:


ktf:
@sy-c there is a genuine compilation issue on macOS. Can you please fix?

ktf:
@sy-c I took the liberty to fix the module so that it is compatible with modules 4.

ktf:
@Barthelemy the fix I did is what I meant when discussing the other cleanup you are doing.

PR #1916, opened 2019-12-02
bump readout v1.2.4

sy-c:


sy-c:
@alisw/infrastructure-admins 
The test build/AliGenerators/prod-latest is failing. I don't think it has anything to do with this PR. Can you please have a look?

PR #1918, opened 2019-12-03
Fix LLVM symbol collision in ROOT and use only a single LLVM

davidrohr:


clang.sh
```diff
@@ -1,7 +1,7 @@
 package: Clang
-version: "7.0.0"
-tag: v7.0.0-alice1
-source: https://github.com/alisw/clang
+version: "v7.1.0"
+tag: "llvmorg-7.1.0"
+source: https://github.com/llvm/llvm-project
```
sawenzel: Is this providing the clang tools and extra tools as well? If not, it's likely breaking the codechecker.

davidrohr:
Not sure, this is what I want to try, if not we need to create a new tag in our fork.

clang.sh
aalkin: By default it should build all projects, including clang and its tools. Otherwise you can explicitly set `-DLLVM_ENABLE_PROJECTS="clang;clang-tools-extra"`

davidrohr:
the llvm-build went fine locally with the new official mirror, had to fix few other things, hope it works now.
But my patch is failing in the ROOT CI, so perhaps this is in vain and we cannot fix it this way. Let's see.

davidrohr:
@ktf : Forgive my ignorance, but I do not understand the problem reported by travis linter in arrow.sh and clang.sh.
Could you clarify what I need to fix?

ktf:
i did the work for you. see one of the commits for the explanation .

davidrohr:
Shit, I guess the clang include headers overwrite the macos system includes somehow.
Will have a look.

davidrohr:
New attempt to do it properly. This version does no longer break the ROOT ctests. It requires -fPIC to be set, but we have that anyway unconditionally for all packets.

If this passes the CI, please do not merge yet. I'd probably add something such that my hack is only active for O2 defaults, to avoid any side effects on AliRoot.

This hack should be only temporary until the problem is fixed properly on the ROOT side.

davidrohr:
For reference: This problem has been fixed on the ROOT side. The necessary patch will be ported to 6.18-patches branch after 6th of Januarry. Then we can simply bump ROOT for good. I'll leave this open in the meantime.

davidrohr:
Changed the patch applied to ROOT to the new version provided by Axel, let's see if this works...

davidrohr:
@ktf : Could you check why the build/O2/alidist log is empty?

MichaelLettrich:
@davidrohr no matter what it was, it's gone now. remove WIP and you're good to go.

PR #1919, opened 2019-12-04
Bump clang for codechecker to clang 9

davidrohr:
- Updates clang to clang 9
- Has all the preparations to use this clang instead of the llvm package
- Linter fixes

For the codechecker, this must be used together with AliceO2Group/O2CodeChecker#30 , otherwise the codechecker will break.

@sawenzel : ping

### DIFF
--- a/clang.sh
+++ b/clang.sh
@@ -1,7 +1,7 @@
 package: Clang
-version: "7.0.0"
-tag: v7.0.0-alice1
-source: https://github.com/alisw/clang
+version: "v9.0.0"
+tag: "llvmorg-9.0.0"
+source: https://github.com/llvm/llvm-project
 requires:
  - Python
  - "GCC-Toolchain:(?!osx)"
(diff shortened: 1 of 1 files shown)


sawenzel:
@davidrohr : Before merging, can we also check if the codechecker is running correctly on MacOS? (This is probably not tested in the CI)

davidrohr:
Ok, will do

davidrohr:
@ktf @MichaelLettrich : Build fails again in macOS CI with empty logs (also here #1920)

davidrohr:
@sawenzel : I just tried this on MacOS.
The build works fine, also for the O2CodeChecker.
But then when o2checkcode runs, it has the same problem finding the C headers as I had for clang.

For clang, the fix was to provide the -DDEFAULT_SYSROOT cmake option. I think we have to implement something similar for the codechecker, if we want to fix this.

In any case, I think this is a new independent issue, which has nothing to do with this PR.
I'd wait for the MacOS CI to rerun (no idea why it failed) and then merge this together with AliceO2Group/O2CodeChecker#30

PR #1920, opened 2019-12-04
llvm: find C SYSROOT on macOS in a more robust (?) way

ktf:
The problem is that unless you do `xcode-select --install`, ` /Library/Developer/CommandLineTools/` is not there and one needs to look up headers in the `XCode.app` application folder. This should allow the recipe to work with all the install combinations of `Xcode installed` `DeveloperTools installed`.

### DIFF
--- a/llvm.sh
+++ b/llvm.sh
@@ -18,7 +18,10 @@ unset LDFLAGS
 
 case $ARCHITECTURE in 
   # Needed to have the C headers
-  osx*) DEFAULT_SYSROOT=/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk ;;
+  osx*) 
+    XCODE_PATH=`xcode-select -p`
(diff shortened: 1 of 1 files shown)


ktf:
@davidrohr can you check if this works better?

davidrohr:
@ktf : For me the command returns
```
/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include
```
Shouldn't it be?
```
/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk
```

davidrohr:
@ktf : /Library/Developer/CommandLineTools/SDKs/MacOSX.sdk works for me, I guess you need to strip the /usr/include part from your path.

davidrohr:
Indeed, the macOS CI here failed exactly for this reason.

PR #1922, opened 2019-12-04
Require OpenMP and make it a dependency of O2

sawenzel:
OpenMP is a standard for easy multithreading.
It is already used in places in O2 and an upcoming
development will use it in simulation/digitization code.
This is just formalizing the requirement.

As compilers on Linux ship OpenMP in any case, this
only influences the MacOS builds.


davidrohr:
I think we should add -fopenmp to the compiler options for the system check. Some compilers will hide omp.h if the option is absent.

ktf:
We should also add it to https://github.com/alisw/homebrew-system-deps

sawenzel:
@davidrohr : Done.

sawenzel:
@ktf: https://github.com/alisw/homebrew-system-deps/pull/2

sawenzel:
@MichaelLettrich : We need to install libomp on the MacOS CI machines. Could you take care of that (unless it is happening automatically from the homebrew-system-deps configuration)?

sawenzel:
@MichaelLettrich : Any news here?

sawenzel:
Could we go on here?

ktf:
I see on mac there is still an error. Is that understood?

ktf:
@MichaelLettrich do you understand why libomp is not there, yet?

ktf:
@MichaelLettrich can you please have a look at this?

MichaelLettrich:
@sawenzel @ktf 

aliBuild fails in the doctor phase with the following message:
```bash
ERROR: Package openmp is a system requirement and cannot be found.
ERROR: This is due to the fact that the following script fails:
ERROR: 
ERROR: printf "#include <omp.h>\n" | c++ -xc++ - -fopenmp -c -o /dev/null
ERROR: with the following output:
ERROR: 
ERROR: openmp: clang: error: unsupported option '-fopenmp'
ERROR: openmp: 
ERROR: Please install openmp (libomp) on your system:
ERROR:   * On MacOS systems: brew install libomp
ERROR:   * On Linux it comes with the compiler
ERROR: 
ERROR: 
```
`brew` reports that `libomp` is installed. 
```bash
brew install libomp
Warning: libomp 10.0.1 is already installed and up-to-date
To reinstall 10.0.1, run `brew reinstall libomp`
```
A quick google search yields that this will be a more involved operation with the Apple `clang`. 
https://iscinumpy.gitlab.io/post/omp-on-high-sierra/

ktf:
I modified this to make sure we do not build openmp on mac, I guess that's the best option we have so far. If needed we will adapt the code which uses it.

PR #1923, opened 2019-12-04
Compile Geant4 with GDML support

sawenzel:
This enables Geant4 to import/export the standard
GDML file format for detector description.
Needed for geometry related developments and improvements
and in the interest of building a generic Geant4 distribution.

### DIFF
--- a/xercesc.sh
+++ b/xercesc.sh
@@ -1,0 +1,43 @@
+package: xercesc
+version: Xerces-C_3_2_2
+tag: Xerces-C_3_2_2
+source: https://github.com/apache/xerces-c
+build_requires:
+  - CMake
+prefer_system: ".*"
+prefer_system_check: |
+  pkg-config --atleast-version=3.2.0 xerces-c 2>&1 && printf "#include \"<xercesc/util/PlatformUtils.hpp>\"\nint main(){}" | c++ -xc - -o /dev/null
+prepend_path:
+  ROOT_INCLUDE_PATH: "$XERCESC_ROOT/include"
+  LD_LIBRARY_PATH: "$XERCESC_ROOT/lib"
+---
+
+cmake $SOURCEDIR                                         \
+      -DCMAKE_INSTALL_PREFIX=$INSTALLROOT                \
+      -DBUILD_SHARED_LIBS=ON                             \
+      -DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE}             \
+      -DCMAKE_CXX_STANDARD=${CXXSTD}                     \
+      -Dnetwork:BOOL=OFF                                 \
+      -DCMAKE_INSTALL_LIBDIR=lib
+
+make ${JOBS:+-j $JOBS}
+make install
+
+# Modulefile
+MODULEDIR="$INSTALLROOT/etc/modulefiles"
+MODULEFILE="$MODULEDIR/$PKGNAME"
+mkdir -p "$MODULEDIR"
+cat > "$MODULEFILE" <<EoF
+#%Module1.0
+proc ModulesHelp { } {
+  global version
+  puts stderr "ALICE Modulefile for $PKGNAME $PKGVERSION-@@PKGREVISION@$PKGHASH@@"
+}
+set version $PKGVERSION-@@PKGREVISION@$PKGHASH@@
+module-whatis "ALICE Modulefile for $PKGNAME $PKGVERSION-@@PKGREVISION@$PKGHASH@@"
+# Dependencies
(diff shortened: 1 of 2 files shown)


xercesc.sh
```diff
@@ -0,0 +1,36 @@
...
+proc ModulesHelp { } {
+  global version
+  puts stderr "ALICE Modulefile for $PKGNAME $PKGVERSION-@@PKGREVISION@$PKGHASH@@"
+}
+set version $PKGVERSION-@@PKGREVISION@$PKGHASH@@
+module-whatis "ALICE Modulefile for $PKGNAME $PKGVERSION-@@PKGREVISION@$PKGHASH@@"
+# Dependencies
+module load BASE/1.0
+# Our environment
+setenv XERCESC_ROOT \$::env(BASEDIR)/$PKGNAME/\$version
```
ktf: Please use set, not `setenv`.

xercesc.sh
```diff
@@ -0,0 +1,36 @@
...
+prepend-path LD_LIBRARY_PATH \$::env(XERCESC_ROOT)/lib
```
ktf: `prepend-path LD_LIBRARY_PATH \$XERCESC_ROOT/lib`

xercesc.sh
```diff
@@ -0,0 +1,36 @@
...
+$([[ ${ARCHITECTURE:0:3} == osx ]] && echo "prepend-path DYLD_LIBRARY_PATH \$::env(XERCESC_ROOT)/lib")
```
ktf: `DYLD_LIBRARY_PATH` is not used anymore. Please drop this line.

xercesc.sh
sawenzel: ok (but same in arrow.sh)

xercesc.sh
sawenzel: ok (but same in arrow.sh)

ktf:
we need to fix arrow as well. the linter did not spot it because the arrow recipe was not changed since the introduction of the new check.

sawenzel:
I consider this PR ready to be merged if the tests go through. @MichaelLettrich, @ktf  : Can you take care of the merge, please.

ktf:
I will do it after today daily, since we are on friday and I want to have at least one build before the weekend.

sawenzel:
There is a persisting problem which I could not yet solve:
```
cling::DynamicLibraryManager::loadLibrary(): libxerces-c-3.2.so: cannot open shared object file: No such file or directory
Error in <TInterpreter::TCling::AutoLoad>: failure loading library libgeant4vmc.so for TG4RunConfiguration
In file included from input_line_9:1:
/mnt/mesos/sandbox/sandbox/O2/Detectors/gconfig/g4Config.C:61:31: error: unknown type name 'TG4RunConfiguration'
  auto runConfiguration = new TG4RunConfiguration("geomRoot", physicsSetup, "stepLimiter+specialCuts",
                             ^
```
which happens during unit testing. I assume that we are missing some environment variable during the build process. Any idea? I already defined LD_LIBRARY_PATH in the new xercesc module as well as alibuild header.

ihrivnac:
Hi @sawenzel ,
Could you try to disable building shared libraries by adding -DBUILD_SHARED_LIBS=OFF in cmake instructions? The build will then create a static library for which you don't need to update the dynamic path. 
I am using an older xerces-c version (3.1.1) which build the static library by default and I don't set an extra path to it anywhere  when using it with G4/Geant4 VMC stand-alone.

sawenzel:
@ihrivnac : Thanks for hint. Let's see if it goes through.

sawenzel:
@ihrivnac : New problems now (at least on CC7). Someone in the compile chain needs to link against libcurl because libG4persistency tries to access these symbols. I am afraid this probably needs patching CMakeFiles which is not immediately in my hands.
Should we try making the shared lib version work instead?

ihrivnac:
@sawenzel: According to Xerces-c documentation, the option -Dnetwork:BOOL=OFF can switch off this dependency. But when I tried this one on my Mac I got another undefined symbols when building Geant4 with this version of xerces-c, so I think it will be better to go back to shared library and try to resolve the path to the library.

sawenzel:
It should work now. The reason was that xercesc decided to install its libs in `lib64` on CC7 rather than `lib`. This is why it worked on ubuntu.

sawenzel:
The MacOS CI seems stuck. 18 hours and counting.

sawenzel:
This is finally green. @ktf , @pzhristov : Can it be merged?

PR #1913, opened 2019-11-28
(O2-932) Cleanup modulefiles

Barthelemy:
- `PACKAGE_VERSION` -> `PACKAGE_REVISION` in guards
- Removal of `$([[ ${ARCHITECTURE:0:3} == osx ]] && echo "prepend-path DYLD_LIBRARY_PATH \$AEGIS_ROOT/lib")`

PR #1924, opened 2019-12-05
Bump QCG to 1.6.7

graduta:


awegrzyn:
@MichaelLettrich maybe you could help out?

MichaelLettrich:
on it

MichaelLettrich:
hi it is the same problem as in #1913 . FairMQ failed to build because of git issues. Hopefully the problem should be fixed by now but the builder is rebuilding a lot of things now, so allow for a bit of delay.

awegrzyn:
All right, thanks !

ktf:
See last commit for linter issue.

PR #1928, opened 2019-12-09
Update DDS & FairMQ to ODC required versions

ktf:


ktf:
@MohammadAlTurany @dennisklein this fails when building DDS as shown below.

I remember seeing this already in the past, I think it mixes our own boost and the system one. Do you remember how to fix it?

```
-- Using BOOST Library dir: /usr/local/opt/boost/lib
-- prerequisite dirs: /build/ci_checks/alidist_checker_macos_o2_ci_1/sw/BUILD/51872b7017f0a3448506346d4f07c7abecfe436d/DDS/dds-user-defaults;/build/ci_checks/alidist_checker_macos_o2_ci_1/sw/BUILD/51872b7017f0a3448506346d4f07c7abecfe436d/DDS/dds-protocol-lib;/build/ci_checks/alidist_checker_macos_o2_ci_1/sw/BUILD/51872b7017f0a3448506346d4f07c7abecfe436d/DDS/dds-intercom-lib;/build/ci_checks/alidist_checker_macos_o2_ci_1/sw/BUILD/51872b7017f0a3448506346d4f07c7abecfe436d/DDS/dds-topology-lib;/build/ci_checks/alidist_checker_macos_o2_ci_1/sw/BUILD/51872b7017f0a3448506346d4f07c7abecfe436d/DDS/MiscCommon/dds_ncf;/usr/local/opt/boost/lib;/build/ci_checks/alidist_checker_macos_o2_ci_1/sw/osx_x86-64/CMake/v3.13.1-2/lib;/build/ci_checks/alidist_checker_macos_o2_ci_1/sw/osx_x86-64/defaults-release/v1-2/lib;/build/ci_checks/alidist_checker_macos_o2_ci_1/python_local/lib;/build/ci_checks/alidist_checker_macos_o2_ci_1/python_local/lib;
[ 86%] Linking CXX executable dds-session
[ 86%] Built target dds-session
WN PKG prerequisite='/usr/local/opt/boost/lib/libboost_atomic-mt.dylib'
WN PKG prerequisite='/usr/local/opt/boost/lib/libboost_chrono-mt.dylib'
WN PKG prerequisite='/usr/local/opt/boost/lib/libboost_date_time-mt.dylib'
WN PKG prerequisite='/usr/local/opt/boost/lib/libboost_filesystem-mt.dylib'
WN PKG prerequisite='/usr/local/opt/boost/lib/libboost_log-mt.dylib'
WN PKG prerequisite='/usr/local/opt/boost/lib/libboost_log_setup-mt.dylib'
WN PKG prerequisite='/usr/local/opt/boost/lib/libboost_program_options-mt.dylib'
WN PKG prerequisite='/usr/local/opt/boost/lib/libboost_regex-mt.dylib'
WN PKG prerequisite='/usr/local/opt/boost/lib/libboost_system-mt.dylib'
WN PKG prerequisite='/usr/local/opt/boost/lib/libboost_thread-mt.dylib'
-- 
warning: cannot resolve item 'libicudata.64.dylib'

  possible problems:
    need more directories?
    need to use InstallRequiredSystemLibraries?
    run in install tree instead of build tree?

WN PKG prerequisite='libicudata.64.dylib'
CMake Error at /build/ci_checks/alidist_checker_macos_o2_ci_1/sw/SOURCES/DDS/2.5-odc/2.5-odc/cmake/modules/DDS_CollectPrerequisites.cmake:33 (file):
  file COPY cannot find
  "/build/ci_checks/alidist_checker_macos_o2_ci_1/sw/BUILD/51872b7017f0a3448506346d4f07c7abecfe436d/DDS/libicudata.64.dylib".


make[2]: *** [CMakeFiles/wn_bin] Error 1
make[1]: *** [CMakeFiles/wn_bin.dir/all] Error 2
```

dennisklein:
Was this the case they introduced the environment variable `DDS_LD_LIBRARY_PATH` for?! Not sure, maybe I am confusing things.

@AnarManafov @AndreyLebedev Can you comment on this?

AnarManafov:
@dennisklein is right. When building DDS worker packages we use cmake's methods to enumerate all dependencies. If you use a custom compiler (for example) or some lib paths are not properly propagated to cmake, then you can use DDS_LD_LIBRARY_PATH to inject additional look up directories.

ktf:
Boost and icu come from the system, so I am surprised this does not happen already. Also, it used to be that only `make wn_bin` required that, but now it seems to be an issue even with just `make install` is that expected?

ktf:
There seems to be also a compilation issue on linux:

```
[ 28%] Building CXX object fairmq/plugins/DDS/CMakeFiles/FairMQPlugin_dds.dir/DDS.cxx.o
In file included from /mnt/mesos/sandbox/sandbox/sw/SOURCES/FairMQ/v1.4.10/v1.4.10/fairmq/plugins/DDS/DDS.cxx:9:0:
/mnt/mesos/sandbox/sandbox/sw/SOURCES/FairMQ/v1.4.10/v1.4.10/fairmq/plugins/DDS/DDS.h:16:10: fatal error: DDS/dds_env_prop.h: No such file or directory
 #include <DDS/dds_env_prop.h>
          ^~~~~~~~~~~~~~~~~~~~
compilation terminated.
```

does it ring a bell?

AnarManafov:
> Boost and icu come from the system, so I am surprised this does not happen already. 
Can be, but cmake doesn't see it. Honestly I don't know else how to resolve this. This is why we introduced DDS_LD_LIBRARY_PATH to help user to inject additional search path. Just add the location of libicudata.64.dylib to DDS_LD_LIBRARY_PATH before before issuing the cmake command. 

> Also, it used to be that only `make wn_bin` required that, but now it seems to be an issue even with just `make install` is that expected?

Yes, this is expected. We learned, that users mostly build dds from source, rather than using a fixes released versions (where WN packages are build by us). Therefore we have added wn_bin to the ALL target as users need to build it in anyway.

Regarding DDS/dds_env_prop.h and the compilation problem.
I thought you supposed to use a 2.5-odc tag instead of DDS 3.0 (master)?
DDS 3.0 has its API headers adjusted to a common pattern to improve user expirience. It's reflected in the release notes as a breaking change.
FairMQ v1.4.10 is still using the old naming for DDS API.

ktf:
Regarding the WN, please establish back the old behaviour, we do not want to build WN packages on mac since at most we will use it to run on the same node. I've no control on where the installation of libicu will be and so I cannot fix it using DDS_LD_LIBRARY_PATH. 

Regarding the error, I can revert to 2.5.0-odc, although the request I had was "use latest version possible" that's why I moved to 3.0.0. I will revert to `2.5.0-odc`.

AnarManafov:
> Regarding the WN, please establish back the old behaviour, we do not want to build WN packages on mac since at most we will use it to run on the same node. I've no control on where the installation of libicu will be and so I cannot fix it using DDS_LD_LIBRARY_PATH.

WN bin is required for all plug-ins, even for the localhost plug-in one. The concept is the same for all types of plug-ins.
I can, of course, remove wn_bin target from ALL. But it won't solve the problem. Instead, it will be delegated to end-user.

> Regarding the error, I can revert to 2.5.0-odc, although the request I had was "use latest version possible" that's why I moved to 3.0.0. I will revert to `2.5.0-odc`.
The request is actually correct. With DDS it's preferable to use the most recent version possible. And usually you won't get problems as we don't change API often. The case here is that a tagged FairMQ v1.4.10 doesn't yet know about DDS 3.0 breaking changes as it was tagged one week before the DDS release.

AnarManafov:
@ktf , sorry for the confusion. Actually the localhost plug-in uses the lightweight WN package and thus doesn't need wn_bin target. I somehow forgot that we implemented the lightweight packaging feature some time ago and that the localhost plug-in benefits from it.

You are right, it make sense to remove wn_bin from the ALL target. It will be done in the master right now.

ktf:
Ok. Thank you for the clarification. I will wait for a new tag of FairMQ / DDS and in the meanwhile I will simply try with `1.4.10` + `2.5-odc`. @MohammadAlTurany is that fine for you?

AnarManafov:
The wn_bin is removed from ALL in the master now.
If FairMQ can make a couple of adjustments of names of DDS API headers they use, I can make a new production tag.

MohammadAlTurany:
@ktf Yes, it is fine with me, @dennisklein can you please create a new tag which support DDS 3.0

dennisklein:
FairMQ [`v1.4.11`](https://github.com/FairRootGroup/FairMQ/tree/v1.4.11) now supports and *requires* DDS `3+`.

AnarManafov:
@ktf , you can enable now dds 3.0 for mac os. I've re-tagged 3.0 on the master to include the minor change regarding wn_bin (it's removed from the ALL target).

PR #1929, opened 2019-12-09
Use new configure flags for GMP and MPFR for fastjet versions >= 3.3.3

marcovanleeuwen:
See above

This is to fix an issue with libraries in the fastjet version used by Rivet.


ktf:
Thanks. I took the liberty to do some cleanups for the upcoming modules4 migration.

PR #1942, opened 2019-12-16
Don't reset XRootD location in ROOT recipe

Atlantic777:
Leftovers from migrating XRootD into a standalone package.


ktf:
Fast tracking to avoid rebuilding root too many times in the CI.

PR #1944, opened 2019-12-16
GLFW: Fix modulefile creation

ktf:


ktf:
Current setup breaks CVMFS deployments. Merging.

PR #1947, opened 2019-12-16
DebugGUI: fix issue when GLFW is not present

ktf:


ktf:
Tested locally on mac. Merging since it fixes an actual issue on linux when GLFW is not installed.

PR #1950, opened 2019-12-17
[WIP] Remove ApMon-CPP dependency from Monitoring recipe

awegrzyn:


awegrzyn:
@ktf Can I ask you to do usual fixes?

Barthelemy:
I have made the changes , let's see if it pleases the linter

ktf:
How do you plan to support Monalisa in Monitoring?

monitoring.sh
```diff
@@ -4,7 +4,6 @@ tag: v2.6.3
 requires:
   - boost
   - "GCC-Toolchain:(?!osx)"
-  - ApMon-CPP
```
ktf: Maybe we can start by dropping it in the `o2-dataflow` default?

awegrzyn:
Indeed, let's wait for WP8 meeting to discuss it

ktf:
For the record, @Atlantic777 is doing some fix in #1381 which should allow supporting newer Redhat flavoured distributions.

PR #1954, opened 2019-12-18
Backport ROOT Module file fix to AliRoot 52 patches

Atlantic777:
Backporting the ROOT Module file XRootD loading fix.


ktf:
Before merging this we need to change the dailies to use a tag rather than a branch.

Atlantic777:
@ktf this is still relevant, the 53-patches and master branch are already fixed, but this should be backported to 52 branch, too.

PR #1955, opened 2019-12-18
Bump libwebsockets version to v3.0.1

Atlantic777:
Users confirmed that upgrading to libwebsockets v3.0.1fixes issues with linking to OpenSSL on some MacOS machines. Affects only JAliEn builds.


jolopezl:
+1 for this PR, it helps building O2 which was blocking my developments.

ktf:
Was already tested. Merging.

PR #1960, opened 2019-12-19
Build AliPhysics/JAliEn with ROOT v6-16-00

Atlantic777:
This is to realign AliRoot/JAliEn with the AliRoot 52a patches for testing on the Grid and comparing the results.

### DIFF
--- a/defaults-jalien.sh
+++ b/defaults-jalien.sh
@@ -49,6 +49,8 @@ overrides:
       - GEANT4_VMC
       - Vc
       - JAliEn-ROOT
+  ROOT:
+    version: "%(tag_basename)s_XRD4"
 ---
 # This file is included in any build recipe and it's only used to set
 # environment variables. Which file to actually include can be defined by the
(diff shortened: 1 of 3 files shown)


Atlantic777:
@ktf ROOT builds with --defaults jalien are built with XRootD 4, while the regular ROOT builds come with XRootD. However, we don't add any suffix to such ROOT versions.

Do you think we should add `version: "%(tag_basename)s_JALIEN"` like we did for AliPhysics and AliRoot?

ktf:
Yes, please.

Atlantic777:
Added "_XRD4" suffix to ROOT packages where needed.

Atlantic777:
@ktf it looks good to me but please double check. It looks like this:

```
Using cached build for CMake
Using cached build for AliEn-Runtime
Using cached build for XRootD

==> Building ROOT@v6-18-04_XRD4
```

Atlantic777:
@ktf this is outdated, ROOT 6-16-xx is no longer relevant, we have JAliEn daily builds and ROOT 6 `_XRD4` version suffix is not needed because soon all ROOT 6 builds will be JAliEn + XRootD 4. I suggest closing this PR without merging.

PR #1963, opened 2019-12-19
Remove the .la file from ApMon-CPP

Atlantic777:
As discussed with @ktf .

### DIFF
--- a/apmon-cpp.sh
+++ b/apmon-cpp.sh
@@ -22,7 +22,7 @@ fi
 make ${JOBS:+-j$JOBS}
 make install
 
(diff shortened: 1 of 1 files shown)


apmon-cpp.sh
```diff
@@ -22,6 +22,8 @@ fi
 make ${JOBS:+-j$JOBS}
 make install
 
+find $INSTALLROOT -name '*.la' | xargs rm -v
```
ktf: -delete

ktf:
Tested on my own mac. Merging.

PR #1966, opened 2020-01-01
LLVM: attempt at better detection of macOS SDK

ktf:


ktf:
@davidrohr can you check if this behaves better for you? In principle I think from Catalina on:

DEFAULT_SYSROOT=`xcrun --show-sdk-path`

is sufficient / the correct thing to do, I leave the rest, at least for now, to make sure we do not break older platforms.

davidrohr:
@ktf : Thx, that works for me. Could you apply the same patch also to clang.sh, which suffers from the same problem.

Happy new year
David

ktf:
@davidrohr #1967 should do the job.

PR #1969, opened 2020-01-07
Embed own XercesC in the Geant4 installation [O2-1004]

ktf:
This is needed because Geant4 will otherwise put an unrelocatable full path in some of the cmake macros / helpers, breaking builds which do not happen in the exact same location as the initial cached binary.

### DIFF
--- a/geant4.sh
+++ b/geant4.sh
@@ -4,8 +4,8 @@ tag: "v10.4.2"
 source: https://gitlab.cern.ch/geant4/geant4.git
 requires:
   - "GCC-Toolchain:(?!osx)"
-  - xercesc
 build_requires:
+  - xercesc
   - CMake
   - "Xcode:(osx.*)"
 prepend_path:
@@ -33,6 +33,12 @@ env:
 # if this variable is not defined default it to OFF
 : ${GEANT4_BUILD_MULTITHREADED:=OFF}
 
+# We embed Xerces-C directly into GEANT4 to avoid relocation issues and
+# incompatibility with other possible instances of XercesC
+if [ ! "X$XERCESC_ROOT" = X ]; then
+  rsync -av $XERCESC_ROOT/ $INSTALLROOT/
+fi
+
 cmake $SOURCEDIR                                             \
   -DGEANT4_INSTALL_DATA_TIMEOUT=2000                         \
   -DCMAKE_CXX_FLAGS="-fPIC"                                  \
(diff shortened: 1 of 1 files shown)


geant4.sh
```diff
@@ -33,6 +33,12 @@ env:
 # if this variable is not defined default it to OFF
 : ${GEANT4_BUILD_MULTITHREADED:=OFF}
 
+# We embed Xerces-C directly into GEANT4 to avoid relocation issues and
+# incompatibility with other possible instances of XercesC
+if [ ! "X$XERCES_ROOT" = X ]; then
+  rsync -av $XERCES_ROOT/ $INSTALL_ROOT/
+fi
```
sawenzel: I don't understand why this is necessary. It seems like another hack. Can't we use XercesC just as any other package/library? (Your previous fix was much shorter)

ktf: The previous commit does not actually fully work, given CMake does not expand environment variables. It was only ok for `geant4-config`.

The problem is that Geant4 embeds the absolute path to the Xerces installation in some of its scripts, which then cannot be automatically relocated (because the Geant4 root is different from the Xerces root). This is actually the only package which does it, i.e. the only one which pretends to know where its dependencies are installed, rather than delegating it  to the dependency itself. A proper solution would be to generate a CMake config file for XercesC and use that rather than `geant4-config` or the Cmake hardcoded value. An alternative would be to patch by hand all the occurrences of XercesC in Geant4 and use a relative path.

This solution, i.e. bundling xerces inside geant4, (once I managed to get it to work) is probably the cleanest and will also have the advantage to avoid compatibility issues.

sawenzel: Ok. But can't we directly refer to the final XercesC installation (after relocation) when building Geant4? Why did it work on the CI?

ktf:
You can refer to `XERCESC_ROOT` in the build recipe of geant4, and it will work to build geant4 itself. What one cannot do is to cache the relocated absolute path of `XERCES_ROOT` in `geant4-config` and use it in O2. Maybe a better solution is actually to make sure that O2 does not use `Geant4` to understand where XercesC is located, but `XercesC` itself. Let me check.

The CI did not catch it because the absolute path in `geant4-config` will be valid as long as the build path of xerces is the same as the one cached in the geant4 binary tarball. In order to catch this we should build always in a pristine folder, but that means we need to reinstall all the cached binaries every time, meaning a few extra minutes vs a few extra seconds.

ihrivnac:
In Geant4 10.6 there is added a cmake option GEANT4_INSTALL_PACKAGE_CACHE which should allow to disable the external dependencies path embedding, see more details at http://geant4-userdoc.web.cern.ch/geant4-userdoc/UsersGuides/InstallationGuide/html/installguide.html#advanced-options).
So we can use the "hack" proposed by Giulio for the time being and try the new option when we move G4 10.6.

Barthelemy:
this is really getting annoying as we are forced to build with o2-dataflow which takes forever (no cache). When do you think to have a fix ?

ktf:
I discussed with @sawenzel. We came to the same conclusion of @ihrivnac so we go ahead with this. I fixed a typo.

ktf:
Separately I am working on making the caching available for mac and other builds, BTW.

Barthelemy:
> Separately I am working on making the caching available for mac and other builds, BTW.

🎉  that would be fantastic !

ktf:
This is not requested anymore, is it?

PR #1970, opened 2020-01-07
Recipe for ROOT (T)Object diff utility

sawenzel:


sawenzel:
@pzhristov : This will allow building ObjCmp as part of our toolchain.

ktf:
Can we please host it on `https://github.com/alisw` or `https://github.com/AliceO2Group`?

ktf:
Also we agreed in WP3 at some point we should use a tagged version for all our packages, to avoid unwanted rebuilds, so can you put an hash rather than "master"?

sawenzel:
I can host it somewhere when its stable enough. For now it should be good enough. Will provide a tagged version.

sawenzel:
Tagged version provided

ktf:
Thanks.

PR #1973, opened 2020-01-07
yoda: depend on Python rather than Python-modules

ktf:


maireiphc:
Hi Giulio,

  thanks for this commit.
I take note about this change.
I interpret it as YODA will now prefer to pick up the Python installation from the system. Is that correct ?

ktf:
That's unchanged behaviour, it still depends on the Python package and aliBuild might or might not pick up the system version, if deemed compatible. 

The difference is that since it was not depending on any of the modules provided by `Python-modules` we simplify the builds and reduce the number of things that have to be rebuilt when `Python-modules` changes.

maireiphc:
Ok. Clearer.
Then does it mean it will understand naturally which python version to be pick-up between version 2.7 and 3.6, i.e. discover what YODA needs ?
Or do we have specify it explicitly (l.10+11 with the preprend-path and pythonpath, I assume...)

ktf:
If you use `alienv`, that should have always been the case.

PR #1975, opened 2020-01-08
RapidJSON: make it build_requires only

ktf:
It does not actually have any runtime.

### DIFF
--- a/alo-o2.sh
+++ b/alo-o2.sh
@@ -70,10 +70,10 @@ module-whatis "ALICE Modulefile for $PKGNAME $PKGVERSION-@@PKGREVISION@$PKGHASH@
 # Dependencies
 module load BASE/1.0 O2/$O2_VERSION-$O2_REVISION
 # Our environment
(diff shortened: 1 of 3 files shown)


ktf:
@aphecetche do you actually need ALO_O2_ROOT as an environment variable at runtime?

alo-o2.sh
```diff
@@ -70,10 +70,10 @@ module-whatis "ALICE Modulefile for $PKGNAME $PKGVERSION-@@PKGREVISION@$PKGHASH@
 # Dependencies
 module load BASE/1.0 O2/$O2_VERSION-$O2_REVISION
 # Our environment
-setenv ALO_O2_ROOT \$::env(BASEDIR)/$PKGNAME/\$version
+set ALO_O2_ROOT \$::env(BASEDIR)/$PKGNAME/\$version
```
aphecetche: this variable is not used AFAIK, so it can indeed be removed.

PR #1976, opened 2020-01-08
Make JAlien the default

ktf:
* Switch to new XRootD version by default
* defaults-release.sh adapted to use the legacy alien / xrootd.
* Move aliroot to depend on JAlien


ktf:
@Atlantic777 @costing @pzhristov as discussed today.

PR #1980, opened 2020-01-09
root: Improve support for own libpng

ktf:


ktf:
This should also fix [O2-1013]

PR #1981, opened 2020-01-10
Disable testCheckWorkflow (O2-1014)

Barthelemy:


ktf:
Currently broken. Merging.

PR #1983, opened 2020-01-10
DDS: fix ovelaps with ROOT / BOOST [O2-1013]

ktf:


ktf:
Failed in O2 with an OoM error, but DDS compiled fine and so did FairMQ, so merging.

PR #1986, opened 2020-01-10
python: fix building on macOS

ktf:


ktf:
Building our own Python on macOS got broken at some point. This resurrects things.

PR #1987, opened 2020-01-10
geant4: disable GDML support

ktf:


ktf:
Until we find out a proper way to build it.

PR #1989, opened 2020-01-11
Update python.sh

ktf:


ktf:
Currently broken. Not sure how it could pass CI. Merging.

PR #1990, opened 2020-01-11
Drop DDS from O2 on macOS

ktf:
It will always come through FairMQ, which we will always depend on.


ktf:
Error unrelated, merging.

PR #1999, opened 2020-01-15
O2: add module dependency DebugGUI [O2-1011]

ktf:
Also keep the linter happy.


awegrzyn:
@ktf Can we merge this?

PR #2003, opened 2020-01-16
Print a warning if user's rootrc file could prevent loading the ROOT Grid plugins

Atlantic777:
Alibuild creates system.rootrc file that is required for loading ROOT Grid plugins. If a user has a rootrc file in his home directory that rewrites ROOT PluginPath variable, then the Grid plugins might not be loaded properly. We already had such issues so I'm adding this warning message to detect such situation early.

### DIFF
--- a/alien-root-legacy.sh
+++ b/alien-root-legacy.sh
@@ -55,6 +55,9 @@ proc ModulesHelp { } {
   global version
   puts stderr "ALICE Modulefile for $PKGNAME $PKGVERSION-@@PKGREVISION@$PKGHASH@@"
 }
+
+system test -e ~/.rootrc && grep -iq PluginPath ~/.rootrc && echo "WARNING: Your local rootrc config might overwrite ROOT TGrid plugin path"
+
 set version $PKGVERSION-@@PKGREVISION@$PKGHASH@@
 module-whatis "ALICE Modulefile for $PKGNAME $PKGVERSION-@@PKGREVISION@$PKGHASH@@"
 # Dependencies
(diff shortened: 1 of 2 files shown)


root.sh
```diff
@@ -181,6 +181,13 @@ JOBS=$((${JOBS:-1}*3/5))
 [[ $JOBS -gt 0 ]] || JOBS=1
 cmake --build . --target install ${JOBS:+-- -j$JOBS}
 
+if test -f "$HOME/.rootrc" && $(grep -q PluginPath "$HOME/.rootrc"); then
```
ktf: No one reads those warnings and they will not be printed for people who download the binaries. So we need to find a better solution.

Atlantic777:
Maybe "exit 1" then?

ktf:
Again, it will not affect people using binaries (and in general it's a bad idea to kill compilation because tests fails).

Atlantic777:
@ktf here is a second take on this. Now the check is done whenever the JAliEn-ROOT/AliEn-ROOT-Legacy modulefile is loaded.

ktf:
This will exit with 1 if the file is not there. Are you sure that will not break modulefile sourcing?

ktf:
I think this needs to be handled in a different way. Even if we leave the test, nothing prevents the user from adding some breaking change after the build / loading of the environment. I think this is not something to have in the recipe / modulefile but it should be handled at runtime.

Atlantic777:
Ok, it's not that important anyways. Updating troublshooting docs and a standalone "jalien check setup" script would work, too. I think we can drop this PR.

PR #2009, opened 2020-01-17
Get Centos8 to work

ktf:


ktf:
This is required to build the package on Ubuntu and finally support binary distributions for it.

ktf:
@Atlantic777 looks like we still have some "AliEn-Runtime" induced mess. In particular o2-dataflow disables it so that zlib is not picked up from the correct place by libpng. Was this intended?

ktf:
yeah! Seems to work (autotools message spurious, somehow).

ktf:
Now we need to get autotools work on both production and centos 8 environment.

Atlantic777:
@ktf I don't think that was intended for AliEn-Runtime and lipng. 

Does it help if AliEn-Runtime stays enabled, and disable only some of it's dependencies, OpenSSL, ApMon or whatever was the reason to disable AliEn-Runtime for o2-dataflow?

Btw, shouldn't libpng pick up zlib from $ALIEN_RUNTIME_ROOT instead of from $ZLIB_ROOT?

ktf:
@pzhristov @awegrzyn do you understand the issue with python in ROOT for the `alidist-dataflow` test? See:

https://ali-ci.cern.ch/alice-build-logs/alisw/alidist/2009/7be7c280bcebcdf993925b785d5b8a59a969d481/build_O2_alidist-dataflow/fullLog.txt

the issue on macOS is understood and already fixed by @pzhristov in O2.

ktf:
@pzhristov https://github.com/alisw/alidist/pull/2009/commits/692b514e7be05899759eb6b0846822549ec7c16b seems indeed to be the problem. CMake is case sensitive...

ktf:
Error unrelated. Merging.

PR #2011, opened 2020-01-21
Fix OpenSSL loading in ROOT recipe

Atlantic777:
Minor fix with loading the OpenSSL in the ROOT recipe. in ROOT recipe we get OpenSSL from AliEn-Runtime. To do thet, we unset OpenSSL version variable, but this should be OPENSSL_REVISION, instead of OPENSSL_VERSION.


ktf:
Currently broken. Merging.

PR #2014, opened 2020-01-22
Bump AliECS and Go

teo:


teo:
Pushed AliECS v0.12.1 as v0.12.0 had a broken binary

PR #2016, opened 2020-01-22
Fix AliPhysics loading

fcatalan92:
After #1913 I got an error trying to load AliPhysics:
```
AliPhysics/latest-master-user-next-root6(54):ERROR:102: Tcl command execution failed: prepend-path PATH $ALICE_PHYSICS/bin
```
This PR should fix the issue.


ktf:
Currently broken. Merging.

PR #2017, opened 2020-01-22
Delete defaults-pwgmmtest.sh

ktf:


ktf:
Do we still need this? @maireiphc

PR #2021, opened 2020-01-23
Bump Readout to v1.3.0 and ReadoutCard to v0.16.0

kostorr:


sy-c:
I have pushed a new version which includes a change for the osx test which was complaining (a minor warning for a switch statement syntax, which was perfectly good).

This was actually building fine on the lab osx machine we have (and already the initial version 1.3.0), so it might take many iterations if I can not reproduce the issues, especially if it complains on issues one by one. I think there are still major improvements to do on this time-consuming  #@!#% test system.

ktf:
@sy-c I sympathise with the frustration with the test system, however:

* The reason why you get errors for warnings comes from #1754, which I guess was agreed at some point in WP3.
* There seems to be a bug either in cmake or in `Readout/CmakeLists.txt` which breaks compilation completely when `-Werror` is introduced.

kostorr:
@ktf Why can't I merge?

PR #2025, opened 2020-01-23
Fix LLVM symbol collision in ROOT and use only a single LLVM

ktf:


root.sh
```diff
@@ -103,8 +103,74 @@ else
 fi
 
 unset DYLD_LIBRARY_PATH
+
+mkdir -p ./src_tmp
```
ktf: This means that  ROOT development package will stop working. Can we merge the patch in our own fork?

davidrohr: I was copy and pasting how other recipes were modifying the source code. Is there a better way?
What do you mean by "merge the patch in our own fork"?

ktf: Yeah, but for ROOT we should keep the development packages functional. This will break things because the rsync decouples the local checkout from the one used for building. Can you do a PR on top of https://github.com/alisw/root/tree/alice/v6-18-04?

davidrohr: It requires 2 patches, the first is in master but no in 6.18.4-patches, and the second is not even merged in master. We can of course create an 6-18-04-alice tag in our fork, and use that. Then we can do away with the life-patching.
But I cannot write to the alisw fork. Or do you want me to do it in my own fork for the time being?

ktf: I can give you access, one sec.

ktf: can you try now? you should have write access to alisw/root

davidrohr: ok, worked, tag is there, I am creating a PR to alidist

davidrohr: I have updated #1918 with the new root recipe that does not do life patching. But we have to wait for the CI again.

ktf: Ok, thanks. I will merge it as soon as it passes.

davidrohr:
@ktf : why did you close this?

davidrohr:
ah, ok, this is the other PR. Mine is green now, can be merged.

davidrohr:
just figured that I have merge permissions meanwhile after a positive review, so just merged it myself.

ktf:
yes, this was my manual conflict resolution branch. thank you for merging.

PR #2027, opened 2020-01-27
arrow: add Clang to dependencies

ironMann:
Fixes linking when building from devel directory:

```
/.../bin/ld: warning: libLLVMCore.so.9, needed by /home/.../arrow/v0.14.1-1/lib/libgandiva.so.14, not found (try using -rpath or -rpath-link)
/.../bin/ld: warning: libLLVMMCJIT.so.9, needed by /home/.../arrow/v0.14.1-1/lib/libgandiva.so.14, not found (try using -rpath or -rpath-link)
/.../bin/ld: warning: libLLVMipo.so.9, needed by /home/.../arrow/v0.14.1-1/lib/libgandiva.so.14, not found (try using -rpath or -rpath-link)
```

### DIFF
--- a/arrow.sh
+++ b/arrow.sh
@@ -108,7 +108,8 @@ proc ModulesHelp { } {
 set version $PKGVERSION-@@PKGREVISION@$PKGHASH@@
 module-whatis "ALICE Modulefile for $PKGNAME $PKGVERSION-@@PKGREVISION@$PKGHASH@@"
 # Dependencies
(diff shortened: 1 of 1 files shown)


ktf:
The change is fine, however the modulefile is not used when building, so I guess there might be some other adjustment needed.

ironMann:
This happens only when building from the local directory, after `alienv enter DataDistribution/latest`. 
`aliBuild build ...` works fine.

PR #2031, opened 2020-01-27
O2-1074 Workaround for CMake in Readout

MichaelLettrich:
It seems that newer versions of CMake (>3.15) cause problems when building Readout on macOS with `-Werror` enabled. Lets drop `-Werror` on the macOS CI for now until it is fixed in CMake.

### DIFF
--- a/readout.sh
+++ b/readout.sh
@@ -32,7 +32,12 @@ esac
 
 # Enforce no warning code in the PR checker
 if [[ $ALIBUILD_O2_TESTS ]]; then
-  CXXFLAGS="${CXXFLAGS} -Werror -Wno-error=deprecated-declarations"
+  if [[ $ARCHITECTURE != osx ]] ; then
+    # there seems to be a bug in CMake with -Werror which adds unwanted 
+    # includes that lead to failing builds. skip it for now.
+    # https://alice.its.cern.ch/jira/browse/O2-1074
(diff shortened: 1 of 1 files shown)


readout.sh
```diff
@@ -32,7 +32,12 @@ esac
 
 # Enforce no warning code in the PR checker
 if [[ $ALIBUILD_O2_TESTS ]]; then
-  CXXFLAGS="${CXXFLAGS} -Werror -Wno-error=deprecated-declarations"
+  if [[ $ARCHITECTURE != osx ]] ; then
```
ktf: Did you actually test it? I think this will not work.

```
case $ARCHITECTURE in 
  osx*) ;;
  *) CXXFLAGS="${CXXFLAGS} -Werror -Wno-error=deprecated-declarations" ;;
esac
```

is probably what you want (which has also the advantage of being posix compliant).

MichaelLettrich: Oups, just on the linux commandline thb. That's what you get for being sloppy I guess 😅

ktf: To be clear, the issue is that the architecture name is `osx_x86-64`, not `osx`.

MichaelLettrich:
Seems to compile just fine on macOS now however 
```bash
The following tests FAILED:
	 56 - DataFormats/MemoryResources/test/testMemoryResources.cxx (Failed)
	 74 - Utilities/O2Device/test/test_O2Device.cxx (Failed)
```
Is this a general problem @ktf ?

PR #2032, opened 2020-01-29
arrow: Depend on Clang module

sawenzel:
This is needed as otherwise we have unresolved
libraries in O2 (when doing alienv enter O2)


sawenzel:
@mfasDa This should fix the error reported by you.

sawenzel:
Closing as duplicate of #2027 .

PR #2036, opened 2020-01-30
o2suite: add DataDistribution

ironMann:
jira #O2-936

### DIFF
--- a/o2suite.sh
+++ b/o2suite.sh
@@ -13,6 +13,7 @@ requires:
   - qcg
   - QualityControl
   - Run2ESDConverter
+  - DataDistribution
 valid_defaults:
   - o2
   - o2-dataflow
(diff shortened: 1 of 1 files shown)


ironMann:
Expected problems with MacOS ... 

@ktf is macos builder in Jenkis operational?

ktf:
You mean for the PR checker? It's not in jenkins. I think @MichaelLettrich was having more issues...

ironMann:
I was attempting to run https://alijenkins.cern.ch/job/build-any-ib/ on a osx_x86-64-large instance: 8355

ktf:
Mmm... I think the label is wrong. Did you make it up or did you select it from a combobox? I think the machine kind is called "FLP OSX" or something like that. @awegrzyn or @Barthelemy might know.

You can however exclude packages from a given architecture using the:

```
 - Package:<arch-matching-regex>
```

format as described [here](http://alisw.github.io/alibuild/reference.html#the-header).

In this particular case:

```
- DataDistribution:"(?!osx)"
```

just like we do for GCC a few lines above.

PR #2037, opened 2020-01-30
O2-1074 Workaround for -Werror cmake bug on macOS

MichaelLettrich:
Adding `-Werror` to cmake code under certain circumstances adds an additional 
```cmake
-isystem /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.15.sdk/usr/include 
```
which causes compilation to fail with the compiler unable to find cmath functions such as `std::isnan()`
A bugreport exists and I have been able to reproduce this bug outside of our build environment. See [Kitware gitlab](https://gitlab.kitware.com/cmake/cmake/issues/19845#note_689595)

As long as this error persists we need to exclude `-Werror` on macOS.

### DIFF
--- a/alf.sh
+++ b/alf.sh
@@ -18,7 +18,13 @@ incremental_recipe: |
 
 # Enforce no warning code in the PR checker
 if [[ $ALIBUILD_O2_TESTS ]]; then
-  CXXFLAGS="${CXXFLAGS} -Werror -Wno-error=deprecated-declarations"
+  # there seems to be a bug in CMake in macOS with -Werror which adds unwanted 
+  # includes that lead to failing builds. skip it for now.
+  # https://alice.its.cern.ch/jira/browse/O2-1074
+  case $ARCHITECTURE in 
+    osx*) ;;
+    *) CXXFLAGS="${CXXFLAGS} -Werror -Wno-error=deprecated-declarations" ;;
+  esac
 fi
 
 cmake $SOURCEDIR                                                      \
--- a/alf.sh
+++ b/alf.sh
@@ -16,17 +16,6 @@ incremental_recipe: |
 ---
 #!/bin/bash -ex
 
-# Enforce no warning code in the PR checker
-if [[ $ALIBUILD_O2_TESTS ]]; then
-  # there seems to be a bug in CMake in macOS with -Werror which adds unwanted 
-  # includes that lead to failing builds. skip it for now.
-  # https://alice.its.cern.ch/jira/browse/O2-1074
-  case $ARCHITECTURE in 
-    osx*) ;;
-    *) CXXFLAGS="${CXXFLAGS} -Werror -Wno-error=deprecated-declarations" ;;
-  esac
-fi
-
 cmake $SOURCEDIR                                                      \
       -DCMAKE_INSTALL_PREFIX=$INSTALLROOT                             \
       ${COMMON_O2_REVISION:+-DCommon_ROOT=$COMMON_O2_ROOT}             \
--- a/infologger.sh
+++ b/infologger.sh
@@ -22,7 +22,13 @@ esac
 
 # Enforce no warning code in the PR checker
 if [[ $ALIBUILD_O2_TESTS ]]; then
-  CXXFLAGS="${CXXFLAGS} -Werror -Wno-error=deprecated-declarations"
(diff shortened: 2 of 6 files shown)


davidrohr:
@MichaelLettrich : You removed the `fi` in monitoring.sh, therefore the CI is failing.
Could you fix this?

MichaelLettrich:
Done, thanks for noticing.

davidrohr:
@ktf : Can we merge this

ktf:
yes, i was waiting for the daily. will do once i am on a computer

ktf:
I changed it so that we drop the hack altogether. We should fix the `CMakeLists.txt` not introduce hacks in alidist.

MichaelLettrich:
For consistency this then applies to readout.sh as well

davidrohr:
I am not 100% sure about this. It might get us code with warnings pushed into o2, which we have to clean up later.

MichaelLettrich:
I don't like the idea either - this should not be merged without adapting the CMakelists of the packages before, which should not be done without a discussion with the codeowners...

davidrohr:
I would not change the CMakeLists. This is a temporary workaround that shall be reverted later. So better have it in one place.

ktf:
no, this is not a work around. CI behaving differently from anyone else is a mistake and should be avoided.

davidrohr:
I do not agree here. The reason the CI is different is that everyone has a different setup, so enabling -Werror everywhere, will lead to many compile failures for the user.
Therefore, we had decided to disable it for the users (if I remember correctly).
But we have it in the CI to enforce a no-warning build at least under our default setup.
Personally, I would leave it like this.

MichaelLettrich:
@davidrohr  I do not think that @ktf  wants to scrap `-Werror` for the CI alltogether but somehow move it into the CMakeLists.txt (e.g. via a debug flag) but not for a general case.
I am not fully convinced by this approach, as this represents a shift in paradigm which, if executed cleanly in my oppinion would also mean, that we would need to internalize other parts of the receipes as well e.g. building for the o2 gpu container.

To me the CMakeLists that should give us the blueprint on how to build a specific software package ( comparable to a class definition) however the buildsystem (i.e. alibuild) has the responsibility to configure it  for the different platforms we are building for (i.e. instanciate a specific object and make sure its in a valid state). 

Anyhow, this should be an interresting topic for morning coffee :)

davidrohr:
In general, I like the idea of setting CXX_FLAGS in the CMakeLists, not in alidist. But right now, it is actually globally set in the defaults.
And in any case, as explained above, we should enable the -Werror only for CI builds, not for users, otherwise the code that passes the CI will not compile for users with a different compiler, which is a pain in the ass. What @ktf was mentioning, is that CI should not behave differently than anyone else. But I think that actually in the case the CI should behave differently.

davidrohr:
Btw, after digging a bit deeper, I am not convinced if the problem comes from CMake itself.
It appears only when one links to ZLIB, and I think the FindZLIB module (which in turn is shipped with CMake) creates the offending include...

davidrohr:
ok, I think I found the bug in CMake. It is stupid like hell.....

FindModule(zlib) finds zlib.h in [foo]/usr/include and adds that as system include path (which it shouldn't in the first place, but later CMake scane for .../usr/include system paths, and gets rid of that.

Now, since 3.15, they added a logic to not hard-remove /usr/include, but to query the compiler for the default include path, in case people like MacOS move it somewhere else.

For this, they call the compiler, passing the normal compile flags. But their compiler invocation triggers warnings, so when there is the -Werror compile flag, the compiler invocation fails, and CMake cannot find the [...]/usr/include path, which it filters out of its include paths.

Hence, if you link to zlib and use -Werror, you end up with the bogus path... Now I just need to find out how to fix this.

ktf:
Nice. Maybe what we should do is that we add `-Werror` programmatically in CMake to the `o2_add_executable` / `o2_add_library` / `o2_add_test` helpers, rather than at top level?

davidrohr:
I fixed this in CMake and created an 3.16.3-alice1 tag in my fork.
We should report this to CMake, indeed a bug on their end. But not tonight.
In the meantime one cannot use the brew CMake of course...
Test is running here #1814 . Let's see whether it succeeds.

ktf:
Closing this as not needed anymore.

PR #2038, opened 2020-01-30
Use Python-system on mac

ktf:


ktf:
Apparently we had a few places where Python-system was not used.

ktf:
It might, actually.

PR #2042, opened 2020-01-31
Use alisw/Cmake

ktf:
This should not actually trigger any rebuild.


ktf:
@davidrohr done. ;-)

davidrohr:
Thx, btw actually we should disallow system cmale on macos for now, since brew doesn't have the fix.

Kind Regards
David Rohr

PR #2044, opened 2020-02-02
CMake 3.16 fix for CUDA compilation

davidrohr:
This fixes the problem in the O2 GPU CI like in AliceO2Group/AliceO2#2876

@ktf : Can you move the tag over to alisw/CMake (or grant me write access) and change the git URL here accordingly. Opening this PR as is in order for the CI to test it.


ktf:
Moved to alisw. There is now a branch `alice/v3.16.3` which you can use to propose more changes, hopefully not needed.

PR #2049, opened 2020-02-04
Fix typo in jalien-root finding libwebsockets

raymondEhlers:
The brew package name is `libwebsockets` (was missing the "s").

Note that this seems to then show an additional issue which hasn't shown up until now because it hasn't be possible to take this from brew:

```
In file included from /Users/re239/alice/sw/BUILD/77319b21e45ce6835c0b8e705d5179abbe5f0b1d/JAliEn-ROOT/src/TJAlien.cxx:12:
In file included from /Users/re239/alice/sw/BUILD/77319b21e45ce6835c0b8e705d5179abbe5f0b1d/JAliEn-ROOT/inc/TJAlien.h:53:
In file included from /Users/re239/alice/sw/BUILD/77319b21e45ce6835c0b8e705d5179abbe5f0b1d/JAliEn-ROOT/inc/TJAlienConnectionManager.h:14:
/usr/local/opt/libwebsockets/include/libwebsockets.h:180:10: fatal error: 'uv.h' file not found
#include <uv.h>
         ^~~~~~
```

I think how I see how I could hack around this, but I'm wondering if there is an easier or better solution

### DIFF
--- a/jalien-root.sh
+++ b/jalien-root.sh
@@ -19,7 +19,7 @@ append_path:
 case $ARCHITECTURE in
   osx*) 
 	[[ ! $OPENSSL_ROOT ]] && OPENSSL_ROOT=$(brew --prefix openssl)
-	[[ ! $LIBWEBSOCKETS_ROOT ]] && LIBWEBSOCKETS_ROOT=$(brew --prefix libwebsocket)
+	[[ ! $LIBWEBSOCKETS_ROOT ]] && LIBWEBSOCKETS_ROOT=$(brew --prefix libwebsockets)
   ;;
 esac
 


ktf:
I guess we are effectively missing a dependency on libUV, then.

raymondEhlers:
I think so. From a quick glance, it looks like it can be built with or without libuv. It's just that the homebrew version is built with libuv

ktf:
IMHO, the correct solution is to setup LIBUV_ROOT=`brew --prefix libuv` when we are on mac and we know we are picking up libwebsockets from the system. Can you try that?

raymondEhlers:
I'm trying it out, but I'm not sure of the proper way of setting this in the build recipe. Looking at some other examples, I came up with [this](https://github.com/alisw/alidist/pull/2050), but it still fails to find the include because it looks like the CXX_FLAGS are overwritten in the JAliEn-ROOT CMakeLists. See [here](https://gitlab.cern.ch/jalien/jalien-root/blob/master/CMakeLists.txt#L16). I suppose this will need to be fixed to proceed

PR #2050, opened 2020-02-04
[WIP] Add libuv dependence for JAliEn-ROOT on macOS

raymondEhlers:
libwebsockets depends on libuv when installed via homebrew. However, this won't yet work because the CXX_FLAGS are overwritten in the [CMakeLists of JAliEn-ROOT](https://gitlab.cern.ch/jalien/jalien-root/blob/master/CMakeLists.txt#L16). Also, I suppose I'll probably need to update the LDFLAGS to cmake, but didn't even get there yet.

### DIFF
--- a/jalien-root.sh
+++ b/jalien-root.sh
@@ -20,14 +20,22 @@ case $ARCHITECTURE in
   osx*) 
 	[[ ! $OPENSSL_ROOT ]] && OPENSSL_ROOT=$(brew --prefix openssl)
 	[[ ! $LIBWEBSOCKETS_ROOT ]] && LIBWEBSOCKETS_ROOT=$(brew --prefix libwebsockets)
+	[[ ! $LIBUV_ROOT && $LIBWEBSOCKETS_ROOT ]] && LIBUV_ROOT=$(brew --prefix libuv)
   ;;
 esac
+
+if [[ "$LIBUV_ROOT" != '' ]]; then
+  export LDFLAGS="$LDFLAGS -L$LIBUV_ROOT/lib -luv"
+  export LD_LIBRARY_PATH="$LIBUV_ROOT/lib:$LD_LIBRARY_PATH"
+  export CXXFLAGS="-I$LIBUV_ROOT/include $CXXFLAGS"
+fi
 
 rsync -a --exclude '**/.git' --delete $SOURCEDIR/ $BUILDDIR
 rsync -a $ALICE_GRID_UTILS_ROOT/include/ $BUILDDIR/inc
 
 cmake $BUILDDIR                                          \
       -DCMAKE_INSTALL_PREFIX="$INSTALLROOT"              \
+      -DCMAKE_CXX_FLAGS="$CXXFLAGS"                      \
       -DROOTSYS="$ROOTSYS"                               \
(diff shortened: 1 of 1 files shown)


jalien-root.sh
```diff
@@ -20,14 +20,22 @@ case $ARCHITECTURE in
...
+	[[ ! $LIBUV_ROOT && $LIBWEBSOCKETS_ROOT ]] && LIBUV_ROOT=$(brew --prefix libuv)
   ;;
 esac
 
+if [[ "$LIBUV_ROOT" != '' ]]; then
+  export LDFLAGS="$LDFLAGS -L$LIBUV_ROOT/lib -luv"
+  export LD_LIBRARY_PATH="$LIBUV_ROOT/lib:$LD_LIBRARY_PATH"
+  export CXXFLAGS="-I$LIBUV_ROOT/include $CXXFLAGS"
+fi
+
```
ktf: Why do you need this? Can't cmake detect what to do from `LIBUV_ROOT`?

raymondEhlers: As I said, I was just following other recipes. Are you suggesting just adding `-DLIBUV="$LIBUV_ROOT"` after `-DLWS...`? If so, I just tried it, but cmake doesn't know what to do with it

ktf: Indeed this is probably something which should be taken into account by JAlien itself.. @Atlantic777 do you actually have the proper find_libuv statements in JAlien-ROOT?

Atlantic777: Adding @yuw726, he might have more info.

Atlantic777:
@raymondEhlers could you share some failing JAliEn-ROOT build logs? Thanks!

raymondEhlers:
Log file is here: [log.txt](https://github.com/alisw/alidist/files/4153929/log.txt). Note that there's some spurious note about LIBUV_ROOT because I was trying Giulio's suggestion, but it can be ignored - it has no effect anyway.

Atlantic777:
So, there are issues on several levels here. When we build libwebsockets from alidist, then we have all the external event loop libraries disabled and use only on the internal LWS event loop.

On the other hand, brew compiles LWS with libuv and libevent which would be fine, but they leak their compile time dependencies. This problem would be avoided if libwebsockets was using forward declaration to hide the libuv includes from the users, but they don't do that. 

We could add libuv/libevent discovery to JAliEn-ROOT CMake, but we can't use libwebsockets 3.2 which comes from brew. The problem here is that our code base currently works with libwebsockets up to 3.0, while brew delivers 3.2 . We are updating our code-base so we can make the switch, but at the moment that we can't use 3.2.

I'm not sure what's the best choice here, maybe we could just add a version check to the libwebsockets alidist recipe and compile it ourselves everywhere. There are multiple problems to be sorted and I would resolve them one by one.

ktf:
i would start by requiring libwebsocket 3.0 in the prefer_system.

raymondEhlers:
@Atlantic777 @ktf Thanks for looking into this! I really need a working install right now, so I'll just have alibuild bring libwebsockets. In any case, it sounds like this PR can be closed because it will be addressed in a more comprehensive manner

PR #2051, opened 2020-02-04
JAliEn-ROOT improvements (O2-1118)

Atlantic777:
- JAliEn-ROOT currently requires libwebsocket version between 2.8 and 3.0, add strict check
- Fix missing GSHELL_ROOT variable in xalienfs (AliEn-ROOT-Legacy)
- Fix XRootD / libxml2 dependency
- bump xjalienfs (JAliEn-ROOT): add alien_mirro and improve alien_cp

### DIFF
--- a/libwebsockets.sh
+++ b/libwebsockets.sh
@@ -9,7 +9,8 @@ build_requires:
   - zlib
 prefer_system: "osx"
 prefer_system_check: |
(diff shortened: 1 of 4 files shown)


ktf:
@Atlantic777 I tried to avoid a too long line for a check.

PR #2055, opened 2020-02-07
lzma: build as static library (O2-1127)

ktf:
Centos7 comes with an incompatible version, so we simply
make our own invisible to the system tools.

### DIFF
--- a/lzma.sh
+++ b/lzma.sh
@@ -9,13 +9,12 @@ prefer_system: "(?!slc5)"
 prefer_system_check: |
   printf "#include <lzma.h>\n" | c++ -xc++ - -c -M 2>&1
 ---
-#!/bin/bash -e
-
 rsync -a --delete --exclude '**/.git' --delete-excluded $SOURCEDIR/ ./
 ./autogen.sh
 ./configure CFLAGS="$CFLAGS -fPIC -Ofast" \
             --prefix="$INSTALLROOT"       \
-            --disable-static              \
+            --disable-shared              \
+            --enable-static               \
             --disable-nls                 \
             --disable-rpath               \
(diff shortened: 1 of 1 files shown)


ktf:
@maireiphc looks like something is wrong again with aligenerators. Do you have the same problem?

maireiphc:
@ktf 

  do you want me to test a compilation of the full AliGenerators under my alidock with the modification you introduced to lzma.sh, right ?
i.e. you want me to validate the fail of the build-check found here.
Or do I miss sthg in your question ?

ktf:
Looks like fastjet simply does not compile because it uses `std::auto_ptr`

maireiphc:
It is not the first time I see that in the compilation log of Rivet when fastjet is invoked.
So far, it has just been notes or warnings. It seems to me this is still the case here with [-Wdeprecated-declarations]

The problem seems to come from the compilation of a given CMS analysis (CMS_2013_I1223519) in Rivet.
(ctrl+F "Error" in the check build log)
The variable Et is ambiguous, there is apparently a namespace conflict.

It has been fixed by Rivet authors in later version of the master branch/ later tags
https://gitlab.com/hepcedar/rivet/-/commits/master/analyses/pluginCMS/CMS_2013_I1223519.cc
- https://gitlab.com/hepcedar/rivet/commit/0d5f91a4c8e66f5c6cbd7d231c73a46be9c7109a
- https://gitlab.com/hepcedar/rivet/commit/ced494d4016dde5dcd24567a276942fa0f64c446

Why it triggers sthg only now and not in earlier builds is not clear to me.

ktf:
@awegrzyn can you check if with this yum is back to working?

Issue #2045, opened 2020-02-03
rsync not required as system package


davidrohr:
Basic containers that don't have rsync installed failed in aliBuild, but this is not reported by aliDoctor

ktf:
well rsync is only needed when you use the rsync backend for the repository. Moving to S3 will remove the need for it.

davidrohr:
Well, but some of our alidist recipes use rsync. When you try to build O2 without rsync, the dependencies will fail.

ktf:
ah, right. good point. Then we need to add a recipe for it and add a build_requires.

ktf:
https://alice.its.cern.ch/jira/browse/O2-1110?filter=13341

ktf:
I've added the recipe. We should now add it to places which use it.

davidrohr:
Ok, I can do that later in order to share the work if you want.

ktf:
Thanks. Probably we should do it as we change the recipes for other reasons, so that we do not trigger a rebuild just for this.

davidrohr:
I thought about that as well, but then, will we always think of it? We can also do it once and let it rebuild over the weekend.
Or we create a dedicates pr for each package now, and then merge together when we merge something else.

ktf:
Given I just merged lzma which was breaking FLP installations, we might as well do this? Actually I just realised that all the recipe with a module file require that because that's how we copy them in place, so we might want to do it for a few base packages (e.g. zlib, lzma, bzip2) and let the transitive closure do the rest...

davidrohr:
+1

ktf:
Will consider this done for now. We might add the `build_requires` in a few other place as we get to change low level packages.

PR #2059, opened 2020-02-07
rsync: add system requirement

ktf:


ktf:
I've added this as a system requirement. If really needed we can make it a real recipe, but I am surprised it's not in the base system.

ktf:
We should add it to build_requires of anything which uses it.

ktf:
#2045

PR #2060, opened 2020-02-07
Add rsync as system dependency

ktf:


ktf:
This should take care already of a good fraction of the packages via transitive closure.

[foo.pdf](https://github.com/alisw/alidist/files/4172661/foo.pdf)

PR #2061, opened 2020-02-10
bump readout v1.3.6

sy-c:


kostorr:
Please don't merge this yet as we need an RPM of the previous Readout version linking to a (still unreleased) ReadoutCard version.

kostorr:
> Please don't merge this yet as we need an RPM of the previous Readout version linking to a (still unreleased) ReadoutCard version.

OK this can be merged now. Thank you.

PR #2063, opened 2020-02-10
datadist: Limit number of jobs for jenkins and bump the version

ironMann:
O2-1129
O2-1105


ktf:
Might work for jenkins, albeit we actually want to move away from it. Merging for now, will have to be cleaned up.

PR #2064, opened 2020-02-10
Bump geant4_vmc to v3-6-p6:

ihrivnac:
This version contains the latest fixes and also the development for mixing
physics lists via biasing.

### DIFF
--- a/geant4_vmc.sh
+++ b/geant4_vmc.sh
@@ -1,6 +1,6 @@
 package: GEANT4_VMC
 version: "%(tag_basename)s"
-tag: "v3-6-p3"
+tag: "v3-6-p6-inclxx-biasing-p2"
 source: https://github.com/vmc-project/geant4_vmc
 requires:
   - ROOT
--- a/geant4_vmc.sh
+++ b/geant4_vmc.sh
@@ -1,6 +1,6 @@
 package: GEANT4_VMC
 version: "%(tag_basename)s"
-tag: "v3-6-p6-inclxx-biasing-p2"
+tag: "v3-6-p6"
 source: https://github.com/vmc-project/geant4_vmc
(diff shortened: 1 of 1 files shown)


ktf:
@ihrivnac the build issue on mac is genuine. Can you look into it?

ihrivnac:
This is due to building with geant4_vmc with c++17 with Geant4 built with c++11; with adding the development for mixing Geant4 physics list ("inclxx-biasing-p2") we depend on G4 includes which we did not depend in previous version, what explains why we did not get this error before.

@sawenzel , do we need to stay with Geant4 10.4.px with O2? The most straightforward solution would be to move to Geant4 10.5 or 10.6, which both support c++17.

ihrivnac:
I changed the PR in WIP, as we need v3-6-p6-inclxx-biasing-p2 for production; and v3-6-p6 or newer version for O2. So let's wait with merge until we agree on the version for O2.

ktf:
Waiting for @sawenzel to greenlight.

ihrivnac:
@ktf , as I wrote above, this is not exactly what we need; I changed the tag to  v3-6-p6 to see if  it passes testing, but we we need the development in "inclxx-biasing". 
To get a different version for production than for o2, should I put back v3-6-p6-inclxx-biasing-p2  in geant4_vms.sh and v3-6-p6 in defaults-o2.sh  ?

ktf:
@ihrivnac yes, that's correct. However why not just using v3-6-p6-inclxx-biasing-p2 for o2 as well?

ihrivnac:
@ktf, v3-6-p6-inclxx-biasing-p2 does not compile with c++17, that's why I changed the tag. We would have to build it with c++11.

PR #2066, opened 2020-02-11
Bump gRPC to v1.27.1

teo:
Related: https://alice.its.cern.ch/jira/browse/O2-1125

### DIFF
--- a/grpc.sh
+++ b/grpc.sh
@@ -1,6 +1,6 @@
 package: grpc
 version: "%(tag_basename)s"
-tag:  v1.19.1
+tag:  v1.27.1
 requires:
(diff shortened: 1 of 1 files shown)


ktf:
Notice this will not be enough. You will need the fix proposed in AliceO2Group/Configuration#51 as well.

teo:
@ktf not sure I understand, I'm not using Travis with gRPC at all?

ktf:
Do you build on macOS with travis? The issue is only there.

teo:
@awegrzyn can you advise on this? Do we need to change the CMake of our packages that use gRPC?

PR #2074, opened 2020-02-12
gfortran-system: add recipe to check for system gfortran

ktf:


MichaelLettrich:
I think this is not needed as `gcc` is automatically installed when installing the prerequisites on macos

ktf:
we should still check, since people tend not to follow instructions.

ktf:
I mean, this can be used to verify via aliDoctor that people did follow instructions.

PR #2075, opened 2020-02-13
Adding POWHEG r3693

mfasDa:
In addition move from LHAPDF5 to LHAPDF6 in order to allow usage of more modern PDF sets.

### DIFF
--- a/powheg.sh
+++ b/powheg.sh
@@ -1,11 +1,12 @@
 package: POWHEG
 version: "%(tag_basename)s"
-tag: "r3178-alice1"
+tag: "r3693-alice1"
 source: https://github.com/alisw/POWHEG
 requires:
   - fastjet
   - "GCC-Toolchain:(?!osx|slc5)"
-  - lhapdf5
+  - lhapdf
+  - lhapdf-pdfsets
 ---
 #!/bin/bash -e
 
@@ -35,7 +36,11 @@ proc ModulesHelp { } {
 set version $PKGVERSION-@@PKGREVISION@$PKGHASH@@
 module-whatis "ALICE Modulefile for $PKGNAME $PKGVERSION-@@PKGREVISION@$PKGHASH@@"
 # Dependencies
-module load BASE/1.0 lhapdf5/$LHAPDF5_VERSION-$LHAPDF5_REVISION ${GCC_TOOLCHAIN_ROOT:+GCC-Toolchain/$GCC_TOOLCHAIN_VERSION-$GCC_TOOLCHAIN_REVISION} ${FASTJET_REVISION:+fastjet/$FASTJET_VERSION-$FASTJET_REVISION}
+module load BASE/1.0 \
+            lhapdf/$LHAPDF_VERSION-$LHAPDF_REVISION \
+            lhapdf-pdfsets/$LHAPDF_PDFSETS_VERSION-$LHAPDF_PDFSETS_REVISION \
+            ${GCC_TOOLCHAIN_ROOT:+GCC-Toolchain/$GCC_TOOLCHAIN_VERSION-$GCC_TOOLCHAIN_REVISION} \
+            ${FASTJET_VERSION:+fastjet/$FASTJET_VERSION-$FASTJET_REVISION}
 # Our environment
 set POWHEG_ROOT \$::env(BASEDIR)/$PKGNAME/\$version
 setenv POWHEG_ROOT \$POWHEG_ROOT


mfasDa:
Marking PR as WIP: In order to use POWHEG from the ALICE fork I have opened a PR in alisw/POWHEG (https://github.com/alisw/POWHEG/pull/2). Before merging the corresponding tag should be created in the alisw POWHEG repository and the version/source in the build recipe should point to this tag.

powheg.sh
```diff
@@ -1,11 +1,11 @@
 package: POWHEG
 version: "%(tag_basename)s"
-tag: "r3178-alice1"
-source: https://github.com/alisw/POWHEG
+tag: "r3693-alice1"
+source: https://github.com/mfasDa/POWHEG
```
maireiphc: I see you there as current code owner from your personal github.
How you want to proceed before the merge ? i.e. : 

1. should we leave that way for a while so that the chain is exercised to its end (cvmfs deployment... + execution test) while you still have the full hand on it ?
2. or do you want push alisw/POWHEG from your github to the central one before that merging this recipe to alisw/alidist ?

Note that I am fine personally with either option.

mfasDa: Hi @maireiphc, as mentioned above ultimately the imported POWHEG version should appear as separate branch in alisw/POWHEG. As this is a 1-1 import of the POWHEG sources from the official svn repository it should normally be fine to import POWHEG directly into alisw/POWHEG and take the tag from there. 

I have already been running productions with the new POWHEG version built via this build recipe on lxplus.

mfasDa:
@maireiphc @qgp I update the PR taking the current head of the patches branch of the official ALICE fork. In case the build test runs fine you can do the tag, and I will update the tag so the PR can be merged.

qgp:
@mfasDa ok, ping me if/when you want me to do the tag

mfasDa:
@qgp  I think you can do the tag

qgp:
@mfasDa Created `r3693-alice1`, please check if everything is as you need it.

mfasDa:
OK, build looks fine. I changed the version to the tag. The PR would be ready to merge now.

powheg.sh
```diff
@@ -1,11 +1,11 @@
 package: POWHEG
 version: "%(tag_basename)s"
-tag: "r3178-alice1"
+tag: "r3693-alice1"
 source: https://github.com/alisw/POWHEG
 requires:
   - fastjet
   - "GCC-Toolchain:(?!osx|slc5)"
-  - lhapdf5
+  - lhapdf-pdfsets
```
qgp: @mfasDa Sorry, I only spotted this now and maybe I am overlooking something. Dropping lhapdf as a requirement here essentially results in it being only *build_*required since lhapdf-pdfsets only build_requires it. The latter is actually inconsistent with loading lhapdf from the lhapdf-pdfsets modulefile. If I remember correctly, not having lhapdf as a runtime requirement for lhapdf-pdfsets is done to avoid having to redeploy the pdfsets in case of an lhapdf update, correct? In this case, lhapdf should stay as a dependency here and we should fix the modulefile for lhapdf-pdfsets. If instead there is no good reason to have lhapdf only as build_requirement for lhapdf-pdfsets we should fix that. Could you. comment?

mfasDa: @qgp Sorry, I overlooked this. As lhapdf-pdfsets loads LHAPDF in the modulefile I expected it was sufficient to set the dependency on lhapdf-pdfsets, but indeed lhapdf is only a build requirement for lhapdf-pdfsets, and the requirement here should be lhapdf (but no longer lhapdf5) + the lhapdf-pdfsets modulefile should be fixed. Maybe in this case also the herwig and SHERPA build recipes - which also use lhapdf6 - should be checked.

qgp: @mfasDa How should we go ahead? Do you want to update the dependency here such that this PR can be merged?

Then, we fix the **lhapdf-pdfsets** modulefile and the dependencies for both **herwig** and **SHERPA** (I confirmed both list only lhapdf-pdfsets as requirement).

mfasDa: HI @qgp, yes, let's do like this. I will fix the dependecies here.

qgp:
@mfasDa Don't you need both `lhapdf` and `lhapdf-pdfsets` here?

mfasDa:
@qgp  POWHEG itself would only requires `lhapdf` to run, but then it has no PDFSET. One could either make `lhapdf-pdfsets` a direct dependency or load `lhapdf-pdfsets` by hand. In any case the lhapdf-pdfsets also needs reassessment, as it comes only with the minimum number of pdfsets which might not be adequate for some cases. On the other side one does not want to make the package too big. One could think of modularising the pdfsets in core, NLO, NNLO, Nuclear, ... . Such an approach would support loading pdfsets manually based on the user choice. Setting the dependency to lhapdf-pdfsets in the build recipe would make the handling easer and more failsafe.

mfasDa:
Re-established dependency on lhapdf-pdfset in order to automatically load the pdfsets. In case of restructuring the pdfsets the build recipe can be adapted in the future.

qgp:
@mfasDa Ok, you are right, we might have to split the PDFsets or add more to the existing package. I think having lhapdf and (a basic set of) pdfsets as dependency here is the least troublesome and ensuring a usable POWHEG.

Good to merge, I'd say.

qgp:
@maireiphc Anything to add here? Or should this get merged?

maireiphc:
Hi @qgp, nothing further to be mentioned to my knowledge. The build 0²suite seems to fails while the logs are not explicit. Giulio pinged Barthelemy on this, for another of our AliGenerators fronts. 
(JEWEL : https://github.com/alisw/alidist/pull/2139)
Likely a fake fail ?

Good to merge I would say

mfasDa:
How is the status with this PR?

PR #2078, opened 2020-02-13
[O2-1158] prefer patched alibuild cmake on macos

MichaelLettrich:
We need macOS to pick up out patched version of CMake as long as this [CMake bug](https://gitlab.kitware.com/cmake/cmake/issues/19845#note_689595) is not fixed in a release version.

### DIFF
--- a/cmake.sh
+++ b/cmake.sh
@@ -5,7 +5,7 @@ source: https://github.com/alisw/CMake
 build_requires:
  - "GCC-Toolchain:(?!osx)"
  - make
-prefer_system: .*
+prefer_system: "(?!osx)"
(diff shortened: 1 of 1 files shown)


cmake.sh
```diff
@@ -5,7 +5,7 @@ source: https://github.com/alisw/CMake
 build_requires:
  - "GCC-Toolchain:(?!osx)"
  - make
-prefer_system: .*
+ prefer_system: (?!osx)
```
ktf: You added an extra space...

MichaelLettrich: -.-

PR #2083, opened 2020-02-14
Exclude 3rd party files from code checker

davidrohr:
@ktf @sawenzel : We don't need to exclude the DebugGUI files any more, since they are not in O2.
Can we instead exclude anything that has 3rdparty in its name (For the Standalone build of the GPU visualization, I have to ship some 3rd party files, like @ktf did before for the DebugGUI). I think 3rdparty is generic enough to be excludable. If not, I can also find another way to hide my files from the codechecker, in principle I can just zip them or so.

### DIFF
--- a/o2checkcode.sh
+++ b/o2checkcode.sh
@@ -82,7 +82,7 @@ COPYRIGHT="$(cat <<'EOF'
 EOF
 )"
 COPYRIGHT_LINES=$(echo "$COPYRIGHT" | wc -l)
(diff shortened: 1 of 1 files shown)


davidrohr:
@ktf : Can you merge if this is OK with you?

ktf:
I fully agree, however are you sure it's not "^3rdparty/" what we need?

davidrohr:
No, I want to exclude every 3rdparty folder

Kind Regards
David Rohr

ktf:
k

PR #2087, opened 2020-02-17
Fix alo-aliroot module load

dstocco:


Barthelemy:
Merging as the tests pass and it is approved.

PR #2091, opened 2020-02-18
Force OpenCL2 for GPU CI

davidrohr:


davidrohr:
This PR makes OpenCL compilation mandatory in the GPU CI, we should merge it once the CI with the new GPU container becomes green.

PR #2092, opened 2020-02-18
Add npm installation prefix to qcg recipe

Barthelemy:
make sure we install in the right place


Barthelemy:
@ktf Because node.js needs to know where it is (@awegrzyn can correct me). Would having a different name be ok ?

ktf:
Does it need to know about it at runtime? Why not simply relying on LD_LIBRARY_PATH?

Barthelemy:
As far as I can see, in the binding definition for node.js (to access QC from JS), it needs the include directories : 
```
      'include_dirs': ["<!@(node -p \"require('node-addon-api').include\")", "<@(qc_root)/include", "<@(root_include)", "$(CONFIGURATION_ROOT)/include", "$(COMMON_O2_ROOT)/include", "$(BOOST_ROOT)/include" ],
```
Thus when building qcg it requires to know these variables.

ktf:
A different name is not ok, because it will still prevent people who installed boost from the system to pick it up correctly, since `BOOST_ROOT` is not set by anyone else. How are the bindings generated?

awegrzyn:
@Barthelemy `BOOST_ROOT` is only required during compilation, not execution.

ktf:
`BOOST_ROOT` is guaranteed to be present at build time the package is built via aliBuild and boost from the system is not compatible with the one we expect in O2.
 
What do you mean by manual installation?

Barthelemy:
@ktf We discussed with adam and I am dropping the line I added to boost.sh. I keep the other fix that I need. Adam will think how to make it work in a proper way.

awegrzyn:
@ktf Could you merge?

ktf:
Related to this, can you tell me what `otool -h someBindingLibrary.so` returns on mac? I suspect you have been hit by a bug in aliBuild which was not correctly relocating the RPATH for bundles (as opposed to dylibs).

Issue #2086, opened 2020-02-17
Yoda needs cython with fastjet 3.3.3


marcovanleeuwen:
Dear experts,

I am not sure what triggers this, but when build Herwig with the defaults=dev, i.e. fastjet 3.3.3,  I suddenly find that YoDa wants to have cython present.

I dug a bit trrough the config scripts, and this is how it seems to check whether cython is present:

```
prog="from __future__ import print_function; import Cython.Compiler.Version; print(Cython.Compiler.Version.version)"
CYTHON_VERSION=`$PYTHON -c "$prog" 2>&5`

py_cv_cython=no
if test "x$CYTHON_VERSION" != "x"; then
   py_cv_cython=yes
fi
```
When I try this with the system python, I get 

```
[marcovl@mole alisoft]$ python3 -c "from __future__ import print_function; import Cython.Compiler.Version; print(Cython.Compiler.Version.version)"
Traceback (most recent call last):
  File "<string>", line 1, in <module>
ModuleNotFoundError: No module named 'Cython'
```
which may not be caught by the configure script, although that would be strange... Of course I should check this also with the aliBuild-installed python, but not 100% sure how to do that...

One solution may be to switch off the python extensions for Yoda...

Marco.

marcovanleeuwen:
I looked a bit more at this, and it seems that when I do

aliBuild Herwig
there is no dependence on Python-modules

while
aliBuild Herwig --defaults=dev
adds Python-modules to the dependences

This is a bit puzzling, because in both cases, boost is required, and boost depends on Python-modules according to  the boost.sh recipe.

Could it be that there is a problem with the dependency parsing?

(NB: even if that is the case, fixing it will not solve this problem, but I guess we should start at the start...)

Best regards,

Marco.

marcovanleeuwen:
I tried removing the Python-modules dependence from boost. That fixes the issue with YODA, but the library
libboost_python 
is not produced:
```
ls sw/slc7_x86-64/boost/v1.70.0-4/lib/*boost_python*
ls: cannot access sw/slc7_x86-64/boost/v1.70.0-4/lib/*boost_python*: No such file or directory
```
same when we use Python instead of Python-modules as the dependence...

marcovanleeuwen:
OK, maybe now I get it. I was not sure which defaults file is used when you specify none, but maybe it is defaults-release.sh ? This one indeed disables the Python support in boost:
'''
  # Minimal boost with no Python & co.
  boost:
    requires:
[...]

PR #2096, opened 2020-02-19
Remove dependence on Python-modules

marcovanleeuwen:
Hi Guilio,

This is the next step, removing the Python-modules dependence (from boost). With this, I could compile aligenerators with '--defaults dev'.

This also solves (or works around) https://github.com/alisw/alidist/issues/2086 

Marco.


ktf:
We can add it as a workaround, but is this something we want in the long term? We should investigate if we use boost python anywhere.

PR #2099, opened 2020-02-20
Bump xjalienfs to 1.0.5

Atlantic777:
- tab complete with filenames
- user defined aliases
- command utilities scripts generated by setup.py
- current working directory persistent across sessions
- improved SSL compatibility

### DIFF
--- a/xjalienfs.sh
+++ b/xjalienfs.sh
@@ -16,7 +16,10 @@ env PYTHONUSERBASE="$INSTALLROOT" ALIBUILD=1 pip3 install --user file://${SOURCE
 XJALIENFS_SITEPACKAGES=$(find ${INSTALLROOT} -name site-packages)
 
 ALIEN_PY=$(find ${INSTALLROOT} -path '*/bin/*' -name alien.py)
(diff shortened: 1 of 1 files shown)


xjalienfs.sh
```diff
@@ -15,11 +15,11 @@ requires:
 env PYTHONUSERBASE="$INSTALLROOT" ALIBUILD=1 pip3 install --user file://${SOURCEDIR}
 XJALIENFS_SITEPACKAGES=$(find ${INSTALLROOT} -name site-packages)
 
-ALIEN_PY=$(find ${INSTALLROOT} -name alien.py)
+ALIEN_PY=$(find ${INSTALLROOT} -path '*/bin/*' -name alien.py)
```
ktf: How do you guarantee there is only one?

Atlantic777: Well, we install alien.py using setup.py and pip and there will be just one bin/alien.py for sure. But, I agree, I should change this to an explicit path.

ktf: ok, but can't you simply use `${INSTALLROOT}/bin/alien.py`, then?

PR #2104, opened 2020-02-24
Disable arrow (not needed for AliRoot; does not compile on sl6?)

marcovanleeuwen:
Hi Guilio,

Another update for the defaults-dev for aligenerator: disable arrow since it does not seem to compile on SL6 and is not needed.

Best regards,

Marco.

### DIFF
--- a/defaults-dev.sh
+++ b/defaults-dev.sh
@@ -4,6 +4,10 @@ env:
   CXXFLAGS: "-fPIC -g -O2 -std=c++11"
   CFLAGS: "-fPIC -g -O2"
   CMAKE_BUILD_TYPE: "RELWITHDEBINFO"
+
+disable:
+  - arrow
+
 overrides:
   fastjet:
     version: "v3.3.4_1.042"


ktf:
Mmm... I think we need to change the name of the default, then. The idea was for "dev" to be the next version of the default that everyone uses, not a crippled one. ;-) Let me add a defaults-generators or something similar.

marcovanleeuwen:
(I had replied by mail, but that does not seem to have arrived here)

This is OK for me as well. However, judging from the other defaults, it seems that e.g. arrow is never part of the AliRoot builds (probably because it does not compile on sl6? Or the dependencies are incomplete?), so another option is to have an 'o2-dev.sh' for O2. Or to make arrow compile on the SL6 build server...

Anyways, I don't feel strongly about this, so please go ahead as you see fit!

The simplest way to move ahead to my mind is to approve this pull request; that way we will at least see whether AliGenerators compiles on the build server, and we can test whether Herwig works again...

ktf:
Sorry, lagging behind. I will try to look into this later tonight.

marcovanleeuwen:
Hi @ktf . Just a quick ping to ask if you could have a look and approve or suggest an alternative (e.g. making the defaults-generators.sh).

I checked the build tests above, and the failure with O2 seems to be due to not being able to connect the ccdb.

Marco.

PR #2111, opened 2020-02-27
Port installation of compile_commands.json

mfasDa:
Port installation of the compile_commands.json in
development mode from o2.sh to qualitycontrol.sh.
Having compile compile_commands.json in the source
dir is helpful as modern IDEs can make use of it.

### DIFF
--- a/qualitycontrol.sh
+++ b/qualitycontrol.sh
@@ -77,6 +77,17 @@ if [[ $ALIBUILD_O2_TESTS ]]; then
   ctest --output-on-failure -LE manual -E "(testWorkflow|testCheckWorkflow)" ${JOBS+-j $JOBS}
 fi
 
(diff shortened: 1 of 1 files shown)


mfasDa:
This is a 1-1 copy of the corresponding lines in o2.sh. Having compile_commands.json in the source directory when working in development mode is very useful as modern IDEs can use it to configure the project.

Barthelemy:
Very good, thank you. Please note that Piotr has also been working on how to use Clion with alibuild : https://github.com/AliceO2Group/AliceO2Group.github.io/blob/master/advanced/ides.markdown 
If you have instructions to be able to use compile_commands.json please do not hesitate to add them there (by means of a PR).

mfasDa:
@Barthelemy Thanks a lot for merging! My primary intention was on the usage of vscode which can configure a project using compile_commands.json. I have no experience with Clion.

PR #2112, opened 2020-02-27
build clang for x86 only

MichaelLettrich:
Currently clang is being built for all available architectures however we only need x86. this small change should bring down the compile time for clang significantly.
Reference: https://llvm.org/docs/CMake.html


MichaelLettrich:
Integrated in #2113

PR #2114, opened 2020-02-28
Update MC packages and FairRoot versions:

ihrivnac:
GEANT3: v3-3
  GEANT4: v10.5.1
  GEANT4_VMC: v5-0-p2
  VGM: v4-7
  FairRoot: 4ab2d2b1f0392 (includes a fix in FindROOT.cmake)
- Keep previous  versions for builds with ROOT 5:
  GEANT3: v2-7-p1
  VGM v4-4
  GEANT4_VMC v3-6-p6-inclxx-biasing-p2
  GEANT4 v10.4.2
This should hopefully fix the JIRA issues O2-1029, O2-1044.


aliroot.sh
```diff
@@ -14,6 +14,14 @@ build_requires:
   - "Xcode:(osx.*)"
 env:
   ALICE_ROOT: "$ALIROOT_ROOT"
+overrides:
```
ihrivnac: @ktf , @pzhristov , 
This seems to be a wrong place where to override versions for building AliRoot for run2. Could you, please, tell me where these definitions should go? Thank you.

ihrivnac:
In Geant4 10.5 the URL for data sets is changed from http:// to https:// which causes the error:
`'Protocol "https" not supported or disabled in libcurl'`
Could someone of the @alisw/infrastructure-admins check this? Thanks.

ihrivnac:
@ktf, could you, please, take a look why curl on the server does not support https protocol? See the error extracted from the log above. Thank you.

ihrivnac:
@ktf, 
this PR needs an administrator's intervention. Can you check?

aliroot.sh
ihrivnac: The last build/AliRoot/alidist  (see the log below) now fails because the new versions are not compatible with ROOT 5.x. So we need to keep the old versions for building AliRoot. However, as I asked above, I didn't find where these should be declared. @pzhristov , could you help with this? Thank you.

ihrivnac:
Hi,
After the commit ae9c28d, the problem with downloading Geant4 data is gone.
The PR is now blocked due to aliDoctor error; is anyone (@ktf, @davidrohr , @pzhristov ) able to fix these ? We need to keep the old versions for all builds with ROOT 5. 
Thank you,

ihrivnac:
In build/O2/alidist, we get a list of failing macros, where we get errors about incomplete types, eg:
```
In file included from input_line_9:1:
In file included from /mnt/mesos/sandbox/sandbox/O2/macro/o2sim.C:11:
/mnt/mesos/sandbox/sandbox/O2/macro/build_geometry.C:48:24: error: unknown type name 'FairRunSim'
void finalize_geometry(FairRunSim* run);

```                       

To compare these with a successful build, we need a log from a recently merged PR. At the last WP3 meeting, @ktf mentioned that this is possible with the current system. @ktf, could you, please, point me at such a log from a recently merged PR?

ktf:
Just click on a commit and then on the details of a given test. If it's still there, you will get the log, otherwise you will get a "304" error.

ihrivnac:
Thanks. I tried https://github.com/alisw/alidist/pull/2154/commits, then after clicking on the x for the 3rd commit, I get the logs. However here, we have "2 errored and 3 successful checks" but the PR was merged anyway. 

I was able to find a successful build and its log for this machine in the PR 
https://github.com/alisw/alidist/pull/2153. Here, all tests pass ok while in our case, 24 tests, all with just one exception are tests run from a macro. For all these tests, the MC change is irrelevant, so the only cause possible is the FairRoot upgrade.

@sawenzel , @aphecetche , would you have an idea what could be common to these failing macros (as the log is very long, I put them also below):
	 93 - Detectors/PHOS/testsimulation/PutPhosInTop.C (Failed)
	 95 - Detectors/PHOS/testsimulation/drawPHOSgeometry.C (Failed)
	103 - Detectors/CPV/testsimulation/drawCPVgeometry.C (Failed)
	137 - Detectors/EMCAL/testsimulation/PutEmcalInTop.C (Failed)
	139 - Detectors/EMCAL/testsimulation/drawEMCALgeometry.C (Failed)
	141 - Detectors/EMCAL/testsimulation/run_sim_emcal.C (Failed)
	159 - Detectors/TOF/prototyping/drawTOFgeometry.C (Failed)
	241 - Detectors/MUON/MID/Simulation/macros/drawMIDGeometry.C (Failed)
	280 - Detectors/TPC/simulation/macro/laserTrackGenerator.C (Failed)
	287 - Detectors/TPC/monitor/macro/RunCompareMode3.C (Failed)
	317 - Generators/share/external/extgen.C (Failed)
	318 - Generators/share/external/extgen.C_compiled (Failed)
	319 - Generators/share/external/pythia6.C (Failed)
	321 - Generators/share/external/QEDLoader.C (Failed)
	322 - Generators/share/external/tgenerator.C (Failed)
	350 - macro/build_geometry.C (Failed)
	356 - macro/eventDisplay.C (Failed)
	357 - macro/eventDisplay.C_compiled (Failed)
	362 - macro/o2sim.C (Failed)
	374 - macro/run_clus_itsSA.C (Failed)
	376 - macro/run_clus_mftSA.C (Failed)
	378 - macro/run_clus_tof.C (Failed)
	380 - macro/run_clus_tpc.C (Failed)
	410 - macro/run_trac_mft.C (Failed)

ktf:
It was merged because the error was unrelated.

ihrivnac:
Finally, I was able to reproduce the error on CC7. In the LD_LIBRARY_PATH, there is:
`/root/alice/sw/slc7_x86-64/FairRoot/4ab2d2b1f0-1/lib`
while the FairRoot installs libraries in lib64:
`/root/alice/sw/slc7_x86-64/FairRoot/4ab2d2b1f0-1/lib64`

After linking by hand lib64 to lib, the failing tests pass. 
So we need to tell somewhere in dependencies, the correct lib path. 
@aphecetche , would you know where to put this fix?

ihrivnac:
@ktf , the build/O2/alidist fails on some python issues, while there is nothing related python in this update. Then, the build/AliRoot/alidist log is systematically empty with each branch update. Could you take a look?

ihrivnac:
@aphecetche ,
The build on Mac is now broken while it was passing fine before your CMake upgrade:  https://github.com/AliceO2Group/AliceO2/commit/7a6b188c32a9b4faf4fd3c31879dbbe6e13485fb

Adding
```
# FIXME: this should be properly done by Geant4 itself instead 
find_package(VGM)
set_property(TARGET geant4 APPEND PROPERTY INTERFACE_LINK_DIRECTORIES $<TARGET_FILE_DIR:XmlVGM>)

```
in FindGeant4.cmake is wrong as Geant4 does not use VGM at all; VGM is used only by Geant4 VMC.  After removing this code, the build goes fine on my MacOS, as well as o2sim_G4 test.

Can we remove this code? No other modifications will be needed for building with the new versions proposed in this PR.

aphecetche:
@ihrivnac will check further later on, as it may indeed be the case that those lines would be better in some FindGeant4_VMC.cmake but the point is that I did need them to have the o2sim_G4 tests running fine on Mac. And what I call running fine is in _all_ of those situations : 

- with or without SIP disabled
- within an env ALIBUILD_O2_TESTS=1 aliBuild --defaults build O2 run
- from a BUILD/O2-latest/O2 directory and just doing ctest -R o2sim_G4 without setting any specific environment

ihrivnac:
@aphecetche , on my side (on Mac with SIP disabled, I believe) the `ctest -R o2sim_G4` runs ok in the build directory without setting environment. Could you try out your tests with the change proposed above with this PR branch and let me know if you see any issues?
Note that in the new VMC packages versions included in this PR have CMake files improved.

aphecetche:
@ihrivnac will do. Might take some time as it means quite a few packages to rebuild ;-)

aphecetche:
@ihrivnac. I tested w/ SIP enabled and no env. and the o2sim_G4 tests indeed work well. So yes, you can remove those lines from `FindGeant4.cmake`

```
# FIXME: this should be properly done by Geant4 itself instead 
find_package(VGM)
set_property(TARGET geant4 APPEND PROPERTY INTERFACE_LINK_DIRECTORIES $<TARGET_FILE_DIR:XmlVGM>)
```

ihrivnac:
@aphecetche , thanks for checking. The PR is now open: https://github.com/AliceO2Group/AliceO2/pull/3395

aphecetche:
> @aphecetche , thanks for checking. The PR is now open: [AliceO2Group/AliceO2#3395](https://github.com/AliceO2Group/AliceO2/pull/3395)

@hrivnac I thought you'll include that removal into this PR, not a separate one... As the other has been merged, now the ctest (at least on my Mac) is broken, e.g. the g4Config.C macro tests fails with : 

```
DEBUG:O2:O2:0: cling::DynamicLibraryManager::loadLibrary(): dlopen(/Users/laurent/alice/pr/sw/osx_x86-64/GEANT4_VMC/v3-6-p6-2/lib/libgeant4vmc.dylib, 9): Library not loaded: @rpath/libXmlVGM.dylib
DEBUG:O2:O2:0:   Referenced from: /Users/laurent/alice/pr/sw/osx_x86-64/GEANT4_VMC/v3-6-p6-2/lib/libgeant4vmc.dylib
DEBUG:O2:O2:0:   Reason: image not found
```

I guess this is not a big deal as long as this one does not take too long to be merged as well, but just to note that it also -currently- prevents a simplification of the `o2.sh` recipe (where we could remove a number of export variables)

aphecetche:
> @aphecetche , thanks for checking. The PR is now open: [AliceO2Group/AliceO2#3395](https://github.com/AliceO2Group/AliceO2/pull/3395)

@hrivnac I thought you'll include that removal into this PR, not a separate one... As the other has been merged, now the ctest (at least on my Mac) is broken, e.g. the g4Config.C macro tests fails with : 

```
DEBUG:O2:O2:0: cling::DynamicLibraryManager::loadLibrary(): dlopen(/Users/laurent/alice/pr/sw/osx_x86-64/GEANT4_VMC/v3-6-p6-2/lib/libgeant4vmc.dylib, 9): Library not loaded: @rpath/libXmlVGM.dylib
DEBUG:O2:O2:0:   Referenced from: /Users/laurent/alice/pr/sw/osx_x86-64/GEANT4_VMC/v3-6-p6-2/lib/libgeant4vmc.dylib
DEBUG:O2:O2:0:   Reason: image not found
```

I guess this is not a big deal as long as this one does not take too long to be merged as well, but just to note that it also -currently- prevents a simplification of the `o2.sh` recipe (where we could remove a number of export variables)

aphecetche:
@ihrivnac Actually it's even worse : I tested the o2sim_G4 but the issue I first encountered was on g4Config.C (my bad). I've rechecked this PR and I have the failure describe in my previous comment for the test g4Config (Mac with SIP enabled). So I'd say the original lines should be put back in, in some way. Maybe in FindGeant4VMC.cmake, like so : 

```
if(APPLE)
# FIXME: this should be properly done by Geant4VMC itself instead
find_package(VGM)
if(TARGET XmlVGM)
set_property(TARGET geant4vmc APPEND PROPERTY INTERFACE_LINK_DIRECTORIES $<TARGET_FILE_DIR:XmlVGM>)
endif()
endif()
```

would that cause a problem for this PR ?

ihrivnac:
@aphecetche , 
The PR was proposed separately as it concerned a different git repository.
When this code was present, it broke compilation on Mac, both on my side and in CI. 
Instead of a work-around in O2 it would be better to fix the Geant4 VMC cmake.  I believed that everything is already here with the updates started by Benedikt in geant3 and then applied by me in Geant4 VMC and VGM. But we did not put INTERFACE_LINK_DIRECTORIES. 
I will try out to add your suggested lines in Geant4 VMC cmake and test with both SIP options.

ihrivnac:
@aphecetche ,
I was confused whether the failures which you report here concern the build with this PR branch or with alidist master. From the error report you posted, this seems to be with master and not with this PR.
I tested on my side with SIP enabled and the test with o2sim_G4 run from the build directory without calling aliEnv works still fine.
So the only problem that we have now is that we do not get log from AliRoot build. 
@ktf, @MichaelLettrich , what happens here?

aphecetche:
@ihrivnac no, the error is with this PR also. And the failing test is the g4Config.C macro test.

MichaelLettrich:
@ihrivnac : logs say there is a typo when building packages
`ERROR: VGM should be spelt vgm`

ihrivnac:
@MichaelLettrich , thanks ! I update the branch with a fix. Why was the log empty?

ihrivnac:
@aphecetche , on my side g4Config.C works fine, too:

```
[MacBook-Pro-4] ~/work/projects/aliceO2/sw/BUILD/O2-latest/O2 > csrutil status
System Integrity Protection status: enabled.
[MacBook-Pro-4] ~/work/projects/aliceO2/sw/BUILD/O2-latest/O2 > ctest -R Detectors/gconfig/g4Config.C
Test project /Users/ivana/work/projects/aliceO2/sw/BUILD/O2-latest/O2
    Start 337: Detectors/gconfig/g4Config.C
1/1 Test #337: Detectors/gconfig/g4Config.C .....   Passed    4.27 sec

100% tests passed, 0 tests failed out of 1

```
Can you post the error which you get when testing with this PR branch?

MichaelLettrich:
@ihrivnac the log is empty because the failure occurred before aliBuild actually started building something and is expected behavior because we don't want to give away any security relevant data by accident.

aphecetche:
> Can you post the error which you get when testing with this PR branch?

will do, but I have to redo it as I "reused" the directory for another branch and so must recompile a bunch of packages, including, unfortunately, Root ...

aphecetche:
ok, here we go, at least I can reproduce it. The relevant part is : 

```
cling::DynamicLibraryManager::loadLibrary(): dlopen(/Users/laurent/alice/dev/sw/osx_x86-64/GEANT4_VMC/v3-6-p6-2/lib/libgeant4vmc.dylib, 9): Library not loaded: @rpath/libXmlVGM.dylib
  Referenced from: /Users/laurent/alice/dev/sw/osx_x86-64/GEANT4_VMC/v3-6-p6-2/lib/libgeant4vmc.dylib
  Reason: image not found
Error in <TInterpreter::TCling::AutoLoad>: failure loading library libgeant4vmc.dylib for TG4RunConfiguration
```

and the full `aliBuild --defaults o2 build O2` log for this attempt below.

[log.txt](https://github.com/alisw/alidist/files/4516468/log.txt)

ihrivnac:
@aphecetche , there still v3-6-p6-2 in your error report above, while the geant4_vmc version in this PR is v5-0-p2. Could it be that you have Geant4 VMC as a development package and then you have to update to the tag manually?

aphecetche:
@ihrivnac sorry, I've taken the information from the wrong set of terminals  (but it does not change the conclusion) : here are the ones I meant to send : 

```
cling::DynamicLibraryManager::loadLibrary(): dlopen(/Users/laurent/alice/cmake/sw/osx_x86-64/GEANT4_VMC/v5-0-p2-2/lib/libgeant4vmc.dylib, 9): Library not loaded: @rpath/libXmlVGM.dylib
  Referenced from: /Users/laurent/alice/cmake/sw/osx_x86-64/GEANT4_VMC/v5-0-p2-2/lib/libgeant4vmc.dylib
  Reason: image not found
Error in <TInterpreter::TCling::AutoLoad>: failure loading library libgeant4vmc.dylib for TG4RunConfiguration

``` 

And the corresponding 
[log.txt](https://github.com/alisw/alidist/files/4517057/log.txt)

(and no I don't have any dev package apart from O2)

ihrivnac:
@aphecetche , thank you. It is strange that I am not getting the same on my side. In the log, there is an error already before that one you posted:

```
Error in <TSystem::ExpandFileName>: input: $ROOT_DYN_PATH/libRIO, output: $ROOT_DYN_PATH/libRIO
Error in <TSystem::ExpandFileName>: input: $ROOT_DYN_PATH/libCling, output: $ROOT_DYN_PATH/libCling
Error in <TSystem::ExpandFileName>: input: $ROOT_DYN_PATH/libMathCore, output: $ROOT_DYN_PATH/libMathCore

```
and the list of libraries in LD_LIBRARY_PATH is cut. But why this should happen only for g4Config.C test?

aphecetche:
While I would not call the ROOT_DYN_PATH errors normal (don't know exactly what the ROOT_DYN_PATH is) and I  believe there have nothing to do with the error at hand. If you pick any macro and do a `ctest -R macro -V` you should get those as well.

Why only for g4Config ? I'd guess because it's the only (complex) macro using G4 ?

ihrivnac:
I confirm that these lines with Errors are present also in my log when run with -V option. 
Can you run a successful test with this option and check that there is present the path to the VGM libraries? On my side, I have:
```
...
340: /Users/ivana/work/projects/aliceO2/sw/osx_x86-64/GEANT4_VMC/42a651095a-2/lib
340: /Users/ivana/work/projects/aliceO2/sw/osx_x86-64/vgm/v4-7-5/lib
340: /Users/ivana/work/projects/aliceO2/sw/osx_x86-64/GEANT4/v10.5.1-3/lib

```

aphecetche:
@ihrivnac  which test ? the list of libs depends on the test ...

sawenzel:
I think the ROOT error messages concerning $ROOT_DYN_PATH are quite a generic problem which appeared a few weeks ago. There is a JIRA ticket here: https://alice.its.cern.ch/jira/browse/O2-1297. We should put this on the agenda for WP3.

ihrivnac:
@aphecetche , I tried macro/run_clus_tof.C  and Detectors/MUON/MCH/Simulation/macros/drawMCHGeometry.C and the list of libraries below "LD_LIBRARY_PATH to be used:" line is the same, both include the path to vgm libs.

aphecetche:
Then there's something I don't quite understand. For those two macros, the list of library directories to be used that I have is completely different :  

```
380: /Users/laurent/alice/cmake/sw/BUILD/81d29d731b066d3cb66ea3a9b7d585ab1afd9f6e/O2/stage/lib
380: /Users/laurent/alice/cmake/sw/osx_x86-64/FairRoot/4ab2d2b1f0-2/lib
380: /Users/laurent/alice/cmake/sw/osx_x86-64/pythia/v8243-1/lib
```

And in principle the first one of those directorys should be the one from the build dir stage/lib area. Could you share the `test-root-macro.sh` that's been generated in your build area ?

ihrivnac:
Here it is: [test-root-macro.sh.txt](https://github.com/alisw/alidist/files/4519267/test-root-macro.sh.txt)

MichaelLettrich:
@ihrivnac the failure and empty logs in `build/O2/alidist-dataflow`  is related to DNS problems.

aphecetche:
@ihrivnac ok, here's my [test-root-macro.sh.txt](https://github.com/alisw/alidist/files/4520957/test-root-macro.sh.txt) where you'll see the LIBPATH is much shorter by default, which shows that : 

- I forgot to remove the @LD_LIBRARY_PATH@ from the test-root-macro.sh.in 
- you are building with a LD_LIBRARY_PATH which is non-empty, is that right ?

ihrivnac:
@aphecetche , 
No, I have LD_LIBRARY_PATH variable empty in the directory where I build O2; but I can't exclude that it was not set in previous builds. To fix test-root-macro.sh.in, I should just remove the @LD_LIBRARY_PATH@ from the line 12, right?

aphecetche:
@ihrivnac right. 
And/or wipe the BUILD/O2-latest/ dir, ensure you have no LD_LIBRARY_PATH set, and rebuild (and wait 30 minutes ;-) )

ihrivnac:
I tried it without wiping the old build directory and the script was recreated fine and I can reproduce the error. I will investigate if adding the code which you placed in FindGeant4 can fix it.

ihrivnac:
> I think the ROOT error messages concerning $ROOT_DYN_PATH are quite a generic problem which appeared a few weeks ago. There is a JIRA ticket here: https://alice.its.cern.ch/jira/browse/O2-1297. We should put this on the agenda for WP3.

@sawenzel , Indeed, this issue escaped our WP3  lists as it is reported in "Build Infrastructure" component, which is not included in our filters used to build the agenda. The component seems to be created only in March. We will have to see if it was created intentionally and why.

ihrivnac:
@aphecetche , 
Could you try out the fix in: https://github.com/ihrivnac/AliceO2/pull/new/fix-cmake-find-vgm ? If it is ok for you I will propose it as PR.
There is added FindVGM.cmake and also included the fix in  test-root-macro.sh.in.
In this way we amend the VGM targets in the same way as it is done for Geant4 and Geant4 VMC.
When we move to this PR, we can remove INTERFACE_INCLUDE_DIRECTORIES from Find files as these are already provided, however INTERFACE_LINK_DIRECTORIES will be still needed.

aphecetche:
@ihrivnac  using your branch fix-cmake-find-vgm I get a cmake error at configure time : 

CMake Error at dependencies/FindVGM.cmake:19 (set_target_properties):
  set_target_properties Can not find target to add properties to: VGM::XmlVGM
  dependencies/O2SimulationDependencies.cmake:66 (find_package)
  dependencies/O2Dependencies.cmake:108 (include)
  dependencies/CMakeLists.txt:11 (include)
  CMakeLists.txt:45 (include)

ihrivnac:
@aphecetche , does this happen with alidist master or this PR? As in the older versions of VGM there is not VGM:: namespace.

aphecetche:
@ihrivnac with alidist master.
With this PR it's working fine.
But to avoid tricky situation between the two repositories / PRs, it would be better to add a protection in FindVGM.cmake (at least for now) :

```
if(TARGET VGM::XmlVGM)
set_target_properties(VGM::XmlVGM
                      PROPERTIES INTERFACE_INCLUDE_DIRECTORIES
                                "${VGM_INCLUDE_DIRS}"
                                INTERFACE_LINK_DIRECTORIES
                                $<TARGET_FILE_DIR:VGM::XmlVGM>)
endif()
```

don't you think ?

ihrivnac:
@aphecetche ,
But don't you need this setting also for the old versions? I am now testing this:
```
if(${VGM_VERSION} VERSION_LESS "4.7")
set_target_properties(XmlVGM
                      PROPERTIES INTERFACE_INCLUDE_DIRECTORIES
                                "${VGM_INCLUDE_DIRS}"
                                INTERFACE_LINK_DIRECTORIES
                                $<TARGET_FILE_DIR:XmlVGM>)
else()
set_target_properties(VGM::XmlVGM
                      PROPERTIES INTERFACE_INCLUDE_DIRECTORIES
                                "${VGM_INCLUDE_DIRS}"
                                INTERFACE_LINK_DIRECTORIES
                                $<TARGET_FILE_DIR:VGM::XmlVGM>)
endif()

``` 
But if yours works fine, it is simpler.

aphecetche:
@ihrivnac indeed you are right.  But then the test could be a little DRYier, using something like (not tested ) : 

if(TARGET XmlVGM)
set(targetXml XmlVGM)
else if (TARGET VGM::XmlVGM)
set(targetXml VGM::XmlVGM)
endif()

if (targetXml)
set_target_properties(${targetXml}
                      PROPERTIES INTERFACE_INCLUDE_DIRECTORIES
                                "${VGM_INCLUDE_DIRS}"
                                INTERFACE_LINK_DIRECTORIES
                                $<TARGET_FILE_DIR:${targetXml}>)
endif()

ihrivnac:
If VGM is found the target should exist, and if it does not exist it is better to issue an error than do silently nothing. So I'd prefer the test with the version; but the code can be improved, eg.:
```
# VGM uses namespace since 4.7
if(${VGM_VERSION} VERSION_LESS "4.7")
  set(targetVGM XmlVGM)
else()
  set(targetVGM VGM::XmlVGM)
endif()

set_target_properties(targetVGM
                      PROPERTIES INTERFACE_INCLUDE_DIRECTORIES
                                "${VGM_INCLUDE_DIRS}"
                      INTERFACE_LINK_DIRECTORIES
                                $<TARGET_FILE_DIR:targetVGM>)

```

ihrivnac:
The fix in O2 CMake is now available https://github.com/AliceO2Group/AliceO2/pull/3437.  Running new test round after rebasing.

ihrivnac:
Now the only test failing is  O2Suite/alidist_macOS, there is one test failing which however fails also in other PRs, so it's not related to the changes in this PR. The change requested by @MichaelLettrich is applied, also the issue pointed by @aphecetche  is solved (in O2).
@ktf, can you merge?

cmake.sh
```diff
@@ -25,7 +25,7 @@ SET(Java_JAVAC_EXECUTABLE FALSE CACHE BOOL "" FORCE)
 SET(BUILD_CursesDialog FALSE CACHE BOOL "" FORCE)
 
 # OpenSSL is problematic
```
ktf, line 27: Do you really need to comment this out?

fairroot.sh
```diff
@@ -1,6 +1,6 @@
 package: FairRoot
 version: "%(short_hash)s"
-tag: "6af7e848585ce720e96a01d5565b1bd5e3884430"
+tag: "4ab2d2b1f0392ad93b92026d70fe049c38665f7d"
```
ktf, line 3: The FLP people cannot use hashes for the version, because that does not work with their Updateable RPMS paradigm. Can you please tag this?

cmake.sh
ihrivnac: This was added to fix the problem commented on 28 Feb.

fairroot.sh
ihrivnac: There is no tag including this development. The version currently used is also specified by a hash, so thy have already to deal with a hash now. But we can, of course, ask FairRoot developers for a tag.

ktf: I will let @awegrzyn @Barthelemy @vascobarroso to comment, but my understanding is that the updateable RPM paradigm the need, requires properly tagged versions.

Barthelemy: Indeed, as far as I know the hashes are not ok for the RPMs. They confuse the system as it does not know which version is more recent. If there is already a tag, indeed, we are already in a bad place. I would therefore propose the following (if @vascobarroso and @awegrzyn ) agree : keep the hash here to go ahead but in parallel could you ask FairRoot to tag and prepare subsequently a PR with a proper version ?

ihrivnac: @Barthelemy , I asked for a tag yesterday in my PR in FairRoot (this PR includes the update in CMake which we need for the developments in this PR, and which is included in the new hash in this PR). When a tag is available, we will move to it in a new PR.

ihrivnac:
@ktf, @michael,
The changes requested by @MichaelLettrich are applied, the change requested by @ktf is marked as resolved Is it now ok to merge or do you still need any action from me?

ihrivnac:
@ktf , the change you requested is marked as resolved. Can you approve?

ktf:
All tests passed before my minor cosmetic cleanups. Merging.

ktf:
Please, next time if you need to do such a large PR, make sure when you are happy you squash the commits and edit the commit message to the actual content of the PR. This time I did it, please check I did not miss anything.

ihrivnac:
@ktf, thanks for squashing. I use to do it just before merge to keep the list of commits available all time before merging. The commit message seems to me complete.

PR #2119, opened 2020-03-03
Add fastjet dependency to O2

qgp:


ihrivnac:
Hi @qgp ,
Could you give more details why this dependency is needed? Thank you.

qgp:
> Hi @qgp ,
> Could you give more details why this dependency is needed? Thank you.

Hi @ihrivnac,
we are working on the inclusion of jet finding (see AliceO2Group/AliceO2#3041), which will be needed for analysis but also for skimming and eventually triggering (at reconstruction level). Let me know if you need more information.
Cheers,
   Jochen

ihrivnac:
Thank you for the link, I don't need more info.

qgp:
Hi @ihrivnac,
Thanks! Do you know why build/O2/alidist failed (from the log it seems because of  gcc not working). Is there anything I can do concerning that? Or are you looking into it?

ihrivnac:
@qgp , I don't know why this is failing. It needs to be investigated on the builder itself by the administrators.

qgp:
@ktf any suggestion how to proceed here?

qgp:
Before requiring fastjet on macOS, we have to remove the requirement to have SIP disabled, for which in turn we need to verify that this does not break:
- O2
- AliPhysics with ROOT 6
- AliPhysics with ROOT 5
Therefore, I'd suggest to pull in fastjet only for non-macOS for now.

NB: I am working on testing fastjet with SIP but it will take a bit of time to get done all the compiles and tests.

qgp:
By now, I have tested fastjet with SIP enabled and have not observed any problems with it. Thus, I removed the requirement to have SIP disabled from fastjet (alisw/alidist#2128. Now, I suggestto go ahead and add fastjet as dependency to O2.

PR #2121, opened 2020-03-04
Enable rpath for arrow

qgp:


qgp:
> If it works, looks good to me.

Works on Linux, test on macOS is still running ...

PR #2123, opened 2020-03-05
Make xjalienfs paths in modulefile independent of INSTALLROOT

mfasDa:


mfasDa:
I think hard-coded paths in modulefiles (even though they are set at build level) are not the idea but instead they should be relative to $BASEDIR.

mfasDa:
Similar problem in the xrootd build recipe, should be fixed as well.

mfasDa:
@Atlantic777 @adriansev @ktf Please review! The hard-coded PYTHONPATH in the modules breaks portability of the packages.

adriansev:
i will let the experts to answer, i have zero knowledge of alibuild/alidist mechanics (and AFAIK there is no documentation about how these work)

Atlantic777:
@mfasDa the problem I had was that on some machines we build with python3.6 and on others with 3.7, and I discover the location of site-packages at build time and use the relative path. This is the situation when loaded on CVMFS for example:

```
/cvmfs/alice.cern.ch/el7-x86_64/Packages/xjalienfs/1.0.6-1/lib/python3.6/site-packages
/cvmfs/alice.cern.ch/el7-x86_64/Packages/XRootD/v4.11.1-20/./lib/python3.6/site-packages
```

I'm not aware of any problems on any platform tested (centos, ubuntu, macOS). What is the exact problem? What breaks and how to reproduce it?

mfasDa:
@Atlantic777 The problem is that in the modulefile you set the full path. Here an example from the modulefile on my laptop:

```
prepend-path PYTHONPATH /Users/markusfasel/alice/sw/osx_x86-64/xjalienfs/1.0.5-2/lib/python/site-packages
prepend-path PATH /Users/markusfasel/alice/sw/osx_x86-64/xjalienfs/1.0.5-2/bin
```

The result is that I cannot install the package (i.e. from TARS) in a different location (breaking portability). I was already wondering how you got it working on cvmfs, or with the pre-installed packages from the cache. Instead the paths should be relative to BASEDIR as alienv can create the absolute path from the BASEDIR and the relative path. Example after the patch:

```
set XJALIENFS_ROOT $::env(BASEDIR)/xjalienfs/$version
prepend-path PYTHONPATH $XJALIENFS_ROOT/lib/python/site-packages
prepend-path PATH $XJALIENFS_ROOT/bin
```

The workaround for the different python versions I took from a different build recipe (LHAPDF). What is done there is to set a symbolic link `python` to `python*`.

Atlantic777:
@mfasDa alright, I see. It must be that the aliBuild relaction does the right thing here. @ktf might have more info and additional comments on the relocation.

In the case of moving the files without relocating them I see how it breaks. I agree, it's better to use basedir and build the paths at runtime.

My only concern is that we loose Python version from the $PYTHONPATH and I hope that doesn't bring any hidden problems.

PR #2127, opened 2020-03-09
fixes the build error of QC on Mac

Barthelemy:
Fixes errors such as 
```
DEBUG:O2Suite:QualityControl:0: CMake Error at Framework/CMakeLists.txt:282 (add_executable):
DEBUG:O2Suite:QualityControl:0:   Target "testTriggerHelpers" links to target "OpenSSL::Crypto" but the
DEBUG:O2Suite:QualityControl:0:   target was not found.  Perhaps a find_package() call is missing for an
DEBUG:O2Suite:QualityControl:0:   IMPORTED target, or an ALIAS target is missing?
DEBUG:O2Suite:QualityControl:0: 
```

### DIFF
--- a/grpc.sh
+++ b/grpc.sh
@@ -60,6 +60,7 @@ module-whatis "ALICE Modulefile for $PKGNAME $PKGVERSION-@@PKGREVISION@$PKGHASH@
 module load BASE/1.0                                                          \\
             ${GCC_TOOLCHAIN_REVISION:+GCC-Toolchain/$GCC_TOOLCHAIN_VERSION-$GCC_TOOLCHAIN_REVISION} \\
             ${C_ARES_REVISION:+c-ares/$C_ARES_VERSION-$C_ARES_REVISION}        \\
+            ${OPENSSL_REVISION:+OpenSSL/$OPENSSL_VERSION-OPENSSL_REVISION} \\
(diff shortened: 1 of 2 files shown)


Barthelemy:
@knopers8 @awegrzyn @teo 
This was probably introduced when adding the dependency to occ-plugin.

awegrzyn:
I don't think setting `OPENSSL_ROOT` in `qualitycontrol.sh` is good practise as QC itself does not use OpenSSL at all

awegrzyn:
...the same applies to `control-occplugin.sh`

Barthelemy:
ok, I will try without it

awegrzyn:
Okay, so apparently `[[ ! $OPENSSL_ROOT ]] && OPENSSL_ROOT_DIR=$(brew --prefix openssl)` is present in many alidist recipe that do not use it directly.
@ktf Is there anyway to avoid that? Would it be possible to set it once in OpenSSL recipe?

ktf:
No, because that is for the case the OpenSSL recipe is not there.

Barthelemy:
I would propose to merge this PR in order to fix the build failure of QC and to create a dedicated ticket about the more general problem raised by Adam.

Barthelemy:
reminder

Barthelemy:
@alisw/infrastructure-admins I need one more approval...

PR #2131, opened 2020-03-12
Add ROOT_DYN_PATH to dynamic search path for ROOT

qgp:
This addition allows other packages (e.g. AliRoot, AliPhysics) to add paths for ROOT auto-loading of dictionaries and libraries without relying on LD_LIBRARY_PATH.

@ktf for discussion

### DIFF
--- a/aliphysics.sh
+++ b/aliphysics.sh
@@ -88,5 +88,6 @@ setenv ALICE_PHYSICS \$ALICE_PHYSICS
 prepend-path PATH \$ALICE_PHYSICS/bin
 prepend-path LD_LIBRARY_PATH \$ALICE_PHYSICS/lib
 prepend-path ROOT_INCLUDE_PATH \$ALICE_PHYSICS/include
+prepend-path ROOT_DYN_PATH \$ALICE_PHYSICS/lib
 EoF
 mkdir -p $INSTALLROOT/etc/modulefiles && rsync -a --delete etc/modulefiles/ $INSTALLROOT/etc/modulefiles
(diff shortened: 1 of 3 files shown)


ktf:
I was thinking that it might actually be enough to set ALICE_PHYSICS and ALICE_ROOT to empty in the root module, no?

qgp:
> I was thinking that it might actually be enough to set ALICE_PHYSICS and ALICE_ROOT to empty in the root module, no?

That would avoid the error messages for non-defined variables. But I am not sure that I find setting unrelated environment variables in root.sh more elegant. In addition, it introduces a forward dependency to packages using ROOT auto-loading, i.e. should we ever have another package that needs auto-loading, we would have to update root.sh and rebuild ROOT, instead of letting other packages add to ROOT_DYN_PATH.

ktf:
Ok, agreed. I suspect we do not have alternatives :-/

Barthelemy:
could you add 
```
prepend-path ROOT_DYN_PATH \$QUALITYCONTROL_ROOT/lib
```
to qualitycontrol.sh ?

qgp:
@ktf @Barthelemy I would suggest to first merge this as-is since this is what I could test. I would not know what to test for the QC part. We can easily add the setting in qualitycontrol.sh afterwards. Or you promise me that it's ok and I'll add it blindly here. Your call, let me know ..

Barthelemy:
no problem merging it as is and I will add it afterwards

Barthelemy:
@ktf FYI this works for me with SIP

qgp:
> @ktf FYI this works for me with SIP

@Barthelemy That's the whole point of this exercise, thanks for the confirmation :-)

qgp:
@ktf Should we merge this? Or are you still thinking about alternatives?

pzhristov:
It looks like this commit created the issues reported in https://alice.its.cern.ch/jira/browse/O2-1297

qgp:
@pzhristov This was followed up by #2169, are you sure the problem is still seen with up-to-date alidist?

PR #2133, opened 2020-03-13
Bump golang to 1.14

teo:


ktf:
genuine error?

teo:
> genuine error?

No, there's no failure in the log. The recipe in question is just an untar.

teo:
Merged in https://github.com/alisw/alidist/pull/2134

PR #2135, opened 2020-03-13
Fix usage of dependencies for fastjet

qgp:
@ktf @marcovanleeuwen If we don't consider CGAL from the system in `cgal.sh`, we should not fall back to a system installation here either. Therefore, use CGAL only if provided through aliBuild. If this is not what we want, we should implement a `prefer_system_check` in `cgal.sh`.

PS: This came up with people trying to build fastjet without cgal in order to be able to use valgrind. However, `--disable cgal` made fastjet (silently and inconsistently) fall back to the system installation.

### DIFF
--- a/fastjet.sh
+++ b/fastjet.sh
@@ -15,12 +15,14 @@ case $ARCHITECTURE in
   ;;
 esac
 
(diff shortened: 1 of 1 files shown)


qgp:
@marcovanleeuwen signed off (by email), so I think this is good to merge

qgp:
Anything pending/missing here or should we merge this?

PR #2136, opened 2020-03-16
root: bump to 6-20-02-alice1

ktf:


ktf:
@pzhristov this includes your patch.

davidrohr:
@ktf : We need also commit 3509af6ff507fb3bb3664161c788ffc87b5b38a0 : Do not complain about visible llvm symbols from ROOT master

PR #2137, opened 2020-03-16
Add QualityControl to ROOT_DYN_PATH

Barthelemy:
Important to find the libraries on Mac without SIP.


qualitycontrol.sh
```diff
@@ -122,6 +122,7 @@ prepend-path PATH \$QUALITYCONTROL_ROOT/bin
 prepend-path LD_LIBRARY_PATH \$QUALITYCONTROL_ROOT/lib
 prepend-path LD_LIBRARY_PATH \$QUALITYCONTROL_ROOT/lib64
 prepend-path ROOT_INCLUDE_PATH \$QUALITYCONTROL_ROOT/include
+prepend-path ROOT_DYN_PATH \$QUALITYCONTROL_ROOT/lib
```
ktf: Why not lib64?

Barthelemy: no idea :) I have just followed what was done in aliroot and aliphysics. If it should be lib64, we should modify them there as well.

PR #2141, opened 2020-03-20
Update apmon-cpp.sh [O2-1211]

costing:
Switch to the version that doesn't need an external rpc header


ktf:
@costing looks like there is some genuine errors. Did you compile with C++17?

costing:
Nope, but this is not different from what was already in place. How come it worked so far?

I'm trying to get rid of all throw-s in the code at the moment, give me a moment.

ktf:
Not sure, actually...

PR #2149, opened 2020-03-23
QC v0.24.1

Barthelemy:


Barthelemy:
@ktf @MichaelLettrich there seems to be a problem with the build/O2/alidist

MichaelLettrich:
I will check. Without looking at it I can say though that last week I had a couple of timeouts when checking out CERN hosted repositories and had to try again multiple times till they were checked out properly. Seems CERN infrastructure is  not always able to take the load that the massive telework is putting on it.

ktf:
Did you take my changes? Looks like it still fails...

Barthelemy:
@ktf  yes : https://github.com/AliceO2Group/QualityControl/compare/v0.23.1...v0.24.0

ktf:
Yeah, I missed one place, apparently... :-/ Will do a new PR.

ktf:
I took the liberty to tag 0.24.1 with the fix. Sorry about that.

Barthelemy:
There is another PR automatically generated when releasing QC.

PR #2158, opened 2020-03-27
Receipt for Delphes fast simulation framework

preghenella:
This PR adds a new receipt to build the Delphes fast simulation tool within the ALICE build system.

### DIFF
--- a/delphes.sh
+++ b/delphes.sh
@@ -1,0 +1,51 @@
+package: Delphes
+version: "%(tag_basename)s"
+tag: master
+requires:
+  - ROOT
+  - HepMC
+  - pythia
+build_requires:
+  - CMake
+  - "GCC-Toolchain:(?!osx)"
+source: https://github.com/alisw/delphes.git
+#prepend_path:
+#  LD_LIBRARY_PATH: "$O2_ROOT/lib"
+#  DYLD_LIBRARY_PATH: "$O2_ROOT/lib"
+#  ROOT_INCLUDE_PATH: "$O2_ROOT/include"
+---
+#!/bin/bash -ex
+cmake $SOURCEDIR -DCMAKE_INSTALL_PREFIX=$INSTALLROOT       \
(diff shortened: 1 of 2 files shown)


preghenella:
ciao @ktf 
I see the commit where you added `Delphes` to the `pagmc` codeowners.
I actually don't know what this means and how the file is treated, perhaps it is just a description?

In any case, it is fine with me if it is kept with `pagmc`, but I believe it would be more appropriate if put under `simulation`.

Roberto

preghenella:
Dear @pzhristov @ktf , 
it seems to me that the failure in the build checks is not related to this PR.
Can this be merged?
Thanks,
Roberto

delphes.sh
```diff
@@ -0,0 +1,51 @@
...
+# Dependencies
+module load BASE/1.0                                                                                \\
+            ${GCC_TOOLCHAIN_REVISION:+GCC-Toolchain/$GCC_TOOLCHAIN_VERSION-$GCC_TOOLCHAIN_REVISION} \\
+            ${PYTHIA_REVISION:+pythia/$PYTHIA_VERSION-$PYTHIA_REVISION}                             \\
+            ${HEPMC_REVISION:+HepMC/$HEPMC_VERSION-$HEPMC_REVISION}                                 \\
+            ${ROOT_REVISION:+ROOT/$ROOT_VERSION-$ROOT_REVISION}                                     \\
+
+# Delphes environment:
+set DELPHES_ROOT \$::env(BASEDIR)/$PKGNAME/\$version
+setenv DELPHES_ROOT \$DELPHES_ROOT
```
ktf, line 43: Do you really need this?

preghenella: yes, this is the way one will be able to point the location of the Delphes installation.
It will be needed to be able to find the Delphes "card" in case of running with provided cards.
It is necessary for developments that require Delphes classes.

ktf: I guess since we will never pick this up from the system, it's fine...

PR #2161, opened 2020-03-30
[O2-1263] Do not remove the lib/cmake from boost installation

aphecetche:
The net effect of this PR should be to solve a long standing issue : being able to compile our stack with our own boost, even if a system boost is installed (which has not been possible, at least on macOS, for a while now). That issue resurfaced in disguise in [O2-1263].
 
The key change is to "un-cripple" our version by _not_ removing in `boost.sh` the lib/cmake directory. As otherwise a "crippled" version (w/o lib/cmake directory) is never chosen by the CMake internal FindBoost module if a "not crippled" version exists instead ([more details](https://alice.its.cern.ch/jira/browse/O2-1263?focusedCommentId=245338&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-245338)).

The `-DBUILD_SHARED_LIBS=ON` additions in a couple of package is to be sure the boost system component is correctly found when using BoostConfig.cmake (which is now used as a result of the change above). I don't think it can cause problem, but maybe @sawenzel and @dennisklein can confirm ?

### DIFF
--- a/arrow.sh
+++ b/arrow.sh
@@ -78,8 +78,7 @@ cmake ./src_tmp/cpp                                                             
       ${PROTOBUF_ROOT:+-DProtobuf_PROTOC_LIBRARY=$PROTOBUF_ROOT/lib/libprotoc.$SONAME}              \
       ${PROTOBUF_ROOT:+-DProtobuf_INCLUDE_DIR=$PROTOBUF_ROOT/include}                               \
       ${PROTOBUF_ROOT:+-DProtobuf_PROTOC_EXECUTABLE=$PROTOBUF_ROOT/bin/protoc}                      \
-      ${BOOST_ROOT:+-DBoost_DIR=$BOOST_ROOT}                                                        \
-      ${BOOST_ROOT:+-DBoost_INCLUDE_DIR=$BOOST_ROOT/include}                                        \
+      ${BOOST_ROOT:+-DBoost_ROOT=$BOOST_ROOT}                                                       \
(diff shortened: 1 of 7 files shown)


aphecetche:
Seems I've missed quite a bit of instances where -DBUILD_SHARED_LIBS=ON is needed... I'm on it...

aphecetche:
moving to WIP while I check how many -DBUILD_SHARED_LIBS=ON I forgot...

aphecetche:
@ktf seems it's now ok in my local tests (can't do the full o2suite though because of qcg). But given the current number of failed alidist PRs, maybe it's not the right time to un-WIP this one ?

ktf:
Keep in mind that while WIP, it will not test it.

aphecetche:
Understood. 
But I guess it can wait a couple of days, right ? (and the merging of the PRs for e.g. CMake 3.17 and now the boost 1.72 ;-) )

ktf:
For the record boost 1.72 and cmake are now in.

aphecetche:
I've tried to build from scratch but : 

- now that we require boost 1.72 aliDoctor is happy to take it from the system, which I did not notice, so I'll have to redo it and explicitly forbid to take boost from the system...
- I get an error in FairMQ related to some asio stuff. 

so I guess I'll need to investigate more (next week)

aphecetche:
@ktf Redid a test and all seems fine on my side now. Removing the WIP to see if the CI confirms that.

pzhristov:
@ktf @Barthelemy @MichaelLettrich @knopers8 @aphecetche I think these changes are OK, should we merge this pull request?

PR #2164, opened 2020-03-31
Bump QCG version 1.6.11

graduta:


graduta:
@ktf  Seems like the `build/AliRoot/alidist` has no logs but it's failing. I checked the tag and the release of qcg and the name and version are correct. 

Any help would be appreciated. Thank you

awegrzyn:
@ktf Can I ask for force-merge?

PR #2166, opened 2020-03-31
Set CGAL_ROOT environment variable

preghenella:
This PR sets `CGAL_ROOT` enviroment variable. See discussion in 
https://github.com/alisw/rivet/issues/6.

### DIFF
--- a/cgal.sh
+++ b/cgal.sh
@@ -83,6 +83,7 @@ module-whatis "ALICE Modulefile for $PKGNAME $PKGVERSION-@@PKGREVISION@$PKGHASH@
 module load BASE/1.0 ${BOOST_REVISION:+boost/$BOOST_VERSION-$BOOST_REVISION}
 # Our environment
 set CGAL_ROOT \$::env(BASEDIR)/$PKGNAME/\$version
+setenv CGAL_ROOT \$CGAL_ROOT
 prepend-path PATH \$CGAL_ROOT/bin
 prepend-path LD_LIBRARY_PATH \$CGAL_ROOT/lib
 EoF


ktf:
We can merge this for now. But it's a bad idea, IMHO, as it will prevent us from picking up cgal from the system.

pzhristov:
OK, I assumed we do not want CGAL from the system and this was the motivation. @preghenella If this is not the case, why do you need this change?

preghenella:
this change was supposed to fix the issue reported in
https://github.com/alisw/rivet/issues/6

If a system has `CGAL`, will the `CGAL` package be loaded?

I would have expected not, so my reasoning is that
- if a system has `CGAL`, the `CGAL` package distributed with ALICE software will not be loaded
- if the ALICE `CGAL` is not loaded the variable `CGAL_ROOT` is not set

I give you the following example, based on my system.
My system has `GLFW`. It is not built by `aliBuild` because it is found on my system.
When loading `o2` enviroment, the package is not loaded and the `GLFW_ROOT` variable is not set.

In the `glfw.sh` receipt the `GLFW_ROOT` variable is set
```
set GLFW_ROOT \$::env(BASEDIR)/$PKGNAME/\$version
```

This looks to me like a similar mechanics as for the `CGAL` case modification of this PR.
Is this not the case?

Cheers,
Roberto

preghenella:
actually, when running `aliDoctor AliGenerators` on my system I find that `cgal` is in the list of packages that have to be anyway built 
```
 The following packages will be built by aliBuild because
     usage of a system version of it is not allowed or supported, by policy:
```
I don't know where is this policy set, but this sees to mean that in any case `cgal` from the system is not being used in any case, isn't it?

PR #2167, opened 2020-03-31
Re-enable FairMQ examples, they must be present in RPMs

teo:


MichaelLettrich:
Error on macOS CI is the same as in  #2168

PR #2171, opened 2020-04-02
Add build recipe for local CCDB application

mfasDa:


localccdb.sh
```diff
@@ -0,0 +1,72 @@
+package: localccdb
+version: "%(tag_basename)s"
+tag: "current"
+requires:
+  - JDK
+---
+#!/bin/bash -e
+
+URL="http://alimonitor.cern.ch/download/local.jar"
```
ktf, line 9: Can we at least have the version in the URL, so that we automatically refetch it when the version change in the recipe?

costing: The current version is http://alimonitor.cern.ch/download/local-1.0.5.jar . +1 to use this URL instead.

mfasDa: Of course I am also more in favour of using a versioned package. It is fixed now. Also the tag is now specified to v1.0.5

mfasDa:
If the PR is OK could you please merge it? I having the local CCDB as part of alidist could be helpful for other developers as well which are working with local test setups.

PR #2175, opened 2020-04-06
Fix for boost compilation

aalkin:


pzhristov:
Hi @aalkin,
I think we do not need anymore the part 
# This is to work around an issue in boost < 1.70 where the include path misses
# the ABI suffix. E.g. ../include/python3 rather than ../include/python3m.
# This is causing havok on different combinations of Ubuntu / Anaconda
# installations.
case $ARCHITECTURE in
  osx*)  ;;
esac

I removed it and the compilation worked on Ubuntu and MacOS.
Cheers,
Peter

aalkin:
@pzhristov it seems to have no effect on ubuntu, but on Fedora it is still needed.

pzhristov:
OK, but then I propose to replace the $(python3 -c 'import sysconfig; print(sysconfig.get_path("include"))')" ;; by the same expressionas proposed in root.sh $(python3-config --includes | cut -f1 -d' ' | cut -c3-)

pzhristov:
@aalkin On the other hand, we always can unify these scripts later. I will test on Ubuntu and Mac your patch and merge it.

PR #2176, opened 2020-04-06
[O2-1012] Fix for python mismatch on macOS 10.15.2 when compiling ROO…

pzhristov:
…T. Bump ROOT to v6-20-02-alice3

### DIFF
--- a/root.sh
+++ b/root.sh
@@ -1,6 +1,6 @@
 package: ROOT
 version: "%(tag_basename)s"
-tag: "v6-18-04-alice1"
+tag: "v6-20-02-alice3"
 source: https://github.com/alisw/root
 requires:
   - arrow
@@ -42,7 +42,7 @@ COMPILER_CC=cc
 COMPILER_CXX=c++
 COMPILER_LD=c++
 case $PKGVERSION in
-  v6-18*) 
+  v6-20*) 
      ENABLE_VMC=1
      [[ "$CXXFLAGS" == *'-std=c++11'* ]] && CMAKE_CXX_STANDARD=11 || true
      [[ "$CXXFLAGS" == *'-std=c++14'* ]] && CMAKE_CXX_STANDARD=14 || true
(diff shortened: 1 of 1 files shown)


root.sh
```diff
@@ -42,7 +42,7 @@ COMPILER_CC=cc
 COMPILER_CXX=c++
 COMPILER_LD=c++
 case $PKGVERSION in
-  v6-18*) 
+  v6-20*)
```
ktf, line 45: How about doing `v6-18*|v6-20*)` so that we can backport this, if needed?

pzhristov: I think we should completely abandon v6-18.

ktf: ok

davidrohr:
@pzhristov : I had a look at the failing tests, but these are not normal ctests failing, this is ROOT failing to interprete / compile macros.
This was working before, so this might be a regression in 6.20?
I would not simply disable them, but I'd try to understand what code fails to compile in 6.20 that worked with 6.18. Perhaps this needs a fix in ROOT.

pzhristov:
@davidrohr I made a pool request concerning the 4 failing O2 tests (out of 510), see 
 https://github.com/AliceO2Group/AliceO2/pull/3299
In the three cases the proble was in the O2 macro. What concerns macros/run_trac_its.C, there we include some files and end up using a huge json.h, where a template instantiation in swap doesn't work. I switched temporarily this test off.

davidrohr:
@pzhristov : I see, but I would rather try to fix this, perhaps getting rid of the json.h and moving it to a cxx. Because if the macro doesn't work with the new ROOT, it also means we cannot run it, which might affect ITS reconstruction.

pzhristov:
@davidrohr I propose to merge the changes and then work on the fix since it may take some time.

davidrohr:
ok, you are not breaking my code :) But who is working to fix it?

pzhristov:
I will try myself and contact the original authors in case I do not find quickly any solution.

PR #2177, opened 2020-04-06
Bump QCG patch version 1.7.1

graduta:


ktf:
Did we update readout? It seems broken.

awegrzyn:
@ktf I can't find any error in Readout? There's missing log from `AliRoot` and `O2` stops on FairMQ

PR #2179, opened 2020-04-07
Possibility to use defaults-o2 with AliRoot/AliPhysics

pzhristov:


defaults-o2.sh
```diff
@@ -5,13 +5,14 @@ env:
   CFLAGS: "-fPIC -O2"
   CMAKE_BUILD_TYPE: "RELWITHDEBINFO"
   CXXSTD: "17"
-  GEANT4_BUILD_MULTITHREADED: "ON"
+  GEANT4_BUILD_MULTITHREADED: "OFF"
   ENABLE_VMC: "ON"
```
ihrivnac, line 9: @pzhristov , this change will disable the O2 test with Geant4 MT. Can you try to put this flag back to ON, and change in the AliRoot g4Config.C macro(s):

    TG4RunConfiguration* runConfiguration 
      = new TG4RunConfiguration("geomRoot", 
                                "FTFP_BERT_EMV+optical", 
                                "specialCuts+stackPopper+stepLimiter",
                                 true);
into
```
    TG4RunConfiguration* runConfiguration 
      = new TG4RunConfiguration("geomRoot", 
                                "FTFP_BERT_EMV+optical", 
                                "specialCuts+stackPopper+stepLimiter",
                                 true, false);
```
The last TG4RunConfiguration argument meaning is "mtApplication", and setting it to "false" will invoke usage of Geant4 in sequential mode.  So in this way we can keep Geant4 libraries built in MT mode and run AliRoot in sequential.

pzhristov: @ihrivnac Yes, I will try. I switched the multithreading off since it caused a crash in the AliRoot tests.

PR #2180, opened 2020-04-08
Bump qualitycontrol to v0.25.0

Barthelemy:


ktf:
Issue on mac is legit.

ktf:
I think the best would be to use the proper int64_t formatting. See the first answer of:

https://stackoverflow.com/questions/9225567/how-to-print-a-int64-t-type-in-c

for a quick example. https://en.cppreference.com/w/cpp/types/integer has the full table.

Barthelemy:
@ktf thank you, I am trying to fix it here : https://github.com/AliceO2Group/QualityControl/pull/367
I tried to disable the errors for the formatting as a workaround but it did not work.

Barthelemy:
closing in favour of https://github.com/alisw/alidist/pull/2189

PR #2181, opened 2020-04-08
root: fix typo when using cmake

ktf:


ktf:
Fast merging because it's currently broken.

PR #2183, opened 2020-04-08
ROOT: handle corner case on centos7.

ktf:


ktf:
Currently broken. Merging.

PR #2184, opened 2020-04-08
root: roll back change to PYTHON_INCLUDE_DIRS

ktf:


ktf:
No, according to CMake documentation PYTHON_INCLUDE_DIR is what is to be used.

PR #2185, opened 2020-04-09
root: set CamelCased version of the CMake Python variable.

ktf:
This is either a bug in ROOT or in CMake, apparently.


ktf:
Fast merging since currently broken.

PR #2187, opened 2020-04-09
To build with G4 VMC v3-6-p6-inclxx-biasing-p2 tag

chiarazampolli:


chiarazampolli:
Ciao @ktf , 

I guess this change will then build the latest tag from 44 with the G4 VMC that we need. Right? 
Note that the last log of the branch mentions 44a, not 44e, and was by Dario. 

Cheers,

Chiara

chiarazampolli:
BTW, why the Travis CI failed?

PR #2191, opened 2020-04-15
Attempt at fixing python issue in the GPU build

ktf:


ktf:
@davidrohr let's see how this goes. I suspect the problem was introduced by Cython 3.0rc1 which was releases on the 12th and indeed is not pinned.

davidrohr:
@ktf : I tried this locally and it doesn't work. Cython is set to 2.29.16, and that one seems to be installed at first, but then I guess RootInteractive brings in a higher Cython requirement, and it tries to install 3.0rc.

ktf:
As far as I can tell root interactive is not setting any explicit requirement, so I do not understand. Also I guess you meant 0.29.16?

davidrohr:
> Also I guess you meant 0.29.16?

Sure... :)

The behavior I get is exactly the same as the failing AliRoot CI here. It installs the old cython and then the new one.

MichaelLettrich:
Can reproduce the error on Ubuntu 18.04 in a virtual environment trying to install `rootinteractive`. Seems the installer brings in packages that are not compatible.

ktf:
Seems that pre-installing cython with the working version does the trick...

python-modules.sh
```diff
@@ -34,6 +34,8 @@ mkdir -p $PYTHON_MODULES_INSTALLROOT
 # See also:
 #   https://github.com/scikit-garden/scikit-garden/issues/23
 grep RootInteractive requirements.txt && env PYTHONUSERBASE="$PYTHON_MODULES_INSTALLROOT" pip3 install --user -IU numpy
+# Do not move cython from 0.29.06 for now since 3.0.0rc1 breaks on GPU
```
davidrohr, line 37: For reference, it doesn't only fail in the GPU CI, but for me it also fails in the alisw/slc7-builder when I do not use precompiled builds.

ktf: Ok. Does this fix the issue for you?

ktf:
Ok, I tried also updating RootInteractive but that does not work as it introduces a dependency on ROOT. Will do the minimal fix for now.

PR #2192, opened 2020-04-16
Receipt to build STARlight

preghenella:
This is the receipt to build the STARlight event generator.

### DIFF
--- a/starlight.sh
+++ b/starlight.sh
@@ -1,0 +1,41 @@
+package: STARlight
+version: "%(tag_basename)s"
+tag: master
+build_requires:
+  - CMake
+  - "GCC-Toolchain:(?!osx)"
+source: https://github.com/alisw/STARlight.git
+---
+#!/bin/bash -ex
+cmake $SOURCEDIR -DCMAKE_INSTALL_PREFIX=$INSTALLROOT       \
+                 ${CMAKE_GENERATOR:+-G "$CMAKE_GENERATOR"} \
+                 -DCMAKE_BUILD_TYPE=$CMAKE_BUILD_TYPE      \
+                 -DCMAKE_SKIP_RPATH=TRUE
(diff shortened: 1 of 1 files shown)


starlight.sh
```diff
@@ -0,0 +1,42 @@
...
+  puts stderr "ALICE Modulefile for $PKGNAME $PKGVERSION-@@PKGREVISION@$PKGHASH@@"
+}
+set version $PKGVERSION-@@PKGREVISION@$PKGHASH@@
+module-whatis "ALICE Modulefile for $PKGNAME $PKGVERSION-@@PKGREVISION@$PKGHASH@@"
+# Dependencies
+module load BASE/1.0                                                                                \\
+
+# Delphes environment:
+set STARLIGHT_ROOT \$::env(BASEDIR)/$PKGNAME/\$version
+setenv STARLIGHT_ROOT \$STARLIGHT_ROOT
```
ktf, line 36: Please drop this.

preghenella: hi @ktf 
ok, I will drop it for the moment. But I am not sure it will be needed in the future if it emerges that the location of the library is needed for some reason.
In case I will open a new PR.

preghenella: no, sorry.
I already found a place where I need to link the {{STARlight}} library.
Please, hold on a sec and I check if I can actually deal with it without knowing the location of it.

ktf:
How about something like;
cat << EOF > $INSTALLROOT/bin/statlight-config
echo $INSTALLROOT/lib/libStarlight.so
EOF

preghenella:
well, this will make my life difficult in the `CMakeLists.txt` hat I need to build the `TStarLight` interface.

ktf:
you do not need it in the modulefile, which is for runtime environment. If you depend on STARLight at build time you will have all the variables you need coming from etc/profile.d/init.sh, which is automatically generated and sourced by alibuild.

PR #2194, opened 2020-04-19
Update STARlight receipt with config script. Install HepMC related files

preghenella:
This PR updates the `STARlight` receipt adding in the `INSTALL/bin` directory a script that can be sourced to set the `STARLIGHT_ROOT` environment variable, needed for several purposes
```
$ . startlight-config
```

It also installs header files and the scripts needed for conversion of `STARlight` output ASCII file into `HepMC2` format, for injection into the `o2` simulation.

### DIFF
--- a/starlight.sh
+++ b/starlight.sh
@@ -16,7 +16,15 @@ mkdir -p $INSTALLROOT/bin
 cp ./starlight $INSTALLROOT/bin/.
 mkdir -p $INSTALLROOT/lib
 cp ./libStarlib.a $INSTALLROOT/lib/.
-cp -r $SOURCEDIR/config $INSTALLROOT
+cp -r $SOURCEDIR/config $INSTALLROOT/.
+cp -r $SOURCEDIR/include $INSTALLROOT/.
+cp -r $SOURCEDIR/HepMC $INSTALLROOT/.
+
+#ConfigFile
+cat << EOF > $INSTALLROOT/bin/starlight-config
+echo $INSTALLROOT
+EOF
+chmod +x $INSTALLROOT/bin/starlight-config
 
 #ModuleFile
 mkdir -p etc/modulefiles


starlight.sh
```diff
@@ -16,7 +16,14 @@ mkdir -p $INSTALLROOT/bin
 cp ./starlight $INSTALLROOT/bin/.
 mkdir -p $INSTALLROOT/lib
 cp ./libStarlib.a $INSTALLROOT/lib/.
-cp -r $SOURCEDIR/config $INSTALLROOT
+cp -r $SOURCEDIR/config $INSTALLROOT/.
+cp -r $SOURCEDIR/include $INSTALLROOT/.
+cp -r $SOURCEDIR/HepMC $INSTALLROOT/.
+
+#ConfigFile
+cat << EOF > $INSTALLROOT/bin/starlight-config
```
ktf, line 24: Don't you need to print it, rather than exporting it? You also need to make it executable. Otherwise it would be just like `$INSTALLROOT/etc/profile.d/init.sh` which we already have.

preghenella: uhm, the fact that I don't have an environment variable that tells me where the installation is makes my life a bit complicated.
But I guess I can print and the eval the output of the script.
Or do something similar... need to try.

preghenella: ok, I have done it as you suggested.
I can always do the following at a later stage
```
export STARLIGHT_ROOT=$(starlight-config)
```

ktf: Yes, that's what well behaved software usually does.

What i still do not understand is what is your use case. Are you trying to build outside aliBuild? Are you trying to build on as the job lands on the worker node on the grid?

preghenella: for the time being I do not need to build.
I need to be able to copy some files that are installed under `$INSTALLROOT` to be able to run the full machinery. 

You can check this PR for the use case
https://github.com/AliceO2Group/AliceO2/pull/3382

Specifically this script
https://github.com/AliceO2Group/AliceO2/pull/3382/files#diff-c91eaf81f12db6c7dc8a87f5bf32bfbf

preghenella:
ciao @ktf 
is this PR good to go?
Thanks

PR #2195, opened 2020-04-20
Fix AliPhysics JAliEn dailies

Atlantic777:
Sync again defaults-next-root6 and defaults-jalien, and revert autotools version to v1.5.0 as a temporary solution for AliPhysics JAliEn daily builds.

### DIFF
--- a/defaults-jalien.sh
+++ b/defaults-jalien.sh
@@ -6,6 +6,7 @@ env:
   CXXFLAGS: "-fPIC -g -O2 -std=c++11"
   CFLAGS: "-fPIC -g -O2"
   CMAKE_BUILD_TYPE: "RELWITHDEBINFO"
+  USE_AUTORECONF: "false"
 overrides:
   AliRoot:
     version: "%(tag_basename)s_JALIEN"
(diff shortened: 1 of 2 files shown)


ktf:
Currently broken. Merging.

ktf:
Why WIP?

Atlantic777:
It still doesn't compile, more fixes needed (help2man).

Atlantic777:
@ktf please have another look.

defaults-jalien.sh
```diff
@@ -6,45 +6,16 @@ env:
   CXXFLAGS: "-fPIC -g -O2 -std=c++11"
   CFLAGS: "-fPIC -g -O2"
   CMAKE_BUILD_TYPE: "RELWITHDEBINFO"
+  USE_AUTORECONF: "false"
```
ktf, line 9: Is USE_AUTORECONF actually used? IIRC env does not apply for current package, but only for the dependent ones.

Atlantic777: It looks like it works, it's defined in a defaults file and seems to be visible for all packages.

```
DEBUG:autotools:autotools:v1.5.0: + source /alice/sw/SPECS/slc7_x86-64/autotools/v1.5.0-1/autotools.sh
...
DEBUG:autotools:autotools:v1.5.0: ++ USE_AUTORECONF=false
DEBUG:autotools:autotools:v1.5.0: ++ echo 'Building ALICE autotools. To avoid this install autoconf, automake, autopoint, texinfo, pkg-config.'
```

PR #2197, opened 2020-04-21
Always build Protobuf static

teo:


teo:
@ktf am I correct in assuming that the macOS issue is unrelated?

ktf:
What was the error?

teo:
@ktf it was thrown by the gRPC build, "unidentified symbol" linker errors for gRPC's abseil-cpp dependency. The dependency itself is built in, so if this next build (1.27.3 as before) succeeds, it means gRPC 1.28.1 doesn't build on Mac.

PR #2198, opened 2020-04-22
Bump qualitycontrol to v0.25.2

Barthelemy:


ktf:
Dataflow works, so merging this. Will have to investigate why the full O2 is taking so long (maybe just recompiling stuff due to more recent changes to autotools)?

aphecetche:
@Barthelemy thanks for the bump.
@ktf btw, do we have a way to monitor the build time(s) evolution ?

PR #2202, opened 2020-04-22
Bump ms_gsl

davidrohr:
@ktf @aphecetche

### DIFF
--- a/ms_gsl.sh
+++ b/ms_gsl.sh
@@ -1,6 +1,6 @@
 package: ms_gsl
-version: "1"
-tag: b014508
+version: "2.1.0"
+tag: v2.1.0
 source: https://github.com/Microsoft/GSL.git
 prepend_path:
   ROOT_INCLUDE_PATH: "$MS_GSL_ROOT/include"


aphecetche:
well, the bump might require some changes to AliceO2 first ... 

In file included from /Users/laurent/alice/pr/sw/SOURCES/O2/v1.2.0/0/Utilities/O2Device/include/O2Device/O2Device.h:33:
/Users/laurent/alice/pr/sw/SOURCES/O2/v1.2.0/0/Utilities/O2Device/include/O2Device/Utilities.h:92:22: error: use of undeclared identifier 'headerBufferSize'; did you mean 'headerBuffer'?
    span::index_type headerBufferSize{0};
                     ^~~~~~~~~~~~~~~~
                     headerBuffer

davidrohr:
FYI: I had tried 2.1 locally before on Linux, never on Mac, but while bumping I saw they recently released 3.0. Let's wait how the linux CI concludes.
And I do not understand this error message actually. This is a variable declaration, so why `undeclared identifier`.

davidrohr:
Apparently span::index_type was removed between ms_gsl 2.1 and 3.0. With 2.1 it is still working for me.
Shall I change the bump to 2.1 only, or do we want to fix it in O2?

ktf:
I would start with 2.1, since in the end only O2 depends on it.

davidrohr:
@sawenzel : Apparently, before the old ms_gsl was pulling in the headers needed for uint data type on macos, which is no longer the case. Could you update the MCTruthContainer to include all required data types by itself?

sawenzel:
ok

sawenzel:
https://github.com/AliceO2Group/AliceO2/pull/3428

PR #2203, opened 2020-04-23
root: Force dependency on glibc-headers for RPMS (O2-1369)

ktf:


ktf:
Waiting for the arrow bump before merging.

PR #2209, opened 2020-04-25
Fix Xdevel requirements on Ubuntu 20.04

preghenella:
In case someone wants to upgrade to Ubuntu 20.04.
This fixes the Xdevel cflags that cannot be obtained via `freetype-config` anymore.


preghenella:
ps: not sure it works on `cc7` or other systems though.
This PR is just to put ahead a potential solution for `Ubuntu 20.04`.

pzhristov:
Your fix works on MacOS, CC7, CC8, Ubuntu 20.04 and Ubuntu 19.10.

pzhristov:
The problem on Mac is not related to this fix.

PR #2213, opened 2020-04-27
root: bump to v6-20-02-alice4

ktf:


ktf:
Separate this from the arrow update.

ktf:
Merging this as ROOT itself works. Issues coming from CCDB.

PR #2216, opened 2020-04-28
fairmq: bump to v1.4.14

ktf:


ktf:
Required to register the shmem regions.

ktf:
All errors in CCDB tests. Merging since 1.4.4 has the callback which will allow registering shmem regions to use DMA to GPU.

PR #2224, opened 2020-04-30
autotools: get slc6 to build

ktf:


ktf:
Entered slightly different with a different PR.

PR #1842, opened 2019-09-26
o2checkcode: enable aliceO2-namespace-naming check

anerokhi:
@sawenzel I'd like to apply the following patch to `o2codechecker` after it is switched to clang 8 and before this PR is merged, so that namespaces outside `::o2::` are not checked.
```diff
diff --git a/aliceO2/NamespaceNamingCheck.cpp b/aliceO2/NamespaceNamingCheck.cpp
index 3f020dd..016b095 100644
--- a/aliceO2/NamespaceNamingCheck.cpp
+++ b/aliceO2/NamespaceNamingCheck.cpp
@@ -9,6 +9,7 @@
 
 #include "NamespaceNamingCheck.h"
 #include "clang/AST/ASTContext.h"
+#include "clang/ASTMatchers/ASTMatchersMacros.h"
 #include "clang/ASTMatchers/ASTMatchFinder.h"
 #include <regex>
 #include <string>
@@ -17,6 +18,13 @@
 using namespace clang::ast_matchers;
 
 namespace clang {
+namespace ast_matchers {
+AST_MATCHER_P(UsingDirectiveDecl, nominatedNamespace,
+              internal::Matcher<NamespaceDecl>, InnerMatcher)
+{
+  return InnerMatcher.matches(*Node.getNominatedNamespace(), Finder, Builder);
+}
+} // namespace ast_matchers
 namespace tidy {
 namespace aliceO2 {
  
@@ -42,18 +50,26 @@ bool isOutsideOfTargetScope(std::string filename)
 
 void NamespaceNamingCheck::registerMatchers(MatchFinder *Finder) {
   const auto validNameMatch = matchesName( std::string("::") + VALID_NAME_REGEX + "$" );
+  const auto inO2NSMatch = matchesName("^::o2::");
 
   // matches namespace declarations that have invalid name
   Finder->addMatcher(namespaceDecl(allOf(
+    inO2NSMatch,
     unless(validNameMatch),
     unless(isAnonymous())
     )).bind("namespace-decl"), this);
   // matches usage of namespace
-  Finder->addMatcher(nestedNameSpecifierLoc(loc(nestedNameSpecifier(specifiesNamespace(unless(validNameMatch)
-    )))).bind("namespace-usage"), this );
+  Finder->addMatcher(nestedNameSpecifierLoc(loc(nestedNameSpecifier(specifiesNamespace(allOf(
+    inO2NSMatch,
+    unless(validNameMatch)
+    ))))).bind("namespace-usage"), this);
[...]

PR #2230, opened 2020-05-05
Prevent LLVM double-link on linux with arrow 0.17

aalkin:
* Build only a single LLVM library in Clang
* Patch arrow linker version script to hide llvm symbols in gandiva.so


aalkin:
@ktf this solves the issue with "LLVM ERROR: inconsistency in registered CommandLine options"
@davidrohr does this affect GPU stuff that uses Clang/LLVM?

davidrohr:
It will not affect the GPU stuff, but it might affect the code checker @sawenzel .

davidrohr:
The O2 codechecker CI seems broken, e.g. in AliceO2Group/AliceO2#3533. It fails with `Unable to run clang-tidy.` This could be related to this change. ping @ktf @sawenzel

ktf:
When did the problem started to appear? Today? There is also https://github.com/alisw/alidist/pull/1842 which could be related and was merged yesterday (rather than 2 days ago).

davidrohr:
Not sure, I saw it today for the first time, but that does of course not exclude that the problem was already there yesterday. But you are right, #1842 could be the cause as well.

ktf:
It looks like it's indeed #1842. Reverting it brings back the tests to green. We should probably have some trivial test in o2codechecker.sh itself to make sure what is compiled actually runs.

PR #2232, opened 2020-05-05
bump readout version

sy-c:
v1.3.10


ktf:
Issue in readout seems a valid one. Can you check?

sy-c:
I don't see anything wrong in the logs. The build reaches 100% and installs fine

ktf:
You are right. It was a warning. Bumping.

PR #2235, opened 2020-05-07
Changes to allow fotran packages to build with gcc 10

aalkin:
* Adds `-std=legacy` to gfortran flags
* Adds relevant `allow` flags


pzhristov:
@aalkin The new options do not work with some of the old compilers. What is the platform that moved to gcc 10?

aalkin:
> @aalkin The new options do not work with some of the old compilers. What is the platform that moved to gcc 10?

I apologize, this was supposed to stay in draft state. Fedora 32 has gcc10 and I was adapting everything to it.

ktf:
What do we want to do with this? Is it ready to be reviewed?

aalkin:
I did not have time to update it, but it needs a version check so that new options are only used with gfortran in gcc10. It is not complicated, but I did not put a priority on it.

ktf:
Ok, closing this for now.

PR #2239, opened 2020-05-10
Add trigger rpm qcg

graduta:
Added a workflow in GH Actions which will be triggered if there are any chages on branch master on file `qcg.sh`

If there are any changes a request will be sent to Jenkins to trigger a built with provided parameters.

Compared to existing job, I changed parameter `--negociate` to `--anyauth` due to CURL dropping the option in the version that GH Action is using.


graduta:
Seems like `KERBEROS_HOSTNAME` is not defined in the secrets of the repository. 
@ktf Could we get this added?

awegrzyn:
I think we could have single workflow to trigger jenkins jobs: https://github.com/alisw/alidist/blob/master/.github/workflows/jenkins.yml
If you add `qcg.sh` to `paths` it will trigger `O2Suite` (which includes qcg)

graduta:
> I think we could have single workflow to trigger jenkins jobs: https://github.com/alisw/alidist/blob/master/.github/workflows/jenkins.yml
> If you add `qcg.sh` to `paths` it will trigger `O2Suite` (which includes qcg)

Hi, @awegrzyn.  I looked at that initially but I thought that running only `qcg` job would use less resources which would bring a benefit to the whole team. 

If that is not the case, I am more than happy to move the changes to the common file.

graduta:
Ok, so I merged the 2 jobs into one.

The only thing left to do is have `KERBEROS_HOSTNAME` added as a secret

awegrzyn:
@ktf Do you know what may happened to `KERBEROS_HOSTNAME` GitHub secret?

ktf:
I need to find out what we need for KERBEROS_HOSTNAME. Please create a jira ticket and assign it to me, so that I do not forget.

awegrzyn:
@ktf is that okay with you?

PR #2246, opened 2020-05-12
fairlogger: bump to 1.6.2

ktf:
Now depends on fmt

### DIFF
--- a/fairlogger.sh
+++ b/fairlogger.sh
@@ -1,21 +1,30 @@
 package: FairLogger
 version: "%(tag_basename)s"
-tag: v1.5.0
+tag: v1.6.2
 source: https://github.com/FairRootGroup/FairLogger
 build_requires:
  - CMake
  - "GCC-Toolchain:(?!osx)"
+ - "fmt"
 incremental_recipe: |
   cmake --build . --target install ${JOBS:+-- -j$JOBS}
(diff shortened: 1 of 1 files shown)


ktf:
@teo this fails on mac because of:

```
-- Configuring done
CMake Error at CMakeLists.txt:404 (add_library):
  Target "FairMQPlugin_OCC" links to target "fmt::fmt" but the target was not
  found.  Perhaps a find_package() call is missing for an IMPORTED target, or
  an ALIAS target is missing?
```

this now comes from FairLogger. Any idea on how to fix it?

teo:
No idea except adding the relevant `find_package` to OCC I'm guessing.

ktf:
@rbx @dennisklein could it be that you are not correctly exporting the `fmt::fmt` target?

dennisklein:
> No idea except adding the relevant `find_package` to OCC I'm guessing.

> @rbx @dennisklein could it be that you are not correctly exporting the fmt::fmt target?

See https://github.com/FairRootGroup/FairLogger#usage. Basically, we do intentionally not call `find_package(fmt)` in our CMake package because the user might have a direct dependency on fmt himself and might want to add stronger/different criteria to the `find_package` call.

teo:
I've added the `find_package` in `Control-OCCPlugin`, should be sorted out with https://github.com/alisw/alidist/pull/2247

ktf:
@sy-c it looks like you have the same issue in readout. Can you do the same fix?

PR #2251, opened 2020-05-14
o2: revert  cleanup of old by-products

ktf:
This is apparently creating issues in some corner case (modified CMakeList.txt?).


ktf:
@davidrohr

PR #2253, opened 2020-05-15
cgal: disallow Qt5

sawenzel:
* following same logic as for Qt3, Qt4
* prevents compilation error when Qt5 is installed and picked up


ktf:
+1000

PR #2260, opened 2020-05-19
Update cub.sh: Fix tag

sawenzel:
tag v1.8.0 does not exist in this repo. only 1.8.0.


ktf:
restating the builders which are failing.

PR #2261, opened 2020-05-20
Add recipes for libjalien packages

yuw726:
This PR introduces two libjalien packages, that can be used for websockets connection management and JAliEn credentials management.

- libjalienws allows the client to connect to JCentral services and to send JAliEn commands using websockets protocol.
- libjalienO2 takes care of loading JAliEn tokens and user certificates, searches for CA's and provides SSL context to let the client connect to the SSL-enabled endpoint.
- curl is no longer taken from the system. It is compiled using official curl build procedure to ensure compatibility with OpenSSL package, provided by aliBuild. This is done to allow libjalienO2 to correctly modify SSL context when using curl for making http requests.


libwebsockets.sh
```diff
@@ -23,11 +22,11 @@ cmake $SOURCEDIR/                                                   \
       -DLWS_WITH_STATIC=ON                                          \
       -DLWS_WITH_SHARED=OFF                                         \
       -DLWS_WITH_IPV6=ON                                            \
+      -DLWS_WITH_ZLIB=OFF                                           \
```
ktf, line 25: Why?

yuw726: We don't use the data compression in websockets, hence no need for this dependency. But just today I had a discussion with @costing that we should enable the compression. Maybe we should keep zlib here for future.

yuw726: Returned zlib back to facilitate enabling the compression in future.

ktf: Thanks. I think libwebsockets is also used somewhere else.

curl.sh
```diff
@@ -1,10 +1,39 @@
...
-system_requirement_missing: |
-  curl and its development package are missing from your system.
-   * RHEL-compatible systems: you will probably need "curl" and "curl-devel" packages.
-   * Ubuntu-compatible systems: you will probably need "curl" and "libcurl4-openssl-dev" (or "libcurl4-gnutls-dev").
-system_requirement: ".*"
-system_requirement_check: |
-  curl --version > /dev/null; if test $? = 127; then exit 1; else printf "#include <curl/curl.h>\nint main() {}\n" | cc -xc -lcurl - -o /dev/null || exit 2; fi; exit 0
+version: "7.70.0"
+tag: curl-7_70_0
+source: https://github.com/curl/curl.git
```
ktf, line 4: @sawenzel @Barthelemy is this fine for the CCDB API?

yuw726: Here I have to comment, that the whole libjalien development was started to let CCDB client use curl + JAliEn SSL features (tokens, CAs). I've tested this case and I am ready to make a PR to AliceO2 as soon as these recipes are good to be merged.

One more point here is that curl executable is used in scripts like freetype.sh:16 - it also works well (in my local and jalien-root CI tests), though there were some issues during development, now solved.

7.70.0 is the latest tagged curl version at the moment.

Barthelemy:
@ktf no idea. But the tests pass so I would say yes.

ktf:
Ok... Looks like we have something which breaks the builds... :-/ It was green a while ago... Investigating.

ktf:
macOS issue looks genuine. Did you test there?

yuw726:
> macOS issue looks genuine. Did you test there?

Good point, thanks. Added a fix for curl and one more for openssl variables in libjalienO2. However I couldn't build O2 because of XRootD error:

```
DEBUG:O2:XRootD:v1.2.0: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/cc -sdk macosx clang -arch x86_64 -DNDEBUG -g -fwrapv -O3 -Wall --std=c++0x -fPIC -O2 -isysroot /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk -I/Users/vyurchen/alice/sw/BUILD/72488163ff60a1e6149be2ce3f8d0cae7d5a1001/XRootD/src -I/Users/vyurchen/alice/sw/BUILD/72488163ff60a1e6149be2ce3f8d0cae7d5a1001/XRootD/build/src -I/Applications/Xcode.app/Contents/Developer/Library/Frameworks/Python3.framework/Versions/3.7/include/python3.7m -c /Users/vyurchen/alice/sw/BUILD/72488163ff60a1e6149be2ce3f8d0cae7d5a1001/XRootD/bindings/python/src/PyXRootDFile.cc -o build/temp.macosx-10.14-x86_64-3.7/Users/vyurchen/alice/sw/BUILD/72488163ff60a1e6149be2ce3f8d0cae7d5a1001/XRootD/bindings/python/src/PyXRootDFile.o -g
DEBUG:O2:XRootD:v1.2.0: clang: error: unknown argument: '-sdk'
DEBUG:O2:XRootD:v1.2.0: clang: error: no such file or directory: 'macosx'
DEBUG:O2:XRootD:v1.2.0: clang: error: no such file or directory: 'clang'
```

I get this error even with alidist master.

ktf:
There seems to be a merge issue. Can you please have a look?

yuw726:
Hi @ktf, I can't reproduce the error in build/O2/alidist check, and my attempts to blindly fix it are not successful. Can you please help me to debug it? How can I get logs of building libwebsockets and zlib? Is it possible to check where libz.so is located on the CI machine?

Also build/O2/alidist-centos8 complains about libInfoLogger, but I haven't touched this package at all. 
PR was rebased against master yesterday.

yuw726:
@ktf, I noticed that the error message point to the wrong zlib version:
`No rule to make target '... zlib/v1.2.8-1 ...'`
but
`DZLIB_ROOT=... zlib/v1.2.8-2`

Is it possible to clean up sw/* and restart the build from the scratch?

o2.sh
```diff
@@ -216,7 +218,9 @@ module load BASE/1.0 \\
             ${LIBUV_REVISION:+libuv/$LIBUV_VERSION-$LIBUV_REVISION}                                 \\
             ${FASTJET_REVISION:+fastjet/$FASTJET_VERSION-$FASTJET_REVISION}                         \\
             ${GLFW_REVISION:+GLFW/$GLFW_VERSION-$GLFW_REVISION}                                     \\
-            ${AEGIS_REVISION:+AEGIS/$AEGIS_VERSION-$AEGIS_REVISION}
+            ${AEGIS_REVISION:+AEGIS/$AEGIS_VERSION-$AEGIS_REVISION}                                 \\
+            ${LIBJALIENO2_REVISION:+libjalienO2/$LIBJALIENO2_VERSION-$LIBJALIENO2_REVISION}
+>>>>>>> Introduce libjalien websocket and O2 packages
```
ktf, line 223: Did you forget to remove a conflict?

yuw726: Sorry! The conflict was resolved, but I missed this message to remove.

yuw726:
@ktf @sawenzel @Barthelemy could you have a look at libInfoLogger build error? I know very little about this package and cannot understand how my change could make an impact on it.

ktf:
@Barthelemy @awegrzyn why is Common trying to download stuff?

awegrzyn:
InfoLogger lib had some issues with Common as transitive dependency, apparently it downloads `Common` directly from CMake (https://github.com/AliceO2Group/InfoLogger/blob/5edc7db171d45a8fd128b2d30dcdad92e5982adb/CMakeLists.txt#L58) instead to having it as dependency in alidist.

ktf:
Can we try to fix the actual issue?

awegrzyn:
I think this was @sy-c decision to go this way, so it'd be great he comments on that (but he's away atm).

awegrzyn:
I can confirm that this breaks on my C8 machine with SSL error: `error:14098044:SSL routines:ssl3_send_client_key_exchange:internal error`. I understand that InfoLogger misuses CMake to download Common package, but the actually issue is probably on OpenSSL side.

yuw726:
@awegrzyn @sy-c, do you have an idea how to proceed with libInfoLogger issue?

curl.sh
```diff
@@ -1,10 +1,41 @@
...
+proc ModulesHelp { } {
+  global version
+  puts stderr "ALICE Modulefile for $PKGNAME $PKGVERSION-@@PKGREVISION@$PKGHASH@@"
+}
+set version $PKGVERSION-@@PKGREVISION@$PKGHASH@@
+module-whatis "ALICE Modulefile for $PKGNAME $PKGVERSION-@@PKGREVISION@$PKGHASH@@"
+# Dependencies
+module load BASE/1.0 ${OPENSSL_REVISION:+OpenSSL/$OPENSSL_VERSION-$OPENSSL_REVISION}
+# Our environment
+prepend-path LD_LIBRARY_PATH \$CURL_ROOT/lib
```
awegrzyn, line 37: @yuw726 You are missing `set CURL_ROOT \$::env(BASEDIR)/$PKGNAME/\$version` before defining `LD_LIBRARY_PATH`.

awegrzyn:
https://github.com/alisw/alidist/pull/2261/files#r450054159

curl.sh
ktf: or you could use the automated generation of modulefiles... See the alibuild reference docs.

yuw726: Thank you for the comments, I've added corresponding lines.

yuw726:
Looks like the infoLogger issue is gone, but now python-modules have problems with finding matplotlib. @ktf

libwebsockets.sh
yuw726: I had to disable zlib support because of the build problems on ci machines. I suggest that I make a new pull request to enable zlib again later, but now merge this PR to have basic libjalien functionality.
libwebsockets is used only in JAliEn-ROOT, it won't affect other packages.

yuw726:
@ktf could you merge this PR please? I believe matplot error is not related to my changes

yuw726:
@ktf, help me understand the error:
```
gcc  -g -I./dim -Wall -Wextra -DMIPSEL -DPROTOCOL=1 -Dunix -Dlinux -fPIC -I./src/did -I/usr/local/include -I/usr/X11R6/include  linux/did.o linux/dui_util.o -L/mnt/mesos/sandbox/sandbox/sw/BUILD/d5700799aeadb3c2dcffdffad6eb34dc9fed667b/dim/dim_v20r26/linux -ldim -L/usr/local/lib -L/usr/X11R6/lib64  -lpthread -lMrm -lpthread -lXm -lXt -lX11 -o linux/did
/lib64/libpng16.so.16: undefined reference to `inflateValidate@ZLIB_1.2.9'
```

Does dim depend on libpng? If so, the recipe should have it in build_requires. System libpng points to a different zlib version, other than is built by us.

curl.sh
```diff
@@ -1,10 +1,36 @@
...
+
+./buildconf
+./configure --prefix=$INSTALLROOT ${OPENSSL_ROOT:+--with-ssl=$OPENSSL_ROOT} --disable-static
+make ${JOBS:+-j$JOBS}
+make install
+
+# Modulefile
+mkdir -p etc/modulefiles
+alibuild-generate-module > etc/modulefiles/$PKGNAME
+cat > etc/modulefiles/$PKGNAME <<EoF
```
ktf, line 28: This will delete the modulefile that you generate above. I guess it's not intended?

yuw726: Thanks, my intention is to append new lines to modulefile.

yuw726:
Hi Giulio, this has been completed and ready for a review since a long time. I've squashed the commits as you asked.

ktf:
So the issue here is that the curl you use is incompatible with the one the system git is using. Do you really need to use a custom curl on CC8?

yuw726:
A custom curl is required to remain compatible with OpenSSL v1.0.2, which is used by all packages in alidist. It is especially important for libjalienO2. The goal here is to use libjalien openssl features in O2 OCDB client, that uses curl.

How can I check the actual error? The log files on this github page show only "cmake libInfoLogger" errors.

ktf:
It's probably just a `git clone`.

ktf:
Ok, I've added system-curl and system-openssl to try to work around this. What I do not understand is where do you actually need curl? Nothing depends on it here.

yuw726:
O2 depends on curl. Since we would like to enable OpenSSL features for curl in O2 (by using libjalienO2), we need to make sure it is compiled with the same OpenSSL version as libjalienO2.

ktf:
ok, right. Then we need to add

```bash
requires:
  - "curl:(?!slc8)"
  - "system-curl:slc8.*"
```

pretty much everywhere we have curl...

yuw726:
Hi @ktf, you mentioned that you added system-curl and system-openssl, but I cannot find these scripts. Am I missing something?

PR #2264, opened 2020-05-22
bump readout v1.4.0

sy-c:


ktf:
@awegrzyn @sy-c dropping the "requires:" in this PR broke completely the CI because it treated the multiline list below as a tag, with very bad side effects. I will add a protection on alibuild side, but was it dropped by accident or by some automatic update of the tags?

sy-c:
This was obviously an accident editing the file in the browser... sorry !

ktf:
No worries. I have added https://alice.its.cern.ch/jira/browse/O2-1458 to track the underlying issue in aliBuild.

PR #2270, opened 2020-05-27
Bump AliECS to v0.13.94

teo:


ktf:
@aphecetche do you understand what is going on here?

aphecetche:
@ktf apart from the fact that the arrow people seems to have an issue with capitalisation ;-), no, not quite (looking at the build/O2/alidist log)). 

But as a first guess I would rule out the Arrow vs arrow as the issue here. As far as I remember I tested our changes on both Mac and on Linux to rule out case sensitivity issues.

What I really don't understand is the line : 
CMake Warning at dependencies/Findarrow.cmake:13 (find_package):

we no longer have such a file in our code in principle ?

ktf:
Ah, wait, could it be a leftover from a different PR, maybe?

aphecetche:
@ktf then it would mean the _source_ part of the PRs can also be mixed-up ?? I thought only the build area was causing problem (because of incremental building) ?

ktf:
I wonder if the problem is not the cmake cache...

PR #2273, opened 2020-05-28
datadist: bump to v0.7.6

ironMann:
GPU builder and FLPSuite (fmq) changes


ktf:
Merging, since we do not run DataDistribution on mac and the change was anyway trivial.

PR #2274, opened 2020-05-29
root: bump to v6-20-02-alice7

ktf:


ktf:
@jgrosseo the root patch seems to actually break our builds... Investigating...

ktf:
@MichaelLettrich could you please check what is going on with the macOS builder?

PR #2277, opened 2020-06-02
libuv: force installation in lib

ktf:


ktf:
Should fix issues on linux...

PR #2279, opened 2020-06-02
qualitycontrol: make sure libuv is picked up

ktf:
Will be needed by the DPL soon. Not having it prevents AliceO2Group/AliceO2#3686 being merged.

### DIFF
--- a/qualitycontrol.sh
+++ b/qualitycontrol.sh
@@ -32,10 +32,15 @@ incremental_recipe: |
 #!/bin/bash -ex
 
 case $ARCHITECTURE in
-    osx*) 
-		  [[ ! $BOOST_ROOT ]] && BOOST_ROOT=$(brew --prefix boost)
-		  [[ ! $OPENSSL_ROOT ]] && OPENSSL_ROOT_DIR=$(brew --prefix openssl)
-    ;;
+  osx*) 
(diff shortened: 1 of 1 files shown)


ktf:
@aphecetche do you understand why this is actually needed at all? Shouldn't the transitive closure of O2 dependencies be available to QualityControl CMake?

aphecetche:
@ktf it should, yes. 

A remark (not necessarily linked to the problem at hand though...) on libuv : I see that this lib is providing a pkgconfig file, so we may want to use that (using pkg_check_modules) instead of cooking our own Find module completely from scratch.

will try to have a look tomorrow

ktf:
Do you have an example for that?

ktf:
@Barthelemy @knopers8 could you please merge this so that I can continue with  AliceO2Group/AliceO2#3686 and  AliceO2Group/AliceO2#3689?

Barthelemy:
As discussed offline, `libuv` should also be loaded in the modules part at the bottom. @ktf proposed to add it to O2 which is fine with me.

aphecetche:
@ktf ok, things are moving fast ;-) is there still an actual problem to be solved here (what was the origin issue btw) or all is settled ?

ktf:
If you have an example with the pkg-config stuff, I would be happy to update the CMakeLists, but it's not urgent. This unblocks me so it's ok for now.

PR #2280, opened 2020-06-03
o2: add libuv to the runtime environment

ktf:


ktf:
Not actually being tested. Force merging.

PR #2287, opened 2020-06-05
Bump FairMQ to 1.4.19

davidrohr:
The alignment patch doesn't seem to work for me.
Still, let's test it with the CI and let @ktf confirm whether the shared memory issues he saw are fixed.


ktf:
@MichaelLettrich do you have any idea of what `continuous-integration/travis-ci` is?

PR #2288, opened 2020-06-08
Disable libpng support in freetype (O2-1497)

ktf:


ktf:
QualityControl issue unrelated.

PR #2294, opened 2020-06-11
Force dependency on setuptools and bump PyYAML

ktf:


ktf:
not really required anymore. Will bump when there is need for it.

PR #2295, opened 2020-06-11
glfw: treat like any other package (O2-1390)

ktf:
In any case GL needs to be there for ROOT, so we simply build
it, no need to make gymnastic which then breaks on non standard
installations.


ktf:
Checked locally that this builds fine also under alidist-dataflow now. Merging with the hope that it does not break centos8 builds, which are still ongoing. In any case it should be an easy fix.

PR #2302, opened 2020-06-16
Upgrade Rivet to 3.1.1

preghenella:
This PR upgrades Rivet to v3.1.1.
It comes also with an upgraded YODA.

### DIFF
--- a/defaults-generators.sh
+++ b/defaults-generators.sh
@@ -1,7 +1,7 @@
 package: defaults-dev
 version: v1
 env:
-  CXXFLAGS: "-fPIC -g -O2 -std=c++11"
+  CXXFLAGS: "-fPIC -g -O2 -std=c++14"
   CFLAGS: "-fPIC -g -O2"
   CMAKE_BUILD_TYPE: "RELWITHDEBINFO"
 
--- a/rivet.sh
+++ b/rivet.sh
@@ -1,6 +1,6 @@
 package: Rivet
 version: "%(tag_basename)s"
-tag: "2.7.2-alice8"
+tag: "3.1.1"
 source: https://github.com/alisw/rivet
 requires:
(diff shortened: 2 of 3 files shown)


preghenella:
the `build/AliGenerators/prod-latest` check fails.
From the log it seems to be that it is not taking the defaults for `generators`.
Could this be the case?

maireiphc:
Hi Roberto,

   if the build checks are in line with the official build, it should read alidist/default-generator.sh.
See for instance the build parameters used today : https://alijenkins.cern.ch/job/DailyBuilds/job/DailyAliGenerators/320/parameters/ (DEFAULTS field).

Could it be that the compiler used (gcc 7.3), despite normally compatible for c++14, is not "activated" or "set up properly" in our case with desired option(s) to deal with host c++14 features ?

My 2 cents

preghenella:
Hi @maireiphc ,
on alidock this works and only building with the updated `defaults-generators.sh`.
The complaint about `c++14` looks like the `--defaults generators` are not set in this check.

It is not clear to me that these checks are done with the same configuration as the one used for the Jenkins daily builds.

@ktf , can you please help us figure out what is the actual setting of the `build/AliGenerators/prod-latest` check? Thanks.

preghenella:
thanks to @ktf that pointed out to me that failed check was running with `--defaults prod-latest`.
They were in fact not including the `c++14` flag which should allow to build the upgraded Rivet.

@ktf modified the check to run with `--defaults generators`.
The check chain is supposed to start automatically.

preghenella:
well, I guess I have to trigger this somehow... 
the new checks are not starting

ktf:
They just take long because it's doing a full rebuild of all the PRs. Now done. Merging.

PR #2307, opened 2020-06-17
Fixes to build new FairLogger

davidrohr:
This must be merged before we bump the version, probably good to have it in already.


davidrohr:
@ktf : Can we merge this?

PR #2308, opened 2020-06-17
readout: allow for system installed zeromq

ktf:
Modulefile is not needed in that case.


ktf:
Currently broken when ZeroMQ is picked up from the system. Merging.

PR #2309, opened 2020-06-18
Bump FairLogger to 1.7.0 and required dependencies (Readout + DataDistribution)

davidrohr:


ktf:
@sy-c @davidrohr I guess we need a new tag for Readout.

davidrohr:
There is a working tag for readout already. I am waiting for the new tag of DataDistribution, which is needed as well. Then I'll change this PR to bump the 3.

davidrohr:
Please don't merge yet, this is just to test the CI, the explicit commit id in data distribution should be replaced by a tag.

davidrohr:
@ktf @sawenzel  : What do we do here. The mac compiler gives a warning which is treated as an error:
```
/System/Volumes/Data/build/ci_checks/alidist_checker_macos_o2_ci_0/sw/SOURCES/O2/2309/0/Utilities/DataFlow/src/TimeframeValidatorDevice.cxx:50:7: error: add explicit braces to avoid dangling else [-Werror,-Wdangling-else]
      LOG(ERROR) << "Expecting at least 2 parts\n";
      ^
/System/Volumes/Data/build/ci_checks/alidist_checker_macos_o2_ci_0/sw/osx_x86-64/FairLogger/v1.7.0-2/include/fairlogger/Logger.h:375:13: note: expanded from macro 'LOG'
```
The LOG macro expands to some if... else without braces. It is correct as is, and it should work in all cases. Actually, the compiler will only complain in cases where we are missing braces in O2, i.e.
```
if (foo)
    LOG(debug) << "foo";
```
So that is forbidden by the coding guidelines anyway. Should we just clean this up in O2 generally? Or should I try to find a way around in FairLogger. I believe I can do it somehow using `while` or `for` instead of `if`, or shall we disable this warning in the CI?

ktf:
I would just clean that up. Chances are that we do not have many in any case.

davidrohr:
> I would just clean that up. Chances are that we do not have many in any case.

OK, I am now building locally in a MacVM to find all occurrences. In any case good to fix them.

davidrohr:
@ktf : It finally works, would you be so kind :)

ktf:
I was on the phone and it does not let me to merge with missing approvals...

davidrohr:
@ktf @rbx : There is a stupid problem with FairLogger 1.7:
Apparently, FairLogger uses `LOGF` for `fmt::format` and `LOGP` for `fmt::printf` while O2 was so far using `LOGF` for `fmt::printf` and `LOGP` for `fmt::format`.
Actually, O2 was so far not using the FairLogger `LOGF` and `LOGP` at all, since FairLogger was at 1.5, but we were using @ktf 's fallbacks.
So I think one of the packages has to adapt and swap all `LOGF` / `LOGP`.
Any opinions? And either we have to do this soon, or we should revert the bump in the meantime.

davidrohr:
Alternatively, we can for now use `FAIR_NO_LOGP` and `FAIR_NO_LOGF` and swap them in Framework/Logger, but I would only consider this a hot fix.

rbx:
Initial development for this was triggered somewhere around this ticket:
https://github.com/AliceO2Group/AliceO2/pull/2161
where the first sentence of the ticket defines LOGF as the printf variant ("former"). This at least makes sense to me, but only because the **f**s align^^ (LOG**F**, print**f**). Similarly, I took the LOGP name from that ticket, following the first sentence, although without a reasoning for the meaning of 'P' (perhaps P as **P**ython?).

When implementation in FairLogger was added we also asked for Giulio's comments:
https://github.com/FairRootGroup/FairLogger/pull/20 

Anyway, if there is a good fundamental reason which methods fits better to which macro name, I am prepared to change it in FairLogger, even if it means breaking whoever might be using it. Otherwise I would like to keep it as it already is.

rbx:
Ok, now I see the problem, it is a typo in FairLogger code.

davidrohr:
I think it was LOGP for `fmt::**p**rint` (which is the same as `fmt::format` I guess) and LOGF for `fmt::print**f**`. This is also what was done in AliceO2Group/AliceO2#2161. In FairRootGroup/FairLogger#20 it is the other way around, but apparently noone noticed :)
Your implementation is also consistent with your documentation in https://github.com/FairRootGroup/FairLogger#2-additional-macros

I just checked, it is used ~400 times in O2, but I think it is rather easy to swap. Don't know how many people use the FairLogger syntax. So one of us has to swap it, don't have a strong opinion. Let's wait for @ktf to comment.

rbx:
Now I remember, the 'P' was for **p**rintf and 'F' was for **f**mt. Looks like I forgot it meanwhile, and the names are not obvious to me. I am fine with any solution you prefer.

rbx:
FairLogger syntax is not used in FairMQ and afaik not in FairRoot.

ktf:
`P` was python syntax, `F` print`f`..

rbx:
`P` for `P`ython-like syntax makes more sense than aligning the letters to a function name. I'll change it in FairLogger.

rbx:
FairLogger v1.8.0 aligns both macro names with O2.

PR #2310, opened 2020-06-18
Add missing fmt dependencies

davidrohr:


davidrohr:
fairlogger needs fmt for the formatting, and apparently they do not expose the dependency by design. I have asked what is the reasoning behind, let's see. But if they do not change it, we have to add it ourselves.

ktf:
alibuild adds it. It's FairLogger cmake file which does not export it.

davidrohr:
ah, it does, then I think we don't even need these changes here, except for the ones for fairlogger.sh for system fmt.

ktf:
Since it's an explicit dependency, I would keep it.

PR #2311, opened 2020-06-19
datadist: bump for fmt change

ironMann:
@davidrohr this should do it


davidrohr:
@ironMann : thx!
@ktf : Can you merge, there seems to be another problem with includes in Readout... I'll follow that up independently in #2309

davidrohr:
Ah, apparently I can merge after approved by admins :|

ktf:
You can only merge after being approved by a codeowner. Maybe we should review the codeowners for DataDistribution?

PR #2325, opened 2020-06-25
Bump and patch gRPC to v1.30.0-alice1

teo:
also minor Protobuf bump

### DIFF
--- a/grpc.sh
+++ b/grpc.sh
@@ -1,6 +1,6 @@
 package: grpc
 version: "%(tag_basename)s"
-tag:  v1.27.3
+tag:  v1.30.0-alice1
 requires:
   - protobuf
   - c-ares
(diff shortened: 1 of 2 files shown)


awegrzyn:
Is it really necessary to force-merge when 3 most important jobs fail? :(

teo:
@awegrzyn 1 of those was a real failure which I failed to recognize. My bad.

https://github.com/alisw/alidist/pull/2331 should fix it.

awegrzyn:
I see 3 "X" which means 3 real failures.

awegrzyn:
@ktf Is there any way to fix `test_Framework_test_AnalysisDataModel`? It failed on both CC7 and C8.

ktf:
It's already fixed.

PR #2326, opened 2020-06-25
Adapt pyroot feature string based on ROOT version

mfasDa:
Add backward support for pyroot feature for old root
version (< v6.20). In order to determin root version
properly split PKG_VERSION into MAJOR, MINOR, PATCH
and ALICE version. Select "pyroot" in case of ROOT6
and version >= 6.20 and "python" in case of ROOT6
and version <= 6.18.

### DIFF
--- a/root.sh
+++ b/root.sh
@@ -84,10 +84,25 @@ if [[ $ALIEN_RUNTIME_VERSION ]]; then
 fi
 [[ $SYS_OPENSSL_ROOT ]] && OPENSSL_ROOT=$SYS_OPENSSL_ROOT
 
+#determine ROOT version for PYTHON check
+ROOT_VERSION_MAJOR=$(echo $PKGVERSION | cut -d '-' -f1)
+ROOT_VERSION_MAJOR_NUM=${ROOT_VERSION_MAJOR:1:1}
+ROOT_VERSION_MINOR=$(echo $PKGVERSION | cut -d '-' -f2)
+ROOT_VERSION_PATCH=$(echo $PKGVERSION | cut -d '-' -f3)
+ROOT_VERSION_ALICE=$(echo $PKGVERSION | cut -d '-' -f4)
+PYROOT_FEATURE=
+if [ $ROOT_VERSION_MAJOR_NUM -eq 6 ]; then
+  if [ $ROOT_VERSION_MINOR -ge 20 ]; then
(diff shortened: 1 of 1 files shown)


mfasDa:
@ktf There were problems observed with the RooUnfold package leading to reproducible crashes since root version 6.18, which were leading to the failing unit tests which are by now disabled. The problem is clearly in the RooUnfold package itself and not in user code. The RooUnfold package is quite fundamental for jet analyses and the root6 defaults, which were deleted when going to root 6.20, should be re-established with root version 6.16. This commit, adapting the root build recipe, is a preparation for this, because builds using 6.16 fail due to the pyroot feature check.

ktf:
What do we need ROOT 6.16 for?

mfasDa:
Explained in my previous comment: RooUnfold is not properly work with ROOT v6.18 and higher. Users relying on RooUnfold need to be able to downgrade to v6.16 for their local installation in order to be able to use RooUnfold there. This means restoring the (user-)root6 defaults.

ktf:
I think this should be brought up in the computing board and discussed there. We cannot keep going back to old ROOT because some analysis cannot adapt their code. Is the actual problem with RooUnfold being investigated?

mfasDa:
I am not asking to deploy a build with ROOT 6.16 on the grid but to adapt the build recipe in a way that users can again build AliPhysics with this ROOT version, because the current build recipe is incompatible with that version. Also re-establishing the root6 defaults - which existed until half a year ago - should be only for local purpose. I don't see why this is not going to work - I use this with the latest AliPhysics master tag without problems. What concerns the maintenance the package is poorly maintained and it would be on us to fix the problem. RooUnfold is part of our software stack! Also here again: I can hack the build recipe as much as I want locally, but there are also less experienced users which rely on RooUnfold and cannot work anymore because RooUnfold is crashing all the time.

pzhristov:
@mfasDa Let me try go debug the issue with RooUnfold in Root 6.20.02. As I wrote in the mail exchange, it is better to fix the issue than to rely on an old (unsupported) Root version.

ktf:
@mfasDa if people can run it locally, they will ask for a production with it.

mfasDa:
FIx in RooUnfold seems to work, so the support for old root versions in build recipe is no longer needed. The PR can be closed.

PR #2328, opened 2020-06-26
QC v0.26.4 - Fix for compatibility with O2

Barthelemy:


Barthelemy:
@ktf @knopers8

ktf:
I assume the fixes which went into 0.26.3 / 0.26.4 are also in 1.0.0?

Barthelemy:
yes, you were correct to close this one, we did not need it in the end.

PR #2331, opened 2020-06-26
Disable C sharp extension for gRPC

teo:


ktf:
Force merging since it's currently broken.

PR #2334, opened 2020-06-27
grpc: fix building on macOS

ktf:


ktf:
@teo, this should fix gRPC on mac. 

I noticed that you use embedded copies of some gRPC dependencies. In general we should avoid relying on them and use properly packaged ones (eventually as archive libraries, if you want to remain self contained for the AliECS stuff).

ktf:
Merging without waiting for approval, since the tests are green and the without it we have macOS broken.

mfasDa:
Something is screwed up with the module file:
```
Dez 01 18:16:42 pc079 bash[303057]: WARNING: not updating modulefiles
Dez 01 18:16:42 pc079 bash[303057]: Use alienv list to list loaded modules.
Dez 01 18:16:42 pc079 bash[303057]: Module ERROR: can't read "ABSEIL_ROOT": no such variable
Dez 01 18:16:42 pc079 bash[303057]:   In '/software/markus/qcg/sw/MODULES/ubuntu1804_x86-64/abseil/20200225.2-1'
```
You use ABSEIL_ROOT but define GRPC_ROOT

ktf:
Fixed #2652

PR #2336, opened 2020-06-29
Patched pythia8 had a typo, bump to tag with fix

preghenella:


ihrivnac:
@preghenella ,
The pythia version is overridden in defaults-o2.sh; so it needs to be changed also here to be taken into account when building O2.
Cheers,

preghenella:
Hi @ihrivnac ,
the patch we got from the authors apply to Pythia8.2 series.
The version used in the O2 (8.3 series) would probably need a different patch.
We are not taking care of that because the issue shows only in a very specialised use of Pythia8 features.

ChristianTackeGSI:
8.3 can be patched quite the same. At least for 8301 and 8302 the patch applies with fuzz 2 and is fine.

PR #2341, opened 2020-07-01
xjalienfs.sh:: update tag to 1.2.0

adriansev:
Quite a long list of changes and internal change of python module location name prompted to update version to 1.2.0
Moreover now the package is also found in pypi under the name of `alienpy`


ktf:
@alisw/alien-experts to sign off...

PR #2349, opened 2020-07-03
Request to add Future python package to module list

KalanaDananjaya:
ApMon needs `past.utils` function from the `future` package to run. As this package is not available on the nodes, it is not possible to run python implementation of ApMon on the nodes at the moment. This PR is a request to add this package to the python modules list in CVMFS in order to get ApMon Python implementation get working on the nodes.
@Atlantic777 can you check this please?

### DIFF
--- a/python-modules-list.sh
+++ b/python-modules-list.sh
@@ -25,6 +25,7 @@ env:
     dryable==1.0.3
     responses==0.10.6
     RootInteractive==0.0.10
+    future==0.18.2
 ---
 # Modulefile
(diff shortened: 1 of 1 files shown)


Atlantic777:
Yup, looks fine. cc @costing @zensanp

ktf:
Can't we simply fix ApMon to use Python3?

KalanaDananjaya:
Closing because ApMon was fixed to work with Python 3 without external dependencies

PR #2350, opened 2020-07-03
Bump HepMC3 to 3.2.2

preghenella:
This PR bumps `HepMC3` to tag version 3.2.2.
It comes with an incompatibility with the current `o2` software, so the check will fail.
Please, refer to https://github.com/AliceO2Group/AliceO2/pull/3941 for further details.

For this PR to be successful, merging with the above `o2` PR must be simultaneous.

### DIFF
--- a/hepmc3.sh
+++ b/hepmc3.sh
@@ -13,7 +13,8 @@ build_requires:
 cmake  $SOURCEDIR                          \
        -DROOT_DIR=$ROOT_ROOT               \
        -DCMAKE_INSTALL_PREFIX=$INSTALLROOT \
-       -DCMAKE_INSTALL_LIBDIR=lib
+       -DCMAKE_INSTALL_LIBDIR=lib          \
+       -DHEPMC3_ENABLE_PYTHON=OFF
 
 make ${JOBS+-j $JOBS}
 make install
--- a/hepmc3.sh
+++ b/hepmc3.sh
@@ -1,6 +1,6 @@
 package: HepMC3
-version: "v3.0.0-git_%(short_hash)s"
-tag: d43693ce0e7731e9b787dbd6176cb6245fd770b3
+version: "%(tag_basename)s"
+tag: 3.2.2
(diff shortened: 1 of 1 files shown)


preghenella:
the incompatibility issue of the `o2` is resolved and as soon as https://github.com/AliceO2Group/AliceO2/pull/3941 will be merged this PR should succeed. At that point one should re-trigger the checks.

ktf:
It also looks like it installing stuff on the system:

```
-- Installing: /System/Volumes/Data/build/ci_checks/alidist_checker_macos_o2_ci_0/sw/INSTALLROOT/e9c24da4f16830bd9da39856bd7925b81b8a0320/osx_x86-64/HepMC3/3.2.2-1/include/HepMC3/ReaderRoot.h
CMake Error at python/cmake_install.cmake:44 (file):
  file cannot create directory: /Library/Python/2.7/site-packages/pyHepMC3.
  Maybe need administrative privileges.
  cmake_install.cmake:104 (include)
```

can you check?

preghenella:
ciao @ktf ,
it does not happen on my system (alidock).
Perhaps this is because my alidock is not picking Python from the system?

Still, it does not look good, I will look into it.

preghenella:
hi @ktf ,

looks like HepMC3 is trying to install a Python module in the Python master directories (wherever Python is installed).
I don't find a way to steer away from this behavior, there is discussion on the GitLab of the project about this with people not happy, but apparently the solution is just to disable Python support.

Anyway, I don't think we are using HepMC3 Python support (which is a new feature), so it is ok.

preghenella:
looks like we are green here.
@maireiphc can you check and approve in case?
thanks

ktf:
Ping @maireiphc

preghenella:
this is not green anymore, but it looks like the failure in `build/O2/alidist-centos8` is unrelated.
@maireiphc can you please approve the PR if fine with you?
I unfortunately cannot self-approve it

ktf:
Yes, it's indeed unrelated.

PR #2353, opened 2020-07-06
Fix tracy in case freetype comes from the system

ktf:


ktf:
@aalkin @davidrohr

davidrohr:
@ktf : `TBB= ` doesn't disable tbb for me

aalkin:
It still needs the correct include path for system freetype. Adding '-ltbb'  to EXTRA_LIBS works for me, but it may be not useful for systems without it.

ktf:
In principle tbb is not a requirement, so I'd rather disable it. I think TBB= does not work because it's considered as not set and therefore not overriding the default.

davidrohr:
Also `TBB=off` doesn't work for me. And I also have the same issue @aalkin just reported for the system freetype headers

ktf:
This should fix the system freetype issue. For TBB and GLEW I guess we still need something better...

PR #2355, opened 2020-07-07
DebugGUI: remove need for glew

ktf:


ktf:
@aalkin @davidrohr could you check if this drops the need for GLEW?

davidrohr:
@ktf : This fixes the GLEW issue. The only problem I am left with now is the TBB one.

ktf:
Currently broken. Merging.

ktf:
@davidrohr re TBB I've still to find where it is used. Grepping does not help. I suspect we can similarly just disable support for it, since in any case we use libuv for the multithreading...

aalkin:
@ktf ROOT has its own libtbb, can that be a problem? Compilation of tracy fails at link stage, not because it can't find tbb includes, so I assume the problem that it is linked to a lib, that already links to tbb, and not uses it directly.

davidrohr:
@ktf:
So, what happens is:
`tracy/server/TracySort.hpp` has
```
#if ( defined _MSC_VER && _MSVC_LANG >= 201703L ) || __cplusplus >= 201703L
#  if __has_include(<execution>)
#    include <algorithm>
#    include <execution>
#  else
#    define NO_PARALLEL_SORT
#  endif
#else
#  define NO_PARALLEL_SORT
#endif
```
GCC 9's `execution` header includes GCC's `pstl/parallel_backend.h`, which provides the parallel backend via `pstl/parallel_backend_tbb.h` if available (checked via `__has_include(...)`, and disables the parallel execution otherwise.

That brings the tbb headers in, which overload some memory allocation stuff, which is then finally missing at linking stage if -ltbb is not provided.

I have no idea why gcc is not linking to -ltbb if it pulls in the headers by itself...
This is either a bug in the tracy makefiles or in GCC. Just disabling the `<execution>` code path above fixes it for me, but I think this is not a proper fix.

And, actually, the tracy makefiles are already trying to detect -ltbb and add it if available, but apparently they are doing it incorrectly.
`tracy/profiler/build/unix/build.mk` does `ld -ltbb -o /dev/null` which fails, because _start is not defined.
Probably the correct thing to do is to fix the tracy build file here.

aalkin:
@davidrohr adding '-ltbb' to EXTRA_LIBS will work, I have it as a workaround. `ld -ltbb` would still return 0 even if it complains that _start is not defined.

ktf:
@davidrohr agreed. I will fix the build file in a bit.

davidrohr:
Ah, didn't check that in detail, but if it finds -ltbb, why isn't it adding the lib correctly?

aalkin:
Maybe supplying $LIBS variable overrides the check

PR #2357, opened 2020-07-09
[TEST] List install/lib directory after build

preghenella:


preghenella:
@ktf this PR modified the YODA receipt, it should then trigger a new build on Jenkins (I guess).
It add a printout of the content of the install/lib directory.

yoda.sh
```diff
@@ -22,6 +22,9 @@ make -j$JOBS
 make install
 )
 
+### list the content of lib directory
+ls -ld $INSTALL_ROOT/lib/*
```
ktf, line 26: `ls -ld $INSTALLROOT/lib`

preghenella: thanks @ktf , my bad.
Although I fear that missing * will not list the files in the directory

PR #2359, opened 2020-07-10
Drop GH actions triggering RPM rebuild

awegrzyn:


ktf:
Why?

ktf:
You mean you are not using this?

awegrzyn:
Sorry did not see the first message. Actually once we have O2Suite RPM "trigger" then this will be useful.

PR #2360, opened 2020-07-13
Add support for env variables that will force enabling CUDA or HIP GPU builds

davidrohr:


davidrohr:
@ktf : Are these env variables ok for you? I need something to steer the compilation on the different test servers... and to give to people to force enabling this, and hacking the o2.sh is a bit failure-prone.

ktf:
While I do not like the idea, I do not have a better solution to propose at the moment...

PR #2361, opened 2020-07-13
Bump to 3.1.1-alice1 patched version, plus other fixes

preghenella:
This PR bumps Rivet to the tagger version `3.3.1-alice1`.
This specific version contains a fix for a few relocation issues that are encountered when running `rivet-build`.

Another issue that is fixed by this PR is the missing `$GMP_ROOT` variable, because none of the modules are loading `GMP`.
So, to make sure of their presence, both `CGAL` and `GMP` modules are explicitly loaded.

Another issue that is fixed is due to the potential missing presence of `kpsewhich`, which makes the loading of the module fail.
For instance, in `alidock` it is not there. A check for the presence of it is added.

### DIFF
--- a/rivet.sh
+++ b/rivet.sh
@@ -1,6 +1,6 @@
 package: Rivet
 version: "%(tag_basename)s"
-tag: "3.1.1"
+tag: "3.1.1-alice1"
 source: https://github.com/alisw/rivet
 requires:
   - GSL
@@ -87,7 +87,7 @@ proc ModulesHelp { } {
 set version $PKGVERSION-@@PKGREVISION@$PKGHASH@@
 module-whatis "ALICE Modulefile for $PKGNAME $PKGVERSION-@@PKGREVISION@$PKGHASH@@"
 # Dependencies
-module load BASE/1.0 ${GSL_REVISION:+GSL/$GSL_VERSION-$GSL_REVISION} YODA/$YODA_VERSION-$YODA_REVISION fastjet/$FASTJET_VERSION-$FASTJET_REVISION HepMC/$HEPMC_VERSION-$HEPMC_REVISION
+module load BASE/1.0 ${GSL_REVISION:+GSL/$GSL_VERSION-$GSL_REVISION} ${CGAL_REVISION:+cgal/$CGAL_VERSION-$CGAL_REVISION} ${GMP_REVISION:+GMP/$GMP_VERSION-$GMP_REVISION} YODA/$YODA_VERSION-$YODA_REVISION fastjet/$FASTJET_VERSION-$FASTJET_REVISION HepMC/$HEPMC_VERSION-$HEPMC_REVISION
 # Our environment
 set RIVET_ROOT \$::env(BASEDIR)/$PKGNAME/\$version
 setenv RIVET_ROOT \$RIVET_ROOT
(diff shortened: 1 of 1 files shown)


preghenella:
it looks like the failure in build/O2/alidist-centos8 is unrelated to this PR.
@ktf any chance that this can be moved forward in any case?
@maireiphc can you please also approve this one?

ktf:
It's fine with me. Waiting for @maireiphc.

maireiphc:
Hi Roberto, 

thanks for taking care of this. Now approved.
About kpsewhich command part of the latex utilities, ...

- alidock

This is indeed not part of alidock.
The user has to set up one's own docker image for this with extra "yum install packages", since LaTeX is not there by default.
I have bugged Matteo Concas with this some time ago :
[ https://github.com/alidock/alidock/issues/145 - Use custom container with alidock]
Compare to this thread, I have reduced in the meantime now a more minimalist dockerfile (no need to download relsize.sty, which should come along with Rivet).

- lxplus

Kpsewhich should be known. I asked CERN IT to deploy it on lxplus back then to make sure users can be autonomous on lxplus (rivet-build -> rivet-cmphistos -> make-plots).

preghenella:
thanks @maireiphc 
if you can also please have a look at https://github.com/alisw/alidist/pull/2350 and see if you can approve it it would be cool ;)

re. latex on alidock.
I imagine that indeed one can install the latex utils there as well.
I just wanted to have the receipt working also without it, otherwise it would be let's say correct to have an explicit dependency (better than being able to build the code just to eventually find that the module does not load).

Cheers,
Roberto

mfasDa:
@preghenella Not sure why I just see it now but I have problems with the rivet module when loading Herwig on ubuntu 18.04:

```
Rivet/3.1.1-alice1-3(28):ERROR:102: Tcl command execution failed: set Old_TEXINPUTS [exec which kpsewhich > /dev/null 2>&1 && kpsewhich -var-value TEXINPUTS]
```
Any idea?

Cheers

Markus

preghenella:
@mfasDa 
don't know.
can you check what is the output of
```
which kpsewhich > /dev/null 2>&1 && echo "I DO have kpsewhich" || echo "I DO NOT have kpsewhich"
```
on your system?

mfasDa:
@preghenella I am trying to debug a bit. The problem is not connected to kpsewhich, it is installed on the system.The bash command is OK, no problem running it in bash (or zsh as I am using), the problem happens when going to tclsh. I can evaluate the two commands

```
exec which kpsewhich > /dev/null 2>&1
exec kpsewhich -var-value TEXINPUTS
```

separately and in this case they work, but once I combine them with && to have the full command it breaks with 

```
child process exited abnormally
```

tcl version is 8.6.

Cheers

Markus

preghenella:
I am no expert in `tcl`.
any chance that putting the `exec "this && that"` command within quotes helps?

mfasDa:
Me neither ...

No, doesn't help :(

preghenella:
maybe this works
```
exec bash -c "this && that"
```

mfasDa:
> maybe this works
> 
> ```
> exec bash -c "this && that"
> ```

Looks good!

maireiphc:
Hello Markus,

  what would be the next action to take on this ? Is your use case so specific (tclsh) that we leave the recipe as is or you would call for an update of the recipe ? 
Spontaneously, I lean to the first option but you may tell us.

preghenella:
hi @maireiphc @mfasDa 
I would go for an update of the receipt.
I don't see a drawback for that.
@mfasDa given that you have it and tested it, can you please take care of that?
Thanks

mfasDa:
Hi @maireiphc,

this is a problem in the build recipe as modules are using tcl, consequently if the tcl command fails the module evaluation fails. This happened in my case. We only went to tclsh directly in order to debug where the problem comes from. So the build recipe creating the modulefile must be fixed because the tcl command used in the modulefile doesn't work on a wide spread platform.

@preghenella yes I can prepare a PR.

Cheers

Markus

PR #2362, opened 2020-07-14
xjalienfs:: update to 1.2.2

adriansev:


adriansev:
@Atlantic777 i would say that for xjalienfs all build checks should be removed ... i do not get why are there in the first place, it's just a waste of cpu cycles.. OTOH now we know that O2 does not build on centos 8 :)))

adriansev:
@ktf it seems that the error of O2/alidist-centos8 is unrelated to this PR as i see in the logs:  
```
Finished processing dependencies for matplotlib==3.0.3
++ unset PYTHONPATH
++ env PYTHONPATH=/mnt/mesos/sandbox/sandbox/sw/INSTALLROOT/18eb5561ab364ab0bbb710759c8996012b614a4a/slc8_x86-64/Python-modules/1.0-1/share/python-modules/lib/python/site-packages python3 -c 'import matplotlib'
Traceback (most recent call last):
  File "<string>", line 1, in <module>
ModuleNotFoundError: No module named 'matplotlib'
```

ktf:
I suspect something changed on the matplotlib side and we now have broken cc8 builds... :-/

PR #2365, opened 2020-07-14
geant4: minimize usage of environment variables

ktf:
This seems to create issues under certain conditions on CC8.


ktf:
@davidrohr I am not sure if this is the issue you witnessed, but in any case this change seems cleaner.

PR #2374, opened 2020-07-20
replace pip3 calls with "python3 -m pip"

jezwilkinson:
Replaces pip3 calls in python-modules.sh and xjalienfs.sh with "python3 -m pip", to avoid possible errors with local pip installation ("No module named 'pip._internal'")

### DIFF
--- a/python-modules.sh
+++ b/python-modules.sh
@@ -30,15 +30,16 @@ fi
 PYTHON_MODULES_INSTALLROOT=$INSTALLROOT/share/python-modules
 mkdir -p $PYTHON_MODULES_INSTALLROOT
 # Install setuptools upfront, since this seems to create issues now...
-env PYTHONUSERBASE="$PYTHON_MODULES_INSTALLROOT" pip3 install --user -IU setuptools
+env PYTHONUSERBASE="$PYTHON_MODULES_INSTALLROOT" python3 -m pip install --user -IU setuptools
+env PYTHONUSERBASE="$PYTHON_MODULES_INSTALLROOT" python3 -m pip install --user -IU wheel
 # FIXME: required because of the newly introduced dependency on scikit-garden requires
 # a numpy to be installed separately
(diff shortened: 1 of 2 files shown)


ktf:
Actual issues seems to be bound to o2-dataflow and the fact it has a reduced set of dependencies which does not allow for importing wheel packages (and no, adding wheel does not seem to help).

@MichaelLettrich I think the Darwin builder died again...

jezwilkinson:
Added new commit to get rid of the conflict with https://github.com/alisw/alidist/pull/2378 [removal of matplotlib lines in python-modules]

awegrzyn:
Our job @ macOS fails on `python3 -m pip install --user -IU setuptools` with `/usr/local/opt/python/bin/python3.7: No module named pip`.

ktf:
As far as I understand it pip comes with `brew install python3`. If you run, without alibuild:

```
python3 -m pip -V
```

I get:

```
pip 20.1.1 from /usr/local/lib/python3.8/site-packages/pip (python 3.8)
```

maybe your homebrew installation needs a `brew upgrade python3`?

awegrzyn:
Ok, I see some issues in our setup, we're using `python3` coming from the system:
```
% python3 --version
Python 3.7.7
% which python3
/usr/local/bin/python3
```

`pip3` is there, but under different path:
```
% which pip3   
/usr/bin/pip3
```

and `python3 -m pip -V ` fails with `/usr/local/opt/python/bin/python3.7: No module named pip`

awegrzyn:
I'll figure it out, no worries...

ktf:
You seem to have python3 from brew (it's in `local/bin`, not in `bin`) but that copy does not have an embedded pip (which they claim is always the case https://docs.brew.sh/Homebrew-and-Python#setuptools-pip-etc). I would simply do a `brew upgrade python3`.

PR #2372, opened 2020-07-17
Update MC packages to the latest versions:

ihrivnac:
- Versions compatible with Root 5 in the production defaults
  (defaults-prod-latest.sh, defaults-release.sh)
  GEANT3: v2-7-p2
  GEANT4_VMC: v3-6-p6-inclxx-biasing-p2
  GEANT4: v10.4.2
  vgm: v4-4

- Versions compatible with Root 6 in the packages defaults 
  (xyz.sh):
  GEANT3: v3-5
  GEANT4_VMC: v5-0-p5
  GEANT4: v10.5.1
  vgm: v4-7

PR #2376, opened 2020-07-20
Fix geant4 recipe: use XERCESC_REVISION instead of XERCESC_VERSION

ihrivnac:


ihrivnac:
As suggested by @ktf in https://github.com/alisw/alidist/pull/2372

ktf:
Trivial and errors unrelated. Merging.

ktf:
Of course I cannot reproduce the error on neither centos8 not dataflow :-/

PR #2379, opened 2020-07-21
Use default gRPC build for ODC

AndreyLebedev:


teo:
Now that the special gRPC version is cut out, @AndreyLebedev @ktf do we even need a separate defaults for ODC? Building with `o2-dataflow` should be equivalent in this case.

ktf:
@teo for me, the less defaults the better.

teo:
Then I'd propose to eliminate `defaults-odc.sh` altogether, and update the ODC aliBuild instructions to use `--defaults o2-dataflow`. @AndreyLebedev would this be ok for you?

AndreyLebedev:
> Then I'd propose to eliminate `defaults-odc.sh` altogether, and update the ODC aliBuild instructions to use `--defaults o2-dataflow`. @AndreyLebedev would this be ok for you?

Yes, I'll do it.

PR #2382, opened 2020-07-22
Add -LGMP_ROOT/lib to LD_LIBRARY_PATH

mfasDa:
Should fix error concerning missing -lgmp with current herwig builds


mfasDa:
@preghenella This is the only way I can imagine to make the build aware of the libgmp location. If it still doesn‘t find libgmp afterwards it is not there. Strange, because it is a requirement (maybe should be changed to build requirement). I took the path from the fastjet build recipe

PR #2384, opened 2020-07-22
Remove libRivet.la as it creates relocation problems in ThePEG

preghenella:
This PR should fix a issue observed in `ThePEG` which derives from `Rivet`.
By removing the `libRivet.la` file from the `Rivet` installation dir we avoid that.


ktf:
@MichaelLettrich I think macOS is down again.

PR #2389, opened 2020-07-23
Update GEANT4_VMC: v3-6-p6-inclxx-biasing-p3

ihrivnac:
This version includes a fix in FindROOT.cmake which should hopefully fix the build problem in AliPhysics v5-09-02h.


ktf:
Seems to still fail https://alijenkins.cern.ch/job/build-any-ib/8916/console

ihrivnac:
Thanks; the problem seems to be in using the cmake keyword LESS_EQUAL which is not supported in the cmake version used in this build. I will pack a fix in a new geant4_vmc tag tomorrow.

PR #2397, opened 2020-07-27
bump readout v1.4.6

sy-c:


awegrzyn:
@sy-c is it intended for FLP Suite v0.9.0?

sy-c:
> is it intended for FLP Suite v0.9.0?
@awegrzyn, if still possible, yes I would like it in v0.9.

PR #2398, opened 2020-07-28
Correctly handle the finding of libUV for O2 dependencies

dstocco:
This PR complements the one in O2: https://github.com/AliceO2Group/AliceO2/pull/3933


libuv.sh
```diff
@@ -33,5 +33,6 @@ module-whatis "ALICE Modulefile for $PKGNAME $PKGVERSION-@@PKGREVISION@$PKGHASH@
 module load BASE/1.0 ${GCC_TOOLCHAIN_ROOT:+GCC-Toolchain/$GCC_TOOLCHAIN_VERSION-$GCC_TOOLCHAIN_REVISION}
 # Our environment
 set LIBUV_ROOT \$::env(BASEDIR)/$PKGNAME/\$version
+setenv LIBUV_ROOT \$LIBUV_ROOT
```
ktf, line 36: This is not needed.

dstocco: Ciao @ktf ,
I did not try from a fresh install, but if I remove both .build_succeeded and CMakeCache.txt and compile without the line it seems to work.
I modified the PR accordingly.
Thanks,
Cheers,
Diego

ktf: For the record, environment files do not enter the compilation part, as they are meant for runtime only.

PR #2406, opened 2020-07-29
Bump CMake to 3.18

davidrohr:
Has some new CUDA related features I'd like to use

### DIFF
--- a/cmake.sh
+++ b/cmake.sh
@@ -1,6 +1,6 @@
 package: CMake
 version: "%(tag_basename)s"
-tag: "v3.17.0"
+tag: "v3.18.2"
 source: https://github.com/Kitware/CMake
 build_requires:
  - "GCC-Toolchain:(?!osx)"
--- a/glfw.sh
+++ b/glfw.sh
@@ -10,7 +10,7 @@ prefer_system_check: |
   printf "#if ! __has_include(<GLFW/glfw3.h>)\n#error \"GLFW not found, checking if we can build it.\"\n#endif\n" | cc -xc++ -std=c++17 - -c -o /dev/null
 ---
 
-cmake $SOURCEDIR -DCMAKE_INSTALL_PREFIX=$INSTALLROOT \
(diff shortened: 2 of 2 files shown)


davidrohr:
@ktf : Could you check what breaks the GLFW build with CMake 3.18
@kostorr : I see that this and several other alidist PRs fail in CentOS 8 in building ALF/src/AlfServer.cxx: Could you check what is the problem?

awegrzyn:
Until issue with ALF @ C8 is not fixed: https://github.com/alisw/alidist/pull/2410

davidrohr:
@ktf : Could you check what is the issue with GLFW?

ktf:
Cannot reproduce on lxplus. I suspect the issue is a missing X11 package in the docker container which is now required by the FindX11 macro in CMake...

davidrohr:
@ktf : One issue that I have seen is anyway that the glfw / opengl / debuggui recipes are not consistently checking for all required dependencies. I had it several times that on some servers it would find all depencencies it checks for, but then it would still fail the build since other dependencies (which were not checked) were missing. I think this should be cleaned up, probably gradually.

davidrohr:
@ktf : hm, I don't understand this. build/O2/alidist uses the slc7-builder container, right?
I just tried this locally with slc7-builder and it worked...

ktf:
Me neither. I did try and was unable to reproduce. However when I reset the tests here, it still fails...

ktf:
> build/O2/alidist uses the slc7-builder

yes.

davidrohr:
@ktf : I tried to debug what is the problem in glfw so I added --debug-output to cmake in glfw.sh. For whatever reason, with --debug-output it compiles correctly. I tried 2 times with and without, and the behavior is consistent. Are you OK if we merge this as is, shouldn't be a problem except for some glfw cmake debug output written to the logs, or do you have an idea how we can debug it?
I need CMake 3.18 for some new CUDA features.

ktf:
Yes, it's fine with me. I suspect an actual CMake bug... I added a comment and merged.

PR #2408, opened 2020-08-01
Workflow to cherry pick commits via github actions

ktf:


chiarazampolli:
Ciao @ktf ,
Is this for the AliRoot/AliPhysics tags? Something I could learn about?
Thanks,
Chiara

ktf:
only for the alidist ones, for now. it allows you to cherry-pick using github actions.

chiarazampolli:
Ah ok. I will need that you show me how to use it (in case it is needed).

PR #2413, opened 2020-08-03
(ALF-42) ReadoutCard always depends on PDA for linux

kostorr:


awegrzyn:
I'm not an expert in this but most of other recipies use:
```
- "Python:slc.*"
-"Python-system:(?!slc.*)"

```

kostorr:
I opted for the `(?!osx)` bit because I understand that now we have builds for Ubuntu as well? @ktf can weigh in.

ktf:
we have plenty of ubuntu users, but we do not plan to use it in production, so either solution is fine.

mfasDa:
Sorry, but on ubuntu 18.04 this doesn't work because of an issue in the kerne-devel build recipe. The linux sources there are directly under /usr/src, so kernel-devel doesn't find them. I assume the search path /usr/src/kernel is for CentOS.

PR #2420, opened 2020-08-11
[WIP] Bump llvm/clang to 10.0.1

davidrohr:
@pzhristov @ktf @aalkin : It turns out arrow 1.0.0 supports also clang10, 0.17.1 doesn't.
But apparently, we have some customization in our arrow. Could this be rebased to arrow 1.0.0, then we can try to bump clang.

For reference, I tried this locally with my system installation of clang 10, and building arrow failed (even though clang 10 should be supported), but this might be due to the way the system installation is built.


aalkin:
I will try to port arrow customizations to 1.0, will require few days.

ktf:
Can you open a PR with the arrow people as well?

ktf:
Closing this in favour of #2423

PR #2423, opened 2020-08-12
Bump Clang to 10.0.1

aalkin:


aalkin:
Can be merged only after #2424

PR #2424, opened 2020-08-12
Bump arrow to 1.0.0

aalkin:


arrow.sh
```diff
@@ -26,6 +27,7 @@ case $ARCHITECTURE in
     [[ ! $BOOST_ROOT ]] && BOOST_ROOT=$(brew --prefix boost)
     [[ ! $LZ4_ROOT ]] && LZ4_ROOT=$(dirname $(dirname $(which lz4)))
     [[ ! $PROTOBUF_ROOT ]] && PROTOBUF_ROOT=$(dirname $(dirname $(which protoc)))
+    [[ ! $UTF8PROC_ROOT ]] && UTF8PROC_ROOT=$(brew --prefix utf8proc)
```
ktf, line 30: I am not sure I added the prefer_system. Could you add it?

arrow.sh
```diff
@@ -90,6 +90,7 @@ cmake ./src_tmp/cpp
       ${PROTOBUF_ROOT:+-DProtobuf_PROTOC_EXECUTABLE=$PROTOBUF_ROOT/bin/protoc}                      \
       ${BOOST_ROOT:+-DBoost_ROOT=$BOOST_ROOT}                                                       \
       ${LZ4_ROOT:+-DLZ4_ROOT=${LZ4_ROOT}}                                                           \
+      ${UTF8PROC_ROOT:+-DUTF8PROC_ROOT=${UTF8PROC_ROOT}}                                            \
```
ktf, line 93: I think it needs to be lowecase

aalkin:
@ktf I'm not entirely sure the 'prefer_system_check' is correct, can you have a look?

utf8proc.sh
```diff
@@ -5,6 +5,10 @@ source: https://github.com/JuliaStrings/utf8proc
 build_requires:
  - "GCC-Toolchain:(?!osx)"
  - alibuild-recipe-tools
+prefer_system: "(?!osx)"
+prefer_system_check: \
+  brew --prefix utf8proc > /dev/null;
+  if [ $? -ne 0 ]; then printf "Use brew install utf8proc"; exit 1; fi
```
ktf, line 11: I would have a test which compiles utf8proc.h, like done elsewhere... Something along the lines of:

```bash
printf "#include <utf8proc.h>\n" | c++ -c -I$(brew --prefix utf8proc)/include -xc++ - -o /dev/null 2>&1
```

arrow.sh
```diff
@@ -112,19 +113,8 @@ make install
 MODULEDIR="$INSTALLROOT/etc/modulefiles"
 MODULEFILE="$MODULEDIR/$PKGNAME"
 mkdir -p "$MODULEDIR"
+alibuild-generate-module > "$MODULEFILE"
 cat > "$MODULEFILE" <<EoF
```
ktf, line 117: >>

aalkin: Missed that, fixed.

PR #2427, opened 2020-08-14
utf8proc: init at v2.5.0

ktf:


ktf:
@aalkin we should integrate this in the arrow 1.0.0 recipe.

PR #2430, opened 2020-08-19
arrow: Make sure we pick up our own Re2 version.

ktf:


ktf:
Failing test seems to be CCDB backend related. @knopers8 @Barthelemy could you take a quick look?
Merging.

Barthelemy:
having a look

Barthelemy:
Weird. I propose to wait and see if it happens again in another PR.

PR #2435, opened 2020-08-23
ALF: remove until issues with DIM is fixed

ktf:


awegrzyn:
okay, could you create `alisw/dim` repo, I can push the files there and swap the URL.

ktf:
@awegrzyn done.

PR #2436, opened 2020-08-24
[O2-1652] Mirror DIM package in alisw and add back ALF

awegrzyn:


awegrzyn:
@ktf I think build fails because:
- currently ZIP source code has been download during each build and deleted after build
- aliBuild does not pull source code from Git as it thinks that it is already there

Barthelemy (approved):
good

kostorr (approved):

awegrzyn:
So IMO smooth transition would be:
- bring back ZIP download (keep it together with GH clone)
- remove ZIP download in a while

ktf:
@awegrzyn agreed. I can probably restart the build container as well, but that will introduce delays as well.

PR #2438, opened 2020-08-24
Add -ltbb to EXTRA_LIBS

preghenella:
I needed to explicitly add `-ltbb` to `debuggui.sh` receipt to have it successfully compiled on `ubuntu2004_x86-64`.
I am not sure this is the right way to do it, opened the PR anyway.

### DIFF
--- a/debuggui.sh
+++ b/debuggui.sh
@@ -22,7 +22,7 @@ case $ARCHITECTURE in
     ;;
     *) 
       DEFINES="-DIMGUI_IMPL_OPENGL_LOADER_GL3W -DTRACY_NO_FILESELECTOR"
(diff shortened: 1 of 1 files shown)


qgp:
For what it's worth, I have done the same for Gentoo ...

davidrohr:
@preghenella @ktf : We discussed today that we need to fix this, but we cannot add -ltbb unconditionally, since it is not available on every server. I'll open a PR today which checks for the presence of the library and adds it then.

preghenella:
thanks @davidrohr 
I just copied from your #2443 and force-pushed here.
I assume you are taking care of that also for `debuggui.sh` .
If not, I can reopen this PR

PR #2453, opened 2020-08-27
Modify Geant4 recipe to allow an external directory for data sets, clean-up:

ihrivnac:
- Added GEANT4_DATADIR variable which allows to override the default installation
  path for data sets (CMAKE_INSTALL_DATAROOTDIR)
- Replaced auto discovery of data installation paths (not working when data for more Geant4
  versions are installed in the same directory) with use of geant4-config script
- Removed G4INSTALL, G4INSTALL_DATA, G4SYSTEM env variables (not needed)
- Removed appending ROOT_INCLUDE_PATH with Geant4 includes (not needed)

### DIFF
--- a/geant4.sh
+++ b/geant4.sh
@@ -28,7 +28,7 @@ env:
   G4SAIDXSDATA : "`${G4INSTALL}/bin/geant4-config --datasets | grep G4SAIDXSDATA | sed 's/[^ ]* [^ ]* //'`"
 
 ---
-# if this variable is not defined default it to OFF
+# If this variable is not defined default it to OFF
 : ${GEANT4_BUILD_MULTITHREADED:=OFF}
 
 # Data sets directory:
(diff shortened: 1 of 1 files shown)


ihrivnac:
Motivation for this update:
Using a dedicated data sets directory can save compilation time and disk space, as some data sets are shared by several Geant4 versions and data sets need not to be downloaded again with each rebuilding of Geant4, what is the case when the default installation path is used.

ihrivnac:
@MichaelLettrich , the build/alidist/O2Suite/o2/macOS build is in Expected state since 1 day; is it normal or is the builder stuck? Thank you.

ihrivnac:
The two failures are nor relevant to this PR:
The one on build/alidist/O2Suite/o2/macOS is caused by no space left:
\TOFCalibration.dir/G__O2TOFCalibration.cxx.o
fatal error: error in backend: IO failure on output stream: No space left on device
And the one in build/O2/alidist-dataflow is non related compilation error.

@sawenzel , should I update the branch to trigger new build, or is it ok to approve? Thank you.

ktf:
@MichaelLettrich looks like macOS is still having troubles.

ihrivnac:
All checks passed ok. Ping @sawenzel , @ktf .

sawenzel (approved):

PR #2458, opened 2020-08-28
[TEST - Dont merge!] Bump FairMQ to 1.4.22

davidrohr:


davidrohr:
@rbx : Could you have a look at this? In the alidist-dataflow setup FairMQ seems to trigger a GCC compiler bug. That is probably not your fault, but we should work around this I guess.

rbx:
What is the actual error? Could the machine have run out of memory?

davidrohr:
Well it says
```
c++: internal compiler error: Killed (program cc1plus)
Please submit a full bug report,
with preprocessed source if appropriate.
```
But you are right, perhaps this was an ooo killer. Let's see when the CI reruns whether it passes.

rbx:
Still crashing. Could we run dmesg after the crash to see a hint of why it was killed? Is the machine building anything else successfully?

davidrohr:
I don't have access to that node. Only thing I can do is pull the docker container and try to rerun locally with that setup, can do that later, just in a meeting with US.

davidrohr:
I have built it locally in the slc7-builder container for o2-dataflow defaults and that worked...

rbx:
Did you notice any unusual memory usage when running it the container? Perhaps, if it is indeed a compiler bug, it just causes excessive memory usage, which brings the test machine down?

ktf:
I've moved the builder to 20GB, rather than 16, and restarted the build. As far as I can tell, using any of boost::signals, boost::state_machine, boost::spirit or related programs is simply too memory hungry and should be avoided to be exposed in the interface.

ktf:
20G not enough... Trying with 24...

ktf:
There is something weird... The builder has now 24GB and it still fails, memory usage is low until it get compiling the SDK and then it explodes almost immediately. Trying now without the SDK to verify the problem is there. It indeed looks like a (system) compiler bug...

ktf:
(it is killed by the OOM, BTW).

davidrohr:
What seems really strange to me: why does it happen only with the o2-dataflow defaults?

rbx:
We did not add any new public includes since the last tag:
https://github.com/FairRootGroup/FairMQ/compare/v1.4.21...v1.4.22

davidrohr:
retrying with -j1, should show whether it needs infinite memory or just much, and where exactly it fails.

davidrohr:
Now it fails in control which uses FairMQ... We should go through the header changes and check what causes the memory increase and try to forward declare I guess.

ktf:
dataflow uses system gcc. we could also check if there is any update available for it. DDS (well, boost spirit, actually) has always been extremely memory hungry with system GCC.

davidrohr:
Which system gcc? System gcc of slc7-builder is 4.8 isn't it?

ktf:
software collection one (7.3 something)

--
Ciao,
Giulio

davidrohr:
@rbx : I have tried not to reduce the number of parallel jobs. Unfortunately, it works only with -j 1, already with -j2 we run out of memory. And this is not restricted to FairMQ, but basically to all packages that include the FairMQ headers, i.e. control-ocp and readout. I think we should try to fix this at the source. Would it be possible to clean some of the includes from the boost headers and use forward declarations?
I think we need a solution soon, since this is a total blocker for bumping FairMQ.
What still puzzles me: Why does it affect only the o2-dataflow default? @ktf: What is the difference in memory between the alidist-dataflow builder and the normal alidist o2 build?

ktf:
o2-dataflow uses system GCC (i.e. the one coming from Developer Toolset 7), o2 uses our own build of GCC.

davidrohr:
> o2-dataflow uses system GCC (i.e. the one coming from Developer Toolset 7), o2 uses our own build of GCC.

I don't think it uses the system GCC. In the log I see it uses `/mnt/mesos/sandbox/sandbox/sw/slc7_x86-64/GCC-Toolchain/v7.3.0-alice2-1/bin/c++` and also the 'GCC_TOOLCHAIN_ROOT' environment variable is set.

rbx:
I'm all for minimizing boost headers exposure, which we have done in the past and since then do not expose more unless really necessary. I'll take another look if anything else can be hidden.

But I don't see why this is an issue now. Like I said, there were no new public includes added since the last tag. So 1.4.21 should've been failing already. iirc, last time boost includes in public headers were touched is more than a year ago. SDK component might've been modified more recently, but also many months back.

ktf:
I've been checking. Actually the compiler used is the same. I will investigate if there is any difference in the container memory requirements...

ktf:
both containers have the same memory requirements...

Barthelemy:
It is indeed puzzling. `o2-dataflow` disables lots of packages which should, of course, not hurt. It uses O2 dev, but compilation does not even reach O2. 

PS: Mac build fails due to "no space left on device"

Barthelemy:
I have tried on a CC7 server and it worked with `o2-dataflow`.

rbx:
I replaced a couple of public boost includes with forward declarations in [this branch](https://github.com/rbx/FairMQ/tree/shm-options-alt-2). On my system (gcc 9.3.0) this had no effect on the compile time. Both this branch and 1.4.22 take around 3.7 minutes to compile and peak memory consumption during compilation with -j8 is 6GB. But perhaps there will be difference for gcc 7.3. Give it a try to see if it makes a difference.

The changes touch MemoryRessources, so the places where you use it might need extra includes with this patch.

Besides these headers I don't think there are any more we can remove easily now.

davidrohr:
I changed this PR to your branch, let's see

ktf:
@MichaelLettrich also here macOS dies.

MichaelLettrich:
> @MichaelLettrich also here macOS dies.

Thanks for the hint. Again, corrupted repo. Don't know what happened there.

davidrohr:
@ktf: any idea what is the problem with 
build/O2/alidist has empty log
build/O2/alidist-dataflow fails in control ocp but doesn't have an error.
mac os seems to be broken in general? (@MichaelLettrich )

In any case @rbx : With this branch, the CI could build FairMQ with the dataflow defaults, so apparently cleaning up the headers helped. What is the status of merging the branch into dev and tagging?

ktf:
Yes. Major outage of the CI infrastructure over the weekend. Hopefully recovered.

davidrohr:
ok, now we are a bit further, but now QualityControl fails because of the changed includes. I'll fix that and open a PR to QC.

rbx:
@davidrohr I made another test PR, bumping FairMQ to 1.4.22 (without my added changes), because I don't think the problem is with our includes and the build seems to have passed:
https://github.com/alisw/alidist/pull/2465

But anyway the include changes are already submitted and won't hurt. The PR will be merged once reviewed and then I can make a tag.

davidrohr:
@rbx : alidist/dataflow CI has also failed in #2465

davidrohr:
The QC problem have been fixed and have have triggered the CI here again, should pass now.
Could you create a 1.4.23 tag with the include fixes?

davidrohr:
@rbx @ktf : ok, the new FairMQ with the include fixes should finally work, everything green here except for the known errors.
As soon as  @rbx provides a new tag, I'll update this PR with the proper version (right now it uses Alexey's private branch).

rbx:
1.4.23 is released.

ktf:
Thanks. Bumped.

ktf:
@ironMann @davidrohr any reason **not** to upgrade?

ironMann:
LGTM for DataDistribution build

ktf:
Ok. Merging this since I think @davidrohr was also ok (but on holidays today) and in any case all the tests are green. We can always have a further bump / revert if there is some regression.

PR #2460, opened 2020-09-01
The DelphesO2 receipt

preghenella:
The receipt to build the fast-simulation interface between `Delphes` and the `o2`.

### DIFF
--- a/delpheso2.sh
+++ b/delpheso2.sh
@@ -1,0 +1,49 @@
+package: DelphesO2
+version: "%(tag_basename)s"
+tag: master
+requires:
+  - ROOT
+  - Delphes
+  - O2
+overrides:
+  O2:
+    version: "%(short_hash)s%(defaults_upper)s"
+    tag: dev
+build_requires:
+  - CMake
+  - "GCC-Toolchain:(?!osx)"
+source: https://github.com/preghenella/DelphesO2.git
+---
+#!/bin/bash -ex
+cmake $SOURCEDIR -DCMAKE_INSTALL_PREFIX=$INSTALLROOT       \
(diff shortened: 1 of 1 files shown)


preghenella:
the failure of the `build/O2/alidist-dataflow` check seems to be unrelated to this PR

ktf:
As discussed, improvements for next round:
* Use `-DCMAKE_INSTALL_LIBDIR=lib` to get rid of lib64
* Use `alibuild-recipe-tools` to remove the boilerplate for the modulefile, e.g.:

```bash
...
build_requires:
  - alibuild-recipe-tools
...
---
...
mkdir -p $INSTALLROOT/etc/modulefiles
alibuild-generate-module > $INSTALLROOT/etc/modulefiles/$PKGNAME
```

ktf:
For now `alibuild-generate-module` generates only the basic "module" dependencies. All the rest is still up to you. So I would suggest something like:

```bash
cat << EOF >> $INSTALLROOT/etc/modulefiles/$PKGNAME
set DELPHESO2_ROOT \$::env(BASEDIR)/$PKGNAME/\$version
prepend-path PATH \$DELPHESO2_ROOT/bin
prepend-path LD_LIBRARY_PATH \$DELPHESO2_ROOT/lib
EOF
```

preghenella:
Thanks @ktf ,
it works. 
For the `O2DPG` case in #2462 you have commented that

> Given it is our own recipe, I do not care about O2DPG_ROOT.

This is also our own recipe, so I would keep the `DELPHESO2_ROOT` env variable even if we discussed to remove it.
Unless you disagree, in which case I have no objections and I will remove it.

ktf:
As long as you will not want to pick it up from the system, it's fine.

PR #2461, opened 2020-09-01
Force setuptools 49.6.0

ktf:
50.0.0 seems to give issues on some platforms (ubuntu and CC8).

### DIFF
--- a/python-modules-list.sh
+++ b/python-modules-list.sh
@@ -2,6 +2,7 @@ package: Python-modules-list
 version: "1.0"
 env:
   PIP_REQUIREMENTS: |
+    setuptools==49.6.0
     requests==2.21.0
     ipykernel==5.1.0
     ipython==7.4.0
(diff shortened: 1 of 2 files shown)


benedikt-voelkel:
Just tried that and the following seems to happen:

`python3 -m pip install --user -IU setuptools` installs latest setuptools (50.0.0)
(https://github.com/alisw/alidist/blob/master/python-modules.sh#L48)

`python3 -m pip install --user -IU -r requirements.txt` afterwards however doesn't seem to overwrite the `setuptools` version even if an older version of that is specified therein.
(https://github.com/alisw/alidist/blob/master/python-modules.sh#L56)

maybe putting `wheels` and `setuptools==49.6.0` in a separate list in `python-modules-list.sh` and installing that before the actual modules of interest?

ktf:
@MichaelLettrich "no space left on device" on macOs.

MichaelLettrich (approved):
Fixes my problems.

MichaelLettrich:
@ktf  macos builds are back on track.

PR #2462, opened 2020-09-03
The o2 DPG receipt

preghenella:
The PR adds the receipt for the `o2` DPG software

### DIFF
--- a/o2dpg.sh
+++ b/o2dpg.sh
@@ -1,0 +1,28 @@
+package: O2DPG
+version: "%(tag_basename)s"
+tag: master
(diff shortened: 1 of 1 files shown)


preghenella:
hi @ktf ,
I have just copied from what is used for `AliDPG`.
I guess you would like me to modify the modules part to use `alibuild-recipe-tools`, will do that asap.

Re. `O2DPG_ROOT` environment variable, I would tend to keep it set as for the `AliDPG` case, unless you have strong objections for that.

ktf:
Given it is our own recipe, I do not care about O2DPG_ROOT.

PR #2465, opened 2020-09-06
test pr, do not merge

rbx:
test


davidrohr:
@ktf @awegrzyn : We seem to have a general problem with the alidist-centos8 CI, which seems to fail building clang due to insufficient memory.

ktf:
I will be investigating after the WP4+WP14 meeting.

PR #2471, opened 2020-09-08
[O2-1630] Adding recipe for ADAPRO

chiarazampolli:


chiarazampolli:
Hello @ktf , 

To run the example in O2-1630, I need to have $ADAPRO_ROOT defined, but when I load the environment, it is not set. Can you check if there is something I forgot?

Thanks!

Chiara

systemd.sh
```diff
@@ -0,0 +1,9 @@
+package: systemd
+version: "237"
+system_requirement_missing: |
+   Please install systemd develpment package on your system.
+   If they are there, make sure you have them in your default path or check you have `which` installed.
+system_requirement: ".*"
```
ktf, line 6: At least `(?!osx)` since systemd does not even exist there. We should avoid the dependency completely.

ktf (changes requested):

systemd.sh
chiarazampolli: Done, with the fix for the modulefile.

adapro.sh
```diff
@@ -0,0 +1,40 @@
+package: ADAPRO
+version: "%(tag_basename)s"
+tag: master
```
ktf, line 3: Please, use a tagged version.

systemd.sh
```diff
@@ -0,0 +1,9 @@
+package: systemd
```
ktf, line 1: Before this can be included in O2, you will have to get rid of this dependency.

ktf (approved):
If it simplifies integration we can merge this as is. You will have to walk on my dead body before I allow this to be part of O2 with the systemd dependency. ;-)

chiarazampolli:
Ciao @ktf , 

It was not so urgent to merge this, maybe I should have added WIP. I opened the PR following your advice, to be able to discuss the possible implementation of the recipe. 
Anyway, as discussed, if we want to include this, the client part should be detached from the server one. In principle, I need only 2 classes. On Friday I discussed with Ruben, and he was even thinking of creating our own because the existing ones in ADAPRO seem a bit too complicated. But since it will be a repetition of code, I would first give it a try with ADAPRO. 
One question: @shahor02  commented that it is is possible probably to simply use the includes without building the whole package. Is it the case?

Cheers,

Chiara

ktf:
For the record, I agree with @shahor02. Duplication of code cannot be as bad as having a dependencies on systemd. Code reusage is highly overrated. No idea if the includes would suffice.

chiarazampolli:
In the discussion with @shahor02 , it was more about the fact that the classes allocate more than needed, but since this is not something that detectors will store, it is probably not so bad. The dependency on systemd is indeed bad, we discussed it also in JIRA. Duplicating is a bit "risky" in the sense that if something is changed by DCS (no idea if it can happen or if the code can be considered frozen enough - last change was 3 years ago, see https://gitlab.cern.ch/adapos/adapro), then we will need to synchronize. 
As I said, I created the recipe only because I wanted to start with what DCS provides, the PR came because I was suggested to open one for discussion. My fault that I did not put WIP, but it also does not harm much, till the package is not used. 
We can check with Peter Bond. Creating a couple of classes won't be an issue.

PR #2472, opened 2020-09-08
lhapdf should disable python, if python3 is installed

martenole:
Hi @ktf ,

probably I am wrong, but isn't there an error in the check for the python version in lhapdf.sh? In case python3 is detected, python should be disabled, but I think it is the other way around currently.

Cheers,
Ole

### DIFF
--- a/lhapdf.sh
+++ b/lhapdf.sh
@@ -26,7 +26,7 @@ rsync -a --exclude '**/.git' $SOURCEDIR/ ./
 
 export LIBRARY_PATH="$LD_LIBRARY_PATH"
 
-if python -c 'import sys; exit(0 if sys.version_info.major >=3 else 1)'; then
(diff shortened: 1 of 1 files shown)


preghenella:
it looks to me that the code is correct as it is.
you can check running it with
```
preghenella@RobyThinkPad:~$ if python -c 'import sys; exit(0 if sys.version_info.major >=3 else 1)'; then echo "PYTHON version is > 3"; fi
```
which does not `echo` anything, whereas 
```
if python3 -c 'import sys; exit(0 if sys.version_info.major >=3 else 1)'; then echo "PYTHON version is >= 3"; fi
PYTHON version is >= 3
```

preghenella (commented):
to me the code looks ok as it is

martenole:
Sorry about that, you are right. I will close this.

PR #2481, opened 2020-09-16
Bump FairRoot / DDS

davidrohr:
Let's see if this breaks something. The new tags will be needed for new ROOT / new boost.

### DIFF
--- a/dds.sh
+++ b/dds.sh
@@ -1,6 +1,6 @@
 package: DDS
 version: "3.5.1"
-tag: "3.5.1"
+tag: "3.5.2"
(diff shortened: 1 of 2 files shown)


davidrohr:
@ktf @sawenzel @pzhristov  : Any reservations? Afterwards we could try new root / boost.

sawenzel:
@davidrohr : Did you test this locally? I'd say fine if it doesn't break the sim/reco workflows.

davidrohr:
I ran a quick simulation and tpc reco locally with latest DDS/FairRoot but I didn't try e.g. the full system test.

PR #2495, opened 2020-09-23
Fix tcl command for TEXINPUTS in modulefile

mfasDa:
On certain platforms the previous version
of the command leaded to a child error
in the command execution. Specifying that
the command should be executed as bash
command fixed the problem.

### DIFF
--- a/rivet.sh
+++ b/rivet.sh
@@ -106,7 +106,7 @@ prepend-path LD_LIBRARY_PATH \$RIVET_ROOT/lib
 # (TEXMFHOME, HOMETEXMF, TEXMFCNF, TEXINPUTS, LATEXINPUTS)
 # Here trying to keep the env variable changes to their minimum, i.e touch only TEXINPUTS, LATEXINPUTS
 # Manual prepend-path for TEX variables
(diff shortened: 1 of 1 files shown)


mfasDa:
Fixing problem discussed in #2361. Credits to @preghenella

ktf:
Can you use sh rather than bash? I fear bash is on the way out on mac.

mfasDa:
> Can you use sh rather than bash? I fear bash is on the way out on mac.

Done

ktf:
No way to test this in any case...

mfasDa:
@preghenella I was too optimistic :(

```
Rivet/3.1.1-alice1-6(28):ERROR:102: Tcl command execution failed: set Old_TEXINPUTS [exec sh -c "which kpsewhich > /dev/null 2>&1 && kpsewhich -var-value TEXINPUTS"]
```

Surprisingly it had worked for me in the first test, and on the tclsh. I am not a tcl expert. @ktf any idea?

mfasDa:
System is ubuntu 18.04

PR #2498, opened 2020-09-24
o2checkcode: enable aliceO2-namespace-naming check

anerokhi:


o2checkcode.sh
```diff
@@ -63,7 +64,10 @@ aliceO2-member-name\
 ,modernize-use-uncaught-exceptions}"
 
 # Run C++ checks
-run_O2CodeChecker.py -clang-tidy-binary $(which O2codecheck) -header-filter=.*SOURCES.* ${O2_CHECKER_FIX:+-fix} -checks=${CHECKS} 2>&1 | tee error-log.txt
+run_O2CodeChecker.py ${JOBS+-j $JOBS} \
+	-clang-tidy-binary $(which O2codecheck) \
+	-clang-apply-replacements-binary "$CLANG_ROOT/bin-safe/clang-apply-replacements" \
```
sawenzel, line 69: do we actually want the replacements binary? Now that the codebase is clean it might be sufficient to merely enable the check?

anerokhi: Are you proposing to also remove `O2_CHECKER_FIX`? How someone would add new checks and autofix code if it is not clean from the new checks POV?

PR #2499, opened 2020-09-24
add curl in monitoring/ppconsol requirements

AllaMaevskaya:


awegrzyn:
`curl` is optional `Monitoring` dependency, do you need to have `http` backend enabled by default?

ktf (approved):

awegrzyn (changes requested):
`ppconsul` indeed requires curl but `monitoring` does only for optional `http` backend. Could you explain why you need it enabled?

AllaMaevskaya:
this was the only way to build monitoring, otherwise it crashed with the same complains as Roberto published

sawenzel:
I confirm that the dependency in monitoring is also needed.

sawenzel (approved):

awegrzyn:
@sawenzel Do you know why it is needed? The CMake of Monitoring does not make it required: `find_package(CURL MODULE)`

sawenzel:
I am afraid I don't know. Just confirming that it solves the issue https://alice.its.cern.ch/jira/browse/O2-1723. Could it be that ppconsul is using curl in header files?

awegrzyn:
Nope it does not, could you just paste me what CMake reports when compiles Monitoring?

awegrzyn (approved):
Not to block, we can investigate later on

awegrzyn (approved):

PR #2500, opened 2020-09-25
Bump O2 to v1.3.0

awegrzyn:


awegrzyn:
@ktf Is that okay with you?

ktf (approved):

PR #2502, opened 2020-09-27
boost: attempt version 1.73 to fix various compatibility bugs

ktf:


davidrohr:
Why don't we try 1.74?

ktf:
done.

boost.sh
```diff
@@ -1,7 +1,7 @@
 package: boost
 version: "%(tag_basename)s"
-tag: v1.72.0-alice2
-source: https://github.com/alisw/boost.git
+tag: v1.74.0
+source: https://github.com/boostorg/boost.git
 requires:
```
pzhristov, line 5: As far as I can see, the tag is called boost-1.74.0 in https://github.com/boostorg/boost.git. I tried to use this name, but boost still does not compile.

awegrzyn: @pzhristov Do you want to try with small modification I did?

awegrzyn:
It looks like `O2` build fails because `CommonUtils/ShmManager.h` does not include `<atomic>`

PR #2503, opened 2020-09-28
xjalienfs:: tag 1.2.5

adriansev:
* fix delays : ungarded sleep conditions; to many checks when recovery procedures are in place
* add env ALIENPY_TIMING and enable detailed timing
* timing_ms:: if key exists, print it at the end of output
* time:: move timeing info after command output
* RestoreSession:: when restoring session AND session variables force the cd, do not check against already restored session variables
* implement exit/quit/logout

### DIFF
--- a/xjalienfs.sh
+++ b/xjalienfs.sh
@@ -1,6 +1,6 @@
 package: xjalienfs
 version: "%(tag_basename)s"
-tag: "1.2.4"
+tag: "1.2.5"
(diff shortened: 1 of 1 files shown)


Atlantic777 (approved):

yuw726 (approved):

adriansev:
@ktf is it normal for O2/alidist to take >1h? Thanks!

ktf:
@adriansev it might be if large rebuilds are happening before yours.

PR #2506, opened 2020-09-28
Add recipe for legotrain helpers

mfasDa:
Add recipe for legotrain helpers, a lightweight
package running traintest locally with the same
configuration as in a regular train test.

### DIFF
--- a/legotrain_helpers.sh
+++ b/legotrain_helpers.sh
@@ -1,0 +1,28 @@
+package: legotrain_helpers
+version: "%(tag_basename)s"
+tag: "v1.0.0"
+source: https://github.com/mfasDa/legotrain_helpers
+---
+#!/bin/bash -e
+
+rsync -a --exclude='**/.git' --delete --delete-excluded $SOURCEDIR/ $INSTALLROOT/
+
+# Modulefile
+MODULEDIR="$INSTALLROOT/etc/modulefiles"
+MODULEFILE="$MODULEDIR/$PKGNAME"
+mkdir -p "$MODULEDIR"
+cat > "$MODULEFILE" <<EoF
+#%Module1.0
+proc ModulesHelp { } {
+  global version
+  puts stderr "ALICE Modulefile for $PKGNAME $PKGVERSION-@@PKGREVISION@$PKGHASH@@"
+}
+set version $PKGVERSION-@@PKGREVISION@$PKGHASH@@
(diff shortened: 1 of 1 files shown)


mfasDa:
Something useful?

legotrain_helpers.sh
```diff
@@ -0,0 +1,28 @@
+package: legotrain_helpers
+version: "%(tag_basename)s"
+tag: "v1.0.0"
+source: https://github.com/mfasDa/legotrain_helpers
```
ktf, line 4: If this is useful, we should move it under alisw / AliceO2Group, I guess (and give you maintain rights)

legotrain_helpers.sh
```diff
@@ -0,0 +1,28 @@
...
+---
+#!/bin/bash -e
+
+rsync -a --exclude='**/.git' --delete --delete-excluded $SOURCEDIR/ $INSTALLROOT/
+
+# Modulefile
+MODULEDIR="$INSTALLROOT/etc/modulefiles"
```
ktf, line 11: You could use the `alibuild-recipe-tools` to generate the module. See http://alisw.github.io/alibuild/reference.html#runtime-environment for documentation.

legotrain_helpers.sh
mfasDa: The package is intended for the current train, therefore alisw would be the most appropriate. The run3 train test structure might be different. The scripts could also go to AliPhysics directly in case it would get a bin directory. This would eventually be even easier.

The goal of the package is to download the train configuration and run the test locally. This is intended for debugging and could be useful for many people in the collaboration in order to develop and test their code locally on existing train configurations. I am using this myself for debugging and would like to share this with whoever is interested.

legotrain_helpers.sh
mfasDa: I was not aware of this - I will take a look. The module is adapted from the aligenmc module, which is also a bash-script-only package.

legotrain_helpers.sh
ktf: If the only scope of the package is AliPhysics, then I would propose to add it there, actually...

ktf:
I assume this is not needed anymore. Closing.

mfasDa:
The scope was to make available some simple tools to run legotrain test locally, which I use for debugging of train tests, to the collaboration so that they can be used by other users for their debugging. I agree that AliPhysics would be a better place for this, as it would come for free.

PR #2521, opened 2020-10-06
Add NLO process support to Herwig package

mfasDa:
- Add package for MadGraph external loop provider
- Build Herwig with OpenLoops and MadGraph support

### DIFF
--- a/madgraph.sh
+++ b/madgraph.sh
@@ -1,0 +1,24 @@
+package: madgraph
+version: "%(tag_basename)s"
+tag: "v2.8.1"
+source: https://github.com/alisw/MadGraph
+requires:
+ - "Python:slc.*"
+ - "Python-system:(?!slc.*)"
+build_requires:
+  - alibuild-recipe-tools
+---
+#!/bin/bash -e
+
+rsync -a --exclude='**/.git' --delete --delete-excluded $SOURCEDIR/ $INSTALLROOT/
+
+#ModuleFile
+mkdir -p $INSTALLROOT/etc/modulefiles
+alibuild-generate-module > $INSTALLROOT/etc/modulefiles/$PKGNAME
+
+cat << EOF >> $INSTALLROOT/etc/modulefiles/$PKGNAME
+set MADGRAPH_ROOT \$::env(BASEDIR)/$PKGNAME/\$version
+setenv MADGRAPH_ROOT \$MADGRAPH_ROOT
+prepend-path PATH \$MADGRAPH_ROOT/bin
+
+EOF
--- a/herwig.sh
+++ b/herwig.sh
@@ -8,6 +8,8 @@ requires:
   - ThePEG
   - lhapdf
   - lhapdf-pdfsets
+  - Openloops
+  - madgraph
 build_requires:
   - autotools
 ---
@@ -22,6 +24,8 @@ export LDFLAGS="-L$LHAPDF_ROOT/lib -L$CGAL_ROOT/lib -L$GMP_ROOT/lib $LDZLIB"
 ./configure                        \
     --prefix="$INSTALLROOT"        \
     --with-thepeg="${THEPEG_ROOT}" \
+    --with-openloops=${OPENLOOPS_ROOT} \
+    --with-madgraph=${MADGRAPH_ROOT} \
     --with-gsl="${GSL_ROOT}"
 
 make ${JOBS:+-j $JOBS}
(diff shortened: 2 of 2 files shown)


mfasDa:
The madgraph build recipe is currently using a repository under my userspace - I will fix this once the PAGMC has created a repository for the official ALICE fork of madgraph.

preghenella (approved):

madgraph.sh
```diff
@@ -0,0 +1,28 @@
+package: madgraph
+version: "%(tag_basename)s"
+tag: "v2.8.1"
+source: https://github.com/mfasDa/MadGraph5
```
ktf, line 4: Please do not use a personal repository.

madgraph.sh
```diff
@@ -0,0 +1,28 @@
...
+---
+#!/bin/bash -e
+
+rsync -a --exclude='**/.git' --delete --delete-excluded $SOURCEDIR/ $INSTALLROOT/
```
ktf, line 8: Does this mean that this is a binary only distribution?

madgraph.sh
```diff
@@ -0,0 +1,28 @@
...
+
+# Modulefile
+MODULEDIR="$INSTALLROOT/etc/modulefiles"
+MODULEFILE="$MODULEDIR/$PKGNAME"
+mkdir -p "$MODULEDIR"
+cat > "$MODULEFILE" <<EoF
```
ktf, line 14: Can you use the (somewhat) recently introduced `alibuild-recipe-tools.sh` to generate the modulefile?

ktf (commented):
Please fix at least the repository issue. Let me know if you need help to create our own repo.

madgraph.sh
mfasDa: As mentioned above someone with rights needs to create the repository. @preghenella @maireiphc @ktf 

Names could be 
madgraph
madgraph5
MadGraph
MadGraph5

I will adapt to what fits best to the naming convention.

madgraph.sh
mfasDa: Yes, it is a pure binary (python). However for NLO process calculations i.e. within Herwig it produces files which needs to be compiled on-the-fly. Sherpa does something similar. This is different with respect to openloops where the process libraries are static and can be provided with the package. All this is steered from within Herwig, which only needs to know where to find the binary.

For process calculation on the grid as far as I see in local tests the most efficient is to build the processes and run the integration locally - because this is time consuming (waste of time if we repeat it for each job), can be parallelized on clusters, and can be reused from all jobs - and copy the process library and the grids to the grid (working dir), and just run event generation there.

madgraph.sh
mfasDa: OK, is there an example how to use it?

mfasDa:
OK, the last point (usage of alibuild-generate-module) is done

madgraph.sh
preghenella: @mfasDa 
you can check an example in `delpheso2.sh` recipe.

madgraph.sh
preghenella: request to open the repo opened.
https://alice.its.cern.ch/jira/browse/O2-1754

@mfasDa 
is a brand new repo needed? or can you fork from another repo?

mfasDa: Unfortunately they use launchpad, so no git repo we can fork from :(

preghenella: the repo is ready, you can fork it, populate it and open PR.
Please, ask in the PR for the tag if it needs to be created.

mfasDa: @preghenella  Forking is prevented while the repository is still empty - I suspect you need to import it from my repository.

preghenella: uhm... can we give you the write rights in this repo?

preghenella: @mfasDa  I have started the import from your repository.

preghenella: done @mfasDa 
you also have the tag ready there, check it out!
https://github.com/alisw/MadGraph

mfasDa: OK, done as well. The build recipe is using the tag from the alisw repo

madgraph.sh
mfasDa: @preghenella Yes, I have seen it. It is done.

madgraph.sh
mfasDa: I want to add that "binaries" (mg5_aMC) take care about the PYTHONPATH, so this can be skipped.

preghenella:
Hi @mfasDa 
Thanks for the updates.
I think this PR is good to be merged.
But AliGenerators build is failing since yesterday (don't know why) and we might better check that first before introducing this extra degree. I still had no change to look into the cause, might be an intermittent failure

preghenella (approved):
AliGenerators still fails.
Let's try to merge this, perhaps it helps.

PR #2522, opened 2020-10-06
Fix clang compilation on CC7

ihrivnac:
This feix on my side the problem report https://alice.its.cern.ch/jira/browse/O2-1729

### DIFF
--- a/clang.sh
+++ b/clang.sh
@@ -31,7 +31,7 @@ case $ARCHITECTURE in
   ubuntu*) DEFAULT_SYSROOT="" ;;
   debian*) DEFAULT_SYSROOT="" ;;
   unknown*) DEFAULT_SYSROOT="" ;;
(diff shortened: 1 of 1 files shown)


ktf:
The /usr was there because it did not work otherwise in the past. We should do a bit more stress testing before we merge this.

ihrivnac:
@ktf , Andry confirmed in JIRA that the fix works fine on her side, too. Which further stress testing do you propose to validate this?

ktf:
did you try with the devtoolset 7?

ihrivnac:
@ktf, I checked the version now, it should be ok:
```
[root@o2tests02 ~]# yum list installed | grep devtools
devtoolset-7.x86_64                  7.1-4.el7                  @centos-sclo-rh 
...
```

PR #2523, opened 2020-10-06
O2-1722 add gcc as dependency to cmake

Barthelemy:


ktf:
Error in AliGenerators is unrelated. I would claim CMake must indeed depend on GCC at runtime, since it's a C++ executable. Merging.

PR #2537, opened 2020-10-13
Make sure Python-modules are installed in the proper place

ktf:


ktf:
@Barthelemy @awegrzyn I am not sure if this was always broken or if some recent changes in pip broke the behaviour, but this should hopefully fix your python woes.

Barthelemy:
thank you

Barthelemy:
Could we merge this ?

ktf:
Yes. I was waiting for the tests.

PR #2540, opened 2020-10-14
Build ROOT with all cores

Barthelemy:
Given that we build clang.sh with all the cores, I expect that we now can build ROOT the same way. 

If you think that too many people will have an OOM error, we can drop it.


Barthelemy:
Hum... the builder did not take it that well. I presume that I have to do something more clever or just leave it as it is.

ktf:
I think it's actually unrelated...

Barthelemy:
I hope you were right :)

PR #2542, opened 2020-10-15
boost: use own tag and generate modulefile

ktf:
- git submodule init is slow.
- use alibuild-generate-module


ktf:
@TimoWilken can you check what is going on with the alidist-centos8 builder? You can use `aurora task ssh` to login to the machine.

PR #2544, opened 2020-10-15
Fix: temporarily disable libjalienws for JAliEn-ROOT

Atlantic777:
Disabling libjalienws for the moment until the issue with linking to OpenSSL is resolved. JAliEn-ROOT can still work without it.

cc @costing @ktf @odatskov


ktf:
Currently broken. Trusting @Atlantic777 and merging.

PR #2546, opened 2020-10-16
Bump qualitycontrol to v1.3.3

Barthelemy:


Barthelemy:
this should fix the macos build error with the new xcode

awegrzyn (approved):

PR #2547, opened 2020-10-16
Bump Monitoring to v3.4.0 and use alibuild-recipe-tools

awegrzyn:


monitoring.sh
```diff
@@ -38,22 +39,11 @@ if [[ $ALIBUILD_O2_TESTS ]]; then
...
-}
-set version $PKGVERSION-@@PKGREVISION@$PKGHASH@@
-module-whatis "ALICE Modulefile for $PKGNAME $PKGVERSION-@@PKGREVISION@$PKGHASH@@"
-# Dependencies
-module load BASE/1.0                                                                                \\
-            ${BOOST_REVISION:+boost/$BOOST_VERSION-$BOOST_REVISION}                                 \\
-            ${APMON_CPP_REVISION:+ApMon-CPP/$APMON_CPP_VERSION-$APMON_CPP_REVISION}                 \\
-            ${GCC_TOOLCHAIN_REVISION:+GCC-Toolchain/$GCC_TOOLCHAIN_VERSION-$GCC_TOOLCHAIN_REVISION}
-
+alibuild-generate-module > etc/modulefiles/$PKGNAME
```
ktf, line 45: As a note, you can use `alibuild-generate-module --bin --lib` and it will generate the LD_LIBRARY_PATH and PATH automatically. It will not however export the `MONITORING_ROOT`, which you will have to do by hand.

awegrzyn: Ok, cool, I'll do that. On the other hand I think no1 is using the env var anymore so I'll just drop it.

PR #2549, opened 2020-10-19
Bump to v0.0.7-2

mfasDa:


preghenella (approved):

mfasDa:
In the new tag the verbosity is reduced for Herwig. Herwig by default prints the first 10 events to the log file. This leads to log files of the order of 1 MB. For the grid production the log file size needs to be reduced, printing only the first event. This needs to be specified in the Herwig configuration. The Herwig log file should be kept as it contains useful information like the kinematic cuts and the cross section.

Relates to ALIROOT-8558.

PR #2555, opened 2020-10-21
[O2-1736] Add missing python modules to o2-dataflow defaults

Barthelemy:


awegrzyn (approved):

Barthelemy:
it seems to be stuck...

ktf:
@TimoWilken can you check why this one is not being picked up?

TimoWilken:
I'll look into it @ktf!

TimoWilken:
It's the same problem as AliceO2Group/AliceO2#4654. Should be fixed now.

PR #2557, opened 2020-10-22
Change realpath to readlink in xrootd script

marcovanleeuwen:
This is needed for some SL6 systems. The same was done for some other packages, so it seems that readlink works for everyone.

### DIFF
--- a/xrootd.sh
+++ b/xrootd.sh
@@ -15,7 +15,7 @@ build_requires:
 ---
 #!/bin/bash -e
 [[ -e $SOURCEDIR/bindings ]] && XROOTD_V4=True && XROOTD_PYTHON=True || XROOTD_PYTHON=False
-PYTHON_EXECUTABLE=$( $(readlink -f $(which python3)) -c 'import sys; print(sys.executable)')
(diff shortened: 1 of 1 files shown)


xrootd.sh
```diff
@@ -15,7 +15,7 @@ build_requires:
 ---
 #!/bin/bash -e
 [[ -e $SOURCEDIR/bindings ]] && XROOTD_V4=True && XROOTD_PYTHON=True || XROOTD_PYTHON=False
-PYTHON_EXECUTABLE=$( $(realpath $(which python3)) -c 'import sys; print(sys.executable)')
+PYTHON_EXECUTABLE=$( $(readlink -f $(which python3)) -c 'import sys; print(sys.executable)')
```
ktf, line 18: This will not work on mac.

marcovanleeuwen: Ah, I see. somehow the Mac does not like the -f option. Checking again, I see that it is not really needed; I will update the pull request

ktf:
Still fails on macOS (ignore Rivet, it got added back while we were reorganizing the tests to throw more resources at them).

marcovanleeuwen:
Closing this one, I will try a new one to make sure that it removes the -f option

PR #2562, opened 2020-10-24
Change realpath to readlink (without -f this time)

marcovanleeuwen:
Trying again with readlink. This works for other recipes, so would also expect it to work here..

### DIFF
--- a/xrootd.sh
+++ b/xrootd.sh
@@ -15,7 +15,7 @@ build_requires:
 ---
 #!/bin/bash -e
 [[ -e $SOURCEDIR/bindings ]] && XROOTD_V4=True && XROOTD_PYTHON=True || XROOTD_PYTHON=False
-PYTHON_EXECUTABLE=$( $(readlink -f $(which python3)) -c 'import sys; print(sys.executable)')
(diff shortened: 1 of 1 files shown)


pzhristov:
@marcovanleeuwen What is the issue you try to solve? I tested on my mac and in fact realpath works better than readlink:
[...hristov/alisw/alidist] readlink `which python3`
../Cellar/python@3.9/3.9.1_4/bin/python3
[...hristov/alisw/alidist] realpath `which python3`
/usr/local/Cellar/python@3.9/3.9.1_4/Frameworks/Python.framework/Versions/3.9/bin/python3.9

pzhristov:
In addition, SLC6 is not supported anymore, so we have dropped the SLC6 builds since December.

marcovanleeuwen:
Hi Peter,

This change is indeed for SLC6; we have a local farm which still runs SLC6, so I thought it would be useful to keep this working.

Marco.

davidrohr:
For reference, Full CI fails due to https://alice.its.cern.ch/jira/browse/O2-2034

pzhristov (commented):
I propose to move the use of readlink in the case statement only for SLC6, this will not affect MacOS where we have only realpath. It is time to process this PR.

marcovanleeuwen:
This pull request is no longer needed

PR #2563, opened 2020-10-25
Misc Fixes

ktf:


ktf:
@jgrosseo as a side effect of generating the modulefile with proper "recursion protection", this should speedup the alienv time.

ktf:
Hold on this. I understood the issue with libtirc and I would like to fix it before this is merged.

PR #2570, opened 2020-10-28
Bump fmt

davidrohr:
version 6 has a broken assertion which led to code being checked into O2 that will be incompatible with version 7, so I think it is a good idea to just bump.

### DIFF
--- a/fmt.sh
+++ b/fmt.sh
@@ -1,6 +1,6 @@
 package: fmt
 version: "%(tag_basename)s"
-tag: 6.1.2
+tag: 7.1.0
 source: https://github.com/fmtlib/fmt
 requires:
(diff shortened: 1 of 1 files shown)


davidrohr:
@ironMann : There seems to be a problem with the newer libfmt version in DataDistribution. Could you have a look?

ironMann:
@davidrohr fixed. LGTM

davidrohr:
@ironMann : I don't understand, is there a new version of DataDistribution?

ironMann:
> @ironMann : I don't understand, is there a new version of DataDistribution?
I'll push one shortly.

davidrohr:
error unrelated, merging

PR #2585, opened 2020-11-05
Adding recipe for TPC FEC utils v0.1.0

sk3x:


awegrzyn:
I'm not sure that storing GitLab token as clear text is good idea. @ktf Would it be possible to add it as GitHub Secret?

awegrzyn:
@vascobarroso As this will be running at FLP shall we add it to `@alisw/dataflow-experts` in CODEWONERS?

tpcfecutils.sh
```diff
@@ -0,0 +1,72 @@
...
+cp ${BUILDDIR}/compile_commands.json ${INSTALLROOT}
+
+# Reduce number of jobs if invoked by Jenkins
+if [ ! "X$JENKINS_HOME" = X ]; then
+  JOBS=1
+fi
+make ${JOBS+-j $JOBS} install
+
+#ModuleFile
+mkdir -p etc/modulefiles
```
ktf, line 51: I would suggest you use the new `alibuild-generate-module` utility which you can find in `alibuild-recipe-tools`. Have a look at e.g. the `boost.sh` recipe to see how it's used.

ktf: I will merge this, you can fix it later.

ktf:
Merging since nothing depends on it yet.

PR #2587, opened 2020-11-05
Bump AliECS to v0.17.80

teo:


sy-c (approved):

ktf:
I would recommend to move to use alibuild-recipe-tools to generate the modulefiles, both for modules 4 compatibility, correctness, and performance (no recursion in dependencies...).

teo:
> I would recommend to move to use alibuild-recipe-tools to generate the modulefiles, both for modules 4 compatibility, correctness, and performance (no recursion in dependencies...).

Thanks @ktf I'd be happy to, but I struggle to find any documentation on this. Can you point me where to RTFM please?

ktf:
RTFM is at http://alisw.github.io/alibuild/reference.html#runtime-environment, you can probably just copy what is done in boost.sh or python.sh.

teo:
@ktf alright, thanks! I've put something together, currently testing locally. I'll submit it in a separate PR.

PR #2590, opened 2020-11-05
Using alibuild-generate-module

sk3x:


ktf (approved):

tpcfecutils.sh
```diff
@@ -3,12 +3,12 @@ version: "%(tag_basename)s"
 tag: v0.1.0
 requires:
   - boost
-  - "GCC-Toolchain:(?!osx)"
   - Python
   - ReadoutCard
   - LLA
 build_requires:
   - CMake
+  - "GCC-Toolchain:(?!osx)"
```
ktf, line 11: Mmm, no this is most likely wrong, GCC-Toolchain contains the C++ runtime, so most likely you need it as a full requires, unless you had the bright idea (no pun) to use plain C for it.

PR #2593, opened 2020-11-06
Bump to v6-20-08

ktf:


ktf:
@pzhristov

pzhristov:
@ktf It seems we can go directly to v6-22-02. The O2 tests on Ubuntu 18.04 are OK, the AliRoot tests also. I am not sure which of the two tags contains the fixes for bulk reading.

ktf:
Ok, let's try.

pzhristov:
@ktf Unfortunately we have a problem on Mac - some O2 tests were not OK. In addition, libPyROOT.so was renamed and the compilation of AliPhysics fails, so v6-20-08 for the moment is the only option for upgrade of Root.

ktf:
Rolling back...

pzhristov (approved):
All the tests are OK, we probably can merge and expose the new version to the analysis trains and the vertical slice activities.

ktf:
Looks like the mac build is still pending? @MichaelLettrich could you check that?

pzhristov:
I run the O2 tests om macpromat (there I have python@3.9, so some changes in Python-modules-list.sh were needed). everything went smoothly. I also successfully compiled O2Suite.

ktf:
Merging this to have it in tonight build.

PR #2597, opened 2020-11-11
Fix curl build when OpenSSL libs are installed on the system

Atlantic777:
This PR fixes a problem when curl fails to build if OpenSSL is installed on the system, for example in alisw/slc6-builder. The build fails with a message that there is a mismatch between SSL header and library version, even though curl was explicitly configured to use OpenSSL from alidist. 

The problem comes from the curl config script which prefers the system SSL if it's detected, but even then the headers and libs can get mixed up. There is a trace of some discussion in the curl git log and I don't expect that this can be changed upstream. here's that commit message:

```
commit 338f427a24f78a717888c7c2b6b91fa831bea28e
Author: Jay Satiro <raysatiro@yahoo.com>
Date:   Mon Apr 24 03:13:28 2017 -0400

    configure: stop prepending to LDFLAGS, CPPFLAGS

    - Change prepends to appends because user's LDFLAGS and CPPFLAGS should
      always come first so they're searched before ours.

    Bug: https://github.com/curl/curl/issues/1420
    Reported-by: Helmut K. C. Tessarek
```

### DIFF
--- a/curl.sh
+++ b/curl.sh
@@ -17,6 +17,9 @@ fi
 
 rsync -av --delete --exclude="**/.git" $SOURCEDIR/ .
 
+sed -i.deleteme 's/CPPFLAGS="$CPPFLAGS $SSL_CPPFLAGS"/CPPFLAGS="$SSL_CPPFLAGS $CPPFLAGS"/' configure.ac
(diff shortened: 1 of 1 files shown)


ktf:
Merging this, since it probably fixes some issues when running slc7 on ubuntu.

PR #2602, opened 2020-11-12
Improved detection of kernel.h (Ubuntu)

pzhristov:


pzhristov:
@MichaelLettrich Here we have the same issue on Mac - python@3.9 needs additional patch for Python-modules-list.sh. The easiest is to link python@3.8
`brew link --force --overwrite python@3-8`

PR #2608, opened 2020-11-14
Fix for xcode 12.2

ktf:


pzhristov (approved):
For me both clang 11.0.0 and the new version of arrow compile. There are other issues in lhapdf and JAlieEn that have to be fixed in addition, plus the patch of libwebsocket.

ktf:
Verified also on my two macs: Catalina and now BigSur. Merging.

PR #2615, opened 2020-11-16
Fix lhapdf build scripts

TimoWilken:
The source URL now returns a HTTP redirect, so we need to pass `-L` to `curl` to follow it.


ktf:
Merged as obviously broken. @alisw/pagmc those are actually test files, correct? Maybe it makes sense to store them in git...

PR #2617, opened 2020-11-16
Minor fix in the JAliEn module file

Atlantic777:
The xjalienfs package was declared but not loaded in the JAliEn module file. This PR should fix the issue.


ktf:
Please, move to use alice-recipe-tools.

PR #2621, opened 2020-11-18
datadist: revert and bump again to 0.8

ironMann:
@vascobarroso @ktf 
This should do it


ktf:
Merging this as it's currently broken.

PR #2623, opened 2020-11-19
utf8proc: use cmake for building

ktf:
By default this will produce an archive library, reducing our
runtime footprint.


ktf:
@aphecetche this should solve at least one of the problems you reported me about the tarballs.

teo (approved):

aphecetche:
@ktf thanks. Is there a way to know if/when the change has actually been used in some build and deployed to the bin repo ?

ktf:
In principle we keep building and uploading as pr get merged, so it should be there. @TimoWilken we need to find a way to mark the last commit uploaded (e.g. via a tag).

PR #2627, opened 2020-11-20
xjalienfs :: hot-fix update to 1.2.8

adriansev:
* Fix typo in usage of tokens as JALIEN_TOKEN_{CERT,KEY} env vars
* Disable for all user/human communication the return of json keys
* fix global status of process output and with this the return of nowb functions; add showmsg and showkeys SendMsg options

### DIFF
--- a/xjalienfs.sh
+++ b/xjalienfs.sh
@@ -1,6 +1,6 @@
 package: xjalienfs
 version: "%(tag_basename)s"
-tag: "1.2.7"
+tag: "1.2.8"
 source: https://gitlab.cern.ch/jalien/xjalienfs.git
 requires:
  - "OpenSSL:(?!osx)"


costing (approved):

sawenzel (approved):

adriansev:
@ktf there are some strange logs here: 2 are telling that there are no logs and one is complaining about `FATAL: cannot find python3-config in your $PATH. Cannot enable boost_python` 
I was wondering about the usefulness of these tests: this is a standalone python application, it have nothing to do with the actual software. There is https://gitlab.cern.ch/nhardi/jalien-ci which have tests of alien.py ,, if you envision some more checks just tell me what other things should be checked and i will add them to the tests. It is quite pointless to do so many builds and compilations when one can only do `pip install alienpy`

adriansev:
ping @ktf ? can this somehow move forward? thanks a lot!!

ktf:
Merging this, since in any case there are no relevant tests.

ktf:
@adriansev notice that @alisw/alien-expert and @alisw/alidist-maitainers is probably a better route to fast track a PR over the weekend, rather than pinging me directly.

adriansev:
> @adriansev notice that @alisw/alien-experts and @alisw/alidist-maitainers is probably a better route to fast track a PR over the weekend, rather than pinging me directly.

@ktf actually i had no idea how can i mention groups .. when i tried, seeing that there is no completion i assumed that this is a wrong way to do it.

PR #2628, opened 2020-11-23
Don't use CMake >= 3.19.0 and leave a message

davidrohr:
We cannot bump right now because there is some incompatibility with GPU compilation. I'll check if the problem is on our side or a regression in CMake. So long block >= 3.19


ktf (approved):

davidrohr:
@MichaelLettrich @TimoWilken  : Could you check why this fails on MacOS and why there is no failure log?

PR #2634, opened 2020-11-24
Explicitly list AliECS components as dependencies of O2Suite

teo:


teo:
@awegrzyn this should make the automatic bump possible. Can it be merged? I don't know what the build failures are but I suspect them to be unrelated.

awegrzyn (approved):

PR #2554, opened 2020-10-21
Bump FairMQ to v1.4.26

rbx:
Upon request from @AndreyLebedev, related to the ongoing DDS/ODC developments.

@Barthelemy this also includes the changes discussed in https://github.com/AliceO2Group/InfoLogger/pull/52.

PR #2635, opened 2020-11-25
Bump FairMQ to 1.4.27

rbx:
Supersedes #2554.
Includes fix for [O2-1850](https://alice.its.cern.ch/jira/browse/O2-1850).

Requires DDS 3.5.3 (#2636) because FairMQ 1.4.27 requires that as minimum version (which includes multiple sessions support).

### DIFF
--- a/dds.sh
+++ b/dds.sh
@@ -1,6 +1,5 @@
 package: DDS
-version: "3.5.1"
-tag: "3.5.2"
+version: "3.5.3"
 source: https://github.com/FairRootGroup/DDS
 requires:
   - boost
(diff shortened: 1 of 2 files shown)


davidrohr:
@TimoWilken @ktf : Could you check what is going on with the AliRoot/alidist aand O2/alidist CI? They are failing without a build log.

ktf:
The issue is most likely DPMJET repository misbehaving. CERN/IT already informed. Will move it to Github as soon as we have a bit of time.

davidrohr:
Compiled and tried it locally, but for me it still fails as described in the JIRA ticket.

davidrohr (approved):
works now, sorry for the false report before

davidrohr:
@MichaelLettrich @ktf : MacOS seeems to fail due to an autotools error, which seems unrelated. Is this a known issue?

ktf:
no, it's a new one rebuilding autotools, I am afraid.

PR #2636, opened 2020-11-25
Bump DDS to 3.5.3

rbx:
Required by #2635.

### DIFF
--- a/dds.sh
+++ b/dds.sh
@@ -1,5 +1,5 @@
 package: DDS
-version: "3.5.1"
+version: "3.5.3"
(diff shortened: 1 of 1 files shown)


dds.sh
```diff
@@ -1,6 +1,6 @@
 package: DDS
 version: "3.5.1"
```
davidrohr, line 2: Could you change also the version, or use tag_basename?

rbx: Done.

ktf:
Folded in #2635

PR #2638, opened 2020-11-26
Drop permissions to access the repo

ktf:


ktf:
@sk3x can you please either add alibuild as a user or make the repository public?

sk3x:
I added alibuild as a user

ktf:
Can we simply make it public, actually? That would simplify things a lot.

PR #2639, opened 2020-11-26
Drop authentication from TpcFecUtils

ktf:


ktf:
@alisw/dataflow-experts please make sure your repository can be accessed by the alibuild user. I also would prefer that the the submodule is treated as a separate package.

vascobarroso:
@sk3x can you comment on the request from @ktf ?

ktf:
Merging this as it's breaking the CI.

sk3x:
The repository now can be accessed by the alibuild user.

What exactly is the issue with the deploy tokens?
Will removing them non break things for some other non-alibuild users?

ktf:
If they build from scratch, yes. They would need to download it prebuilt from the server.

What is the point for making it private if you then ship the tokens in the recipe?

sk3x:
Ok, I think #2638 and this are one discussion, so maybe let's have it here.

What we are interested in is being able to build RPMs for CentOS on FLPs. The FEC utils depend on some other packages managed here (essentially ReadoutCard and LLA). I wasn't aware there is an alibuild user, deploy tokens were our compromise to get that running successfully.

Can this simply be exempted from the CI?

PR #2641, opened 2020-11-27
Dropping TpcFecUtils from O2Suite for now

ktf:


ktf:
@awegrzyn @sk3x this should still allow you to create an RPM for it, however it will exclude it from the CI.
@TimoWilken this should workaround the issue you see.

awegrzyn:
@ktf We use O2Suite meta package to make sure that we do not install several versions of a dependency. In addition we already instructed our CI to pick up TpcFecUtils version from O2Suite RPM.

Removing it from O2Suite will break our pipeline, summarising, we really need it as O2Suite dependency.

ktf:
@awegrzyn adding this package is breaking everything for the whole collaboration, including ongoing analysis, so we need to find a way to introduce it gracefully. Can we make the private submodule public, given in any case there is a copy of the deploy key in github?

awegrzyn:
All right I see, let's try to find a solution with TPC team

sk3x:
I still don't really understand what the actual problem is, except that something breaks stuff somewhere.

With that deploy key, the repo practically behaves like a public repo.
So what is the real problem? Handling of the git config command? Handling of the submodule itself?
Could you provide some error reports from the system where it is failing?

ktf:
We do not know exactly yet. Can we move the discussion to mattermost?

vascobarroso:
@ktf @sk3x can we revisit this ? The fact that TpcFecUtils is not part of O2Suite is a major nuisance in our release procedure and our CI.

sk3x:
We removed the submodule as we had discussed. I was under the impression that this is fixed now. Is it?

ktf:
yes can you re-add the dependency and we check?

awegrzyn:
The next issue was GL credentials as far as I remember

ktf:
wasnt the submodule itself to have the issue?

PR #2648, opened 2020-12-01
O2: allow direct usage of CGAL

ktf:


ktf:
@davidrohr this is probably what we need. GMP and MPFR are compiled statically to avoid interference with the system one which sometimes gets carried in by GCC.

davidrohr:
@ktf : I guess this is correct, Actually I cannot reproduce the build error of AliceO2Group/AliceO2#4618 in my local CI container, and actually the PR has passed the O2 CI before. So I believe it is some kind of incremental build error.
In any case this PR is correct to make CGAL a direct dependency, so LGTM.

One question though:
Are build_requires dependencies propagated? I.e. CGAL's CMake target CGAL::CGAL has the GMP include folder in its list of include folders, which means every package that builds against CGAL also needs GMP at build time.

davidrohr (approved):

ktf:
IIRC, when indirect, they are not propagated if the direct dependency is already there, by design, so that e.g. you do not download CMake if you do not need it for building, just because one of your dependencies needed it. I think that's the issue (and why you cannot reproduce).

davidrohr:
ok, thx, then let's wait for one CI to pass and then let's merge this, and then I'll see if the other PR finally becomes green :)

PR #2650, opened 2020-12-01
[xjalienfs] Deal correctly with existing virtual env

aphecetche:
Proposing changes to be able to build from within an existing virtual env : can't use the `--user` option for pip, and don't need to play with paths.

### DIFF
--- a/xjalienfs.sh
+++ b/xjalienfs.sh
@@ -11,18 +11,24 @@ requires:
 ---
 #!/bin/bash -e
 
-# env PYTHONUSERBASE="$INSTALLROOT" pip3 install --user -r alibuild_requirements.txt
-env PYTHONUSERBASE="$INSTALLROOT" ALIBUILD=1 python3 -m pip install --ignore-installed --user file://${SOURCEDIR}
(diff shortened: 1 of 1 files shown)


aphecetche:
@alisw/alien-experts as far as I can tell the CI failure is not directly directed to this PR. Can it be reviewed ? Thanks.

aphecetche:
@alisw/alien-experts any comment on this PR ?

costing:
Adding @adriansev to the thread

adriansev:
i have no opinion on this.. as long as locally `alibuild ... build xjalienfs` works then it's ok... more so if the checks are successful (and the seen errors from O2 are completely unrelated) so, within the mentioned conditions, ACK from me

aphecetche:
@adriansev ok, thanks.
Just wanted to know if there was some objections or not.  
But if ok with you I'd propose to merge this early next year only. Just to avoid any unexpected collateral damage during the Christmas break ;-)

pzhristov:
I tested the compilation on MacOS, everything is OK. If everybody agrees, I will merge this PR.

PR #2654, opened 2020-12-02
libxml2: move to our own clone

ktf:
Relying on Gnome.org is simply too flaky.

### DIFF
--- a/libxml2.sh
+++ b/libxml2.sh
@@ -5,7 +5,7 @@ build_requires:
   - autotools
   - zlib
   - "GCC-Toolchain:(?!osx)"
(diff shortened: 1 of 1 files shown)


ihrivnac:
@ktf,
Could you, please, give the reason(s) why we need our own clone of  libxml2? The description in the PR is too vague. Thank you.

ktf:
Because we eliminate yet another (unreliable) moving part. We need github in any case and gitlab.gnome.org has been unreliable and moved / not worked quite a few time in the past. Given we are not changing the external that often, I think it's ok to keep our own copy.

ihrivnac:
Ok, thanks.

PR #2658, opened 2020-12-02
Delete defaults-o2-dev-fairroot.sh

ktf:


ktf:
Not used anymore in any of the tests.

PR #2661, opened 2020-12-03
Add default for FLUKA

ktf:


pzhristov:
Do we really need one more default, especially if we want to reduce their number?

ktf:
No, I agree we should probably have it together with the root6 one and always release with FLUKA support. I can do that if you are ok with it. However, until then I'd rather not change the O2 default because it otherwise requires manual intervention every time someone requests a new tag.

pzhristov:
Ok, in this case probably we can remove root6-next and keep only root6 with the latest certified version v6-20-08-alice1.

ktf:
Yes, agreed.

ihrivnac:
I see now 11 defaults:
defaults-coverage.sh
defaults-generators.sh
defaults-jalien.sh
defaults-next-root6.sh
defaults-o2-dataflow.sh
defaults-o2-dev-fairroot.sh
defaults-o2.sh
defaults-prod-latest.sh
defaults-release.sh
defaults-user-next-root6.sh
defaults-user.sh

Besides the next-root6, are all of them in use?
Can there be added short comments about their purpose?

ktf:
Apart from defaults-o2-dev-fairroot.sh (and maybe defaults-coverage.sh)  which I just deleted they are all in use or referenced in some documentation. We should probably consolidate to a couple of them (say root5 and release) rather than documenting them (which is in any case probably done somewhere).

PR #2626, opened 2020-11-19
Update node-gyp when building node recipe

awegrzyn:

PR #2676, opened 2020-12-09
Bump qualitycontrol to v1.6.1

Barthelemy:


Barthelemy:
@awegrzyn That should fix the warning in QC we see in macos and thus fix the build.

awegrzyn (approved):

Barthelemy:
@awegrzyn npm error it seems, related to python

awegrzyn:
Yes, I cannot reproduce it on my setup so I created a this PR https://github.com/alisw/alidist/pull/2626 that might fix the issue.
The problem is that this PR is "red" for past 20 days and as I'm not able to test it locally I'd like to have all checks green.

ktf:
#2626 has been red for a while due to the issue fixed by this PR, AFAICT. Merging it.

ktf:
(any other issue with #2626 was solved 2 weeks ago, BTW...)

Barthelemy:
I have actually made a full version instead of the patch here: https://github.com/alisw/alidist/pull/2677

PR #2685, opened 2020-12-11
Bring back TpcFecUtils to O2Suite

awegrzyn:


tpcfecutils.sh
```diff
@@ -9,7 +9,7 @@ requires:
 build_requires:
   - CMake
   - "GCC-Toolchain:(?!osx)"
-source: https://gitlab.cern.ch/alice-tpc-upgrade/alice-tpc-fec-utils.git
+source: https://gitlab+deploy-token-1303:ivjQdcMRX9qpxdv4RCcM@gitlab.cern.ch/alice-tpc-upgrade/alice-tpc-fec-utils.git
```
ktf, line 12: This was breaking the builds in the past. Why do you think it will not now?

PR #2692, opened 2020-12-15
Put back token to TpcFecUtils recipe

awegrzyn:


graduta (dismissed):

tpcfecutils.sh
```diff
@@ -9,7 +9,7 @@ requires:
   - LLA
 build_requires:
   - CMake
-source: https://gitlab.cern.ch/alice-tpc-upgrade/alice-tpc-fec-utils.git
+source: https://gitlab+deploy-token-1303:ivjQdcMRX9qpxdv4RCcM@gitlab.cern.ch/alice-tpc-upgrade/alice-tpc-fec-utils.git
```
ktf, line 12: No token, please.

awegrzyn: ah, I thought no token in submodule :(

awegrzyn:
This just brings back the token back to the recipe, it's still out of `O2Suite`

ktf:
This will still break stuff, I am afraid. Didn't we agree to refactor the code so that the private part is not actually needed?

awegrzyn:
As far as I remember the reported issue concerned submodule only, and Stefan proceed as you proposed: `the third option is that the two header files which are needed from the submodule are moved to the main package`.

How it can break stuff if `TpcFecUtils` is not part of `O2Suite` ?

awegrzyn:
@ktf As I mentioned above, `TpcFecUtils` is not part of `O2Suite` anymore, this is just to be able to compile it manually.

PR #2694, opened 2020-12-15
tpcfecutils: build using system python when on mac / ubuntu

ktf:


ktf:
@Barthelemy I suspect this is the reason why the QC builders (and other builders) are now failing.

ktf:
Maybe not, actually, since this is not yet it...:-(

PR #2698, opened 2020-12-16
Use system cURL in Monitorng and ppconsul

awegrzyn:


sy-c (approved):

monitoring.sh
```diff
@@ -5,8 +5,7 @@ requires:
   - boost
   - "GCC-Toolchain:(?!osx)"
   - "ApMon-CPP:(?!osx)"
-  - "system-curl:(slc8)"
-  - "curl:(?!slc8)"
+  - system-curl
```
ktf, line 8: This will break JAlien. So it you want to customise it this way, you should do so in o2-dataflow.

ktf (changes requested):

monitoring.sh
awegrzyn: ok so I'll add `curl` to disabled packages and then how can I enable `system-curl` on all systems?

awegrzyn: On the other side, shouldn't this be enforced in `o2` defaults? Instead of disabling it in `o2-dataflow`?

ktf: You probably want to simply override the dependencies (make sure you override all of them). I would say that the default `o2` should stay as is, i.e. the most comprehensive setup which most people will use.

Barthelemy: I am worried about the fact that JAlien would break when using `system-curl`. Why is that ? Shouldn't JAlien be fixed ?

awegrzyn: Right I think discussion which default is irrelevant.
I just try to resolve it in most reasonable way. These packages do not require "special" curl so if AliEn does, it should enable this in its default no? I thought this has been our policy since beginning?

ktf: @Barthelemy @costing @Atlantic777 @odatskov should comment on your question. I think the issue is that the curl provided on CC7 does not work with the Grid / AliEn authentication mechanism.

@awegrzyn this is not about "importance" as there is no "alien" default. The policy has always been so far:

* o2 is the most comprehensive default, which is deployed on CVMFS and used for the Grid productions
* If people need a stripped down installation, they can have their own default and disable things

This is because disabling foreseen tools is easier than enabling unforeseen ones.

ktf:
Why do you need system curl, BTW?

ktf:
Asking @costing to review.

monitoring.sh
Atlantic777: Hello @costing @ktf @Barthelemy @odatskov @awegrzyn,

Here's some comments on the state of AliEn, JAliEn and SSL. By the way, I think we don't use curl directly but @yuw726 mostly worked on that and @odatskov took over. Here's what I know:
  - ROOT5 / AliEn builds depend on OpenSSL 1.0 becase of XRootD 3
  - Because of that, everything in alidist was compiled against OpenSSL 1.0, even JAliEn-ROOT and libjalien.
  - Mixing software that is linked to different OpenSSL versions leads to nasty segfaults and silent failures.
  - There is nothing really to be fixed on the JAliEn-ROOT side, the last time I tested it worked on any SSL version. I suppose everything JAliEn related could be switched to the new SSL and work without problems with `system-curl` if needed.
  - Keep in mind that AliEn-ROOT-Legacy can use system OpenSSL on CC6, but needs OpenSSL 1.0 from alidist on CC7 and newer and it's not just AliEn, it's anything linking to OpenSSL (Python, XRootD, ROOT, curl).
  - The curl/JAliEn incompatibility appears in O2, and if curl and JAliEn are not compiled with compatible OpenSSL versions then I am not sure what can be done from JAliEn side except to bring them to the same version.

I hope this answers some questions, but maybe we can include @yuw726 as he spent some time working with SSL. @costing @odatskov do you maybe know if there is something more besides this?

davidrohr: Which builds do still use ROOT5? It is only AliRoot with Release defaults, or do I miss something? Perhaps we could swith the default OpenSSL to 1.1.1, and then just fall back to 1.0 for AliRoot / release?
In any case, I have opened a PR to at least test the builds with a new OpenSSL here #2760

PR #2700, opened 2020-12-16
Fix python3 system check on CC7

awegrzyn:
There's no `python-config` but `python3-config` on CC7. I'm not sure whether this does not break other OS.


pzhristov (approved):

ktf:
It used to work because:
* python-config is there in many places due to python2 compatibility
* CC7 always download prebuild packages
* The code itself does not check for python3, merely that pyconfig.h can be included... We should add a check for a python version.

PR #2710, opened 2020-12-17
Always use werror in QC

Barthelemy:


mfasDa:
I had yesterday my issues with a different package (ReadoutCard) that suddenly stopped compiling under *ubuntu* - a wide spread distribution in the collaboration - because of Werror. With this change you will make the package compiler dependent. Even though this might be decided by the O2 management (I don't know what is the push for this): I don't think it is a good idea what you are doing here, even if Werror can be useful to force warnings to be corrected - it might be that your package will stop work on platforms that are wide spread in the collaboration. This is good for a debug setup but not for a production setup.

ktf:
Please fix this in the QC Cmakelists

ktf (changes requested):
Should be done in the CMakeLists.txt, IMHO.

ktf:
@mfasDa I would say we spend more time chasing the kind of issues this trick brings us than the required support for ubuntu (for which we have even CI tests now). If very big issues arise, we can always turn of the spurious warnings.

Barthelemy:
so let's drop this PR and continue happily with using werror in the PRs checks.

Barthelemy:
https://alice.its.cern.ch/jira/browse/QC-512

PR #2712, opened 2020-12-18
Bump qualitycontrol to v1.7.2

Barthelemy:


sy-c (approved):

ktf:
Merging this to fix the QC issues. For the error in QCG we need to either exclude autotools everywhere on osx (assuming nowadays we have a recent enough version to compile all our externals, which was not the case a few years back) or to bump gyp, I guess. The libtool we ship is to date the most recent release.

PR #2713, opened 2020-12-18
O2: Fix XRootD detection

ktf:


ktf:
@TimoWilken I guess we need to restart ubuntu as well?

TimoWilken:
@ktf restarted and marked yellow.

PR #2715, opened 2020-12-18
Drop autotools macOS

ktf:


pzhristov (approved):

pzhristov:
I propose to merge this PR.

MichaelLettrich:
@ktf @pzhristov can we merge this? maybe it allows us to solve the macOS issues in #2738

pzhristov:
@MichaelLettrich OK, merged.

ktf:
This will trigger a massive rebuild...

pzhristov:
Yes, sooner it starts, sooner it will finish :)

ktf:
ok... Once we stabilize alibuild 1.7.0 all those platform specific things will be better done with a platform default (which is like a default, but automatically added on a per platform basis). This will avoid having to touch every single recipe to do platform specific changes.

PR #2716, opened 2020-12-18
Revert "Update simulation.sh (#2683)"

ktf:
This reverts commit b085c454fe5ba3580c57bfa1d7270c6935cb3f1b. Currently the fact that fluka_vmc does not relocate correctly breaks the builds.

### DIFF
--- a/simulation.sh
+++ b/simulation.sh
@@ -4,7 +4,6 @@ requires:
   - GEANT4_VMC
   - GEANT4
   - GEANT3
-  - FLUKA_VMC
 ---
 #!/bin/bash -e
 


sawenzel:
Is someone following the problem with relocation up? Do we have a JIRA ticket?

ihrivnac:
@ktf , @sawenzel , I checked the fluka_vmc CMake files and do not see anything missing wrt Geant3 ones; all the changes for relocability which were applied during the last year seems to be present.
@ktf, can you point at the problems which caused this MR?

ihrivnac:
@ktf , @sawenzel, 
A JIRA ticket is created: https://alice.its.cern.ch/jira/browse/O2-1971 
But we need the feedback for the problems observed.

PR #2719, opened 2020-12-24
Use python3 for o2checkcode

ktf:


davidrohr:
Thx, this seems correct. There is no python symlink but only the python3 binary in the container, so this should fix the reported problem.

PR #2722, opened 2020-12-28
Force SHERPA to build under standard c++11

mfasDa:


mfasDa:
SHERPA with the current version is using features that are removed in c++17, consequently building under the o2 defaults fails. Therefore the c++-standard for building SHERPA must be fixed manually to c++11.

preghenella (approved):

preghenella:
hi @mfasDa @MichaelLettrich @ktf 
is the macOS error relevant? there are not details to inspect.
Thanks

preghenella:
not clear what the error for macOS is.
There are no details.
@MichaelLettrich @ktf do you have any hints?

mfasDa:
@preghenella SHERPA 2.2.8 is using std::unexpected_handler in one class which has been declared deprecated in c++11 and removed in c++17, see https://en.cppreference.com/w/cpp/error/unexpected_handler. clang++ <= 11 and gcc still seem to provide it even for c++17, but clang++12 removed it (error: unknown class ...). So building SHERPA under the o2 defaults (what I usually do for all my packages as I want to have c++17 and I don't want to duplicate packages) fails. Furthermore as the transition from run2 to run3 stack becomes imminent also for generator packages compatibility with the o2 defaults should be considered.

Probably also an upgrade to the latest SHERPA version 2.2.10 should be considered - eventually this already solves c++17 compatibility.

preghenella:
@mfasDa 
we upgrade SHERPA, no problems. Did you try if this is indeed working and solving your problem?
We do not officially build it with o2 defaults, but if this upgrade makes it build with both the generators and o2 defaults, let's upgrade

preghenella:
I would possibly suggest to move to the official SHERPA repo as well, unless there are drawbacks
https://gitlab.com/sherpa-team/sherpa/

mfasDa:
Ideally, yes. however there was a problem in the past that SHERPA built a manual which didn't manage to build, because it was relying on a package we don't ship, so we needed to hack the module make in order to disable building the manual. Though we asked our SHERPA contact whether it would be feasible to make the build of the manual optional I suspect this is not yet the case (at least up to v2.2.8). So I suspect we still need to stay with our fork.

preghenella:
I noticed that, but...
instead of hacking the source code, can we deal with that in the recipe?
It is a single line to be removed in `Makefile.am`.

mfasDa:
> I noticed that, but...
> instead of hacking the source code, can we deal with that in the recipe?
> It is a single line to be removed in `Makefile.am`.

This was the plan. Unfortunately the configure script needs to support this, otherwise we cannot propagate the information. This was not the case at least until the previous version.

preghenella:
@mfasDa 
I am not sure I get your last comment.
What I meant was to tweak the `Makefile.am` in the `sherpa.sh` recipe "before" `./configure`, such that we remove the line with the `Manual` ... 

basically, instead of doing it with a commit in the `sherpa` repo, like this one
https://github.com/alisw/SHERPA/commit/1180d82a02027b4892bfbdfe86b80321e20bd981
we do it by hands before starting the build.
Not 100% sure it works... but why it should not?

mfasDa:
I see, that could work

mfasDa:
Superseeded by #2723

PR #2726, opened 2021-01-04
Bumping to r3693-alice2

preghenella:
to port the patch discussed in https://github.com/alisw/POWHEG/pull/5

### DIFF
--- a/powheg.sh
+++ b/powheg.sh
@@ -1,6 +1,6 @@
 package: POWHEG
 version: "%(tag_basename)s"
-tag: "r3693-alice1"
+tag: "r3693-alice2"
 source: https://github.com/alisw/POWHEG
 requires:
   - fastjet


preghenella:
@mfasDa 
please check if not ok.

mfasDa:
@preghenella this is a duplication of #2724. In my version I however also changed to alibuild-recipe-tools, which take care of dependency handling (+ there is another small fix dealing with include directories). Therefore I would prefer to stay with my version.

I recommend to slowly move also the other packages to alibuild-recipe-tools as it makes sure no package dependency is overlooked during modulefile creation - for Herwig, Sherpa, Openloops and MadGraph I already took care of.

preghenella:
sorry, I did not notice your PR.
Closing this one.

PR #2728, opened 2021-01-04
Bump DebugGUI to v0.4.0

ktf:


ktf:
@MichaelLettrich @TimoWilken I think there is still some issue with the builders.

PR #2732, opened 2021-01-06
DataDistribution: build with 4 cores

ktf:
Right now DataDistribution is the second most expensive package to build, behind O2.


ktf:
This should speed up the alidist builds.

PR #2735, opened 2021-01-06
Automatically open a PR from rc/ branches

TimoWilken:
With alisw/ali-bot#910, daily-tags.sh pushes automatic changes to an `rc/*` branch first to avoid overwriting others' changes. This action creates a PR from that branch.

Tested and working in another repo.


TimoWilken:
I've tested the script and it works (at least in a testing repo I made for the purpose).

@ktf, it needs `ALIBUILD_GITHUB_TOKEN` defined as a secret in alisw, though. That secret so far only exists in AliceO2Group.

PR #2740, opened 2021-01-07
JAliEn-ROOT: add missing ROOT_INCLUDE_PATH

ktf:
ROOT_INCLUDE_PATH is needed also when building, not only at runtime.

### DIFF
--- a/jalien-root.sh
+++ b/jalien-root.sh
@@ -15,6 +15,7 @@ build_requires:
   - Alice-GRID-Utils
 append_path:
   ROOT_PLUGIN_PATH: "$JALIEN_ROOT_ROOT/etc/plugins"
+  ROOT_INCLUDE_PATH: "$JALIEN_ROOT_ROOT/include"
 ---
 #!/bin/bash -e
 case $ARCHITECTURE in


ktf:
@davidrohr this should fix the various:

```
   Missing FileEntry for TJAlien.h
   requested to autoload type TJAlienJDL
Error in cling::AutoloadingVisitor::InsertIntoAutoloadingState:
   Missing FileEntry for TJAlien.h
   requested to autoload type TAliceFile
```

messages.

@jgrosseo I wonder if this is what is causing problems on ubuntu.

davidrohr (approved):
thx, makes sense for me, but didn't check locally if the errors are gone. Let's see what the CI says.

ktf:
What I do not understand is why we have the JAlien dictionary loaded at all at link time, given the only place where we add it it's in the `Framework/AnalysisSupport` plugin. @aphecetche any idea of what might be going on?

ktf:
This seems to have fixed at least some of the issues. Merging.

aphecetche:
> What I do not understand is why we have the JAlien dictionary loaded at all at link time, given the only place where we add it it's in the `Framework/AnalysisSupport` plugin. @aphecetche any idea of what might be going on?

@ktf is there still a question here or has been sorted out already ?

PR #2741, opened 2021-01-07
Update o2.sh

sawenzel:
Require VecGeom


awegrzyn:
Is it ok to have it disabled in `o2-dataflow` defaults?

sawenzel:
@awegrzyn : No, this will be used even outside pure simulation. But we have pre-build binaries and it compiles reasonably fast.

PR #2746, opened 2021-01-08
Set FMT_ROOT env variable

davidrohr:
@ktf : are you OK with this? It is required such that my GPU standalone benchmark can find the alidist fmt.


ktf:
At runtime?

davidrohr:
at compile time, that standalone thing is not built by alibuild. I am doing alienv enter O2/latest fmt/latest CMake/latest to load all dependencies. But if fmt doesn't export FMT_ROOT, it is not picked up.
It seems that without the setenv the FMT_ROOT is only available during the alibuild build of O2, but not after alienv?

PR #2747, opened 2021-01-09
gcc : improve system detection

aphecetche:
Depending on how gcc was installed on the system the cc link (to gcc) might
not exist at all. But the CC environment variable might, so test that
too.

For the record, I've encountered the situation on a CentOS7 installation where gcc is installed via Spack.

### DIFF
--- a/gcc-toolchain.sh
+++ b/gcc-toolchain.sh
@@ -12,7 +12,8 @@ prefer_system: .*
 prefer_system_check: |
   set -e
   which gfortran || { echo "gfortran missing"; exit 1; }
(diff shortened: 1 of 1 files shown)


mfasDa:
I suspect I ran into a similar issue on Ubuntu 18.04, because also on that system with fresh build recipes aliBuild is building gcc though gcc-7.5.0 is installed, which should be OK (for the moment) for AliPhysics and O2.

ktf:
I think we should simply use `c99`, which is required to be present by the standard.

ktf (changes requested):
I switched to use c99 in the 10.2 PR. Would that be ok?

ktf:
Closing this to avoid rebuilding GCC in the CI. Please let me know if c99 is not ok.

PR #2748, opened 2021-01-11
Use tagged version of VecGeom

ktf:


ktf:
@sawenzel moving refs (e.g. master) make builds unreproducible and are not allowed by policy.

awegrzyn:
@sawenzel Would it be possible to use actual version instead of commit? The last tagged version is `v1.1.8`.

PR #2750, opened 2021-01-11
use tagged versions in o2dpg

sawenzel:


sawenzel:
@preghenella @ktf: I guess we need to tag (and bump) o2dpg as well instead of using master.

sawenzel:
On the other hand alidpg.sh uses "master" as well.

preghenella:
@sawenzel 
I think that for automated builds that start when new tag is created use the actual tag and not master.
I would keep this one to `master`

PR #2752, opened 2021-01-11
boost: bump to 1.75.0

ktf:


ktf:
Let's see if we manage to get rid of the deprecation warnings.

davidrohr:
Don't we have to fix the usage of deprecated stuff on our side for that? :)

ktf:
I think the deprecation is completely inside boost.

davidrohr:
@TimoWilken : all builds here have failed without a build log. Could you check?
O2 compilation with boost 1.75 should work in the meantime.

davidrohr:
Ah, I believe the problem is that we are using our own alisw/boost fork, and that doesn't have yet the 1.75 tag. @ktf Could you import the tag?

pzhristov:
May be we should use https://github.com/boostorg/boost ?

pzhristov:
I imported boost 1.75.0, see https://github.com/alisw/boost/pull/5

pzhristov:
Now we have the tag v1.75.0, how can we restart the test?

davidrohr:
It should actually restart automatically. But not sure if it works. Perhaps the build area is in a bad shape after the checkout failed due to the missing tag. @TimoWilken : Could you wipe the build area for this PR and restart the CI builders for it?

TimoWilken:
It seems that the CI builders didn't fetch the new tag automatically. I've done that manually and restarted the tests.

davidrohr:
@TimoWilken (ping @ktf): It seems the full log for the full CI now includes the log file of all packages that were built.
It is not so easy to see, which package actually failed the build.
Could we indicate already in the HTML log which was the actual package that failed?
And in the full log, instead of ordering the log files alphabetically, would it be possible to orther them in build order such that the failing log comes at the end? Or we should at least try to somehow point out which is the log of the failing package.

davidrohr:
@sawenzel : `o2checkcode` fails in this PR with
```
## Begin sw/BUILD/o2checkcode-latest/log
++ cp /mnt/mesos/sandbox/sandbox/o2-fullci-alidist/sw/slc8_x86-64/O2/2752-1/compile_commands.json .
+++ python3 -c 'import json,sys,os; sys.stdout.write( os.path.commonprefix([ x["file"] for x in json.loads(open("compile_commands.json").read()) if not "G__" in x["file"] and x["file"].endswith(".cxx") ]) )'
++ O2_SRC=/mnt/mesos/sandbox/sandbox/o2-fullci-alidist/sw/SOURCES/O2/2752/0/
++ [[ -e /mnt/mesos/sandbox/sandbox/o2-fullci-alidist/sw/SOURCES/O2/2752/0//CMakeLists.txt ]]
++ [[ -d /mnt/mesos/sandbox/sandbox/o2-fullci-alidist/sw/SOURCES/O2/2752/0//.git ]]
++ [[ -n e4f220566f363a604a455d153c5823c916eb7a3b ]]
++ pushd /mnt/mesos/sandbox/sandbox/o2-fullci-alidist/sw/SOURCES/O2/2752/0/
/mnt/mesos/sandbox/sandbox/o2-fullci-alidist/sw/SOURCES/O2/2752/0 /mnt/mesos/sandbox/sandbox/o2-fullci-alidist/sw/BUILD/d53ed24aca75646c68f8a47387b00160602cd5fc/o2checkcode
++ git diff --name-only e4f220566f363a604a455d153c5823c916eb7a3b...17c503ba8fd8882ac38d06b098dfc71f8fadcd54
++ sort -u
++ grep -E '\.cxx$|\.h$'
fatal: Invalid symmetric difference expression e4f220566f363a604a455d153c5823c916eb7a3b...17c503ba8fd8882ac38d06b098dfc71f8fadcd54
++ true
## End sw/BUILD/o2checkcode-latest/log
```
Any idea what happened?

davidrohr:
All required CIs have passed, full CI failed for unrelated reason, merging.

ktf:
@davidrohr I think it was always the case that we included all the available logs.

ktf:
In this case the PR was tested against a full rebuild, so we happen to have more logs available.

TimoWilken:
@ktf before https://github.com/alisw/ali-bot/pull/968, logs from non-dev packages were deleted right after the build finished (default alibuild behaviour), so they didn't get uploaded. Now they are (this was necessary because e.g. o2checkcode logs weren't recognised as dev package logs before), so we usually get lots more logs that we can upload.

TimoWilken:
@davidrohr this isn't entirely trivial to do, I think, so I'll open an issue for that behaviour.

davidrohr:
Thx, then we can discuss on the ticket, or tomorrow in the WP3 meeting.
I agree that not deleting the logs might be the right solution, but then we need some filter what we show, because currently the length of the log is just overwhelming :(

PR #2757, opened 2021-01-12
Use VecGeom tag instead of commit hash

awegrzyn:


sawenzel:
Why is this needed? I would like to be able to use VecGeom independently on the release schedule. The version you mention will not compile.

sawenzel:
To be precise, this will prevent compilation of https://github.com/AliceO2Group/AliceO2/pull/5137 (which compiled fine in the past).

awegrzyn:
Ok I see, we need this for the RPM policy, we want to enforce major version and allow higher minor and patch versions.

awegrzyn:
The other options would be to disable `VecGeom` in `o2-dataflow` defaults

ktf:
So what do we do here?

awegrzyn:
@sawenzel
1. There was `v1.1.9` released yesterday, is that good enough?
2. It is added to `O2` dependencies but not loaded by modulefile and currently `O2` compiles fine without that, so I'd disable it in the `o2-dataflow` defaults and enable back once it is actually needed, is that ok?

sawenzel:
@awegrzyn : I am not in favour of disabling in dataflow. There is an upcoming PR which will use VecGeom in O2 and this will immediately break dataflow CI. Whose responsibility will it be to update the dataflow recipe then? 

I am fine to put the latest tag of VecGeom here (it should be ok). But for me the problem goes deeper and I think we need a formal policy: alibuild allows to use commit hashes and I think we do that in various places - not just here.
So, has there been a formal decision to move **all** packages to a certain tag format? Or is this a requirement which 
is indirectly introduced by FLP? Could you remind me if this was formally discussed in WP3? @ktf @ihrivnac @Barthelemy ?
I am asking this to clarify the situation for future cases like this.

awegrzyn:
Dear @sawenzel
Thanks for you reply.
This was not discussed during WP3, and no one is forcing anything. I'm just asking nicely to use `tag` instead of commit so generated RPMs can have meaningful version. I don't think there's need of discussing this topic and adding new policies (but if you think otherwise please bring it in) and of course you can keep using commit if this is needed.

ktf:
@awegrzyn @sawenzel a solution which should still allow you to use the commit hash and not break (at least formally) the updateable RPMs policy would be to use the commit for `tag:` and a generic mnemonic version like `v1` in the version field.

This is of course is true only formally, as I said a number of times. There is zero guarantees that a given build of a package is binary compatible with the "update" of one of its dependencies, especially if we start using hashes.

pzhristov:
@sawenzel @ktf We can discuss and formally decide that only tags are used in the build system, but this will force us to clone external packages where this practice is probably OK. Is there any problem to tag the packages we are responsible for?

sawenzel:
I like the idea of using the version field to meaningfully annotate tags and I discussed this also with @awegrzyn in private.
That should allow the best of 2 worlds.

davidrohr:
For reference, +1 to the approach with the version field to annotate tags

ktf:
As @pzhristov said, we should probably discuss this in a dedicated meeting. One thing which occurred to me is that we could do the following:

* Use the `version:` field to indicate the name of the O2Suite / FLPSuite release, e.g. v21.03. This will be the same for all packages.
* Use the `tag:` field to indicate the actual tag / commit to pickup.
* aliBuild will keep track of the rebuilds for us, using the `revision` part of the package name. The actual tag used is part of the package contents themselves, something like the `.build-hash` file which we already have.
* Non updateable RPMs stay as they are now. They just get installed in a hierarchy which looks like:

```bash
/sw/<architecture>/O2/v21.03-1/bin   # First iteration depending on ROOT v6.22.00
/sw/<architecture>/O2/v21.03-1/lib
/sw/<architecture>/O2/v21.03-1/...
/sw/<architecture>/O2/v21.03-2/bin  # Second iteration depending on ROOT v6.24.00
/sw/<architecture>/O2/v21.05-1/...     # Based on ROOT v6.24.00 
/sw/<architecture>/ROOT/v21.03-1/...   # First iteration, `tag: v6.22.00`
/sw/<architecture>/ROOT/v21.03-2/...   # Second iteration, `tag: v6.24.00`
/sw/<architecture>/ROOT/v21.05-1/...   # Another rebuild of `tag: v6.24.00`, for the new O2 tag.
```

* Updateable RPMs could drop the package name and the revision part from the path, keeping the revision as RPM version in order to be able to tell what RPM is the "latest". The result from the above would be:

```bash
/sw/<architecture>/v21.03/bin
/sw/<architecture>/v21.03/lib
/sw/<architecture>/v21.03/...
/sw/<architecture>/v21.05/bin
/sw/<architecture>/v21.05/lib
```

Notice that:

* This will require repackaging (not necessarily rebuilding though) everything for each change from vXX.YY to vXX.YY+1. CVMFS take care of that deduplication, aliBuild should not rebuild but simply unpack and repack a prebuilt tarball. Of course some things will be duplicated, like space of users laptops, but again, they can simply delete the old "streams" they do not care anymore.
* Deprecation becomes trivial, just kill the weeks we do not care about.
* Work on a "stream" (e.g. v21.05) begins immediately after the previous one has been "released", rather than being the conclusion of a two week period. A "tag" is simply a point in time in which we declare a given stream as "usable for production" and we could mark it by tagging alidist, O2 and maybe by having a standard release of some top-level package (like O2Suite).
* All of the above (apart from the updateable RPMS) is available already now as a feature of alibuild, it could even be implemented in parallel as a demonstrator via a default.

ktf:
Adding also @jgrosseo and @costing

ktf:
To be clear, here is what the above would require:

https://github.com/alisw/alidist/pull/2797

(of course if we agree, there is no need for the specification in the default itself...)

ktf:
@jgrosseo this also basically allows us to have "continuous releases" for the trains.

sawenzel:
@ktf: Thanks for your thoughts. Wouldn't changing the "version" field of each recipe trigger unnecessary builds all the time, even if the actual tag did not change. In principle the global alidist version "21.03" can come from the alidist release tag without need of further propagation?? And who would be responsible to update the version fields?

ktf:
Changing the version field will trigger a repackaging, IIRC, not an actual rebuild. The old tarball will be unpacked and relocated to the new installation path. So in principle it should be much faster. We could also set up things so that users are exposed to higher granularity tags (to an extreme `v21`) while for production / data taking we have finer grained ones.

To reduce the changes in the recipes we could also have a `%(default_version)s` macro (just like we have `%(date)s` or `%(tag_basename)s`) which picks up the common version from a known file (e.g. the defaults).

davidrohr:
@ktf : I am a bit wondering if we are not overdoing it a bit here. I totally agree on putting a version there when we use a commit hash as tag, as originally suggested. But which problem are we trying to solve by aligning all versions to the next O2 suite release?

I think this might have several caveats, you mention some already:
- Local installation size will increase significantly.
- File catalogue size and RPM repository size will increase a lot.
- Wont't this trigger a lot of rebuilds every time the version is updated? Not everyone has a fast laptop. I mean, won't the recipe hashes change when we change the version, so all subsequent package would rebuild? And what about development packages? They would probably rebuild, too?

Also, it feels to me a bit like abusing the version in the build system for something it was not meant to. Basically, if we override and align the version to a fixed one, it becomes kind of obsolete. Then it should be part of the default, but not of the package.

On the other hand, I don't actually see why we have to couple the recipes to the version of the release.
- We could as well just leave the packages as they are, and we build the RPMs we don't use the version of the package but we just call the RPM  `21.03` etc. irrespective of the recipe.
- Or we can add a simple flag to aliBuild for releasing RPMs, which would ignore the version tag in the recipes and build as the release version.
In this way the normal development would be totally unaffected.

ktf:
I took the liberty to rename your bullets for better replying.

1. Not necessarily if we have that by default people have `v21` as the version (or something similar, a-la ubuntu), while `v21.XX` is kept for the deployed versions.
2. True, however it's much simpler to deprecate and partition them. For example, we could have one repository per "stream". 
3. No, as I was saying to @sawenzel the version is not part of the package hash, only the tag, IIRC. What happens is that if `v21.03` has the same hash of `v21.05`, the binary tarball of the former gets repackaged in the latter.
4. Yes, this is possible. The other alternative is what I was proposing in 1.
5. This is possible already, the flag is just `--defaults o2-single-version` (and the appropriate default on the line of my other PR).

davidrohr:
For 3.: OK, I just tried what happens when I change the version field. It triggers a rebuild, both for development packages and for normal packages, and also for all dependent packages. Don't know if this is a bug in the current alibuild?

For 5.: I don't fully understand. That uses the normal `--defaults` setting, so we could not use the `--defaults o2` in parallel, since only 1 default is possible at a time. Of course we could copy it...

My personal preference would thus be:
- Just use 4.
- Or use a slight modification of 5.: When we create the alidist tag for an o2 version, we don't tag directly, but we add one more commit where we place the correct version in the default, overriding all versions. This commit would not go to dev, but only in the versioned tag.

ktf:
Actually 4. would not work. There are "degenerate" cases for when the revision is the same but not the version. E.g. `ROOT-v6.22.0-1` and `ROOT-v6.22.1.-1` would both produce an RPM which is called `ROOT-v21.03-1` with 4., while what I am proposing will automatically handle it to `ROOT-v21.03-1` and `ROOT-v21.03-2`.

ktf:
Reworded your bullet to have `6.`. Indeed 6 is basically what I am doing with `#2797` with two changes:

* Have a single `stream:` entry available in the default.
* Have a `%(stream)s` expansion variable which expands to either the `stream` found in the default or to the provided tag.

davidrohr:
For 4.: I see your point, but I think this can be fixed easily by appending the full version to the O2 release version, or just appending a patchlevel.

For 6.: What I do not like in #2797 is:
- It modifies our actual defaults. My suggestion was not to change them in alidist/master at all, but just when we tag in an extra commit.
- Also, I don't really like overriding the version for all packages individually. Isn't there away to override all version settings for all packages at once? But is the PR up to date? I don't see `$(stream)` in it.

ktf:
I am not sure I understand your point about 4. If you append the full version, then RPM will still be confused on how to upgrade. There might be cases in which ROOT-v7.00.00 is actually to be preferred to ROOT-v7.00.01 because we discovered a bug and we need to rollback.

Regarding 6.1, agreed, the change where we have `stream: v21.03` would need to go in the O2-v21.03-patches in a separate commit.

Regarding 6.2, that's how it can be done right now. If there is consensus it should be trivial to introduce a `stream` tag for the defaults which, if present, affects all the versions. It's not available yet.

ktf:
The drawback of 5 and 6 is of course that users will have a completely different set of versions compared to the releases.

davidrohr:
4. My idea was that when we release 21.03 with root 6.22.06 and need to patch it with 6.22.08, then we call the first RPM ROOT-21.03.00 and the second one just ROOT-21.03.01 (irrespective of the version in the recipe). Alternatively, we could call them  ROOT-21.03.00-6.22.06 and ROOT-21.03.01-6.22.08.

For 6., ok, with an additional stream tag, where we don't have to touch the version tags of tens of packages, it makes more sense.

In general, I see the difference between the release and the development version, but if the user wants a release, he can take the RPMs, or he can check out alidist-21.03.00 and build that, and that he'll have the release version.
For development with the upstream version, you are anyway not on a release, are you?

PR #2758, opened 2021-01-12
[O2-1954] On CC7 c-ares library is installed in lib64

awegrzyn:


teo (approved):

ktf:
I think it would be better to use `-DCMAKE_INSTALL_LIBDIR="lib"` to avoid polluting LD_LIBRARY_PATH.

ktf:
Tested privately. Merging.

Issue #1420, opened 2018-12-04
Inconsistent local OpenSSL build


matthiasrichter:
After upgrading to Ubuntu 18.04, alibuild dos not pick OpenSSL from the system anymore. I have two issues with the local build. Using defaults o2.

When compiling O2 I get lots of warnings like this one
```
/usr/local/bin/cmake: /home/richter/src/alisw/sw/ubuntu1804_x86-64/OpenSSL/v1.0.2o-1/lib/libssl.so.1.0.0: no version information available (required by /usr/local/bin/cmake)
```

I can not use ssh with the local version
```
/usr/bin/ssh: /home/richter/src/alisw/sw/ubuntu1804_x86-64/OpenSSL/v1.0.2o-1/lib/libcrypto.so.1.0.0: no version information available (required by /usr/bin/ssh)
/usr/bin/ssh: /home/richter/src/alisw/sw/ubuntu1804_x86-64/OpenSSL/v1.0.2o-1/lib/libcrypto.so.1.0.0: no version information available (required by /usr/bin/ssh)
/usr/bin/ssh: relocation error: /usr/bin/ssh: symbol EC_KEY_get0_public_key version OPENSSL_1.0.0 not defined in file libcrypto.so.1.0.0 with link time reference
```

ktf:
I think we need to move to openssl 1.1.0 to fix that, or alternatively use a different output name for the openssl libraries so that they do not get seen by the system. Could you try to see if the version bump is something that can be easily done?

davidrohr:
@ktf @costing : Today I say that I actually have a lot of similar issues with OpenSSL version conflicts during the O2 build, because system libraries are built against the system OpenSSL 1.1.1, while the alibuilt stuff is still at 1.0.x. In particular, 1.0.x and 1.1.0x are unsupported since end of 2019, so I think we have to bump to 1.1.1.
Do we know what prevents us? I remember there was a problem related to alien/grid at some point. Is this fixed with JAlien now, or is there a blocker? And if we cannot bump at the moment, what is our strategy?

costing:
See please @Atlantic777 's comment https://github.com/alisw/alidist/pull/2698#discussion_r545715175 on what the status is. If I understand correctly what pulls this dependency back is compatibility with legacy AliEn + ROOT 5 + Xrootd 3.x .

ktf:
I think we fixed this at some point. Reopen if not the case anymore.

PR #2760, opened 2021-01-12
Bump OpenSSL

davidrohr:
Just to see how badly this breaks (#1420)


davidrohr:
Closing this. We should test this in the containers manually first, or really follow it up. But in my opinion it makes sense to bump this at one point!

PR #2765, opened 2021-01-14
Modify Geant4 recipe to make possible to define an external directory for data sets, plus clean-up: 

ihrivnac:
- This PR restores back ca2863573252a2cefd93b52843e13c11932a1774
- Fix setting Geant4 data sets in build environment (failing in the previous version):
   - Removed setting data sets in env: section in geant4.sh
   - Added setting data sets environment via geant4-config in recipes with Geant4 tests
     (aliroot-guntest.sh, o2.sh)

### DIFF
--- a/geant4.sh
+++ b/geant4.sh
@@ -14,64 +14,55 @@ incremental_recipe: |
   mkdir -p $INSTALLROOT/etc/modulefiles && rsync -a --delete etc/modulefiles/ $INSTALLROOT/etc/modulefiles
 env:
   G4INSTALL : $GEANT4_ROOT
-  G4DATASEARCHOPT : "-mindepth 2 -maxdepth 4 -type d -wholename"
-  G4ABLADATA : "`find ${G4INSTALL} $G4DATASEARCHOPT '*data*G4ABLA*'`"  ## v10.4.px only
-  G4ENSDFSTATEDATA : "`find ${G4INSTALL} $G4DATASEARCHOPT '*data*G4ENSDFSTATE*'`"
-  G4INCLDATA : "`find ${G4INSTALL} $G4DATASEARCHOPT '*data*G4INCL*'`"  ## v10.5.px only
-  G4LEDATA : "`find ${G4INSTALL} $G4DATASEARCHOPT '*data*G4EMLOW*'`"
-  G4LEVELGAMMADATA : "`find ${G4INSTALL} $G4DATASEARCHOPT '*data*PhotonEvaporation*'`"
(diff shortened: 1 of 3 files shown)


ihrivnac:
@ktf, @sawenzel , can you approve ? Thank you.

geant4.sh
```diff
@@ -88,28 +67,11 @@ module-whatis "ALICE Modulefile for $PKGNAME $PKGVERSION-@@PKGREVISION@$PKGHASH@
...
-setenv G4NEUTRONXSDATA ${G4NEUTRONXSDATA:-not-defined}
-setenv G4PARTICLEXSDATA ${G4PARTICLEXSDATA:-not-defined}
-setenv G4PIIDATA ${G4PIIDATA:-not-defined}
-setenv G4RADIOACTIVEDATA  ${G4RADIOACTIVEDATA:-not-defined}
-setenv G4REALSURFACEDATA ${G4REALSURFACEDATA:-not-defined}
-setenv G4SAIDXSDATA ${G4SAIDXSDATA:-not-defined}
-set G4BASE \$GEANT4_ROOT
 prepend-path PATH \$GEANT4_ROOT/bin
-prepend-path ROOT_INCLUDE_PATH \$GEANT4_ROOT/include/Geant4
-prepend-path ROOT_INCLUDE_PATH \$GEANT4_ROOT/include
```
sawenzel, line 113: just to understand: Which module file is now setting these variables (or who is setting these things when we load Geant4 on the GRID?)

sawenzel: Ah I see. You modify the modulefile in a post-step.

sawenzel (approved):

geant4.sh
ihrivnac: The data paths are added in the next step.
Geant4 includes need not to be added in the ROOT_INCLUDE_PATH as we do not generate dictionaries from Geant4 classes.

ihrivnac:
@ktf , can you merge ? Thank you.

sawenzel:
@ktf: Shouldn't simulation experts be able to merge this as per CODEOWNERS? The button is not activated.

ktf (approved):

ktf:
@sawenzel the issue was that since o2.sh and aliroot-gun.sh were modified, it required a review from `alisw@infrastructure-admins`. I added `alisw/pdp-experts` for all the "mostly PDP" things, so next time your review should be enough. Of course I could have added you to alisw@infrastructure-admin (or you can add yourself, since you should be a global admin), however I think it's better if we keep infrastructure-admin separate, so that we have some better chance to notice new recipes / recipes without a proper owner.

PR #2771, opened 2021-01-20
[O2-1976] Fix unloading Python module

awegrzyn:


python.sh
```diff
@@ -115,5 +115,10 @@ alibuild-generate-module --bin --lib > "$MODULEFILE"
 cat >> "$MODULEFILE" <<EoF
 setenv PYTHONHOME \$PKG_ROOT
 prepend-path PYTHONPATH \$PKG_ROOT/lib/python/site-packages
-setenv SSL_CERT_FILE  [exec \$PKG_ROOT/bin/python3 -c "import certifi; print(certifi.where())"] 
+if { [module-info mode load] } {
```
ktf, line 118: Where does "module-info" come from? It does not seem to exists on lxplus.

awegrzyn: https://modules.readthedocs.io/en/latest/modulefile.html#mfcmd-module-info

I tested it on CC7 with modules `3.2.10` and `4.6.1`...

PR #2780, opened 2021-01-22
o2: move to v21.03

ktf:


ktf:
@Barthelemy this probably explains why you were not able to create the QC rpms.

Barthelemy:
could we merge this in order to have the RPMs on Monday ?

PR #2783, opened 2021-01-25
Keep only top packages in O2Suite

awegrzyn:
@ktf @Barthelemy does it make sense?

This also removes `InfoLogger` but keeps `libInfoLogger`. @sy-c is that okay with you?

### DIFF
--- a/o2suite.sh
+++ b/o2suite.sh
@@ -2,15 +2,10 @@ package: O2Suite
 version: "1.0.0"
 tag: "O2Suite-1.0.0"
 requires:
-  - Common-O2
   - coconut
   - Control-Core
   - Control-OCCPlugin
-  - Monitoring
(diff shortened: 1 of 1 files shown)


sy-c:
@awegrzyn I'm not sure to understand the rationale of this. O2 suite should still provide the log-related executables from infoLogger. This will not the be the case if InfoLogger is removed from the list.

awegrzyn:
1. InfoLogger is installed under `/opt/o2-InfoLogger` so it does not come from aliBuild
2. We compile `InfoLogger` in aliBuild infrastructure and generate RPM, but never use it, so why to waste time to build it?

sy-c:
all right, I though it was installed by default and might be used to inject logs from the command line. that's ok for me to remove it.

PR #2788, opened 2021-01-27
Remove FreeType and libpng deps from python modules

awegrzyn:


awegrzyn:
@TimoWilken could you lemme know what happens with `build/O2/alidist`? There are no logs.

ktf:
I am not 100% this will work correctly. Some python modules (IIRC, matplotlib, which is potentially used by users, but not in the relval / ci tests) do require libpng and freetype and this would create troubles, especially on mac / ubuntu which are known to have a different version of libpng.

awegrzyn:
Not an expert on that but it looks to me that these paths are passed to `python` already: https://github.com/alisw/alidist/blob/011ee0d1857cd55a6ca254fc63a2d45fe0c3b398/python.sh#L31
In addition `python` loads them in modulefile which is not the case for `python-modules`.

TimoWilken:
@awegrzyn @ktf build/O2/alidist failed because it tries to source the wrong thing. This is a bug in aliBuild.
```
2021-01-28@04:37:45:DEBUG:O2Suite:O2:build_O2_alidist: ++++ source /mnt/mesos/sandbox/sandbox/o2-alidist/sw/slc7_x86-64/GEANT4/v10.5.1-29/etc/profile.d/init.sh
2021-01-28@04:37:45:DEBUG:O2Suite:O2:build_O2_alidist: /mnt/mesos/sandbox/sandbox/o2-alidist/sw/slc7_x86-64/GEANT4_VMC/v5-0-p5-75/etc/profile.d/init.sh: line 3: /mnt/mesos/sandbox/sandbox/o2-alidist/sw/slc7_x86-64/GEANT4/v10.5.1-29/etc/profile.d/init.sh: No such file or directory
2021-01-28@04:37:45:ERROR:O2Suite:O2:build_O2_alidist: Error while executing /mnt/mesos/sandbox/sandbox/o2-alidist/sw/SPECS/slc7_x86-64/O2/build_O2_alidist-1/build.sh on `alibuild05.cern.ch'.
```

A previous build of the same test succeeded, so I've marked it as green again.

PR #2792, opened 2021-01-28
Disable yaml-cpp in o2-dataflow defaults

awegrzyn:
`yaml-cpp` is optional dependency to `FairRoot` that is required to build `MCConfigurator`.
As `MCConfigurator` requires also `Geant3` and `Geant4` it will never be built with `o2-dataflow` defaults.

### DIFF
--- a/defaults-o2-dataflow.sh
+++ b/defaults-o2-dataflow.sh
@@ -27,6 +27,7 @@ disable:
   - JAliEn-ROOT
   - KFParticle
   - MCStepLogger
+  - yaml-cpp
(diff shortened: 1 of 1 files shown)


kostorr (approved):

ktf:
@sawenzel do we need the MCConfigurator at all?

ktf:
Looks like no?

```
→ git grep MCConfigurator | wc -l
       0
```

sawenzel:
I don't know what MCConfigurator is.

PR #2797, opened 2021-01-28
Implement the proposal for the "stream" based releases

ktf:
Proposal description at:

https://github.com/alisw/alidist/pull/2757

### DIFF
--- a/defaults-o2-dataflow.sh
+++ b/defaults-o2-dataflow.sh
@@ -1,5 +1,5 @@
 package: defaults-o2-dataflow
-version: v1
+version: v21.05
 env:
   CXXFLAGS: "-fPIC -O2 -std=c++17"
   CFLAGS: "-fPIC -O2"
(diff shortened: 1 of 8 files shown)


ktf:
A better way of implementing this would be to use the soon (aliBuild 1.7.2) to be introduce `%(branch_stream)s` expansion. Closing this for now.

davidrohr:
@ktf : Just for my understanding: how will branch_stream work? Will you have to set version to `%(branch_stream)s`? Or will the presence of a branch_stream tag != "" in the default already fix all versions to the same stream? I'd prefer something along the latter, since if we have to set all the versions, there is a chance we miss one package, e.g. when we add new packages.

ktf:
http://alisw.github.io/alibuild/reference.html describes the behavior. It gets activated only when building from the `-patches` branch in alidist and otherwise falls back to what specified in `tag:`.

davidrohr:
ok, thx, I see. But then still, if someone would add a package with some explicit version / change a version, RPMs would be created with the incorrect version? Wouldn't it be simpler if we could just hard-override all version fields for all packages in the defaults? Then we'd need to change one line in the default, and all versions would be aligned, without room for mistakes?

ktf:
We could simply have "%(branch_stream)s" always the default for `version`, and drop `version` everywhere. That way people will have to actively add it if they do not want this behavior, which might be useful for things like built time only packages.

davidrohr:
But then I don't understand what is the advantage? Doesn't that mean the version field becomes kind of obsolete, but if someone uses it accidentally, it will yield incorrect RPM versions. I mean, it is up to you, but in my opinion just overriding the version is still simpler and safer :)

ktf:
The advantage would be that one does not need the extra commit. Anyways we can still discuss this, what is not clear to me is wether there is interest in the proposal or not.

PR #2786, opened 2021-01-26
Only build ReadoutCard on CentOS

kostorr:

PR #2798, opened 2021-01-29
Remove cub (part of CUDA since v11)

davidrohr:


davidrohr:
errors unrelated, merging

davidrohr:
@Barthelemy : I have seen the `testCheckWorkflow` of QC failling in many PRs testing O2Suite. Are you aware of that?

Barthelemy:
@davidrohr I have seen it indeed. I had asked @ktf and @knopers8 if they had an idea of what could be wrong but forgot to come back to it. We have not touched them. We also mentioned it here: https://github.com/alisw/alidist/pull/2786

@ktf @knopers8 any idea ? in the worst case we can disable them but I would really like to understand what is going on. They just started timing out on Mac.

PR #2806, opened 2021-02-02
Revert "Disable yaml-cpp in o2-dataflow defaults"

awegrzyn:
Reverts alisw/alidist#2792

Apparently there are some issue with FairRoot on the build machines:

```
Make Error in DataFormats/common/CMakeLists.txt:
  Imported target "FairRoot::Base" includes non-existent path
...
```

### DIFF
--- a/defaults-o2-dataflow.sh
+++ b/defaults-o2-dataflow.sh
@@ -27,7 +27,6 @@ disable:
   - JAliEn-ROOT
   - KFParticle
   - MCStepLogger
-  - yaml-cpp
 overrides:
   Python-modules-list:
     env:


awegrzyn:
@Barthelemy @ktf

ktf:
Fasttracking since someone else is having the same issue. Indeed I suspect it depends on the packages installed on the system. @dennisklein could it be that fairroot is confused when yaml-cpp is installed on the system (but maybe not entirely?)

dennisklein:
The cmake configure output of FairRoot shows which yaml-cpp is found and used. It follows (should) the standard search procedure as documented here: https://cmake.org/cmake/help/latest/command/find_package.html#search-procedure. I would help more, but there is little context here for me to be more useful.

ktf:
Mmm.. I have someone else who ended up with:

```
2021-02-02@15:57:57:DEBUG:O2:FairRoot:0: --   yaml-cpp             0.6.2                     /
2021-02-02@15:57:57:DEBUG:O2:FairRoot:0: --   
2021-02-02@15:57:57:DEBUG:O2:FairRoot:0: --   COMPONENT           BUILT?  INFO
2021-02-02@15:57:57:DEBUG:O2:FairRoot:0: --   Examples              NO    (enable with -DBBUILD_EXAMPLES=ON)
2021-02-02@15:57:57:DEBUG:O2:FairRoot:0: --   Doxygen              NO     (enable with -DBUILD_DOXYGEN=ON)
2021-02-02@15:57:57:DEBUG:O2:FairRoot:0: --   Unit tests           NO     (default, enable with -DBUILD_UNITTESTS=ON)
2021-02-02@15:57:57:DEBUG:O2:FairRoot:0: --   
2021-02-02@15:57:57:DEBUG:O2:FairRoot:0: -- Configuring done
2021-02-02@15:57:57:DEBUG:O2:FairRoot:0: CMake Error in fairtools/MCConfigurator/CMakeLists.txt:
2021-02-02@15:57:57:DEBUG:O2:FairRoot:0:   Imported target "yaml-cpp" includes non-existent path
2021-02-02@15:57:57:DEBUG:O2:FairRoot:0: 
2021-02-02@15:57:57:DEBUG:O2:FairRoot:0:     "/include"
```

notice the `yaml-cpp             0.6.2                     /`. How can that actually happen?

shahor02:
I had the same problem when installing O2 on ubuntu 20.04. The system yaml-cpp was detected leading to this error. I've fixed this by removing it from the system, then it was installed by the aliBuild and the problem was gone. 
I see that others had similar problem and solved it by telling to yaml-cpp to not prefer the system instance: https://alice-talk.web.cern.ch/t/o2-build-fails-in-fairrot-v18-4-1-on-ubuntu-20-04/799

ktf:
@dennisklein could you make the detection of yaml-cpp enabled only if MCConfigurator is requested via some BUILD_MCCONFIGURATOR flag?

dennisklein:
See https://github.com/FairRootGroup/FairRoot/commit/a57dea74a89bd6c8ce0b47f543e3bca4497dae7d and the linked debian/ubuntu bug for details.

However, what I do not yet understand, is, why libBase is linked to yaml-cpp at all. The MCConfigurator module is only built, when yaml-cpp, geant3 and geant4 are found (which introduces llibBase's dependency on yaml-cpp finally). Wasn't the argument of #2792 that one of the three deps are missing?

dennisklein:
> @dennisklein could you make the detection of yaml-cpp enabled only if MCConfigurator is requested via some BUILD_MCCONFIGURATOR flag?

Yes, this is in the pipeline since a while. Will come, not sure how fast.

ktf:
Ok, thank you. I think the problem is that FindYaml.cmake is somehow broken and it gets confused by (completely unrelated) system installations. The problem for us arised when we removed our own build of yaml-cpp and the system version started to be picked up.

PR #2810, opened 2021-02-05
test to bump fastjet to 3.3.4 and fjcontrib to 1.045

bhohlweger:
Added a new defaults file for fastjet to get a dedicated build to check for any differences in analysis with the upgraded version.


ktf (approved):
This is fine with me and better than an additional default.

PR #2811, opened 2021-02-05
use alibuild-generate-module

Barthelemy:


awegrzyn:
I think you need to add ` alibuild-recipe-tools` to `build_requires`

Barthelemy:
@awegrzyn thank you

abseil.sh
```diff
@@ -20,20 +21,13 @@ cmake $SOURCEDIR                        \
...
-set version $PKGVERSION-@@PKGREVISION@$PKGHASH@@
-module-whatis "ALICE Modulefile for $PKGNAME $PKGVERSION-@@PKGREVISION@$PKGHASH@@"
-# Dependencies
-module load BASE/1.0                                                          \\
-            ${GCC_TOOLCHAIN_REVISION:+GCC-Toolchain/$GCC_TOOLCHAIN_VERSION-$GCC_TOOLCHAIN_REVISION}
+alibuild-generate-module > "$MODULEFILE"
+cat >> "$MODULEFILE" <<EoF
+
 # Our environment
 set ABSEIL_ROOT \$::env(BASEDIR)/$PKGNAME/\$version
```
ktf, line 32: Note that we do not really need those, because abseil does not have any bin folder...

mpfr.sh
```diff
@@ -26,16 +27,9 @@ rm -f $INSTALLROOT/lib/*.la
...
-}
-set version $PKGVERSION-@@PKGREVISION@$PKGHASH@@
-module-whatis "ALICE Modulefile for $PKGNAME $PKGVERSION-@@PKGREVISION@$PKGHASH@@"
-# Dependencies
-module load BASE/1.0
+alibuild-generate-module > "$MODULEFILE"
+cat >> "$MODULEFILE" <<EoF
+
 # Our environment
 set MPFR_ROOT \$::env(BASEDIR)/$PKGNAME/\$version
```
ktf, line 34: Same here...

ktf:
I think this is fine for now. In retrospect we do not need the modulefile at all, since IIRC I removed the bit from the script which was mandating the need for one in order to be published...

Barthelemy:
ok, as we need a new release of these packages, I propose to keep it as it is as it does not hurt.

PR #2812, opened 2021-02-05
Run Full System Test with TF of 5 Pb-Pb events as part of Full CI

davidrohr:
@sawenzel @ktf @TimoWilken : This PR adds a short full system test run to the Full CI. (1 time frame of 5 Pb-Pb events, only CPU, simulattion + raw creation + sync + async reco.)
It will run last, so the other errors would be reported first.
On my laptop it prolongs the full CI by 6.5 minutes, and takes ~10 GB of memory in total, that should be OK for the CI nodes?

I don't see how we can test this before deploying it, since there is no full CI for alidist.
I would thus just deploy it on the weekend, see how it goes, and quickly fix it, if something goes wrong.
Obviously I have tested it locally, where it worked, but the setup of the CI machines could be different than mine.

### DIFF
--- a/o2-full-system-test.sh
+++ b/o2-full-system-test.sh
@@ -1,0 +1,28 @@
+package: O2-full-system-test
+version: "1.0"
+requires:
+  - O2Suite
+force_rebuild: 1
+---
+#!/bin/bash -e
+
+rm -Rf $BUILDDIR/full-system-test-sim
+mkdir $BUILDDIR/full-system-test-sim
+pushd $BUILDDIR/full-system-test-sim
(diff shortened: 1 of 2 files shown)


TimoWilken:
The CI machines have 16 cores and 24 GiB of RAM, so that should be fine.

I'll be a bit busy tomorrow morning, but I should be available in the afternoon in case something goes wrong.

sawenzel (approved):
I think this becomes a necessity. But how would we detect hanging situations just like now?
Maybe we could set a timeout to the taskwrapper?

davidrohr:
I have added a timeout (tested locally) and merged it. Some PRs were running the CI already, but it seems there are random framework ctest failures breaking the CI, so it doesn't even arrive at the full system test.

PR #2821, opened 2021-02-09
Introduction of a meta package for developing on an FLP (O2-1974)

Barthelemy:


flpsuitedevenv.sh
```diff
@@ -0,0 +1,32 @@
...
+tag: "FLPSuiteDevEnv-1.0.0"
+requires:
+  - CMake
+  - O2-customization
+  - alibuild-recipe-tools
+  - ms_gsl
+  - RapidJSON
+  - GMP
+  - MPFR
+  - cub
```
davidrohr, line 12: cub has been removed from alidist, I think it should be removed here

davidrohr (changes requested):

flpsuitedevenv.sh
Barthelemy: Indeed, thank you

flpsuitedevenv.sh
```diff
@@ -0,0 +1,31 @@
...
+  - alibuild-recipe-tools
+  - ms_gsl
+  - RapidJSON
+  - GMP
+  - MPFR
+  - defaults-release
+  - abseil
+  - freeglut-devel
+  - OpenSSL
+  - abseil
```
knopers8, line 16: abseil is twice

flpsuitedevenv.sh
```diff
@@ -0,0 +1,31 @@
...
+
+# Modulefile
+MODULEDIR="$INSTALLROOT/etc/modulefiles"
+MODULEFILE="$MODULEDIR/$PKGNAME"
+
+mkdir -p etc/modulefiles
+alibuild-generate-module --bin --lib > etc/modulefiles/$PKGNAME
+cat << EOF >> etc/modulefiles/$PKGNAME
+prepend-path ROOT_INCLUDE_PATH \$PKG_ROOT/include
+EOF
```
ktf, line 30: Do we really need all this? Isn't the package empty?

Barthelemy: The package is empty. Module stuff is black magic to me. I copy paste the whole shebang from another recipe. Should I remove everything related to modulefile ?

awegrzyn:
@Barthelemy Could you elaborate why this is needed? If you develop QC code you can just build QC and you'll get all dev deps.

Barthelemy:
@awegrzyn This is to be able to build O2 or QC on a machine (FLP) installed with RPMs. We could instruct people to install these packages by hand but it is easier to tell them to build this recipe and load it.

awegrzyn:
Ok I see, what about https://alice.its.cern.ch/jira/browse/O2-1896 ?

Barthelemy:
It is related. On one hand we need to make sure what the procedure is and on the other hand we have to make it easy for the users. So the procedure is outlined now and this meta package should help users to have all the needed packages to follow the procedure easily.

Barthelemy:
Mac build failed for no reason

flpsuitedevenv.sh
ktf: Ok, we should probably drop it in the long run. Merging for now.

ktf (approved):

PR #2830, opened 2021-02-11
Do not use alibuild-generate-module in coconut for the moment

awegrzyn:


Barthelemy:
does it fix a problem ?

teo:
@awegrzyn what is this for? Is it related to last night's Suite pipeline failure on loading `coconut`?

ktf:
I think #2831 might be a better solution. Longer term we should provide the correct set of build requires on a per package level (rather than globally, as now).

Barthelemy:
@ktf but shouldn't we have zero dependencies on this static package ?

ktf:
On this one, yes, but no one is telling alibuild that you will always use only this package... FULL_BUILD_REQUIRES (which alibuild-generate-module uses to determine the build requires) is the list of all the packages which are used ONLY as build requires in a given dependency chain. In this particular case c-ares and protobuf are used as runtime requires and they are dropped from the list. We can change that / add a separate variable, but I am not 100% sure we do not use the above behavior somewhere else were we do want to have those packages treated as if they were build requires (e.g. to generate the modulefile of gRPC).

Barthelemy:
After discussion, we decided to use this workaround and to aim a proper fix later on.

vascobarroso (approved):

PR #2831, opened 2021-02-11
coconut: force build requires to alibuild-generate-module

ktf:
aliBuild does not consider as a build require something which is marked as
runtime require by one of the packages in the chain. This is because it cannot
know if you will use only the toplevel package (coconut, in this case) or one
it's dependencies (nothing stops you from loading gRPC directly, once you have
built coconut). However this does not play well with the automatic generation
of modulefiles. We use the FULL_BUILD_REQUIRES to correct the behavior of
alibuild-generate-module.

### DIFF
--- a/coconut.sh
+++ b/coconut.sh
@@ -2,8 +2,10 @@ package: coconut
 version: "%(tag_basename)s"
 tag: "v0.19.80"
 build_requires:
+  - abseil
   - golang
   - protobuf
+  - c-ares
   - grpc
   - alibuild-recipe-tools
 source: https://github.com/AliceO2Group/Control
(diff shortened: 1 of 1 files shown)


Barthelemy:
Is it fixing the same thing as https://github.com/alisw/alidist/pull/2830 ?

ktf:
Yes. A proper fix would be to make aliBuild create a valid FULL_BUILD_REQUIRES for each package, rather than a toplevel one, but that is not easily doable right now.

ktf:
Or better, it might be possible with just a simple variable name substitution in aliBuild, but I am not sure I want to take the risk.

Barthelemy:
After discussion, we decided to close this one in favour of https://github.com/alisw/alidist/pull/2830

PR #2837, opened 2021-02-12
Fix synchronization of POWHEG module file after move to alibuild-gene…

mfasDa:
…rate-module


mfasDa:
Module file was generated but not synchronised after the build, therefore loading the package failed. @preghenella @maireiphc can you please check! Could this please be merged even though the CI of aligenerators is currently failing?

PR #2840, opened 2021-02-15
Update python-modules-list.sh

sawenzel:
Add psutil python module. To replace some hand-made stuff currently in O2/Utilities/Tools/jobutils.sh.


sawenzel:
@pzhristov

PR #2841, opened 2021-02-15
Bump0.9.1: fix protobuf dep name

ironMann:
@ktf this should fix module loading


ktf:
Currently wrong. Merging.

PR #2848, opened 2021-02-18
Drop curl as optional monitoring dependency

awegrzyn:


vascobarroso (approved):

awegrzyn:
@ktf `build/O2/alidist` crashes with `[ERROR] Unhandled o2::framework::runtime_error reached the top of main, device shutting down. Details follow:`, I had the same issue yesterday compiling O2, what does that mean?

ktf:
@awegrzyn you should drop the implicit curl support also from Monitoring itself. With just this, it seems to still detect the system curl and link against it causing havoc on platforms where we need our own curl.

awegrzyn:
Uhhh :| in such case let's better revert this.

PR #2849, opened 2021-02-18
Disable  defaults-release in dataflow defaults

awegrzyn:


awegrzyn:
@ktf is that fine? I don't see how it could be useful for `dataflow` defaults

ktf:
defaults-release is a special package. You should not touch it.

PR #2850, opened 2021-02-18
Set O2 options in dataflow defaults

awegrzyn:
@ktf I think we can safely pass vars from `defaults` instead of `O2-customization` as I don't see a use-case that someone wants to enable it. Are these vars gonna be picked up?

### DIFF
--- a/defaults-o2-dataflow.sh
+++ b/defaults-o2-dataflow.sh
@@ -5,10 +5,6 @@ env:
   CFLAGS: "-fPIC -O2"
   CMAKE_BUILD_TYPE: "RELWITHDEBINFO"
   CXXSTD: "17"
-  ENABLE_UPGRADES: "OFF" # Disable detector upgrades in O2
-  BUILD_ANALYSIS: "OFF" # Disable analysis in O2
-  BUILD_EXAMPLES: "OFF" # Disable examples in O2
-  BUILD_TEST_ROOT_MACROS: "OFF" # Disable macro tests in O2
 disable:
   - AEGIS
   - AliEn-Runtime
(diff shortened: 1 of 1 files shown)


vascobarroso (approved):

ktf:
Changing defaults will rebuild everything. Apart from that, there is no difference.

defaults-o2-dataflow.sh
```diff
@@ -5,6 +5,10 @@ env:
   CFLAGS: "-fPIC -O2"
   CMAKE_BUILD_TYPE: "RELWITHDEBINFO"
   CXXSTD: "17"
+  ENABLE_UPGRADES: "OFF" # Disable detector upgrades in O2
```
ktf, line 8: This one needs to be done in O2-customizations which will have precedence.

ktf:
Actually, no this will not work the moment we override them in o2-customization. What you need to do is to override o2-customization itself from the default.

ktf (changes requested):
You should override o2-customization in defaults-o2-dataflow.sh and change the variables there.

ktf:
See #2853

awegrzyn:
> Changing defaults will rebuild everything. Apart from that, there is no difference.

Well, as we're overwriting it in the defaults so I guess the full rebuilt is inevitable

awegrzyn:
@ktf This does not work as expected, do we need any changes to O2 recipe? By default CMake does not pick up env vars.

ktf:
you probably need something like `-DBUILD_ANALYSIS=$BUILD_ANALYSIS`. For upgrades it should already be there.

awegrzyn:
Well, not really: https://github.com/alisw/alidist/blob/ba9b61c24eac62c295ca5ed249ac02ba0c1d3366/o2.sh#L190
If `ENABLE_UPGRADES` variable is set, it will be always overwritten by `-DENABLE_UPGRADES=ON`.

PR #2852, opened 2021-02-18
datadist: bump 0.9.4

ironMann:
@ktf please include this in the new CC8 RPMs


ktf:
I started from your branch `ironnmann/flp-suite-v0.15.1` and it seems to be there.

PR #2854, opened 2021-02-18
Make ninja the default tool to build

ktf:


ktf:
@alisw/o2-wp3 any objections to this?

davidrohr:
Fine with me, but I cannot judge if this has any side effects / will fail for some setups.

chiarazampolli:
Hello @ktf , 
Maybe the question is naive, but will the usual compilation flow of people using "make" change? I usually compile from the BUILD directory with "make". 
Chiara

TimoWilken:
No objections from me either, same as David.

aphecetche:
No objection either (I'm only using Ninja anyway ;-)). 
We should maybe just make sure that we don't have user facing documentation that mention make explicitly then ?

@chiarazampolli yes, the workflow would change slightly : you go the build directory and would type 'ninja' instead of 'make', 
You can also go to the build directory and use cmake command that's independent of the underlying build system used (ninja, make, Xcode, etc...) , e.g `cmake --build .` to build everything or `cmake --build . --target some_target` to (re)build something specific.  To build everything it is a bit longer to type than just make or ninja, but to rebuild quickly an executable for instance it might make sense, e.g. `cmake --build . --target o2-dcs-proxy`

awegrzyn:
A CC7 user will need to either manually install `ninja` or use `alienv enter ninja` which is slightly more complex that just running `make`.

aphecetche:
@awegrzyn to be confirmed, but I believe after the (1st) cmake configure stage, the fullpath to the build system program is actually embedded in the CMakeCache, e.g. this is what I see on my (Mac) machine : 

CMAKE_MAKE_PROGRAM:FILEPATH=/Users/laurent/github.com/Homebrew/brew/bin/ninja

so `cmake --build .` from the build directory _should_ work even without `alienv enter ninja` (or module load ninja or whatever).  But indeed directly typing `ninja` won't.

ihrivnac:
It's fine with me, too.  As @aphecetche already pointed, we will then need to update documentation.

Barthelemy:
I propose to add it to the next WP3 for discussion

ktf:
@TimoWilken It looks like the ubuntu tests are failing (https://ali-ci.cern.ch/alice-build-logs/alisw/alidist/2854/4ef8da40fc3b949cf9d28f2ddbb71a8908561e76/build_O2_alidist-ubuntu1804/pretty.html). Can you have a look?

TimoWilken:
@ktf, the fix was only just merged. I've made the ubuntu1804 builder install alibuild@master for now, but maybe it's worth making a new alibuild release.

ktf:
@MichaelLettrich @TimoWilken now it's the mac to fail. Any idea of what is going wrong?

TimoWilken:
@ktf well, this is interesting...
```
DEBUG: Updating reference repository: cd /Volumes/build/build/ci_checks/alice_ci_0/alidist-o2/sw/MIRROR/clang && git fetch -f --tags https://github.com/llvm/llvm-project 2>&1 && git fetch -f https://github.com/llvm/llvm-project '+refs/heads/*:refs/heads/*' 2>&1
DEBUG: fatal: not a git repository (or any parent up to mount point /Volumes)
DEBUG: Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
ERROR: Error while updating reference repos https://github.com/llvm/llvm-project.
++ report_pr_errors --no-comments
```
Seems like something went wrong with the clone there, though I can't find any errors in the logs. I've removed the clone, aliBuild re-downloaded it, and it seems to be fine now. I've reset the failed macOS tests.

MichaelLettrich:
Concerning Ninja builds: 
I think most of our users (me included) don't care about the underlying build system since we are only interacting with CMake and aliBuild directly as long as it does not require them to change anything. IMHO we should
* Announce it, explain the motivation for the change and leave some time for feedback by users
* update the documentation 
* send a round mail before the change and explain the migration path and a FAQ, to contain the amount of angry mails.

PR #2855, opened 2021-02-18
MW2: bump gsl, libxml2, zlib, clang, libxml2

ktf:
There was a bad version of aliBuild which was used to kickstart the cc8
distribution which produced rpms. I had to bump revisions of a few packages in
order to get proper builds.


ktf:
Whitespace changes. Merging.

PR #2857, opened 2021-02-22
Add unzip system recipe as it is required by DIM

awegrzyn:
DIM recipe uses `unzip`...

### DIFF
--- a/dim.sh
+++ b/dim.sh
@@ -7,6 +7,7 @@ requires:
 build_requires:
   - motif
   - system-curl
+  - unzip
 ---
 #!/bin/bash -e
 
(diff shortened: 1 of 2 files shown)


kostorr (approved):

ktf:
Is this really needed? Do we have a case in which unzip is not in the base system?

ktf (commented):
Do we really need this?

awegrzyn:
Yes, default CC7 OpenStack image.

ktf:
Ok, I would suggest we use type, though.

kostorr (approved):

PR #2867, opened 2021-03-02
DelphesO2 now requires AliRoot

preghenella:


ktf:
@preghenella can we move this to non-WIP?

ktf:
Closing this since there was no answer about making it non-WIP.

PR #2869, opened 2021-03-02
Make O2 and DD recipies consistent with provided modulefiles

awegrzyn:
Either what I proposed or simply an entry to the modulefile should be added.


o2.sh
```diff
@@ -28,6 +27,7 @@ requires:
   - KFParticle
   - VecGeom
 build_requires:
+  - FairMQ
```
ktf, line 30: mmm fairmq is definitely needed at runtime. Am I missing something?

awegrzyn: If you look at modulefile it is not loaded, it probably comes with FairRoot so just to make it consistent I wanted to add an entry to modulefile OR as above, move `FairMQ` to `build_requires` (it will be compiled anyway as it comes with FairRoot).

ktf: I think we should move it to the modulefile.

ktf (approved):

PR #2875, opened 2021-03-03
Bump O2 to v21.09

awegrzyn:
We need this to compile O2 with `o2-dataflow` defaults as now it fails due to lack of boost 1.75 fixes.

### DIFF
--- a/o2.sh
+++ b/o2.sh
@@ -1,6 +1,6 @@
 package: O2
 version: "%(tag_basename)s"
-tag: v21.07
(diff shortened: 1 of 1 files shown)


ktf:
This is fine for me, but then I do not understand what's the point of tagging alidist with O2-v21-09.

awegrzyn:
We use master between releases, few days before actual release we create `flp-suite-v0.YY.ZZ` branch. I didn't even know that we do have alidist tags.

PR #2878, opened 2021-03-04
Bump Delphes to v20210304

preghenella:


preghenella:
failure of `build/O2/alidist-centos8` seems unrelated

PR #2881, opened 2021-03-04
Fix missing square bracket for incremental build

aphecetche:
Looks like there's a missing bracket there ?
@ihrivnac btw, is it on purpose that this geant4-config is only called for incremental builds ?


ihrivnac:
@aphecethe, thanks for fixing the bracket. The  geant4-config  sets the environment for Geant4 data sets, so it is needed for running g4-sim ctest.

aphecetche:
@ihrivnac ok, but then why is it called only for the incremental builds ?

PR #2884, opened 2021-03-05
Manually generate modulefile for Control-Core

teo:


sy-c (approved):

ktf:
Why is this needed?

awegrzyn:
Because of https://alice.its.cern.ch/jira/browse/O2-2031
Core RPM does not pull any deps, but modulefile want to load all `requires` in the tree.

ktf:
Ok, but can't you simply mark all of them as BUILD_REQUIRES so that they get skipped? Not nice, but it's one just one additional line.

PR #2886, opened 2021-03-08
Add AEGIS and O2DPG to full-system-test requirements

davidrohr:
As discussed in AliceO2Group/AliceO2#5635


davidrohr:
Closing since we will do it in a slightly different way as discussed today.

PR #2891, opened 2021-03-09
Fix common prefix in o2checkcode

TimoWilken:
For some reason, O2's `compile_commands.json` in CI has started including the following path, in addition to the usual paths under `sw/SOURCES/`:

    .../sw/BUILD/$hash/O2/version/O2Version.cxx

This breaks the common-prefix heuristic here. This patch includes only files under `sw/SOURCES/` in the calculation, as originally intended, I presume.

### DIFF
--- a/o2checkcode.sh
+++ b/o2checkcode.sh
@@ -16,7 +16,7 @@ cp "${O2_ROOT}"/compile_commands.json .
 # We will try to setup a list of files to be checked by using 2 specific Git commits to compare
 
 # Heuristically guess source directory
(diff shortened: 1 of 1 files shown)


TimoWilken:
In this case (in CI, e.g. [fullCI here]), O2 is a dev package. AFAIK, aliBuild symlinks dev packages under `sw/SOURCES/` as well, so this should be correct in either case.

[fullCI here]: https://github.com/AliceO2Group/AliceO2/pull/5627

aphecetche:
For the record the O2Version.cxx thing (generated file in BUILD dir) is coming from this O2 PR merged yesterday (March 8,2021):  https://github.com/AliceO2Group/AliceO2/pull/5636

davidrohr:
Actually, what is the problem that motivates this PR? Does `O2Version.cxx` violate the coding rules?

davidrohr:
OK, I see, let's see whether it works.
Not tested by alidist CI anyway, merging.

PR #2892, opened 2021-03-09
Added O2sim recipe for the O2 simulation meta-package

preghenella:
@davidrohr @sawenzel @shahor02 
as discussed, here is the recipe for the `O2sim` meta-package.
It currently bundles the following packages together
* O2
* O2DPG
* AEGIS

### DIFF
--- a/defaults-o2-dataflow.sh
+++ b/defaults-o2-dataflow.sh
@@ -27,6 +27,8 @@ disable:
   - JAliEn-ROOT
   - KFParticle
   - MCStepLogger
+  - O2sim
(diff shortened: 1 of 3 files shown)


davidrohr:
I would suggest we make O2sim a requirement for o2-full-system-test.sh, and then we should probably disable O2sim and O2-full-system-test ind defaults-o2-dataflow.sh.

davidrohr:
OK, first test will be whether the CI here runs through then.
BTW FYI: Squashing is never needed for alidist since alidist is configured to always squash at merge.

PR #2902, opened 2021-03-11
Recipes for OpenBLAS and Millepede-II

MichaelLettrich:
Millepede is a separate executable used for detector alignment. It's a bit dated and only available on SVN thus was cloned to a git repo using git-svn.
OpenBLAS is a requirement for Millepede-II

### DIFF
--- a/millepede-ii.sh
+++ b/millepede-ii.sh
@@ -1,0 +1,31 @@
+package: Millepede-II
+version: "%(tag_basename)s"
+tag: "V04-09-00"
+source: https://github.com/alisw/MillepedeII.git
+requires:
+ - zlib
+ - openmp
+ - OpenBLAS
+ - "GCC-Toolchain:(?!osx)"
+build_requires:
+ - alibuild-recipe-tools
+---
+#!/bin/bash -e
+
+rsync -av --delete --exclude="**/.git" ${SOURCEDIR}/ .
+
+case $ARCHITECTURE in
+    osx*)
+    [[ ! $ZLIB_ROOT ]] && ZLIB_ROOT=`brew --prefix zlib` ;;
+esac
+
+make pede\
+     ${ZLIB_ROOT:+ ZLIB_INCLUDES_DIR=${ZLIB_ROOT}/include ZLIB_LIBS_DIR=${ZLIB_ROOT}/lib}\
+     ${OPENBLAS_ROOT:+ SUPPORT_LAPACK64=yes LAPACK64=OPENBLAS LAPACK64_LIBS_DIR=${OPENBLAS_ROOT}/lib LAPACK64_LIB=openblas}
+      
+rsync -a --delete pede ${INSTALLROOT}/bin/
+
+#ModuleFile
+mkdir -p etc/modulefiles
+alibuild-generate-module --bin > etc/modulefiles/$PKGNAME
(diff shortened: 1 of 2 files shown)


MichaelLettrich:
@ktf Millepede-II is only available on SVN thus was cloned to a git repo using git-svn and is hosted under my private github. Should I make that an alisw repo?

ktf:
yes, please.

millepede-ii.sh
```diff
@@ -0,0 +1,31 @@
+package: Millepede-II
+version: "%(tag_basename)s"
+tag: "V04-09-00"
+source: https://github.com/MichaelLettrich/MillepedeII.git
+requires:
+ - zlib
+ - openmp
+ - OpenBLAS
+build_requires:
+ - "GCC-Toolchain:(?!osx)"
```
ktf, line 10: Don't you actually need the fortran runtime?

MichaelLettrich:
@ktf moved the repo.

millepede-ii.sh
MichaelLettrich:
```
ldd pede
	linux-vdso.so.1 (0x00007ffe5e7d6000)
	libgfortran.so.5 => /lib/x86_64-linux-gnu/libgfortran.so.5 (0x00007fc8ba86b000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007fc8ba71c000)
	libgomp.so.1 => /lib/x86_64-linux-gnu/libgomp.so.1 (0x00007fc8ba6da000)
	libz.so.1 => /lib/x86_64-linux-gnu/libz.so.1 (0x00007fc8ba6be000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007fc8ba4cc000)
	libquadmath.so.0 => /lib/x86_64-linux-gnu/libquadmath.so.0 (0x00007fc8ba482000)
	libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007fc8ba465000)
	/lib64/ld-linux-x86-64.so.2 (0x00007fc8babfc000)
	libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007fc8ba45f000)
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007fc8ba43c000)
```
yes, looks like I need both the `libc` and `libgfortran`.

davidrohr:
I have one remark. We already have gsl in alidist (nit ms_gsl but gsl), which provides blas routines iirc. Could you check, and if yes, could we use gsl's blas instead of installing yet another blas package?

MichaelLettrich:
@davidrohr thanks for mentioning. I looked into it. Millepede wants `LAPACK` which comes with `OPENBLAS`. I did not find that in the `gsl` repo unfortunately.

MichaelLettrich:
@ktf this should be OK now I guess.

davidrohr:
For reference, I could compile it locally with `https://www.netlib.org/lapack/` and GSL, but ok, anyway.
One point though, it compiles with only a single thread. And compilation fails when I added -j....
OK, also doesn't take that long anyway.

MichaelLettrich:
@davidrohr indeed, I ran into the same problem that is why the recipe uses a single thread only. Thanks for checking on the lapack dependency. I could switch openblas for lapack to avoid having two BLAS packages in alidist.

PR #2905, opened 2021-03-12
Fix libboost* deletion from INSTALLROOT for DDS

kostorr:
The `-delete` statement was only tied to the second part of the "OR" expression specified by `-o`, leaving `libboost_*` files under the installed `lib` dir.

### DIFF
--- a/dds.sh
+++ b/dds.sh
@@ -40,7 +40,7 @@ esac
 
 make -j$JOBS install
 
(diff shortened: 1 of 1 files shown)


awegrzyn:
Dropping `lib64` should be fine as we're now using `-DCMAKE_INSTALL_LIBDIR=lib`

awegrzyn (approved):

ktf (approved):

PR #2913, opened 2021-03-18
xjalienfs :: tag 1.2.9

adriansev:
Brief user-centric changelog:  
- XrootdCp:: add -strictspec and -http options
- XrootdCp::makelist_xrdjobs:: enforce unix authentication 
- MyCopyProgressHandler::end ::  fix deletion of meta file
- XrootD::cp rework the checks and messages to be clear and informative
- XRootD::cp fix the return/message for the single file copies with destionation already present
- add SEqos command and the internal get_qos
- add certkey-match and tokenkey-match commands
- add {cert,token}-verify commands
- XRootD:: uploads will default to 4 replicas ONLY if absolutly no specs given
- XRootD:: uploads will default to 4 replicas for any .sh, .C, .jdl, .xml smaller than 1MiB

### DIFF
--- a/xjalienfs.sh
+++ b/xjalienfs.sh
@@ -1,6 +1,6 @@
 package: xjalienfs
 version: "%(tag_basename)s"
-tag: "1.2.8"
+tag: "1.2.9"
 source: https://gitlab.cern.ch/jalien/xjalienfs.git
 requires:
  - "OpenSSL:(?!osx)"


adriansev:
i see  
```
[ERROR] Requested resource does not exist: http://ccdb-test.cern.ch:8080/Test/pid22237/CCDBPathUnitTest/999/Hello=World/
```

i think this is not related to the PR.. any thoughts? @alien-experts

ktf:
No, indeed , I think it's unrelated. @Barthelemy

Barthelemy:
probably a transient problem

adriansev:
+test

adriansev:
so, is there a way to restart tests? do i need some empty commit to be added to the branch?

PR #2914, opened 2021-03-19
Force building of yaml-cpp on Ubuntu 20.04

vkucera:
To fix the FairRoot build, which fails if yaml-cpp is picked up from the system.

Reported here: https://alice-talk.web.cern.ch/t/o2-build-fails-in-fairrot-v18-4-1-on-ubuntu-20-04/799
and several times before.

### DIFF
--- a/yaml-cpp.sh
+++ b/yaml-cpp.sh
@@ -6,7 +6,7 @@ requires:
   - boost
 build_requires:
   - CMake
-prefer_system: (?!slc5)
+prefer_system: (?!slc5|ubuntu2004)
 prefer_system_check: |
   pkg-config --atleast-version=0.6.2 yaml-cpp && printf "#include \"yaml-cpp/yaml.h\"\n" | c++ -std=c++17 -I`brew --prefix yaml-cpp`/include -I`brew --prefix boost`/include -xc++ - -c -o /dev/null
 ---


vkucera:
@ktf

ktf (commented):
I'd rather fix FairRoot (possibly dropping yaml-cpp).

dennisklein:
Bump FairRoot to v18.4.2? See
* https://github.com/FairRootGroup/FairRoot/releases/tag/v18.4.2
* https://github.com/FairRootGroup/FairRoot/commit/b49705684ae25b74a10d0826dfbf10cb1469b1fb

dennisklein:
> Is there an option to disable YAML-CPP completely in FairRoot?

Since FairRoot `v18.4.0` the component MCConfigurator which requires `yaml-cpp` is only built, if `yaml-cpp` is found. There is no switch currently to disable the actual `find_package(yaml-cpp)` logic.

vkucera:
> > Is there an option to disable YAML-CPP completely in FairRoot?
> 
> Since FairRoot `v18.4.0` the component MCConfigurator which requires `yaml-cpp` is only built, if `yaml-cpp` is found. There is no switch currently to disable the actual `find_package(yaml-cpp)` logic.

@ktf @dennisklein What is the conclusion on this? Should FairRoot be bumped to v18.4.2?

dennisklein:
> Should FairRoot be bumped to v18.4.2?

From the point of view of upstream FairRoot (which I can represent here) I recommend bumping to `v18.4.2` or later. Fixing the issue described here was one reason for the `v18.4.2` patch release after all (Note: There is also [`v18.4.3`](https://github.com/FairRootGroup/FairRoot/releases/tag/v18.4.3) with more fixes (not sure how relevant for alidist though) and also already `v18.6`).

If `v18.4.2` is no option for you or does not work (which I both hope not), I can bump priority on a build flag as discussed in alisw/alidist#2806.

dennisklein:
> If `v18.4.2` is no option for you or does not work (which I hope both are not), I can bump priority on a build flag as discussed in #2806.

Actually, a [package-level switch already exists](https://cmake.org/cmake/help/latest/variable/CMAKE_DISABLE_FIND_PACKAGE_PackageName.html): Passing `-DCMAKE_DISABLE_FIND_PACKAGE_yaml-cpp=ON` to the FairRoot cmake configure should do the trick, too. (Have not tried it)

vkucera:
@dennisklein @ktf Has this been addressed somehow? It's not clear to me from the discussion what the proposed solution is.

pzhristov:
This PR is not needed anymore, closing.

PR #2916, opened 2021-03-19
Fix O2_SRC computation, for dev packages

TimoWilken:
For example in AliceO2Group/AliceO2#5745, `.cxx` files don't have paths in `SOURCES`, but in the development package, such as `.../o2-fullci/O2/Common/MathUtils/src/Cartesian.cxx`.


ktf:
Ok, for now. I think we need to find a better way to find the common path...

ktf (approved):

aphecetche:
@TimoWilken sorry I missed this one when I created #2942 (now merged). I guess this one can be closed then.

TimoWilken:
OK, thanks @aphecetche! Closing.

PR #2917, opened 2021-03-20
Remove HIJING requirement

preghenella:
not sure why HIJING is set as a requirement for the O2.
Removing it, I don't think this will break anything and it should not be there in principle.


preghenella:
error on `build/O2/alidist-centos8` seems unrelated

ktf:
I think it's just because at the time we did not have AliGenerators. Merging.

PR #2918, opened 2021-03-20
Allow AliRoot and AliPhysics to be built using ninja when in developm…

mfasDa:
…ent mode

Mehtod to detect and select ninja for development
mode ported 1-1 from the O2 build recipe.

### DIFF
--- a/aliphysics.sh
+++ b/aliphysics.sh
@@ -36,6 +36,15 @@ if [[ $CMAKE_BUILD_TYPE == COVERAGE ]]; then
   source $ALIROOT_ROOT/etc/gcov-setup.sh
 fi
 
+# Use ninja if in devel mode, ninja is found and DISABLE_NINJA is not 1
+if [[ ! $CMAKE_GENERATOR && $DISABLE_NINJA != 1 && $DEVEL_SOURCES != $SOURCEDIR ]]; then
+  NINJA_BIN=ninja-build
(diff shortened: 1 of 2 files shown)


mfasDa:
Could one please at least agree to enable ninja in development mode i.e. if users request it, and not forbid the user completely to use it? Such setting could be done i.e. via aliBuild if such options are foreseen, or via an environment variable. In case an agreement could be achieved I would modify the pull request accordingly.

ninja is useful in development mode as it can reduce the build time in particular in partial builds, and honestly I don't understand why the O2 developers can profit from this, the AliRoot/AliPhysics developers however are forced to use make. If ninja is not on the system it anyhow falls back to make.

ktf:
Nothing is getting forbidden, just prioritised, since no one ever requested for it nor proposed a solution. Since you now did both, and I assume you are willing to support the solution, I guess we can merge this. BTW, FYI, we decided in WP3 to switch to ninja as default in O2.

mfasDa:
@ktf Thanks a lot! Yes, I am very much in favour - in fact I was building myself with ninja all the time because it is simply faster. I don't expect there will be too much disturbance, but of course in case I am willing to help.

mfasDa:
@ktf There might actually be a problem for AliPhysics on older platform, which is that AliRoot contains Fortran Code, and this is supported by ninja only from v1.10. This is not of relevance for O2 and AliPhysics, which itself don't contain any Fortran code. But this should be possible to handle via a check for the ninja version.

PR #2923, opened 2021-03-23
Workaround issue with python in Catalina

ktf:


pzhristov (approved):
We may need something similar for osx_arm64

ktf:
I think the problem is on Catalina, not Big Sur.

pzhristov:
Yes, the current xrootd version compiles on M1 (Big Sur). I was thinking about xrootd 5...

PR #2926, opened 2021-03-28
Remove AEGIS from O2 requirements

preghenella:
@sawenzel @shahor02 @davidrohr 
this is the second step of the process


davidrohr:
thx, full system test passes.

davidrohr (approved):

PR #2928, opened 2021-03-29
Fix golang detection from system

awegrzyn:
Otherwise it does not fail when `go` is not present at all

### DIFF
--- a/golang.sh
+++ b/golang.sh
@@ -3,7 +3,7 @@ version: "1.15.6"
 build_requires:
   - system-curl
 prefer_system_check: |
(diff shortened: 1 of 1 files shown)


golang.sh
```diff
@@ -3,7 +3,7 @@ version: "1.15.6"
 build_requires:
   - system-curl
 prefer_system_check: |
-  case `go version | sed -e 's/go version go//' | sed -e 's/ .*//'` in 0*|1.[0-9].*) exit 1 ;; esac
+  which go && case `go version | sed -e 's/go version go//' | sed -e 's/ .*//'` in 0*|1.[0-9].*) exit 1 ;; esac
```
ktf, line 6: `type go`

ktf (changes requested):

PR #2929, opened 2021-03-29
Use O2@dev for o2.sh in alidist@master

ktf:


ktf:
There is no guarantee of forward compatibility for any of the O2 tags. If someone needs forward compatibility, they will have to refer to the `O2-<version>` tag and fork from there.

awegrzyn:
Could you elaborate why?
I'd expect AliceO2 checks to use `dev` and `alidist` to use released version.

ktf:
The alidists test are propaedeutic to the **next** O2 tag, not to the previous one. We do not want to spend 2 weeks in checking the old tag vs alidist master to then find out it's broken once we release. If you want a forward compatible branch, based on the old tag, you can use the `O2-<tag>-patches` branch, where we do guarantee that everything will work with the old tag.

awegrzyn:
Why? In order to test features for the next O2 tag there are tests in AliceO2.
I'd expect to have latest (or current) tag in alidist as it is for all other packages.

ktf:
The tests in O2 only check O2, not new stuff that might enter alidist. Say that we want to move to a new version of ROOT, which we actually need to do soon, we want to make sure that the current O2 works with the new ROOT, not the O2 released 2 weeks ago. Same for QC: we want to make sure that 0.14.x works with the new O2 tag, not with the old one. Again, for forward / backward compatibility we have the `-patches` branches.

davidrohr:
I am also in favor of comparing to dev here.
Anyhow, when you check out alidist/master, you don't get the software stack of the last o2 tag, for that you'll have to check out the alidist tag.
And for validating our software, I think it is more important to follow the dev in order to make the next release work (since the previous release was already working). In particular, if we use the latest tag here, it is possible that alidist PRs break O2/dev.
But then, we have actually already discussed this in WP3...

awegrzyn:
Okay understood, in such case we should move all our packages to `dev` to make sure they work.

davidrohr:
> Okay understood, in such case we should move all our packages to `dev` to make sure they work.

Not sure. With the current procedure, the next O2 release will become what is currently tagged in alidist + O2/dev?

awegrzyn:
> (...) Say that we want to move to a new version of ROOT, which we actually need to do soon, we want to make sure that the current O2 works with the new ROOT, not the O2 released 2 weeks ago. (...)

Sure I fully understand, therefore I misunderstood the point of `master` branch.
So we should make sure that such changes do not break other top-packages as well, we should set to dev: QualityControl, Readout, ALF and Control...

ktf:
No, also this was already discussed. The point is that when we are releasing O2, not ROOT, nor Monitoring nor whatever. So we need to have a coherent set of dependencies which we know will work with the current version of O2. This will give the minimum baseline everyone agrees on, and they can build on top (again, in the patches branch).

If we move everything to dev, then every single package will have to be tagged on O2 release date because that's the only "tested ensemble". This is fine for me, but that means that we are synchronising all the releases to O2, which I understood it is not what the FLP project wants to do (hence the request for the bi-weekly tags).

awegrzyn:
All right, could you please explain what would be the steps to release FLP Suite and generate RPMs? What should we use as our base branch to create our release branch? As we don't want to use `O2@dev` (therefore `master` branch`) for our releases, how can we set CI for our release branches?

ktf:
You already have, since a while, `alidist-dataflow` tests which you can use, I guess. Unless I am wrong, the configuration file for it is at:

https://github.com/alisw/ali-bot/blob/master/ci/repo-config/mesosci/slc7/o2-dataflow.env

Unless I misunderstand your request, I think you simply need to change the branch to `O2-<version>-patches` and remove the checkout of AliceO2. @TimoWilken might have more insights.

Barthelemy:
I had put a discussion point for tomorrow's WP3 meeting: https://indico.cern.ch/event/1019879/
I propose to break here the discussion until tomorrow.

Barthelemy (approved):

awegrzyn:
This was already done in #2975

PR #2933, opened 2021-03-30
Update jalien.sh

costing:


adriansev (approved):

adriansev:
@ktf given that this is a java part all these tests (and their errors) have nothing to do with the PR (unless there are specific tests that are unknown to me)

PR #2935, opened 2021-03-30
Initial attempt at onnxruntime

ktf:


ktf:
@saganatt this requires a bit of massaging, but it should be ok to get you started. In particular:

* You will need to have flake installed apparently. You could try dropping it by hand from the CmakeLists.txt
* it will build Eigen and Protobuf internally. Maybe you want to use our own version.

saganatt:
> @saganatt this requires a bit of massaging, but it should be ok to get you started. In particular:
> 
>     * You will need to have flake installed apparently. You could try dropping it by hand from the CmakeLists.txt
> 
>     * it will build Eigen and Protobuf internally. Maybe you want to use our own version.

Thanks a lot!

ktf:
@saganatt is this still useful?

PR #2920, opened 2021-03-21
Require ninja > 1.10 for AliRoot, otherwise fall back to make

mfasDa:
AliRoot contains Fortran code, which requires ninja 1.10.
It the ninja version is too old on the system this will
break the compilation. In case the ninja version is 1.10
the compilation must fall back to make.

PR #2936, opened 2021-03-30
Require ninja >= 1.10 for AliPhysics, otherwise fall back to make

vkucera:
Same fix as https://github.com/alisw/alidist/pull/2920 because of Fortran code in PWGCFfemtoscopyUser.


ktf (approved):

mfasDa:
@ktf Can this be merged? It should be merged as soon as possible as many linux distributions still ship ninja <= 1.8, and AliPhysics still has two Fortran files which would require ninja 1.10.

ktf (approved):

ktf:
I missed that @vkucera is a first time committer. Let's see how the tests go.

PR #2942, opened 2021-04-04
[O2-2212] Change the logic to infer the O2_SRC directory

aphecetche:
Seems to solve [O2-2212](https://alice.its.cern.ch/jira/browse/O2-2212) for me


sawenzel (approved):
Ok for me (if it still works in the first iteration, too).

aphecetche:
@sawenzel AFAICT yes it still works on the first iteration too :wink:

PR #2944, opened 2021-04-06
Update O2 to latest version

Barthelemy:
While we wait for the outcome of the upcoming discussion about whether we should use dev or a fixed tag here, let's at least have an up-to-date version.


Barthelemy:
@alisw/infrastructure-admins It seems that macOS is stuck

Barthelemy:
it is not stuck anymore :)

awegrzyn (approved):

PR #2950, opened 2021-04-08
Fix xjalienfs shebang replacement

TimoWilken:
If we're building in a deeply-nested subdirectory, the scripts' shebangs will be over 127 bytes, and therefore pip will use a hack to embed the python command line, like:

```bash
#!/bin/sh
'''exec' /some/long/path/to/our/python3 "$0" "$@"
' '''
```

We want to replace this to use the system python3, but as that's a short shebang, we can just use a regular `#!` line. The new command detects the above case, deletes the first 3 lines, and inserts a regular `#!/usr/bin/env python3` shebang instead.

If pip generates scripts with a regular shebang like:

```bash
#!/some/shortish/path/to/our/python3
```

...we just replace the shebang with one pointing to the system python3.

### DIFF
--- a/xjalienfs.sh
+++ b/xjalienfs.sh
@@ -19,9 +19,19 @@ fi
 env PYTHONUSERBASE="$INSTALLROOT" ALIBUILD=1 python3 -m pip install --ignore-installed $PIPOPTION file://${SOURCEDIR}
 
 # Make sure all the tools use the correct python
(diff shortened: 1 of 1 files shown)


TimoWilken:
Tested and working inside a long build path on slc7.

adriansev (approved):

adriansev:
@ktf alidist-centos8 errors are not related to this PR, i would say that this can be merged

PR #2956, opened 2021-04-12
DebugGUI: bump to v0.5.2

ktf:


ktf:
@davidrohr can you check with this?

davidrohr:
@ktf : I tried this with Linux. My event display still works, but the DebugGUI itself is broken. It shows me just a single-color greenish window without any content.

ktf:
Ok, I actually forgot to call the removed callback. As seen in AliceO2Group/AliceO2#23, `v0.5.2` should fix things.

PR #2960, opened 2021-04-12
root.sh:: build vmc for any v6

adriansev:


adriansev:
@ktf any opinion on this?

PR #2967, opened 2021-04-16
[QC-527] Add python module to access Consul

Barthelemy:


sy-c (approved):

python-modules-list.sh
```diff
@@ -25,6 +25,7 @@ env:
     responses==0.10.6
     pandas==0.24.2
     scikit-learn==0.20.3
+    python-consul==1.1.0
```
ktf, line 28: Do we really need this everywhere?

Barthelemy: good question actually. We need it on the CCDB. What would you advise ? remove it and install by hand on that machine ?

awegrzyn: I think same applies to `dryable`

awegrzyn: > remove it and install by hand on that machine ?

If you need it for a standalone script this could go to QC Ansible role, and I guess this is needed on head node only (?)

Barthelemy: yes head node for now, and later QCDB node. Ok, good point. I will prepare something in ansible for that and close this ticket.

PR #2668, opened 2020-12-03
Move O2 to v10.2.0-alice2

ktf:

PR #2973, opened 2021-04-20
Protobuf: bump to 3.15.8

ktf:
Needed by Arrow 3.0 on some platforms.

### DIFF
--- a/defaults-o2-dataflow.sh
+++ b/defaults-o2-dataflow.sh
@@ -30,6 +30,8 @@ disable:
   - O2sim
   - O2-full-system-test
 overrides:
+  protobuf:
+    version: v3.14.0
   Python-modules-list:
     env:
       PIP_REQUIREMENTS: |
--- a/coconut.sh
+++ b/coconut.sh
@@ -1,6 +1,6 @@
 package: coconut
 version: "%(tag_basename)s"
-tag: "v0.22.80"
+tag: "v0.22.81"
 build_requires:
   - golang
   - protobuf
--- a/coconut.sh
+++ b/coconut.sh
@@ -1,6 +1,6 @@
 package: coconut
 version: "%(tag_basename)s"
-tag: "v0.22.2"
+tag: "v0.22.3"
 build_requires:
   - golang
   - protobuf
--- a/coconut.sh
+++ b/coconut.sh
@@ -1,6 +1,6 @@
 package: coconut
 version: "%(tag_basename)s"
-tag: "v0.22.81"
+tag: "v0.22.82"
 build_requires:
   - golang
   - protobuf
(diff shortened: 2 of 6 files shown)


teo (approved):

teo (approved):

ktf:
@teo looks like this is creating troubles to `occ.pb.cc`, can you have a look?

ktf:
@ironMann can you take a look at why DD fails with the new version of protobuf? For a similar issue in AliECS @teo had to force C++14 compilation, IIUC.

ironMann:
@ktf it's prob this: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=82461
Seems to be solved as of gcc 7.4

Stepping down from c++17 is unfortunately not an option now for DD. There are also other protobuf users, dds, odc, ...

davidrohr:
Sounds like yet another reason to go to GCC 10.2 :)

ktf:
@davidrohr Yeah... As long as we do not start using C++20, since that will break ROOT...

ktf:
Ok I've bumped #2668 to use GCC 10.2. Seems to work fine on my ubuntu box, including the autotools-archive stuff...`--std=c++2y` here we come! ;-)

ktf:
@vascobarroso @awegrzyn could you comment if it's really required for FLP to stick to the system compiler for cc7? It will require us to do some juggling to get the new arrow in, if that is the case.

vascobarroso (dismissed):

teo:
@ktf can we merge this? We could use it now for the Suite beta release.

defaults-o2-dataflow.sh
```diff
@@ -30,6 +30,8 @@ disable:
   - O2sim
   - O2-full-system-test
 overrides:
+  protobuf:
```
awegrzyn, line 33: @teo I think this needs to be removed before?

teo: Sounds right, is there any impediment to doing so?

vascobarroso: @ktf could you confirm that we can go ahead with this suggestion ?

ktf: Not until you are happy with dropping GCC 7.3, since 5.15.8 will not compile there.

ktf: Alternative @ironMann should find a workaround like you did to use 3.15.8 in DD & the rest of the EPN stack.

teo: I will unbump the minimum version of Protobuf for OCC back to v3.14.0 so we can still build the FLP Suite with defaults `o2-dataflow`.

teo: Fix here https://github.com/alisw/alidist/pull/3014

vascobarroso (approved):

vascobarroso:
@alisw/infrastructure-admins this looks green now. Could it be merged, we need this for the release of FLP Suite v0.18.0 beta.

PR #2981, opened 2021-04-22
Codechecker should not complain aboud cpulimit, which is a 3rdparty tool

davidrohr:


davidrohr:
@sawenzel : either this way, or we move the cpulimit into a folder that has 3rdparty in its name, then we would not need to change the o2checkcode.sh recipe.

PR #2994, opened 2021-04-29
Add FLUKA as an O2 dependency

ktf:


pzhristov (approved):

davidrohr:
Hi  @ktf @pzhristov : Do I understand correctly that this means one needs CERN credentials to do a default build of O2 now? Since Fluka and Fluka VMC are hosted on the CERN gitlib, and my automatic build on got stuck there, waiting for me to enter my cern credentials. I am not really sure whether this is ideal.

awegrzyn:
Same here, can we revert it? It's gonna break a lot of builds. Can we disable it in `o2-dataflow` defaults or it's mandatory for `O2`?

davidrohr:
Btw, even if you have CERN credentials, currently the checkout with aliBuild is kind of broken: alisw/alibuild#681

pzhristov:
@davidrohr @awegrzyn O2 simulation needs Fluka for radiation studies and this is why we put the dependency. FLUKA/FLUKA_VM are in our private gitlab repository at CERN, so it requests credentials. You can disable FLUKA_VMC from the command line "aliBuild build O2 --defaults o2 --disable fluka_vmc", or modify o2-dataflow.

davidrohr:
@pzhristov : I think we should discuss this with @sawenzel and @preghenella .
As far as I remember, the idea was to have the simulation packages not as O2 dependency. 
In order to use them, your build both and then you do alienv enter O2/latest FLUKA_VMC/latest.
I think they should rather be a dependency of O2DPG. For instance we have removed AEGIS from the O2 dependencies.
Or do we need to have FLUKA_VMC at build time?

pzhristov:
@davidrohr This was the easiest way to set the FLUKA_VMC environment in the O2 modulefiles. If other way works, we can use it.

o2.sh
```diff
@@ -25,6 +25,7 @@ requires:
   - cgal
   - KFParticle
   - VecGeom
+  - FLUKA_VMC
```
sawenzel, line 28: I think the dependency is correct but why is it added here instead of simulation.sh?
This should be the correct place.

pzhristov: Ok, I will have a look and make a new PR.

ktf: You are right, my bad fixed in #3005.

PR #2996, opened 2021-04-29
Specify ROOTSYS to current aliBuild ROOT version

mfasDa:
Needed in order to force compilation against ROOT from
alidist in case system ROOT versions are present.

### DIFF
--- a/roounfold.sh
+++ b/roounfold.sh
@@ -10,6 +10,7 @@ cmake $SOURCEDIR                              \
       -DCMAKE_INSTALL_PREFIX=$INSTALLROOT     \
       ${CXXSTD:+-DCMAKE_CXX_STANDARD=$CXXSTD} \
       ${CMAKE_BUILD_TYPE:+-DCMAKE_BUILD_TYPE="$CMAKE_BUILD_TYPE"}    \
+      -DROOT_DIR=$ROOTSYS \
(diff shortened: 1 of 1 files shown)


ktf:
I would use `$ROOT_ROOT`.

mfasDa:
> I would use `$ROOT_ROOT`.

Not sure what you mean, but RooUnfold has its own CMake module for ROOT, and there it discovers ROOT according to the root-config executable which it finds like this: 

```
FIND_PROGRAM( ROOT_CONFIG_EXECUTABLE root-config PATHS ${ROOT_DIR}/bin NO_DEFAULT_PATH )
```

So in order to take the custom root I need to specify -DROOT_DIR, and I suspect the environment variable ROOT_ROOT won't help here too much.

mfasDa:
@ktf I see now what you mean, but who exports $ROOT_ROOT? Our modulefile still export $ROOTSYS. Just checked with the ROOT module and the $ROOT_ROOT environment variable is not set

ktf:
modulefiles are build time only. alibuild automatically defines `<uppercase-package-name>_ROOT`. IMHO this is better than using ROOTSYS, because this way the user environment is less likely to come into play during the build.

mfasDa:
@ktf Thanks for clarification - I was not aware of this. Indeed then that approach is safer.

ktf (approved):

PR #3006, opened 2021-05-01
Bump to GCC 10.2 everywhere except in release defaults

davidrohr:


davidrohr:
for testing, for now.

ktf:
Closing the pull request as rebuilding it slows down the the rest of the CI.

davidrohr:
Makes sense, the failures were with ROOT5 in the ALiRoot CI, which does not compile out of the box with GCC 10. I am not sure whether we want to do anything here @ktf @pzhristov

pzhristov:
I really hope to get rid of Root 5 in very near future (as soon as DPG stops running productions on the old data). Out of curiosity I can try to compile Root 5 with gcc 10.2, but it will take some time to fix all the issues.

PR #3007, opened 2021-05-03
Standalone onnxruntime for O2 ML tests

saganatt:
Ciao @ktf ,

here's a complete working setup. I couldn't force it not to build its own flatbuffers, the rest of dependencies seems not repeated.

### DIFF
--- a/onnxruntime.sh
+++ b/onnxruntime.sh
@@ -1,25 +1,47 @@
 package: onnxruntime
-version: "%(tag_basename)s"
-tag: 1.7.2
-source: https://github.com/microsoft/onnxruntime.git
+version: "1.7.2"
+tag: c073c88ef
+source: https://github.com/saganatt/onnxruntime.git
+requires:
+  - protobuf
+  - re2
 build_requires:
- - "GCC-Toolchain:(?!osx)"
- - CMake
- - Python-modules
- - re2
- - protobuf
+  - CMake
 ---
-# TODO: should be: git clone --recurse-submodules, how to pass git option?
+#!/bin/bash -e
+
+pushd $SOURCEDIR
+  git submodule update --init -- cmake/external/date
+  git submodule update --init -- cmake/external/mp11
+  git submodule update --init -- cmake/external/onnx
+  git submodule update --init -- cmake/external/optional-lite
+  git submodule update --init -- cmake/external/eigen
+  git submodule update --init -- cmake/external/nsync
+  git submodule update --init -- cmake/external/flatbuffers
+  git submodule update --init -- cmake/external/SafeInt
+  git submodule update --init -- cmake/external/json
+popd
 
 mkdir -p $INSTALLROOT
-cmake -Donnxruntime_DEV_MODE=OFF \
-      -Donnxruntime_PREFER_SYSTEM_LIB=ON \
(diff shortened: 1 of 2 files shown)


onnxruntime.sh
```diff
@@ -0,0 +1,58 @@
...
+      -Donnxruntime_USE_PREINSTALLED_NSYNC=OFF \
+      ${RE2_ROOT:+-DRE2_INCLUDE_DIR=${RE2_ROOT}/include}
+
+cmake --build . -- ${JOBS:+-j$JOBS} install
+
+# Modulefile
+MODULEDIR="$INSTALLROOT/etc/modulefiles"
+MODULEFILE="$MODULEDIR/$PKGNAME"
+mkdir -p "$MODULEDIR"
+alibuild-generate-module > "$MODULEFILE"
```
ktf, line 52: `--lib` will avoid you the bit below.

onnxruntime.sh
```diff
@@ -0,0 +1,58 @@
+package: onnxruntime
+version: "1.7.2"
+tag: c073c88ef
+source: https://github.com/saganatt/onnxruntime.git
```
ktf, line 4: why your own repo? do we need to create one in `alisw`?

onnxruntime.sh
```diff
@@ -0,0 +1,58 @@
...
+requires:
+  - protobuf
+  - re2
+build_requires:
+  - CMake
+---
+#!/bin/bash -e
+
+pushd $SOURCEDIR
+  git submodule update --init -- cmake/external/date
```
ktf, line 14: There is no alternative than doing git submodule init?

onnxruntime.sh
```diff
@@ -0,0 +1,58 @@
...
+mkdir -p $INSTALLROOT
+
+# NOTE: It builds its own flatbuffers, eigen, nsync, pybind11
+cmake "$SOURCEDIR/cmake" \
+      -DCMAKE_INSTALL_PREFIX=$INSTALLROOT \
+      -DCMAKE_BUILD_TYPE=$CMAKE_BUILD_TYPE \
+      -Donnxruntime_DEV_MODE=OFF \
+      -Donnxruntime_BUILD_UNIT_TESTS=OFF \
+      -Donnxruntime_ENABLE_PYTHON=ON \
+      -DPYTHON_EXECUTABLE=$(python3 -c "import sys; print(sys.executable)") \
```
ktf, line 34: you should probably depend on our own python...

ktf (approved):

ktf:
Merging for now, since it's not tested in any case. Please make sure you address the comments.

onnxruntime.sh
saganatt: I made slight changes to their CMakeLists.txt. If we used the original, it would build much more.

onnxruntime.sh
saganatt: hm, what only comes to my mind is that the users would need to install these dependencies themselves.

onnxruntime.sh
saganatt: If we don't do git submodule, and we force the users to install dependencies themselves, then it won't matter that much, though. The difference would be just one cmake line adding Flake as dependency.

PR #3015, opened 2021-05-04
bump readout v2.1.0

sy-c:


readout.sh
```diff
@@ -1,6 +1,6 @@
 package: Readout
 version: "%(tag_basename)s"
-tag: v2.0.2
+tag: v2.1,0
```
kostorr, line 3: There's a comma instead of a dot here ;)

sy-c: indeed, thanks !

kostorr (dismissed):

ktf:
I think you are missing an explicit cast to `void*`...

sy-c:
> I think you are missing an explicit cast to `void*`...

@ktf I am tired of loosing time "fixing" things blocking when doing PR and that are not even reported as a warning when building with aliBuild on my straight CC7 system.

ktf:
I fully sympathise, but it's not my fault nor alibuild. ;-) As I think I said already the issue is in the readout.sh recipe:

https://github.com/alisw/alidist/blob/master/readout.sh#L38-L46

Since no one did so far, I opened:

https://github.com/alisw/alidist/pull/3022

but someone from FLP should approve / merge it, since my understanding is that this is actually what is requested.

awegrzyn:
@sy-c I'd recommend using GCC 9 from devtoolset.

This warning is generated because in Readout's CMake `-Wall` is enabled, but indeed GCC 7 from devtoolset does not report it.

sy-c:
> 
> 
> @sy-c I'd recommend using GCC 9 from devtoolset.
> 
> This warning is generated because in Readout's CMake `-Wall` is enabled, but indeed GCC 7 from devtoolset does not report it.

Please document this. I'm using the build system exactly as advertised in the documentation.

sy-c:
> 
> 
> I fully sympathise, but it's not my fault nor alibuild. ;-) As I think I said already the issue is in the readout.sh recipe:
> 
> https://github.com/alisw/alidist/blob/master/readout.sh#L38-L46
> 
> Since no one did so far, I opened:
> 
> #3022
> 
> but someone from FLP should approve / merge it, since my understanding is that this is actually what is requested.

@ktf thanks, I was not aware of this issue about the recipes.

awegrzyn:
> Please document this.

GCC 9 is not officially supported but this is what I'm doing.
The actual problem is that GCC 7 from devtoolset does not report the warning but aliBuild compiled one does.

sy-c:
@alisw/infrastructure-admins the CI test for CentOS 8 failed with an unrelated error. Can you please check?

vascobarroso (approved):

PR #3020, opened 2021-05-05
Bump bookkeeping to v0.16.9

vascobarroso:


awegrzyn:
@vascobarroso CI will not report any error on this as this `Bookkeeping` it not used in any package
@DimitrySteenkamer It looks like `v0.16.9` fails with 
```
/cpp-api-client/src/CreateRunParameters.h:4:10: fatal error: boost/date_time/posix_time/ptime.hpp: No such file or directory
 #include <boost/date_time/posix_time/ptime.hpp>
```
If I take a look at your CMake (https://github.com/AliceO2Group/Bookkeeping/blob/decc0eed681edcdb667ca892d78f06996583c873/cpp-api-client/CMakeLists.txt#L16) you don't mention `date_time` in `COMPONENTS`.

awegrzyn:
Actually the error comes from `CreateRunParameters.h` header? Do you even use it? As far as I can see it's included in `BookkeepingInterface.h` but never used?

awegrzyn (changes requested):
see my comments

DimitrySteenkamer:
I do not know why the error says this. Because bookkeepingApi.cpp uses this header as well.
https://github.com/AliceO2Group/Bookkeeping/blob/decc0eed681edcdb667ca892d78f06996583c873/cpp-api-client/src/BookkeepingApi.cpp#L3

DimitrySteenkamer (approved):

awegrzyn:
@DimitrySteenkamer The error says that because use `Boost::date_time` in public header and in CMake you link against it under `PRIVATE`...

PR #3022, opened 2021-05-06
Make Readout not behave differently in CI from the rest

ktf:


sy-c (approved):

awegrzyn:
I think you would need to get rid of `-Werror` from here as well: https://github.com/AliceO2Group/Readout/blob/master/CMakeLists.txt#L30

ktf:
Why? I think the issue is the different behaviour, not the behaviour itself, no?

sy-c:
> Why? I think the issue is the different behaviour, not the behaviour itself, no?

exactly... I put the explicit -Werror to detect them on my dev system, and avoid bad surprises when doing the osx checks

awegrzyn:
Because the error comes from fact that `o2` defaults are using GCC 10 that reports that warning and GCC 7 does not.

awegrzyn:
Okay, I'm a bit confused, how did it pass?

ktf:
Old readout. The issue is in the new version.

sy-c:
> Because the error comes from fact that `o2` defaults are using GCC 10 that reports that warning and `dataflow` defaults still use GCC 7 that does not.

great, if that's so simple, no problem, I will use o2 defaults instead of dataflow on my dev node.

ktf:
Shall I just close this, then?

sy-c:
I think it's still worth to remove the unnecessary custom part... let me just fix the source code first.

ktf:
@sy-c did you fix the code in the end? Can we merge this?

sy-c:
@awegrzyn yes, we have been using gcc10 everywhere for some time now.

PR #3011, opened 2021-05-04
Bump ReadoutCard to v0.31.20

kostorr:

PR #3026, opened 2021-05-07
Bump FairMQ to 1.4.37

rbx:
@ktf includes proper fixes for the two previous hotfixes for filesystem and the plugin changes.
@ironMann includes the region caching. This also includes a possibility to pass option for UnmanagedRegion creation: https://github.com/FairRootGroup/FairMQ/blob/master/examples/region/Sampler.cxx#L55
The two options will mlock and/or zero the region memory *before the region event callbacks are called*, like we discussed in the GPU meeting. This is non-breaking as it is an optional argument.

### DIFF
--- a/fairmq.sh
+++ b/fairmq.sh
@@ -1,8 +1,6 @@
 package: FairMQ
-# version: "%(tag_basename)s"
-# tag: v1.4.35 TODO switch back to version tags once 1.4.36 is released
-version: "v1.4.35.2"
-tag: v1.4.35_hotfix_alice2
+version: "%(tag_basename)s"
+tag: v1.4.36
 source: https://github.com/FairRootGroup/FairMQ
 requires:
  - boost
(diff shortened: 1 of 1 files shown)


rbx:
Ok, issue with OFI transport interface, fix coming up.

dennisklein:
`unknown location(0): fatal error: in "store_retrieve_TMemFile_templated_test": boost::filesystem::filesystem_error: boost::filesystem::status: Operation not permitted: "153e-f9c4-2ea0-1527/Test/pid9161/"`

@ktf @davidrohr Possibly related to https://github.com/alisw/alidist/pull/3011#issuecomment-832434085 and/or https://github.com/AliceO2Group/AliceO2/pull/6064#issuecomment-833323193?

davidrohr:
@dennisklein : The boost:filesystem error seems to be some incompatibility with boost::filesystem + docker + cc8, no idea what is behind, but we are switching to std::filesystem anyway, so this will hopefully disappear. #6102

PR #3034, opened 2021-05-11
Switch from Herwig v7.2.0 to v7.2.2

maireiphc:
See also the dependency to thePEG v2.2.2, also upgraded from its v2.2.0
Update of ThePEG recipe thepeg.sh : https://github.com/alisw/alidist/pull/3033


mfasDa:
@maireiphc in Herwig we are at version 7.2.2, not 2.2.2 :)

Needed for compatibility with gcc-10

PR #3036, opened 2021-05-11
Bump DebugGUI and use ninja

ktf:
Hopefully solving issues some people have encountered with Ubuntu, when TBB is found.


ktf:
Compiles in full CI and not really tested otherwise.

PR #3044, opened 2021-05-19
Bump FairMQ to v1.4.38

rbx:
A regression in shmem was introduced in the previous tag. Fixed in this tag.

### DIFF
--- a/fairmq.sh
+++ b/fairmq.sh
@@ -1,6 +1,6 @@
 package: FairMQ
 version: "%(tag_basename)s"
-tag: v1.4.37
+tag: v1.4.38
 source: https://github.com/FairRootGroup/FairMQ
 requires:
  - boost


ironMann:
Thanks @rbx for the fast response!

@vascobarroso This should fix instability in the FLP chain that we see on large setups. How soon can we deploy this in the current O2 suite? Thanks

awegrzyn:
Are these changes touch public headers? I'm wondering if we need to recompile everything above or we could get away with replacing library.

ironMann:
From stack traces it looks like the problem is contained within the library.
```
/opt/o2/lib/libFairMQ.so.1.4(_ZN4fair2mq5shmem7Message10DeallocateEv+0x492)[0x7fba764daf42]
/opt/o2/lib/libFairMQ.so.1.4(_ZN4fair2mq5shmem7MessageD1Ev+0x24)[0x7fba764db854]
/opt/o2/lib/libFairMQ.so.1.4(_ZN4fair2mq5shmem7MessageD0Ev+0x9)[0x7fba764dbad9]
StfSender[0x4a16a2]
StfSender[0x49f736]
StfSender[0x456e9f]
```

awegrzyn:
ok, @rbx could you confirm?

rbx:
The latest bug-fixing commit touches only private headers. But this tag contains also commits with publicly visible changes (examples and two new methods in `fair::mq::shmem::Monitor`).

ktf (approved):

PR #3048, opened 2021-05-21
cppzmq: init at 4.7.1

ktf:


ktf:
@shahor02 here it is... The reason why it did not work out of the box is that we still build zeromq with autotools. I changed that and build cppzmq, but I could not verify the rest still builds.

shahor02:
Thanks! Will test.

ktf:
@shahor02 do you still need this?

shahor02:
Sorry, just saw this question. Yes, we do need it and preferably in the master...

ktf:
Error unrelated. Merging.

PR #3054, opened 2021-05-31
Reenable the multinode test

Barthelemy:


Barthelemy:
@knopers8

knopers8 (approved):

Barthelemy:
@ktf There is an analysis test failure. Could you have a look ?

ktf:
@TimoWilken this one seems to have issues on Centos 8

knopers8:
This test uses DPL proxies and checks them in the context of QC. Since the proxies are likely to be modified quite soon, it would be good if this test was enabled before that.

ktf:
Given dataflow seems to work. I am merging this.

knopers8:
Thanks!

PR #3068, opened 2021-06-04
Bump Mesos to v0.11.0

teo:


ktf:
Do we really need to have the dependency? That would increase the build time of O2Suite for no actual reason, no?

teo:
> Do we really need to have the dependency? That would increase the build time of O2Suite for no actual reason, no? In the sense that nothing has a direct dependency on mesos, not even the parts which actually use it, no?

No indeed, there there is no linkage against `mesos`. However we need to ensure Mesos is at its latest version at the time of the O2Suite release, and we need to ensure a Mesos RPM is produced by f2r, so I've been asked to do this.

vascobarroso:
Can this be fast-tracked ? It is one of the blocking pieces for the migration to CS8.

davidrohr:
MacOS error is unrealted and will be fixed in a separate PR.
Full system test fails in simulation, which is obviously unrelated to Mesos.
Merging.

PR #3074, opened 2021-06-08
Update defaults-generators.sh

preghenella:
use latest pythia8 version


preghenella:
hello @ktf 

there is a chance that this PR will break the building of `AliGenerators`.
but there is also a chance that it will fix a issue with `AliGenerators`.
if you would be so kind to approve and merge this before 16:00, we will know on which side the coin will fall.

Thanks

PR #3077, opened 2021-06-09
Protect against alienv enter ofi when building.

ktf:


ktf:
@preghenella looks like your update of yesterday indeed broke stuff in AliGenerators. Can you have a look?

ktf:
@awegrzyn this should fix the issue you saw with the asiofi modulefile.

PR #3089, opened 2021-06-11
Fix dependencies for Mesos

teo:


mesos.sh
```diff
@@ -2,12 +2,16 @@ package: mesos
 version: v1.11.0
 tag: 1.11.0
 source: https://git-wip-us.apache.org/repos/asf/mesos.git
+requires:
+- zlib
+- curl
```
awegrzyn, line 7: @teo Mesos is only package that requires our own `curl` on CS8, could this be replaced with:
```
  - "system-curl:(slc8)" 
  - "curl:(?!slc8)"
```
?

PR #3094, opened 2021-06-14
Add system-* recipes for Mesos dependencies

teo:


awegrzyn:
Thanks, this works on my setup, `.rpm-extra-deps` is not needed anymore

teo:
@awegrzyn aren't they still potentially needed for the runtime deps to end up in the RPM? The `system-*` recipes I've added are only build requirements.

awegrzyn:
You can simply move from `build_requires` to `requires` those recipes that are needed during runtime, this is enough

teo:
How would this work though? I can put something like `system-apr` under `requires`, but the actual RPM that needs to be installed on the target machine isn't `system-apr` because the `system-apr` recipe doesn't provide `apr` at all, it only provides the `aliDoctor` check. `apr` from the main repo remains the package to install, and `system-apr.sh` will not pull it.

awegrzyn:
You need this for alisw RPMs but not for FLP RPMs

awegrzyn (approved):

ktf:
I guess we could actually package apr, at some point.

PR #3095, opened 2021-06-14
Update GEANT4 to v10.7.2:

ihrivnac:
The last patch (p02) includes a fix needed when using the FTFP_BERT physics list with biasing option


ihrivnac:
@sawenzel , the problem  was reported by Gustavo, observed with AliRoot installed with Root 6 after switching to Geant4 10.7.

sawenzel (approved):

PR #3098, opened 2021-06-15
Do not build separate re2 for grpc

awegrzyn:
This is to use our `re2` for `grpc` build. Currently it builds its own version.


teo (approved):

awegrzyn:
@ktf can this be merged?
Calling @davidrohr for failing fullCI

PR #3103, opened 2021-06-16
Use alibuild-generate-module for as many recipes as possible

TimoWilken:
Needs `--root-env` and `--extra` extensions for `alibuild-generate-module`.

The following are not converted yet:

    alice-grid-utils.sh     alidpg.sh           alien-root-legacy.sh
    alien-workqueue.sh      aligenerators.sh    aligenmc.sh
    aliroot-guntest.sh      aliroot-ocdb.sh     alo-aliroot.sh
    alo-o2.sh               ampt.sh             apmon-cpp.sh
    asiofi.sh               asio.sh             aurora.sh
    autotools.sh            cctools.sh          codingguidelines.sh
    cpprestsdk.sh           crmc.sh             datadistribution.sh
    dimrpcparallel.sh       epos.sh             fairmq.sh
    fastjet.sh              fonll.sh            geant3.sh
    geant3_vmc-examples.sh  geantv.sh           geekbench.sh
    generators.sh           glog.sh             golang.sh
    googlebenchmark.sh      hdf5.sh             hepmc3.sh
    hepmc.sh                hijing.sh           jemalloc.sh
    jetscape.sh             jewel.sh            jq.sh
    kfparticle.sh           lcov.sh             lhapdf5.sh
    lhapdf-pdfsets.sh       lhapdf.sh           libffi.sh
    libinfologger.sh        libjalieno2.sh      libjalienws.sh
    librdkafka.sh           libtirpc.sh         libwebsockets.sh
    localccdb.sh            mesos-workqueue.sh  msgpack.sh
    o2codechecker.sh        objcmp.sh           odc.sh
    oniguruma.sh            parquet-cpp.sh      rapidjson.sh
    rivet-hi.sh             rivet.sh            rpm-test.sh
    sacrifice.sh            simulation.sh       sodium.sh
    starlight.sh            template-recipe.sh  thepeg.sh
    therminator2.sh         thrift.sh           treelite.sh

### DIFF
--- a/abseil.sh
+++ b/abseil.sh
@@ -23,11 +23,8 @@ make ${JOBS:+-j$JOBS} install
 
 # Modulefile
 MODULEDIR="$INSTALLROOT/etc/modulefiles"
-MODULEFILE="$MODULEDIR/$PKGNAME"
 mkdir -p "$MODULEDIR"
-alibuild-generate-module > "$MODULEFILE"
-cat >> "$MODULEFILE" <<EoF
-
+alibuild-generate-module --extra > "$MODULEDIR/$PKGNAME" <<EoF
 # Our environment
 set ABSEIL_ROOT \$::env(BASEDIR)/$PKGNAME/\$version
 prepend-path PATH \$ABSEIL_ROOT/bin
--- a/aegis.sh
+++ b/aegis.sh
@@ -8,6 +8,7 @@ build_requires:
   - CMake
   - hijing
   - "Xcode:(osx.*)"
+  - alibuild-recipe-tools
 source: https://github.com/AliceO2Group/AEGIS.git
 prepend_path:
   LD_LIBRARY_PATH: "$AEGIS_ROOT/lib"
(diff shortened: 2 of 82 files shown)


adriansev:
@TimoWilken i see `$`  escaped and not-escaped in the same type of setenv declarations.. is this ok or irrelevant ?

ktf:
thank you for the commendable work, however I would do baby steps and simply do python-modules and alien-runtime (which should account for most of our alienv performance problems), without changing alibuild-generate-module, especially given all the tests failed. We can do the rest in separate steps, as we modify the recipes.

Barthelemy:
I agree with Giulio. I am a bit scared by this large change. From my experience, it requires more effort than just replacing the modules declaration blindly with the call to `alibuild-generate-module` otherwise it breaks things at runtime.

ktf:
Also, I would also drop the --extra feature and simply use cat, so that we do not need to rebuild everything due to the changes in the utility.

teo:
Echoing what @ktf and @Barthelemy said, I'm uncomfortable with this change at least as far as AliECS components are concerned. We already tried using module autogeneration with these Go-based components, we encountered trouble and had to revert.
If and when we revisit this, it should IMO be done recipe by recipe.

ktf:
Closing this to avoid full rebuilds.

PR #3104, opened 2021-06-17
Drop non-existing folders from PATH

ktf:


awegrzyn (dismissed):

ktf:
Not really tested. Merging.

PR #3108, opened 2021-06-17
Use alibuild-generate-module for Python-modules and AliEn-Runtime

TimoWilken:
Replaces https://github.com/alisw/alidist/pull/3103.

### DIFF
--- a/alien-runtime.sh
+++ b/alien-runtime.sh
@@ -11,6 +11,7 @@ build_requires:
  - AliEn-CAs
  - ApMon-CPP
  - UUID
+ - alibuild-recipe-tools
 env:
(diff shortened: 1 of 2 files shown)


TimoWilken:
This doesn't require changes to `alibuild-generate-module`.

ktf (approved):

ktf:
@TimoWilken "Pending — Started Thu 14:49 CET @ mesosci/ci_alisw_slc8-gpu-builder_". Can you check if it is normal?

TimoWilken:
It is still building, and it seems to be rebuilding a lot of stuff, including GCC. Now it's on O2 at 1211/3451. I'm not sure what caused such a big rebuild -- that builder also tested the big PR (3013), but it should've kept the previous artifacts...

PR #3111, opened 2021-06-18
Use correct default when checking JAliEn

ktf:


ktf:
@TimoWilken can you have a look? This should solve one of the issues with the JAliEn PRs...

PR #3112, opened 2021-06-18
Updated ONNXRuntime recipe

saganatt:
Ciao @ktf ,

I updated ONNX so as to integrate it with O2. This PR contains only changes in onnxruntime recipe.
Afterwards, when it will be ok, I would submit also PRs with o2.sh and O2 cmakes update.

As I needed to modify ONNXRuntime cmakes more, we need to use my ONNXRuntime repo.

### DIFF
--- a/onnxruntime.sh
+++ b/onnxruntime.sh
@@ -1,6 +1,6 @@
-package: onnxruntime
+package: ONNXRuntime
 version: "1.7.2"
-tag: c073c88ef
+tag: 4531d55f6
 source: https://github.com/saganatt/onnxruntime.git
 requires:
   - protobuf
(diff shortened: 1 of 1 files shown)


onnxruntime.sh
```diff
@@ -1,6 +1,6 @@
-package: onnxruntime
+package: ONNXRuntime
 version: "1.7.2"
-tag: c073c88ef
+tag: 4531d55f6
 source: https://github.com/saganatt/onnxruntime.git
```
ktf, line 4: Please let's move this to an official alisw repository using the tagging procedures outlined in:

https://github.com/alisw/alidist#guidelines-for-handling-externals-sources

Can you provide me the original github repo to fork it in alisw?

ktf (changes requested):
Let's move the sources to use a repository under alisw.

onnxruntime.sh
saganatt: The original repo is here: https://github.com/microsoft/onnxruntime

saganatt: I based my changes on tag 1.7.2. I can prepare the forked repo myself as well, if you point me to the fork.

ktf: https://github.com/alisw/onnxruntime/tree/alice/v1.7.2 is the branch you should use.

saganatt: Ok, I need one more advice: how I can contribute? It doesn't allow me to push, neither I can make a PR to it...?

ktf: You need to fork and create a PR. How does it fail?

saganatt: I pushed now. Turned out I needed to delete my previous fork, and re-fork it.

ktf:
Not really tested. Merging.

PR #3113, opened 2021-06-18
Updated O2 recipe to use ONNXRuntime

saganatt:
Ciao @ktf 

this is a follow-up to the onnxruntime PRs.
This is requires merging also https://github.com/AliceO2Group/AliceO2/pull/6459

### DIFF
--- a/o2.sh
+++ b/o2.sh
@@ -26,6 +26,7 @@ requires:
   - KFParticle
   - VecGeom
   - FFTW3
+  - ONNXRuntime
 build_requires:
   - GMP
   - MPFR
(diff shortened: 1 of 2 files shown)


saganatt:
Ciao @ktf,
could you merge this PR? The build error from MacOS comes from Python build (not related), the second one does not have any log.
The O2 PR fails because of missing the changes introduced in this alidist PR.

ktf:
Given we now have O2Physics, wouldn't it make sense to make it a dependency there?

saganatt:
O.K., closing, new PR here: #3197

PR #3118, opened 2021-06-21
SYSTEM_VERSION_COMPAT is not needed anymore

pzhristov:


ktf:
@TimoWilken I think something broke for Centos 8. Can you check?

PR #3125, opened 2021-06-25
Drop secret from tpcfecutils.sh

ktf:


ktf:
Having non managed secrets in recipes will break other builds. Dropping it.

vascobarroso:
Pinging @sk3x to review. 
I have the feeling we are going in circles, wasn't this the original problem ?

ktf:
I think the password protected repository is just the submodule that was removed. I just tried on a machine where I do not have credentials and it seems to work.

ktf:
@TimoWilken I just saw this PR going from "all green" to "all yellow", just by reloading the page. Are you sure there is no race condition it the checkers that marks all the checks as "pending" even if only one failed?

ktf:
Now everything is green again. Maybe a github glitch..

ktf:
macOS unrelated and unneeded. Indeed dropping the secret had no issues, so I assume it's fine.

ktf:
ok. apparently not the case. Any objections to forking the repository to AlicePrivateExternals gitlab group? For that one everyone in ALICE, including our CI, has access (but not the rest of the world).

ktf:
@vascobarroso @sk3x

sk3x:
No objections.
Can you set it up to automatically stay in sync?

ktf:
I have to check how to do that. I guess we could setup a gitlab action.

ktf:
can you add my gitlab account to your repository? There is apparently a way to automatically mirror things in gitlab but I need rights.

sk3x:
I just added @eulisse, was that correct?

ktf:
yes.

ktf:
mirroring every 3h should be in place now. Please let me know if you have troubles.

sk3x:
Sounds good, thanks a lot.

PR #3130, opened 2021-06-28
Update o2suite.sh

ktf:


vascobarroso:
@ktf @sk3x we plan to release FLP Suite v0.20.0 later today, it would be nice to have this in place. Can you check if the errors are unrelated and if so force merge ?

sk3x:
Doesn't look like this is directly related the the TpcFecUtils, looks ok to me. @ktf ?

ktf:
No, indeed.

awegrzyn:
On my new set up I get:
```
Cloning into bare repository '/root/sw/MIRROR/tpcfecutils'...
fatal: could not read Username for 'https://gitlab.cern.ch': No such device or address
ERROR: Error while updating reference repos https://gitlab.cern.ch/ALICEPrivateExternals/alice-tpc-fec-utils.git.
```
How do I need to progress?

ktf:
Did you configure your gitlab credentials? I just checked and you are owner of the repository so you should have access.

ktf:
For the record, as discussed privately, instructions on how to setup your credential cache can be found at:

http://alisw.github.io/git-advanced/#setup-a-git-credentials-cache

ktf:
@sk3x do we really need to keep the repository private? From previous discussions, I understood the sensitive bits were actually in the submodule only.

sk3x:
There's no real sensitive bits, it's more a question of going through all the hassle of universities and software licensing questions for something that was never planned to be public and that we simply need deployed for debugging (and unfortunately depends on ReadoutCard and LLA).

We made an effort to remove the because it was claimed that the build system had trouble with authentication. 

Now, I thought the purpose of forking it to ALICEPrivateExternals was that this is something know to be working where ALICE-internal everybody has access?

ktf:
@sk3x, correct, and indeed procedures and instructions are in place to make sure that only ALICE people can access to the repository and so that the CI can handle it gracefully (modulo some changes which @TimoWilken had to introduce to support them in the new way of building for Centos Stream 8). It's just a pain having to deal with private packages, even for experienced developers like @awegrzyn, so I was hoping that we could simply stick the usual O2 copyright statement in front which the rest of our software uses and be done with it.

awegrzyn:
I understand reasoning of @sk3x although if possible I'd second request of having `TpcFecUtils` repo public. Currently this is only package in FLP Suite that requires credentials. As it fairly easy set it up in our release servers (like @ktf did) it affects all FLP Suite developers as they will need to set up their credential stores and Docker images.

PR #3131, opened 2021-06-28
Bump fmt to 8.0.0

ktf:


ktf:
Does not work out of the box. Closing until we have time to look into this.

PR #3132, opened 2021-06-28
datadist: bump to v0.10.20

ironMann:
spdlog fmt 7 compatibility

### DIFF
--- a/datadistribution.sh
+++ b/datadistribution.sh
@@ -1,6 +1,6 @@
 package: DataDistribution
 version: "%(tag_basename)s"
-tag: v0.10.18
+tag: v0.10.20
 requires:
   - "GCC-Toolchain:(?!osx)"
   - boost


ironMann:
@kft @davidrohr Lets try this before we bump fmt to 8.0. There are API changes, and compiler incompatibilities in fmt which might affect ability to build the FLP packages.

ktf:
Currently broken, merging.

davidrohr:
For reference: I have checked in the FullCI that DD 10.20 works with the old fmt 7.
It fails with fmt 8.0.0.

ironMann:
@davidrohr @ktf The blocker for DD atm is spdlog which fails to build with fmt 8.0.0. https://github.com/gabime/spdlog/issues/1975

I guess that will be resolved soon. But, it looks like fmt broke a lot of projects with this release. Maybe we should wait for a point release or something....

davidrohr:
Fine with me, I think it is not urgent to bump.

PR #3144, opened 2021-07-02
Adding scons to python modules

mfasDa:
scons is used in particular in loop generators
in order to compile process libraries on-the-fly.
Therefore it must be available on grid sites.

### DIFF
--- a/python-modules-list.sh
+++ b/python-modules-list.sh
@@ -12,6 +12,7 @@ env:
     PyYAML==5.1
     uproot==3.4.18
     psutil==5.8.0
+    scons==4.1.0
   PIP36_REQUIREMENTS: |
     numpy==1.16.2
     scipy==1.2.1


mfasDa:
@ktf @TimoWilken CI errors unrelated - can this be merged?

ktf:
Does this really need to be part of the Python-modules package? Maybe having a different, self contained, scons package is actually a better idea.

mfasDa:
I have no strong opinion here, but what is your problem / what are the disadvantages having it as part of python-modules? scons should be quite lightweight. Python-modules is anyhow already in the dependency chain of aligenerators

ktf:
scons itself is fine, that would mean that every generator which uses it at build time will require python-modules (which is hefty). Especially if it is used only at build time, it might make sense to separate it. Anyways, for now it's fine.

mfasDa:
First of all it is not only needed at build time, but when running NLO the generators create so-called process libraries on-the-fly, which must also be compiled on the fly, and Openloops-based processes use scons for the compilation of the process libraries. Second, python-modules comes in any case already in as implicit dependency by other packages of aligenerators  via the chain fastjet -> boost -> python-modules, and nearly all pp generators need fastjet for kinematic cuts in case of running pp -> jj processes.

PR #3148, opened 2021-07-05
Fix autotools build on Ubuntu 20.04

TimoWilken:
Texinfo's default encoding has changed, but a file in autotools still uses the old default, so we need to declare the encoding explicitly.

Source: https://github.com/void-linux/void-packages/issues/21163#issuecomment-616497700

### DIFF
--- a/autotools.sh
+++ b/autotools.sh
@@ -49,6 +49,12 @@ fi
 
 # m4 -- requires: nothing special
 pushd m4*
+  # texinfo uses utf-8 by default, but doc/m4.text is still iso-8859-1.
+  # MacOS sed only understands the command with the linebreaks like this.
+  sed -i.bak '1i\
+@documentencoding ISO-8859-1
+' doc/m4.texi
+  rm -f doc/m4.texi.bak
   $USE_AUTORECONF && autoreconf -ivf
   ./configure --disable-dependency-tracking --prefix $INSTALLROOT
   make ${JOBS+-j $JOBS}


TimoWilken:
Updated to fix build error on MacOS.

autotools.sh
```diff
@@ -49,6 +49,11 @@ fi
 
 # m4 -- requires: nothing special
 pushd m4*
+  # texinfo uses utf-8 by default, but doc/m4.text is still iso-8859-1.
+  # MacOS sed only understands the command with the linebreaks like this.
+  sed -i '1i\
```
ktf, line 54: I am not sure this syntax is correct on macOS.

TimoWilken:
```bash
$ seq 5 | sed '1i\
foo
'
```
gives the right result for me on one of the CI Macs (`foo`, then the numbers 1-5).

(The single quotes are very important -- they can't be double quotes here.)

If that doesn't work for you, what error do you get? Or what should the right syntax be?

ktf: I am worried about `-i`.

TimoWilken: Oh, I see. Fixed.

ktf (approved):

TimoWilken:
Almost everything is green, and even in the failed checks, the autotools build succeeded before the usual problems kicked in. Can we merge this?

ktf:
Are we ready to handle the rebuild of our software stack on all the platforms we support?

TimoWilken:
As the alidist commit doesn't matter for the hash calculation, only recipe contents, I've just set all the DailyO2* builds going using this PR branch. Then we could merge once those are done, so that we'd have prebuilt tarballs available as soon as this change is picked up.

ktf:
Yes, good idea. Sounds good to me. Go ahead and merge this when you think the caching is ready.

TimoWilken:
Will do, thanks.

PR #3150, opened 2021-07-06
Adding gosam loop tools

mfasDa:


maireiphc:
Hello @mfasDa,

 can you just give here a bit of context to this new introduction. This is about the AliGenerators world, likely a new dependency/feature open with Herwig generator and likely needed for PWG-JE, as far as I can judge/assume. 
But the offline team may want to know a bit more about it.

. You seem to pick up the official source from GoSam (http://gosam.hepforge.org) in the new recipe gosam.sh. Ok.

madgraph.sh
```diff
@@ -5,12 +5,25 @@ source: https://github.com/alisw/MadGraph
...
+ - lhapdf
 build_requires:
   - alibuild-recipe-tools
 ---
 #!/bin/bash -e
 
-rsync -a --exclude='**/.git' --delete --delete-excluded $SOURCEDIR/ $INSTALLROOT/
+rsync -a --delete --exclude '**/.git' --delete-excluded $SOURCEDIR/ ./
+
+echo "generate e+ e- > u u~" >> test.mg
```
maireiphc, line 18: Hello @mfasDa,

 I am not sure to guess what you are after with these transient lines, you echo sthg into a file and run madgraph for what seems a local test. You want to have a trace somewhere in the build logs ?

You may comment in the bash recipe about your intent.

mfasDa: See recommendation in https://herwig.hepforge.org/tutorials/installation/prerequisite.html (section MadGraph)

mfasDa:
@maireiphc This is to make Herwig+MadGraph (or MadGraph) alone working (MC@NLO), where according to the instruction above GoSam, which is a toolbox for loop providers, is needed.

maireiphc (dismissed):

mfasDa:
There will be one more fix needed in the MadGraph repository + during the MadGraph installation we need to install at least 2 dependencies (collier and oneloop) because they are otherwise installed on-the-fly in the install directory, otherwise I managed to run MadGraph at least locally. Not sure how to deal with the dependency on latex of the gosam installer.

mfasDa:
I think there were recently similar messages on the analysis task force mailing list concerning the same python module. Any idea how to solve this?

```
madgraph requires the six module. The easiest way to install it is to run "pip3 install six --user"
```

Also with madgraph there is another issue with a configuration file in the package keeping track of the tools, in which after the installations all the paths are wrong (local build directories). Any idea how we can fix the paths at install time? They are crucial for madgraph to find the libraries.

ktf:
@TimoWilken could you check why the generators test fails?

TimoWilken:
That builder was restarted, I suspect this check got caught in the crossfire (maybe the CI builder managed to fire off a `report-pr-errors` call before it was killed for good). I've marked the check as yellow.

ktf:
@TimoWilken the generators test does not report any log, could you take a look?

ktf:
Fails to build. It also seems to download stuff on its own. Can it be avoided?

ktf (changes requested):
Fails to build. It also seems to download stuff on its own. Can it be avoided?

ktf:
The error is:

```
Finished building on alientest06 at 2021-10-05-11:54:06
Built commit a7fcf58cfab7afbf617651149b65bd11fd9b86ee from https://github.com/alisw/alidist/pull/3150
To log into the build machine, use:
    aurora task ssh -l root build/mesosci/prod/ci_alisw_slc7-builder_latest/0
The following files (oldest first) are present in the log:
- sw/BUILD/GoSam-latest/log
## Begin sw/BUILD/GoSam-latest/log
++ unset HTTP_PROXY
++ wget http://gosam.hepforge.org/gosam_installer.py
--2021-10-05 09:54:02--  http://gosam.hepforge.org/gosam_installer.py
Resolving gosam.hepforge.org (gosam.hepforge.org)... 129.234.186.186
Connecting to gosam.hepforge.org (gosam.hepforge.org)|129.234.186.186|:80... connected.
HTTP request sent, awaiting response... 302 Found
Location: https://gosam.hepforge.org/gosam_installer.py [following]
--2021-10-05 09:54:03--  https://gosam.hepforge.org/gosam_installer.py
Connecting to gosam.hepforge.org (gosam.hepforge.org)|129.234.186.186|:443... connected.
ERROR: cannot verify gosam.hepforge.org's certificate, issued by '/C=US/O=Let\'s Encrypt/CN=R3':
  Issued certificate has expired.
To connect to gosam.hepforge.org insecurely, use `--no-check-certificate'.
## End sw/BUILD/GoSam-latest/log
```

Please move to use a proper git source or at least an https one with a valid certificate. Please close this if you do not plan to  fix anytime soon.

PR #3151, opened 2021-07-08
Bump qualitycontrol to v1.21.1

Barthelemy:


vascobarroso (approved):

knopers8:
@ktf I think this is why:
```
unknown location(0): [4;31;49mfatal error: in "DataSamplingSimpleFlow": boost::filesystem::filesystem_error: boost::filesystem::status: Function not implemented: "/sw/ubuntu1804_x86-64/O2/build_O2_alidist-ubuntu1804-ubuntu1804_x86-64-local1/share/tests/test_DataSampling.json"[0;39;49m
```
My guess that this function is used in Configuration (invoked by DataSampling), thus @awegrzyn might be interested.
@ktf I suppose that boost::filesystem is not fully implemented for Ubuntu, but it is for other systems.

ktf:
> @ktf I suppose that boost::filesystem is not fully implemented for Ubuntu, but it is for other systems.

Agreed,  however I thought we moved to std::filesystem a while ago though.

knopers8:
Hm, indeed, it was done in Configuration:
https://github.com/AliceO2Group/Configuration/commit/2617e8c9ee2a2a02040340b71453d4d642e9b6ec
..but v2.6.2 was not bumped in alidist. I don't know there is any reason not to do it. I see that @awegrzyn is away this week.

I guess we could just try with a bump PR first and according to the outcome decide what we do with it.

PR #3154, opened 2021-07-09
Fix ROOT compilation on OSX

TimoWilken:
This undoes the change in commit 7748331.

I don't know why this fixes it, but it does.

I've tested it by building O2Suite for PR 6613 on alibuildmac03.

### DIFF
--- a/root.sh
+++ b/root.sh
@@ -23,6 +23,7 @@ build_requires:
   - alibuild-recipe-tools
 env:
   ROOTSYS: "$ROOT_ROOT"
+  SYSTEM_VERSION_COMPAT: "1"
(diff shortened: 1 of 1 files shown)


ktf:
Merging this as currently broken otherwise.

pzhristov:
@ktf @TimoWilken This is wrong - we better switch off the non-updatable builder.

ihrivnac:
@pzhristov , after upgrade of my Mac to BigSur last week I was getting exactly the same errors as reported in https://alice.its.cern.ch/jira/browse/O2-2419 (failures in ctest) and this were gone after applying the changes from this PR.

PR #3155, opened 2021-07-12
test to bump fastjet to 3.4.0 and fjcontrib to 1.045

bhohlweger:


adriansev:
quite strange the output of the failed O2/alidist test (while AliRoot/alidist is ok) .. 
there is a need of infrastructure expert to take a look, as the fastjet is disabled in that test and it has no relation to this PR
the actual error seen in the full log is
```
unknown location(0): [4;31;49mfatal error: in "TestOptionsRetriever": boost::filesystem::filesystem_error: boost::filesystem::status: Function not implemented: "/tmp/alice_o2_configuration_test_file.ini"[0;39;49m
/sw/SOURCES/O2/build_O2_alidist-ubuntu1804-ubuntu1804_x86-64/0/Framework/Core/test/test_ConfigurationOptionsRetriever.cxx(26): [1;36;49mlast checkpoint: "TestOptionsRetriever" test entry[0;39;49m
```

adriansev (approved):
tests green, the failed one is not related

ktf:
We are investigating the issue. Somehow some old boost from the system is being picked up and Configuration still uses boost::filesystem rather than std::filesystem.

bhohlweger:
HI everyone, 

is there a solution around the corner for this test? @ktf

ktf:
The solution has been committed, I have restarted the test.

I guess in any case the PB should bless the change, no?

ktf:
@TimoWilken looks like the ubuntu test is still not being picked up. did something go wrong?

bhohlweger:
@ktf Hi this is only a test tag in order to check the consistency with the fastjet version in the current AliPhysics. Before we request to merge it into default-user-root6 or so we will present at PB afaik.

ktf:
I missed it was for the JALIEN default...Merging.

bhohlweger:
@ktf hi quick follow up, would it be possible to also add jalien for aliphysics to the production of daily tags? It seems this was deprecated sometime in april!

PR #3162, opened 2021-07-15
Bump qualitycontrol to v1.22.1

Barthelemy:


knopers8 (approved):

ktf:
Genuine issue?

knopers8:
Surely not related, but somewhat genuine (in testCcdbDatabase):
```
Fatal error: glibc detected an invalid stdio handle
```
Though I don't understand it where it may come from and what that actually means. I will have to give it a closer look.

knopers8:
As it it unrelated and the issue does not even occur after each build, I would propose to merge it.

In the meanwhile I will add some more debug info (https://github.com/AliceO2Group/QualityControl/pull/769) which hopefully will allow us to understand the problem. So far I know that there is a `std::length_error()` thrown when trying to create a substring from JSON output from CCDB.

PR #3165, opened 2021-07-15
Don't use system flatbuffers, since that breaks FairMQ if it was not …

davidrohr:
…compiled with -fPIC

### DIFF
--- a/flatbuffers.sh
+++ b/flatbuffers.sh
@@ -6,9 +6,6 @@ requires:
 build_requires:
  - CMake
  - "GCC-Toolchain:(?!osx)"
-prefer_system: "(?!slc5)"
-prefer_system_check: |
-  which flatc && printf "#include \"flatbuffers/flatbuffers.h\"\nint main(){}" | c++ -I$(brew --prefix flatbuffers)/include -xc++ -std=c++11 - -o /dev/null
 ---
 cmake $SOURCEDIR                          \
       -G "Unix Makefiles"                 \


dennisklein:
Just for my curiosity, which CentOS 8 Stream system package are you using? My Centos Stream 8 (with EPEL 8 enabled) does not even have one? (I am asking because I use the `flatbuffers-devel` package on Fedora since at least 2 years with FairMQ and there are no problems (and usually the Fedora package is the one that makes it to EPEL).)

davidrohr:
Not CentOS stream but CentOS 8.
Just the version that was installed on the EPNs.

dennisklein:
Can you point me to the docs how the EPN image is set up? Because also my local centos8 (no stream) with epel8 does not seem to have a flatbuffers package?!

davidrohr:
@JohannesLb can probably reply better, I am not in charge of the EPNs

JohannesLb:
We have several cern repositories added to the EPN. Apparently the EOS one has flatbuffers as well and it was installed from there. 
> cat /etc/yum.repos.d/eos.repo \
[eos] \
baseurl = https://storage-ci.web.cern.ch/storage-ci/eos/citrine/tag/el-8/x86_64/ \
gpgcheck = 0 \
name = Cern EOS repo
> cat /etc/yum.repos.d/eos-deps.repo \ 
[eos-deps] \
baseurl = https://storage-ci.web.cern.ch/storage-ci/eos/citrine-depend/el-8/x86_64/ \
gpgcheck = 0 \
name = Cern EOS deps repo

> Available Packages
Name         : flatbuffers
Version      : 1.12.0
Release      : 1.el8
Architecture : x86_64
Size         : 174 k
Source       : flatbuffers-1.12.0-1.el8.src.rpm
Repository   : eos-deps
Summary      : Memory Efficient Serialization Library from Google
URL          : http://google.github.io/flatbuffers/
License      : Apache-2.0
Description  : FlatBuffers is a serialization library for games and other memory constrained apps.
             : FlatBuffers allows you to directly access serialized data without unpacking/parsing
             : it first, while still having great forwards/backwards compatibility.

dennisklein:
Thx, the difference of the eos-deps package to the fedora package is, that it includes `libflatbuffers.a` (which is likely not `-fPIC`) which is exported as CMake target `flatbuffers::flatbuffers` which is picked up by FairMQ first (if it exists). I will change the logic in FairMQ to try to pick up `flatbuffer::flatbuffers_shared` first in the next version.

PR #3173, opened 2021-07-20
Fix ONNXRuntime's dependencies

omasanori:
Sometimes ONNXRuntime fails to build locally without this change, but I am not sure whether this is a proper fix or not.
Any opinions? @ktf @saganatt

### DIFF
--- a/onnxruntime.sh
+++ b/onnxruntime.sh
@@ -5,6 +5,7 @@ source: https://github.com/alisw/onnxruntime.git
 requires:
   - protobuf
   - re2
+  - Python-modules
 build_requires:
   - CMake
 ---
--- a/onnxruntime.sh
+++ b/onnxruntime.sh
@@ -5,7 +5,7 @@ source: https://github.com/alisw/onnxruntime.git
 requires:
   - protobuf
   - re2
-  - Python-modules
+  - Python
 build_requires:
   - CMake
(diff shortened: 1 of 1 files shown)


saganatt:
Thanks @omasanori for checking - hm, ONNXRuntime in general requires some Python, so perhaps it could fail because of Python-modules.
For me locally it always worked and the CI build completed as well.
I have also a PR with joining ONNXRuntime to O2, and builds are ok. Not sure, though, how much the builds test, maybe they don't build from scratch (@ktf ?)

onnxruntime.sh
```diff
@@ -5,6 +5,7 @@ source: https://github.com/alisw/onnxruntime.git
 requires:
   - protobuf
   - re2
+  - Python-modules
```
omasanori, line 8: Actually, `Python-modules` can be reduced to `Python`, because the intention is to let CMake find the proper Python interpreter. The packages `Python-modules` installs are unnecessary to ONNXRuntime itself.

ktf, line 8:
```suggestion
  - Python
```

ktf: I changed it. Notice you can propose changes directly to the PR by clicking on the +/- sign.

ktf (approved):

PR #3185, opened 2021-07-23
FFTW: disabling AVX2

matthias-kleiner:
This might fix the problem `*** Illegal instruction Register dump:….` when using FFTW on alientest* as discussed in the [PR](https://github.com/AliceO2Group/AliceO2/pull/6427).

### DIFF
--- a/fftw3.sh
+++ b/fftw3.sh
@@ -13,8 +13,7 @@ cmake $SOURCEDIR                                   \
       -DCMAKE_INSTALL_PREFIX:PATH="${INSTALLROOT}" \
       -DCMAKE_INSTALL_LIBDIR:PATH="lib"            \
       -DENABLE_FLOAT=ON                            \
-      -DENABLE_AVX=ON                              \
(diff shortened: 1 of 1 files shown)


davidrohr:
I am okay with this for moving ahead now, but I think in general it is quite stupid to unconditionally disable such optimizations. The majority of CPUs today supports AVX2, and by unconditionally disabling it we make the code slower just to support some ancient machines.
I would prefer if we build the software with proper optimization, and also with `-march=native` on our productions servers, e.g. on the EPNs. It really makes no sense to waste performance in this way.
But ok, this is a more general discussion.
@ktf @TimoWilken @ironMann @ihrivnac : Should we perhaps discuss this in one of the WP3 meetings?

ktf:
For "native" we can simply add it to a special "EPN" default which we use to build the release for the cluster.

For anything which goes on the grid we should followup with @costing and Latchezar whether we still have any sizeable non AVX2 resource.

PR #3196, opened 2021-07-28
Use protocolbuffers/protobuf for protobuf

TimoWilken:
google/protobuf seems to have been replaced(?!) with a Dart protobuf library instead of protobuf itself. CI currently fails to build protobuf because the tags we need aren't in the new repo.

### DIFF
--- a/grpc.sh
+++ b/grpc.sh
@@ -1,6 +1,6 @@
 package: grpc
 version: "%(tag_basename)s"
-tag:  v1.34.0-alice1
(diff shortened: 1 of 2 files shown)


TimoWilken:
The problem seems to have been resolved (the redirect from google/protobuf to protocolbuffers/protobuf is back), but it might still be a good idea to merge this.

teo:
@TimoWilken since you're changing the Protobuf upstream, might this also be a good time to bump to the latest tag?

PR #3204, opened 2021-08-02
Use v5-09-57h (Root v6-24-02)

pzhristov:


TimoWilken:
Looks good, but is it necessary to change `defaults-jalien.sh`? That default isn't used in CI, and it's on an older branch (-56 vs -57).

pzhristov:
> Looks good, but is it necessary to change `defaults-jalien.sh`? That default isn't used in CI, and it's on an older branch (-56 vs -57).

I do not want to have different versions of AliRoot/AliPhysics in the defaults. Anyway, defaults-jalien is not used much, so the change is just for convenience.

adriansev (approved):

PR #3205, opened 2021-08-02
Build O2Physics for O2Suite as well

TimoWilken:
This makes sure that changes to O2 don't break O2Physics, once it's factored out into a separate repository.


davidrohr (approved):
Renamed the title, since this is for o2suite not only for the full CI.
In general it makes sense to me, we could think about disabling it in the o2-dataflow defaults.

PR #3214, opened 2021-08-05
use xrootd 5 as default

adriansev:
Since the root was updated to 6.24 we can use the latest xrootd 5.x


adriansev:
@ktf @pzhristov  the failed test was related to a timeout (but now i see that the logs are gone) .. can you review this please?

PR #3221, opened 2021-08-13
Bump FairRoot to v18.4.2

ktf:


ktf:
@dennisklein I think at some point we discussed disabling MCConfigurator completely, since we do not use (AFAICT, @sawenzel to confirm) that component. Is that now possible? That would allow us to drop the yaml dependency, no?

dennisklein:
> @dennisklein I think at some point we discussed disabling MCConfigurator completely, since we do not use (AFAICT, @sawenzel to confirm) that component.

We have more granular component build switches (and a change in the logic from *build-if-dep-found* to *build-if-switched-on-and-fail-fast-otherwise*) on the todo list, but with low priority.

> Is that now possible? That would allow us to drop the yaml dependency, no?

Have you tried `-DCMAKE_DISABLE_FIND_PACKAGE_yaml-cpp=ON`?

ktf:
Thanks. Trying now.

pzhristov:
I tested it locally and it compiled. The yaml-cpp issues is solved.

pzhristov:
The problem on macOS is caused by the use of Python 3.6.10 on this machine. @MichaelLettrich Do you know if there is any particular reason to require such obsolete version? We should move to python@3.9 from brew.

PR #3225, opened 2021-08-16
WIP: use system provided openssl

adriansev:
Given the usage of modern xrootd and root, we can/should use current system-provided openssl.

### DIFF
--- a/alien-runtime.sh
+++ b/alien-runtime.sh
@@ -6,7 +6,7 @@ requires:
 build_requires:
  - zlib
  - libxml2
- - "OpenSSL:(?!osx)"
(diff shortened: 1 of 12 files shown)


adriansev:
not ack by @costing

costing:
I think this should be optional and by default the fixed openssl version should be used. Otherwise the cvmfs deployed binaries will fail to run in different sites (or at a later time when we want to use them again). If users want to use their system-provided openssl then it's their business to debug the compilation.

If the new xrootd, root and the rest can be compiled with a newer openssl, then let's bump the base version instead.

PR #2937, opened 2021-03-31
Move to Apache Arrow 4.0.1

aalkin:

PR #3230, opened 2021-08-17
Bump Arrow to 4.0.1 and Clang to 12.0.1

pzhristov:
Replaces https://github.com/alisw/alidist/pull/2937


pzhristov:
We need updated version of o2codechecker for successful checks.

pzhristov:
@ktf @aalkin @TimoWilken @davidrohr The problems we see are caused by a wrong update in ccdb-test that hopefuly will be fixed today. I propose to merge this PR.

ktf:
In [O2/alidist](https://ali-ci.cern.ch/alice-build-logs/alisw/alidist/3230/b88642c3f70ce0cbe5c4a69e9a2ba063dbcb1c84/build_O2_alidist/pretty.html) I see:

```
## sw/BUILD/arrow-latest/log
--
-- clang-format not found
-- Could NOT find ClangTools (missing: CLANG_FORMAT_BIN CLANG_TIDY_BIN) 
-- infer not found
fatal: not a git repository (or any parent up to mount point /)
Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
-- Found Python3: /sw/slc7_x86-64/Python/v3.6.10-127/bin/python3.6 (found version "3.6.10") found components: Interpreter 
-- Found cpplint executable at /sw/BUILD/0bc021848256bb8a9dec4e9de9e394355efae341/arrow/src_tmp/cpp/build-support/cpplint.py
--
```

is that expected?

arrow.sh
```diff
@@ -101,8 +104,9 @@ cmake ./src_tmp/cpp
       -DARROW_TENSORFLOW=ON                                                                         \
       -DARROW_GANDIVA=ON                                                                            \
       -DARROW_COMPUTE=ON                                                                            \
+      -DARROW_BUILD_STATIC=OFF                                                                      \
       -DCMAKE_INSTALL_RPATH_USE_LINK_PATH=ON                                                        \
-      -DCLANG_EXECUTABLE=${CLANG_ROOT}/bin-safe/clang
```
davidrohr, line 105: just wondering, was there a particular reason we used bin-safe/clang here before? Perhaps that was to avoid a collision with the system clang? If so, perhaps we should keep it as is?

pzhristov: We switched off the compilation of Clang on MacOSX and used the system one on both osx_x86-64 and osx_arm64, while for the platforms where we compile Clang, we use bin-safe. The reason is that Apple has its own patches and the git version does not compile.

davidrohr: ok, and are we sure a system clang with a lower version will not be picked up for arrow?

pzhristov: Concerning the issue with "fatal: not a git repository..." I suspect something is wrong in the local directory. Did we run out of space there?

pzhristov:
There is something else of the O2/alidist builder:
CMake Error at cmake_modules/FindOpenSSLAlt.cmake:33 (message):
  The OpenSSL must be greater than or equal to 1.1.0
  src/gandiva/CMakeLists.txt:43 (find_package)

Apparently on the other systems arrow finds the system openssl which is v1.1.X, while our own version is 1.0.2o. I will try to switch off openssl support from arrow since we do not use it.

arrow.sh
pzhristov: It will fail if the version is below 12.0 - there is a check in arrow.

pzhristov:
Gandiva requires openssl >= 1.1.0 and this cannot be switched off, so this PR cannot be merged without bump of openssl.

arrow.sh
davidrohr: hm, ok, but what with a system installation of clang 13 / clang 14 dev versions? (which I have). Probably arrow will not pick them up anyway I guess, since they usually limit the supported clang version. ok, anyway, if this breaks something, we can still revert to specifying the alibuild clang.

davidrohr (approved):

pzhristov:
Openssl >=1.1.0 breaks many packages on Linux - it is not compatible with the system ssh, so curl does not work, https protocol is broken, python does not find the ssh plugin, and so on.  I will revert the bump and try later arrow 5.0.0 since there the openssl dependency of gandiva is removed.

davidrohr:
> Openssl >=1.1.0 breaks many packages on Linux - it is not compatible with the system ssh, so curl does not work,

for reference, this also breaks the other way around, i.e. for my system which has a modern OpenSSL the old OpenSSL of aliBuild breaks everything. I think we should get rid of the custom OpenSSL in general and just use the system one. If we build our own, I don't see why we cannot build a modern one.

awegrzyn:
This is what we've already been doing on FLPs @ CS8, we force openssl from system (`1.1.1k`), otherwise we hit some system issues.

ktf:
We cannot use the system one on the Grid. OpenSSL is not ABI compatible across versions, and not even within the same version, if different options are used. So, there is no chance that something built on one system will work on a slightly different one, unless they ship exactly the same openssl as vanilla CC7 (which is not guaranteed for the WLCG sites, AFAICT, e.g. because of different kerberos incarnations on different sites). 

That will also prevent doing things like running CC7 on CS8 and other stuff which we have done in the past.

davidrohr:
OK, I see, for binary shipment of packages we need it.
Would it be possible to link it statically, so that we don't put it in ld_library_path.
Or use rpath?

aalkin:
Since arrow 5.0 seems to be already released, and openssl dependency can be switched off with that version, moving to arrow 5 would be the best way, unless it breaks something else.

pzhristov:
@aalkin Yes, we are testing Arrow 5.0.0 now. It requires utf8proc and some changes in O2, but all O2 tests were Ok on MacOSX. Now I am trying on Ubuntu.

pzhristov:
Waiting for https://github.com/AliceO2Group/AliceO2/pull/6886

PR #3242, opened 2021-08-19
Reshuffle where the aligenerators tag gets updated

ktf:


ktf:
@alisw/pagmc any reason for using the defaults rather than the package to specify the tag?

maireiphc:
Hello @ktf,

 sorry I may miss some aspects of the problem.

1. Can you tell what is the difference of usage between AliGenerators and generators ?
To me, AliGenerators.sh and default-generators.sh is what is used to compile and deploy under CVMFS the "standalone" packages for Monte Carlo generators. See Jenkins config : https://alijenkins.cern.ch/job/DailyBuilds/job/DailyAliGenerators/753/parameters/
I am unaware of the use case(s) for generators.sh (which seems to need only Pythia6 and Pythia8)

2. About your exact question, I also miss the intent, sorry.
. What is actually the implication of the tag "vAN-20210812" that you introduced ? 
. Is it a tag in sync with AliPhysics or 0² or it has nothing to do and remain AliGenerator specific here ?

defaults-generators.sh has been introduced by you at some points (~2019, I would say) in a sort of development mode to circumvent the multiple clashes between the configuration needed to compile AliRoot/AliPhysics and the various standalone generators (C++ standards which had to be C++14 at least and newer fastjet version required especially).

Cheers,
   Antonin

PR #3248, opened 2021-08-23
O2DataProcessing will be needed for the full system test once we need some scripts from there

davidrohr:


davidrohr:
error unrelated, merging

PR #3265, opened 2021-08-30
Update o2sim.sh

sawenzel:
Analysis has been factored into O2Physics so we should also depend on it (since we need it during some stages of MC workflows).
Also adding EVTGEN because it is used by some PWGs.

### DIFF
--- a/o2sim.sh
+++ b/o2sim.sh
@@ -1,9 +1,10 @@
 package: O2sim
 version: "v%(year)s%(month)s%(day)s"
 requires:
-  - O2
+  - O2Physics
   - O2DPG
   - AEGIS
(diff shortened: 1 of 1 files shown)


ktf:
Not really tested in the CI. Merging.

davidrohr:
@ktf : For reference, O2sim is build in the FullCI and used for the full system test, si it would actually be tested.
I think in this case the commit cannot break anything since O2Physics is not used in the FST.

ktf:
Ah, ok, sorry about that, I was confused by the fact that even without the addition things seem to work since we split the repository, but indeed adding o2physics might "interfere"...

PR #3279, opened 2021-09-02
Bump ROOT

sawenzel:
Move to tag v6-24-06 since it fixes a threading problem in simulation

### DIFF
--- a/root.sh
+++ b/root.sh
@@ -1,6 +1,6 @@
 package: ROOT
 version: "%(tag_basename)s"
-tag: "v6-24-02"
+tag: "v6-24-06"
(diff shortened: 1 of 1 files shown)


pzhristov (approved):

davidrohr (approved):

pzhristov:
@TimoWilken Do we know why build/AliRoot/alidist and build/O2/alidist are so slow?

TimoWilken:
@pzhristov They were both scheduled on builder 3, which had lots of other stuff running on it as well. I've just added more builders, so they should be picked up by a different one soon.

PR #3281, opened 2021-09-03
Use lib rather than lib64 in FairMQ

ktf:


ktf:
Currently broken. Merging.

PR #3282, opened 2021-09-03
Add incremental_recipe for O2Physics

saganatt:
Ciao @ktf,

O2Physics does not have incremental build enabled yet. In case this wasn't intentional, here is the fix.


ktf:
@Barthelemy there is (new?) error from QC:

```
 1/52 Test  #1: testDbFactory ...........................***Timeout  30.32 sec
```

PR #3288, opened 2021-09-06
Export compile_commands.json in O2Physics

ktf:


ktf:
Not tested in any case.

PR #3289, opened 2021-09-06
Override XRootD default tag for JAliEn

zensanp:
Temporary fix for JAliEn build fails caused by later versions of XRootD.

### DIFF
--- a/defaults-jalien.sh
+++ b/defaults-jalien.sh
@@ -15,6 +15,8 @@ overrides:
     tag: v5-09-57h-01
   fastjet:
     tag: v3.4.0_1.045-alice1
+  XRootD:
+    tag: v4.12.5
 ---
 # This file is included in any build recipe and it's only used to set
 # environment variables. Which file to actually include can be defined by the


costing (approved):

adriansev:
@zensanp @costing is there a reason for this that i can forward to upstream as an issue/bug report/feature request?

zensanp:
Possibly due to extra space in XrdOssCsiPages.hh line 189: not C++ compliant(?) At least not liked by compiler, and build subsequently fails.

adriansev:
@zensanp well, i think that we must investigate further as xrootd is already built non only on our alisw stack successfully (from O2::nightly-20210811-1) but is build in all major rpm distros (centos7/8 and fedora) (i have no information from deb camp) so i kind of doubt an programming error but far from exclude it ... i would suspect a compiler incompatibility and/or missing component/include, let's see what the xrootd expert have to say

costing:
@iraklic for notifications

PR #3297, opened 2021-09-08
Set tag in aligenerators.sh

TimoWilken:
To fix daily builds.


maireiphc:
Thanks Timo, Giulio for the AliGenerators fix. Will see tomorrow if the build succeeds (very likely).

PR #3304, opened 2021-09-13
Bump JAliEn to 1.4.0

zensanp:
Update JAliEn (Java components only). Doesn't affect AliPhysics / O2 in any way.

Also removes the previous XRootD 4.12 entry in defaults, in favour of using the latest master.

### DIFF
--- a/defaults-jalien.sh
+++ b/defaults-jalien.sh
@@ -16,7 +16,8 @@ overrides:
   fastjet:
     tag: v3.4.0_1.045-alice1
   XRootD:
-    tag: v4.12.5
+    source: https://github.com/xrootd/xrootd
+    tag: master
 ---
 # This file is included in any build recipe and it's only used to set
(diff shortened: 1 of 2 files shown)


defaults-jalien.sh
```diff
@@ -16,7 +16,8 @@ overrides:
   fastjet:
     tag: v3.4.0_1.045-alice1
   XRootD:
-    tag: v4.12.5
+    source: https://github.com/xrootd/xrootd
+    tag: master
```
ktf, line 20: I do not think this is a good idea, even for just the jalien default. It will trigger a new build of ROOT every time it changes, potentially clogging our CI. If you really need to change tag that ofter, please just hardcode the commit.

zensanp: Alright, changed to use custom tag in own repo. Would that work?

ktf: Yes.

PR #3306, opened 2021-09-13
bump readout v2.5.2

sy-c:


ktf:
@MichaelLettrich can you check what is going on with the mac builds?
@sawenzel o2Sim seems to fail also on slc7.

PR #3307, opened 2021-09-14
Fix JAliEn XRootD tag

zensanp:
Changes the JAliEn XRootD tag in defaults-jalien.sh

The original commit had a missing "v": 5.3.1A --> v5.3.1A


adriansev (approved):

ktf:
Please also use `-aliceX` as a suffix for your tags.

PR #3312, opened 2021-09-15
Bump FairMQ to 1.4.42

ktf:
Provides support for shallow copy in shared memory.

### DIFF
--- a/fairmq.sh
+++ b/fairmq.sh
@@ -1,6 +1,6 @@
 package: FairMQ
 version: "%(tag_basename)s"
-tag: v1.4.41
+tag: v1.4.42
 source: https://github.com/FairRootGroup/FairMQ
 requires:
  - boost
--- a/fairmq.sh
+++ b/fairmq.sh
@@ -1,6 +1,6 @@
 package: FairMQ
 version: "%(tag_basename)s"
-tag: v1.4.40
+tag: v1.4.41
 source: https://github.com/FairRootGroup/FairMQ
 requires:
  - boost


ktf:
@dennisklein @rbx looks like a genuine compilation issue. What is the suggested change on our side? Is it `#include <fairmq/FwdDecls.h>`? 

If yes, @ironMann can you fix it?

ironMann:
patch from @dennisklein in #3314

ktf:
@davidrohr does it ring any bell or is it a new issue?

```
Detected critical problem in logfile tpcraw.log
tpcraw.log-[INFO] Orbit 279: max. allowed orbit 127 exceeded, Link SubSpec=0x9b6e8a84(245:15:0) feeID=0x7a8f will discard further data
tpcraw.log-[ERROR] The link for SubSpec=0x776674083 was not registered
tpcraw.log:terminate called after throwing an instance of 'std::runtime_error'
tpcraw.log-  what():  data for non-registered GBT link supplied
killing child 9140
killing child 9143
killing child 9145
killing child 9146
killing child 9151
killing child 9154
killing child 10142
/sw/slc8_x86-64/O2/3312-slc8_x86-64-local3/share/scripts/jobutils.sh: line 49:  9140 Killed                  eval ${finalcommand} >> ${logfile} 2>&1
Running: TIME="#walltime %e" /sw/slc8_x86-64/O2/3312-slc8_x86-64-local3/share/scripts/monitor-mem.sh /usr/bin/time --output=tpcraw.log_time './tpcraw.log_tmp.sh'
/usr/bin/time --output=tpcraw.log_time ./tpcraw.log_tmp.sh
[INFO] EP:0 CRU:0x0000 Link:15  FEEID:0x000f SrcID:TPC[3] Packet:0   MemSize:0     OffsNext:0      prio.:0 HS:64 HV:6
[INFO] Orbit:0         BC:0    Stop:0 Page:0     Trg:00000000000000000000000000000000 Par:0     DetFld:0x0000
[INFO] Registered Link SubSpec=0x1e0f0f0f(000:15:0) feeID=0x000f with output to raw/TPC/TPC_alio2-cr1-flp070_cru0_0.raw
[INFO] EP:1 CRU:0x0000 Link:15  FEEID:0x004f SrcID:TPC[3] Packet:0   MemSize:0     OffsNext:0      prio.:0 HS:64 HV:6
[INFO] Orbit:0         BC:0    Stop:0 Page:0     Trg:00000000000000000000000000000000 Par:0     DetFld:0x0000
[INFO] Registered Link SubSpec=0x1e510f50(000:15:1) feeID=0x004f with output to raw/TPC/TPC_alio2-cr1-flp070_cru0_1.raw
```

davidrohr:
no, testing locally now

ktf:
I am restarting the fullCI test.

davidrohr:
OK, the FullCI thing is actually a genuine issue, no idea where it comes from, but it fails for me locally as well (took a while to test since apparently clang was change and I had to rebuild everything). Will try to have a look till tonight.

davidrohr:
Hi @ktf @rbx @shahor02 @sawenzel : It seems this new FMQ version is breaking the simulation / digitization.
Digits produced for the detectors are not correct, and digits to raw conversion fails. It seems somewhat random which detector is affected, I have seen it at least for TPC and for ITS.

I have uploaded a test dataset to https://qon.jwdt.org/nmls/tmp/fmq141-sim-testcase.tar.gz which fails in ITS raw encoding:
```
o2-its-digi2raw --file-for cru -o raw/ITS
```
dies with segfault.

In order to reproduce the full chain, merge this PR locally and run a small full system test:
```
NEvents=5 NEventsQED=100 $O2_ROOT/prodtests/full_system_test.sh
```

I have digit-2-raw-converted some old digitized data using the new O2 version, which worked (I.e. the problem is not digit2raw conversion). In contrast, using the old O2 for the digit2raw conversion of data simulated with the new O2/FMQ crashes, so the issue must happen in simulation / digitization.

I didn't check what exactly fails in the simulation / digitization, and I will also not have time to do so in the next days. Could someone check?

rbx:
I've tried the steps you provide on an epn node with latest alidist+O2. For me it fails in the same way with fmq 1.4.41 and 1.4.40. ITS fails sometimes with `segmentation violation` other times with `free(): invalid pointer`. But seems to be the same for both fmq versions.

It's probably best if someone more familiar with what the code is doing checks it more closely. 

At this point I don't see anything that might cause a crash or invalid data in the new fmq. But ready to check again.

ktf:
is this with or without my to O2?

--
Ciao,
Giulio

rbx:
> is this with or without my to O2?

I'm guessing you mean https://github.com/AliceO2Group/AliceO2/pull/7049 ?
I tried without it, only with O2@dev and alidist@master

ktf:
yes, that's what i meant.

davidrohr:
@rbx : I have now tried 3 times with FMQ140 and 3 times with 141, with 140 it always worked and with 141 it always failed.

What I found out though: I have to wipe the O2 build folder in between switching versions, i.e. when I built with 141, then I just switched FMQ back to 140 and rebuild FMQ/O2, it was still failing. After wiping the build folder and rebuilding, it works.

But in any case I am pretty sure the problem is introduced by the new FMQ.

@ktf : I was testing 141 with and without your AliceO2Group/AliceO2#7049, and it fails in both cases in the same way.

ktf:
> What I found out though: I have to wipe the O2 build folder in between switching versions, i.e. when I built with 141, then I just switched FMQ back to 140 and rebuild FMQ/O2, it was still failing. After wiping the build folder and rebuilding, it works.

I suspect CMake recycles the location of FairMQ if not told otherwise, and since the commit is different, FairMQ gets a different build area. @shahor02 is this the behaviour you were seeing with ROOT? I suspect CMake does not notice that ROOT_ROOT / FAIRMQ_ROOT have changed and does not regenerate the cache. If that is true, I think we need to provide those dependencies explicitly again, via the various `-D...` options, including in the incremental builds.

ktf:
@aphecetche please, have a look at the above and let me know if that makes sense to you as well...

sawenzel:
@davidrohr : Would it be possible to try isolating whether it is in digitization or in transport? Could you just reuse the hits from a previous version (hits are stable) and see if it still fails later on?

davidrohr:
Yes, sure, i can do that later.

dennisklein:
> @aphecetche please, have a look at the above and let me know if that makes sense to you as well...

I would also be interested in your opinion/conclusion. I personally never ever re-use an existing CMake build directory, when the external dependencies change - so far I have no trust at all in CMake handling this correctly.

davidrohr:
@sawenzel : So I ran several times now, some times it actually ran to the end also with 141, but usually it fails either at TPC or ITS.
With 140 it was never failing so far.

I took the data of a run with 140, and reran the digitization and raw creation with 141, which failed in raw creation.
That means at least the digitization part with 141 fails.
I didn't try the other way around, i.e. the simulation part could also be broken, but not sure.

aphecetche:
> > @aphecetche please, have a look at the above and let me know if that makes sense to you as well...
> 
> I would also be interested in your opinion/conclusion. I personally never ever re-use an existing CMake build directory, when the external dependencies change - so far I have no trust at all in CMake handling this correctly.

I believe that _by design_ the `find_package` result is indeed cached. So once a dependency has been found it's not searched for again (unless someone's willing to unset the corresponding `XXX_DIR` variable in between cmake invocations).

To (re)convince myself of that I tested a very simple `CMakeLists.txt` : 

```
cmake_minimum_required(VERSION 3.15)
project(play VERSION 1.2.3 LANGUAGES CXX)
find_package(FairMQ REQUIRED)
```

and tried to configure with two different versions of FairMQ (the fact that those are Spack installed versions is irrelevant here - that's just the thing I have at hand on my Mac- ) : 

```
...etche/cmake-play/build [third-party-dep-cache] $ cmake -DFairMQ_ROOT=/Users/laurent/opt/spack/spack_path_placeholder/spack_path_place/darwin-bigsur-skylake/apple-clang-12.0.5/fairmq-1.4.38-lifhrb72ztjplvziscabrkkdux2qjday ..
-- Configuring done
-- Generating done
-- Build files have been written to: /Users/laurent/github.com/aphecetche/cmake-play/build
...etche/cmake-play/build [third-party-dep-cache] $ rg FairMQ CMakeCache.txt
210://The directory containing a CMake configuration file for FairMQ.
211:FairMQ_DIR:PATH=/Users/laurent/opt/spack/spack_path_placeholder/spack_path_place/darwin-bigsur-skylake/apple-clang-12.0.5/fairmq-1.4.38-lifhrb72ztjplvziscabrkkdux2qjday/lib/cmake/FairMQ-1.4.38
214:FairMQ_ROOT:UNINITIALIZED=/Users/laurent/opt/spack/spack_path_placeholder/spack_path_place/darwin-bigsur-skylake/apple-clang-12.0.5/fairmq-1.4.38-lifhrb72ztjplvziscabrkkdux2qjday

...etche/cmake-play/build [third-party-dep-cache] $ cmake -DFairMQ_ROOT=/Users/laurent/opt/spack/spack_path_placeholder/spack_path_place/darwin-bigsur-skylake/apple-clang-12.0.5/fairmq-1.4.40-5uwzjwhaxtsndtd2dfu3uvvmn2tvtj4e ..
-- Configuring done
-- Generating done
-- Build files have been written to: /Users/laurent/github.com/aphecetche/cmake-play/build

...etche/cmake-play/build [third-party-dep-cache] $ rg FairMQ CMakeCache.txt
210://The directory containing a CMake configuration file for FairMQ.
211:FairMQ_DIR:PATH=/Users/laurent/opt/spack/spack_path_placeholder/spack_path_place/darwin-bigsur-skylake/apple-clang-12.0.5/fairmq-1.4.38-lifhrb72ztjplvziscabrkkdux2qjday/lib/cmake/FairMQ-1.4.38
214:FairMQ_ROOT:UNINITIALIZED=/Users/laurent/opt/spack/spack_path_placeholder/spack_path_place/darwin-bigsur-skylake/apple-clang-12.0.5/fairmq-1.4.40-5uwzjwhaxtsndtd2dfu3uvvmn2tvtj4e
```

So I think I agree with Denis that a build directory should not be reused if/when dependencies change.

Hope I answered the right question :smile:

dennisklein:
> So I think I agree with Denis that a build directory should not be reused if/when dependencies change.

Another circumstance, that makes my stomach hurt, is, that even if the `find_package` logic would be correct, the actual build targets have dependencies expressed in the generated build system, e.g. Makefile, as far as I know they rely on file c/m/a-times to determine if something needs to be relinked/rebuilt. What if I updated dependency package with a manager that uses precompiled binary packages that still contain old enough c/m-times on the installed files.. So, in theory, one could get very messy intermediate unintended states (and even worse they probably work even most of the times). But then, I could be too paranoid 🙈

rbx:
So, I've cleaned everything in my `sw/BUILD`, recompiled with FairMQ 1.4.40 and it still fails for me, on an epn node. 
My plan was to test each commit after 1.4.40 incrementally and see where it starts to fail. But so far I have yet to have a successful run with 1.4.40...
I will do a full rebuild with 1.4.40 now to be certain it's not yet some other environment bug.

Seeing how it sometimes doesn't fail for you, it sounds like a race condition. 
One thing that did change with the 1.4.41 is that aligned allocation performance improved. In my benchmark it is now up to 2x faster. Real world may be a lot less significant, but still may impact timing to some degree. Also, new member was added to the transferred meta data, which may have small negative impact on performance. 
My point is - timing have changed so some race condition may be more or less severe.

davidrohr:
Hm, the race condition would explain that it fails irregularly with 1.41. But then it is strange that also in the CI we haven't seen it before with 1.40. can you tell me where you run it on the epn and your exact commands such that i can try?

Kind Regards
David Rohr

rbx:
I'm running on epn129:
alidist@master 526274986ee53364fb6676ca973f979834d89a5b
AliceO2@dev f7ed096db9359b52deea52170fbd946e4594db4c
FairMQ@v1.4.40 (checked out as development package, but no changes)

Then I compile everything with:
`aliBuild --defaults o2 build O2 O2sim O2DataProcessing DataDistribution Readout --debug --disable FLUKA --disable FLUKA_VMC`
Enter env:
`alienv ^Cter O2sim/latest-o2`
and run `NEvents=5 NEventsQED=100 $O2_ROOT/prodtests/full_system_test.sh` from the sub-directory with the contents of the archive you posted.

davidrohr:
Ah, wait, that is the failure!
If you run from the content I posted, then you have already the corrupt digits, and it will not rerun the digitization.
I posted the files just so people con check without updating fmq.

Can you try to run the command in an empty folder.

Kind Regards
David Rohr

Sent from my mobile. (Excuse the typos!)

On September 17, 2021 8:37:01 PM GMT+02:00, Alexey Rybalchenko ***@***.***> wrote:
>I'm running on epn129:
***@***.*** 526274986ee53364fb6676ca973f979834d89a5b
***@***.*** f7ed096db9359b52deea52170fbd946e4594db4c
***@***.*** (checked out as development package, but no changes)

rbx:
Ah, I see, will retry without.

ktf:
it's by design that they are cached, but it should still check if the
associated variable has changed since the cache. I am pretty sure that if
you pass the correct -DFairMQ_DIR=$FAIRMQ_ROOT, it would have recompiled
correctly, so I think the issue is either a bug in CMake (it should have
noticed that FAIRMQ_ROOT changed) or with some of our wrapper stuff. If
CMake really does not honor changes in the env variables, I think we should
simply add back the explicit dependencies via the -D flags.

Ciao,
Giulio

aphecetche:
As far as I understand the `<PackageName>_DIR` variable should indicate the _full path_ to a cmake package's _config file_. So changing it will indeed find the correct dependency (assuming of course we're dealing with a dependency that provides config file, which FairMQ does all right). 

```
...etche/cmake-play/build [third-party-dep-cache] $ cmake .. -DFairMQ_DIR=/Users/laurent/opt/spack/spack_path_placeholder/spack_path_place/darwin-bigsur-skylake/apple-clang-12.0.5/fairmq-1.4.40-5uwzjwhaxtsndtd2dfu3uvvmn2tvtj4e/lib/cmake/FairMQ-1.4.40
-- Configuring done
-- Generating done
-- Build files have been written to: /Users/laurent/github.com/aphecetche/cmake-play/build
...etche/cmake-play/build [third-party-dep-cache] $ rg FairMQ CMakeCache.txt                                                                                                      
208:FairMQ_DIR:UNINITIALIZED=/Users/laurent/opt/spack/spack_path_placeholder/spack_path_place/darwin-bigsur-skylake/apple-clang-12.0.5/fairmq-1.4.40-5uwzjwhaxtsndtd2dfu3uvvmn2tvtj4e/lib/cmake/FairMQ-1.4.40
...etche/cmake-play/build [third-party-dep-cache] $ cmake .. -DFairMQ_DIR=/Users/laurent/opt/spack/spack_path_placeholder/spack_path_place/darwin-bigsur-skylake/apple-clang-12.0.5/fairmq-1.4.38-lifhrb72ztjplvziscabrkkdux2qjday/lib/cmake/FairMQ-1.4.38
-- Configuring done
-- Generating done
-- Build files have been written to: /Users/laurent/github.com/aphecetche/cmake-play/build
...etche/cmake-play/build [third-party-dep-cache] $ rg FairMQ CMakeCache.txt                                                                                                     
208:FairMQ_DIR:UNINITIALIZED=/Users/laurent/opt/spack/spack_path_placeholder/spack_path_place/darwin-bigsur-skylake/apple-clang-12.0.5/fairmq-1.4.38-lifhrb72ztjplvziscabrkkdux2qjday/lib/cmake/FairMQ-1.4.38
```

But the price to pay (for the calling script/user) is the construction of that full path. While `<PackageName>_ROOT` , `$ENV{<PackageName>_ROOT}` (or `CMAKE_PREFIX_PATH`) are working with prefixes (i.e. no need to find out the `lib/cmake/<PackageName>-version` part prior to calling `cmake`). 

So for instance `-DFairMQ_DIR=$FAIRMQ_ROOT` should not work at all  : 

```
...etche/cmake-play/build [third-party-dep-cache] $ cmake .. -DFairMQ_DIR=/Users/laurent/opt/spack/spack_path_placeholder/spack_path_place/darwin-bigsur-skylake/apple-clang-12.0.5/fairmq-1.4.40-5uwzjwhaxtsndtd2dfu3uvvmn2tvtj4e/
CMake Error at CMakeLists.txt:5 (find_package):
  Could not find a package configuration file provided by "FairMQ" with any
  of the following names:

    FairMQConfig.cmake
    fairmq-config.cmake

  Add the installation prefix of "FairMQ" to CMAKE_PREFIX_PATH or set
  "FairMQ_DIR" to a directory containing one of the above files.  If "FairMQ"
  provides a separate development package or SDK, be sure it has been
  installed.

-- Configuring incomplete, errors occurred!
See also "/Users/laurent/github.com/aphecetche/cmake-play/build/CMakeFiles/CMakeOutput.log".
See also "/Users/laurent/github.com/aphecetche/cmake-play/build/CMakeFiles/CMakeError.log".
```

dennisklein:
> So for instance -DFairMQ_DIR=$FAIRMQ_ROOT should not work at all

Correct, because the **CMake variable** of that name is defined as you state, however, the **environment variable** `<PkgName>_DIR` acts as a package-specific search path like `CMAKE_PREFIX_PATH` :rofl:

dennisklein:
> I think we should simply add back the explicit dependencies via the -D flags.

Please be also aware of the following semantics:

```console
❯ cat CMakeLists.txt
cmake_minimum_required(VERSION 3.20 FATAL_ERROR)
project(test VERSION 1.0.0 LANGUAGES CXX)
find_package(FairMQ REQUIRED)
❯ rm -rf build
❯ cmake -S. -Bbuild
-- The CXX compiler identification is GNU 11.2.1
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Configuring done
-- Generating done
-- Build files have been written to: /home/dklein/projects/FairRoot2/blubb/build
❯ grep FairMQ build/CMakeCache.txt
//The directory containing a CMake configuration file for FairMQ.
FairMQ_DIR:PATH=/lib64/cmake/FairMQ-1.4.41
❯ FairMQ_DIR=/home/dklein/projects/FairMQ2/install cmake -S. -Bbuild
-- Configuring done
-- Generating done
-- Build files have been written to: /home/dklein/projects/FairRoot2/blubb/build
❯ grep FairMQ build/CMakeCache.txt
//The directory containing a CMake configuration file for FairMQ.
FairMQ_DIR:PATH=/lib64/cmake/FairMQ-1.4.41
❯ rm -rf build
❯ FairMQ_DIR=/home/dklein/projects/FairMQ2/install cmake -S. -Bbuild
-- The CXX compiler identification is GNU 11.2.1
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Configuring done
-- Generating done
-- Build files have been written to: /home/dklein/projects/FairRoot2/blubb/build
❯ grep FairMQ build/CMakeCache.txt
//The directory containing a CMake configuration file for FairMQ.
FairMQ_DIR:PATH=/home/dklein/projects/FairMQ2/install/lib64/cmake/FairMQ-1.4.41
```

As you can see, just providing an additional search path input, does not necessarily convince CMake to choose the new candidate at that location. It also appears to depend on whether the new candidate provides a better version ...

rbx:
We've now identified the problem - a regression in `SetUsedSize()` implementation that would "cut off" too many bytes, which explains the observed failures.
I'm testing a fix (looking good so far, across multiple runs of FST) and will publish a new version until the end of Monday.
This has no effect on the shared ownership usage/implementation, so AliceO2Group/AliceO2#7049 should not be affected.

ktf:
Ok, that would explain why the analysis workflows and the unit tests I run
were not affected.
Another possible problem would be that at some point, IIRC, we thought
about overwriting one of the headers shipped by DD. I do not recall if we
actually did it in the end, though.

Ciao,
Giulio

rbx:
v1.4.42 is released that contains the fix for the issue.

davidrohr:
@shahor02 @sawenzel : This fails now in the aod producer of the sim-challenge:
```
Launching task: o2-aod-producer-workflow --aod-writer-keep dangling --aod-writer-resfile AO2D --aod-writer-resmode UPDATE --aod-timeframe-id 1 &> aod.log &
Detected critical problem in logfile aod.log
aod.log-[18586:aod-producer-workflow]: [15:28:23][WARN] Unable to communicate with driver because client is not connected. Continuing connection attempts.
aod.log-[18586:aod-producer-workflow]: 
aod.log:[18586:aod-producer-workflow]:  *** Break *** segmentation violation
```
Is this a known problem?

shahor02:
no, this is something new...

davidrohr:
Other PRs are failing in the same way. So this FMQ version has fixed the fullCI tests.
Merging

PR #3326, opened 2021-09-18
Bump FairLogger to 1.10.0

davidrohr:


ktf:
This will actually require an extensive cleanup, since LOG(ERROR) is deprecated in favor of LOG(error). Why do we need a new version?

davidrohr:
@Barthelemy : The capital severity levels for the logging are deprecated:
```
/sw/SOURCES/QualityControl/v1.30.1/v1.30.1/Framework/include/QualityControl/ExamplePrinterSpec.h: In member function 'virtual void o2::quality_control::example::ExamplePrinterSpec::run(o2::framework::ProcessingContext&)':
/sw/SOURCES/QualityControl/v1.30.1/v1.30.1/Framework/include/QualityControl/ExamplePrinterSpec.h:46:9: error: 'fair::Severity::INFO' is deprecated: Use 'info' instead [-Werror=deprecated-declarations]
   46 |     LOG(INFO) << "Received data";
```
Could you adapt that in QC?

davidrohr:
> This will actually require an extensive cleanup, since LOG(ERROR) is deprecated in favor of LOG(error). Why do we need a new version?

Because I need to have the `important` and `alert` log levels for operation.

The errors I am seeing are in QC, how is it in O2? In any case I think we should clean that up, no reason to allow upper and lower case levels.

ktf:
that’s not limited only to QC, no? Unless fixed everywhere we will get an
avalanche of warnings.

Ciao,
Giulio

Barthelemy:
As it is used everywhere in O2 and there is no error, I wonder if it is a compilation flag that differs from QC to O2. I could just update the qc compilation flags.

davidrohr:
Perhaps o2 is compiled with -Wno-deprecated. For now I tzink changing the flag is a good idea. On the other hand we should adapt in any case, also in O2.

Barthelemy:
I agree that we should adapt. For now switching the flag would be a quick solution. I don't understand why the deprecation warnings are raising errors in QC though. This is in the recipe: 
```
  CXXFLAGS="${CXXFLAGS} -Werror -Wno-error=deprecated-declarations"
fi
```
This is the same in o2.

davidrohr:
Strange, but perhaps we also need -Wno-error=deprecated

Barthelemy:
I have just done a full rebuild on my machine with the flag ALIBUILD_O2_TESTS turned on. QC behaved like O2, i.e. issued warnings but no errors. 

@ktf @TimoWilken Any idea ?

Barthelemy:
Meanwhile I have prepared a PR in QC to adapt to the new interface.

ktf:
That check is only there for the full builds, not for the incremental ones,
apparently...

Ciao,
Giulio

Barthelemy:
@ktf Indeed, here is the fix https://github.com/alisw/alidist/pull/3330

davidrohr:
@Barthelemy : The other CIs ran through now, but the fullCI still fails.
The reason is that the fullCI compiles without the ALIBUILD_O2_TESTS. (The reason is that the ctests take quite long and will anyway run in the other CIs, and also if a ctests would fail in the fullCI, we wouldn't see other errors that come afterwards like FST or codechecker).
Thus your fix does work for the others but not for the fullCI.
I think by default you have `-Werror` enabled, but you have the `-Wno-error=deprecated-declarations` only if `ALIBUILD_O2_TESTS` is set.

Barthelemy:
@davidrohr Thank you for the explanation about fullCI. I did not know. I fail to see where I set `-Werror` by default. What I will do as a quick fix is to set `-Wno-error=deprecated-declarations` in all cases.

Barthelemy:
https://github.com/alisw/alidist/pull/3333

PR #3331, opened 2021-09-20
monalisa.sh

iraklic:
Adding a recipe for MonALISA for automatic deployment to CVMFS

### DIFF
--- a/monalisa.sh
+++ b/monalisa.sh
@@ -1,0 +1,38 @@
+package: MonALISA
+version: "20210818"
+# tag: "1.0.0"
+# source: https://gitlab.cern.ch/jalien/jalien.git
+requires:
+ - JDK
+ - "curl:(?!slc8)"
+ - "system-curl:slc8.*"
+valid_defaults:
+ - monalisa
+---
+#!/bin/bash -e
+
+curl http://alimonitor.cern.ch/download/MonaLisa/MonaLisa-${PKGVERSION}.tar.gz | tar xz
+mkdir $INSTALLROOT/$PKGVERSION-$PKGREVISION
+mv MonaLisa-$PKGVERSION $INSTALLROOT/$PKGVERSION-$PKGREVISION
+
+# Modulefile
+MODULEDIR="$INSTALLROOT/etc/modulefiles"
+MODULEFILE="$MODULEDIR/$PKGNAME"
+mkdir -p "$MODULEDIR"
+cat > "$MODULEFILE" <<EoF
+#%Module1.0
+proc ModulesHelp { } {
+  global version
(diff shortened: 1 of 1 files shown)


monalisa.sh
```diff
@@ -0,0 +1,37 @@
+package: MonALISA
+version: "20210818"
+# tag: "1.0.0"
+# source: https://gitlab.cern.ch/jalien/jalien.git
+requires:
+ - JDK
+ - "curl:(?!slc8)"
+ - "system-curl:slc8.*"
+valid_defaults:
+ - monalisa
```
ktf, line 10: Please, do not introduce another default.

monalisa.sh
```diff
@@ -0,0 +1,37 @@
...
+module-whatis "ALICE Modulefile for $PKGNAME $PKGVERSION-@@PKGREVISION@$PKGHASH@@"
+# Dependencies
+module load BASE/1.0 JDK/$JDK_VERSION-$JDK_REVISION \\
+            XRootD/$XROOTD_VERSION-$XROOTD_REVISION \\
+            xjalienfs/$XJALIENFS_VERSION-$XJALIENFS_REVISION
+# Our environment
+set JALIEN_ROOT \$::env(BASEDIR)/$PKGNAME/\$version
+prepend-path CLASSPATH \$JALIEN_ROOT/lib/alien-users.jar
+prepend-path PATH \$JALIEN_ROOT/bin
+EoF
```
ktf, line 37: I think you should use `alibuild-generate-module --bin --lib > etc/modulefiles/$PKGNAME` and add a `build_requires` on `alibuild-recipe-tools`.

monalisa.sh
```diff
@@ -0,0 +1,37 @@
+package: MonALISA
+version: "20210818"
+# tag: "1.0.0"
+# source: https://gitlab.cern.ch/jalien/jalien.git
+requires:
+ - JDK
+ - "curl:(?!slc8)"
```
ktf, line 7: Do you really need curl at runtime? Also, given you only need it for building, I would simply drop the dependency.

ktf (changes requested):
Please adapt the recipe following the suggestions.

monalisa.sh
iraklic: Sorry, this looks weird, I will have to look up how to properly use github suggestions, I have not committed a suggestion before.

iraklic: somehow your comment doesn't show as a suggestion and mine is little messed up :)

ktf:
Please have a look at #3342. For some reason I could not update this one.

iraklic:
I see you have put in the changes that you suggested. Looks foot do me. I approve that recipe.
Is that PR going to be merged along with the one in `ali-bot` and this one dropped?

PR #3342, opened 2021-09-22
Add recipe for monalisa

ktf:


iraklic (approved):

iraklic:
I see that build on macOS has failed, but this recipe does nothing except download and untar. Do you know what could be failing?

ktf:
It looks like a red herring. We have had troubles with that machine for a few days. @MichaelLettrich can you have a look? I will merge this in any case, give we are not going to need it on mac in any case.

iraklic:
yeah, it's hard to debug, because there are no logs.

PR #3380, opened 2021-10-05
Fix dependency check requiring defaults o2

ktf:


ktf:
This should get away with issue in Build Documentation

PR #3384, opened 2021-10-05
Bump alibuild-recipe-tools to v0.2.3

ktf:


ktf:
> The new speed is fantastic. However, I just encountered new bugs. When doing a combined "enter" with 2 packages, it appears that all packages are loaded twice (in different versions). My sequence is:
> 
> ```
> 1) aliBuild build O2sim
> 2) aliBuild build QualityControl
> 3 alienv enter O2sim/latest, QualityControl/latest
> ```
> 
> Then I see for instance `GEANT4` loaded twice (same for all other packages):
> 
> ```
> GEANT4/v10.7.2-local3              
> GEANT4/v10.7.2-local4
> ```

Are you sure this is a regression? I would have expected also the old version to behave in the same way.

awegrzyn:
> i.e. replacing `'` with `"` in all modulefiles.

Thanks ok, what I meant is that dependencies should remain the same.

ktf:
Yes, dependencies will remain the same. The order in which they are loaded will potentially changed.

PR #3434, opened 2021-10-15
monalisa.sh: Update to version 20211013

iraklic:


ktf:
You still have the JALIEN_ROOT referred, without any mention of JAliEn in the dependencies. Could you fix that?

PR #3435, opened 2021-10-15
Bump Monitoring to v3.8.10 and InfoLogger to v2.3.0

awegrzyn:
@davidrohr This should stop Monitoring from flooding InfoLogger


awegrzyn:
I'd more confident if CS8 checks would pass

ktf:
For the record, the CS8 errors are due to a misconfiguration on the new machines, which I just fixed. Hopefully they should become green soon.

PR #3458, opened 2021-10-19
Bump FairMQ to 1.4.43 - DON't MERGE YET

davidrohr:
Should be merged only when there will be no more FLPSuite before the pilot beam.


ktf:
Reopen when the fixed version is around.

PR #3479, opened 2021-10-25
Do not compile the legacy Python interface

pzhristov:
Should be merged together with
https://github.com/alisw/AliPhysics/pull/18754


ktf:
@TimoWilken is now building the cache with this PR enabled. We can then merge.

PR #3505, opened 2021-11-01
Bump clang to 13.0.0

ktf:


ktf:
@davidrohr I would leave this open until we see the fullCI results and then we merge  #3506.

ktf:
In the meanwhile I will check on hyperloop as well.

ktf:
This seems to break Gandiva...

pzhristov:
> This seems to break Gandiva...

Gandiva was quite Ok with clang 13 on Mac M1. Where do you see the issue?

ktf:
In hyperloop (ubuntu)

ktf:
I am testing again and will report the actual error I see...

ktf:
It looks like hyperloop has unrelated problems right now...

ktf:
Ok, the issue I was having was unrelated. We can probably merge #3506, but we should first create the caches for it. @TimoWilken can you take care?

davidrohr:
For merging #3506, we have to fix the O2CodeChecker issue first. But probably we can close this one. If we bump, we should use #3506.

PR #3506, opened 2021-11-01
Bump clang to 13 + add llvm-spirv translator

davidrohr:
Same as  #3505 but spirv translator added

### DIFF
--- a/clang.sh
+++ b/clang.sh
@@ -51,6 +51,19 @@ cmake $SOURCEDIR/llvm \
 
 cmake --build . -- ${JOBS:+-j$JOBS} install
 
+git clone -b $PKGVERSION https://github.com/KhronosGroup/SPIRV-LLVM-Translator
+mkdir SPIRV-LLVM-Translator/build
+pushd SPIRV-LLVM-Translator/build
+cmake ../ \
+  -G Ninja \
+  -DLLVM_DIR="$INSTALLROOT/lib/cmake/llvm" \
+  -DLLVM_BUILD_TOOLS=ON \
+  -DLLVM_INCLUDE_TESTS=OFF \
+  -DCMAKE_BUILD_TYPE=Release \
+  -DCMAKE_INSTALL_PREFIX:PATH="$INSTALLROOT"
+cmake --build . -- ${JOBS:+-j$JOBS} install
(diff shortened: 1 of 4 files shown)


ktf (dismissed):

davidrohr:
I just tried this locally in the FullCI docker container, and it has build the OpenCL part correctly using the alidist clang 13. So we can get rid of the custom clang in the docker. Will do that with the next update. Still, I'll retain the possibility to add a custom clang again if needed, since we are not free to update alidist clang spontaneously.

ktf:
Looks like it breaks the code checker.

davidrohr:
Ah, oh course. We have to bump O2CodeChecker in addition and update some files there.
@sawenzel @pzhristov : could you have a look, and perhaps open a PR to my PR branch?
(I can also do it, but first have to catch up with some other things this week.)

ktf:
@TimoWilken build/O2/alidist-dataflow breaks.

TimoWilken:
The machine it ran on temporarily had network troubles. Fixed now.

ktf:
@davidrohr @sawenzel it looks like `modernize-return-braced-init-list` started to trigger basically everywhere. It is indeed in the list of checks we explicitly enabled but somehow it silent so far. Do we really care about it?

davidrohr:
THe list has endless repetitions, but there are still 280 unique cases.
Actually I don't care, we can either fix them, or we disable the warning, both is fine with me.
If you want I can take care of fixing it up, shouldn't take that long.

davidrohr:
@ktf @TimoWilken @MichaelLettrich : Could you check why the MacOS CI fails here with
```
find: /Applications/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Library/Developer/CoreSimulator/Profiles/Runtimes/iOS.simruntime/Contents/Resources/RuntimeRoot/private/var/mobile: Permission denied
```

ktf:
Looks like a genuine issue with the machine. @MichaelLettrich @TimoWilken could you have a look?

TimoWilken:
Hm, weird. I'm not sure how it happened, but I've `chmod +r`-ed the directory and restarted the check.

PR #3510, opened 2021-11-02
Backport new root to 57 series

ktf:


ktf:
@pzhristov we need to backport the new root to the 57 branch, if we want to be able to use the new ROOT / python integration in the AliPhysics nightly. Would that actually work?

pzhristov:
@ktf We can also provide a new tag from the AliRoot master.

ktf:
Sounds good to me.

Ciao,
Giulio

PR #3519, opened 2021-11-05
Bump xrootd to v5.3.2

TimoWilken:
This fixes build problems with the latest Xcode on MacOS.

Cc: @Barthelemy


ktf (approved):
Let's make sure the cache is hot with this before we merge it.

PR #3520, opened 2021-11-05
enable warnings

jgrosseo:


ktf:
As discussed privately, maybe it's better to have it in the O2Physics CMake configuration.

PR #3529, opened 2021-11-08
Do not run tests / benchmarks for VecGeom

ktf:


ktf:
@sawenzel any objections to this? It adds ~5 extra minutes every time we rebuild root, which is not much, but if we can avoid it it would not harm. If needed we can make the test dependent on the fact of having a devel package.

PR #3530, opened 2021-11-08
Add jq to o2dataprocessing.sh

ktf:


ktf:
@chiarazampolli as discussed in the meeting.

davidrohr:
I would add it either to O2DataProcessing or to QualityControl, it seems a bit odd here.

ktf:
Given O2DataProcessing is where the scripts actually use it, I added it there.

PR #3533, opened 2021-11-09
Fix of openssl linking in cmake build

jan-busa:
When building latest CMake 3.19.2 against OpenSSL v1.0.2o, cmake looks for the OpenSSL library at wrong place. This should fix it.

### DIFF
--- a/cmake.sh
+++ b/cmake.sh
@@ -28,6 +28,8 @@ SET(Java_JAVAC_EXECUTABLE FALSE CACHE BOOL "" FORCE)
 SET(BUILD_CursesDialog FALSE CACHE BOOL "" FORCE)
 EOF
 
+export OPENSSL_ROOT_DIR=$OPENSSL_ROOT
+
 $SOURCEDIR/bootstrap --prefix=$INSTALLROOT \
                      --init=build-flags.cmake \
                      ${JOBS:+--parallel=$JOBS}


ktf (approved):

ktf:
This is fine, however it will trigger a rebuild of everything, so we first need to warm the cache and then to merge it.

ktf:
@TimoWilken could  you warm up the cache with this PR? I think it's ready to go.

davidrohr:
@ktf @TimoWilken : what is with this PR? It is now pending since 4 months. Shall we merge it?

ktf:
We should merge, but it will trigger a full rebuild, so we did not find a good moment in the last few months..

davidrohr:
@ktf  I want to bump CMake soon to 3.23.1, once that is out, since it has a fix I need. I'd recommend we merge the two things together on Saturday night?

ktf:
Reverting this since we have broken builds and I am out of options.

davidrohr:
Still don't understand which builds are failing. My alidist and O2 PRs both went green after the fix.

ktf:
Frankly speaking, neither do I, but what I see is that now the build is back in the green. The OpenSSL path was clearly where it should have been, yet it was complaining. This was only on cc8 as well, IIRC.

PR #3534, opened 2021-11-09
Fixed root.sh to properly find for libxml2

jan-busa:
When building ROOT v6-24-06 with libxml2/v2.9.3 (--no-system build), proper variable for LIBXML2 libdir and include dir has to be set in ROOT configuration.

Signed-off-by: Jan Busa Jr <busa@jinr.ru>

### DIFF
--- a/root.sh
+++ b/root.sh
@@ -17,6 +17,7 @@ requires:
   - "OpenSSL:(?!osx)"
   - "osx-system-openssl:(osx.*)"
   - XRootD
+  - TBB
 build_requires:
   - CMake
   - "Xcode:(osx.*)"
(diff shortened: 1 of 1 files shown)


root.sh
```diff
@@ -127,6 +127,8 @@ cmake $SOURCEDIR
       ${OPENSSL_ROOT:+-DOPENSSL_ROOT=$OPENSSL_ROOT}                                    \
       ${OPENSSL_ROOT:+-DOPENSSL_INCLUDE_DIR=$OPENSSL_ROOT/include}                     \
       ${LIBXML2_ROOT:+-DLIBXML2_ROOT=$LIBXML2_ROOT}                                    \
+      ${LIBXML2_ROOT:+-DPC_LIBXML_INCLUDEDIR=$LIBXML2_ROOT/include}                    \
+      ${LIBXML2_ROOT:+-DPC_LIBXML_LIBDIR=$LIBXML2_ROOT/lib}                            \
```
ktf, line 131: I am surprised one needs to override the package config hooks. Which distribution are you using?

ktf (commented):
Which distribution are you using?

root.sh
jan-busa: I'm doing build inside CentOS-7 and using alibuild --no-system build. This builds CMake 3.19.2 and inside Modules directory is libXML2 expecting the variable PC_LIBXML_LIBDIR (see slc7_x86-64/CMake/v3.19.2-X/share/cmake-3.19/Modules/FindLibXml2.cmake). LibXML2 is not from system but was built using alibuild.

ktf:
@TimoWilken could you warm up the cache with this PR? I think it's ready to go.

jan-busa:
I would like to add one more thing to the root.sh, can you wait with merge or should I make another pull request? I would like to add possibility to build root with fftw3 support (allowing both external FFTW3 library or enabling builtin-fftw3).

ktf:
Ok, let us know when you are ready.

PR #3537, opened 2021-11-09
Remove unneeded XRootD override in defaults-o2.sh

TimoWilken:
This was masking the updated xrootd.sh for some checks.

Cc: @Barthelemy


ktf (dismissed):
Fine with me, but we should warm the cache as this will rebuild root.

TimoWilken:
Caching is done.

PR #3540, opened 2021-11-10
Bump MCStepLogger to v0.2.0

benedikt-voelkel:


benedikt-voelkel:
Could we merge this @ktf? In case of the failed CI it like an unrelated timeout problem with CCDB

PR #3543, opened 2021-11-10
Fix Clang compilation with newer Apple SDKs

TimoWilken:
`xcrun --show-sdk-path` seems to give the correct result, no need for `find`.

@ktf or @davidrohr: is there a case where this doesn't work that the previous code was guarding against?

In addition, fix the build on `osx_arm64`. Tested on Mac M1.

### DIFF
--- a/clang.sh
+++ b/clang.sh
@@ -20,20 +20,10 @@ unset CXXFLAGS
 unset CFLAGS
 unset LDFLAGS
 
-case $ARCHITECTURE in 
+case $ARCHITECTURE in
   # Needed to have the C headers
-  osx*)
-    XCODE_PATH=`xcode-select -p`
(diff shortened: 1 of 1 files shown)


pzhristov:
@TimoWilken I switched off the compilation of Clang on MacOSX since the system compiler is doing perfectly the job.

pzhristov:
@TimoWilken Of course, it is nice to have the possibility to compile in case we need some ALICe-specific modifications.

TimoWilken:
Yes, good point, I forgot we were only using system packages. Oh well, maybe it'll be useful at some point. At least the `find ...` -> `xcrun --show-sdk-path` changes are needed on Intel Macs as well, especially when the next Xcode update comes along.

pzhristov (approved):

ktf:
@TimoWilken looks good. As usual, could you warm up the cache?

TimoWilken:
> looks good. As usual, could you warm up the cache?

@ktf Done!

PR #3545, opened 2021-11-10
add o2 defaults catered towards system libs

adriansev:
prepared defaults for testing building with system curl and openssl dependencies

### DIFF
--- a/defaults-o2system.sh
+++ b/defaults-o2system.sh
@@ -1,0 +1,47 @@
+env:
+  CFLAGS: -fPIC -O2
+  CMAKE_BUILD_TYPE: RELWITHDEBINFO
+  CXXFLAGS: -fPIC -O2 -std=c++17
+  CXXSTD: '17'
+  ENABLE_VMC: 'ON'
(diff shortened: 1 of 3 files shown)


adriansev:
@ktf this might not build given the target and if the prerequisites listed at https://alice-doc.github.io/alice-analysis-tutorial/building/prereq-centos8.html are not used on testing machines.
I did this defaults for easy building with system libraries for testing and investigations.

ktf:
Didn't we decide that we actually deploy the full installation on the EPNs? Why is this needed now?

adriansev:
@ktf no, it is not needed, this is just an investigative tool on possibility and compatibility with system provided software ... if you tell me that the presence of this setting would have a negative impact i will try to keep it only in my branch

davidrohr:
In my opinion the pure presense of the defaults obviously doesn't break anything, but it could lead people into using it and expecting support. So in case it is only for your private tests, I would actually indeed prefer if it could be kept in a private fork.
Particularly, when it is used only by you there cannot be merge conflicts.

adriansev:
got it!

PR #3550, opened 2021-11-11
Bump FairMQ to 1.4.43

davidrohr:


ktf:
@rbx does this already have the channel-config option for the Send timeout?

davidrohr:
> @rbx does this already have the channel-config option for the Send timeout?

No, this is several days old. It supports querrying the number of peers. We couldn't merge it before since it deprecated some functions used by DD, which was bumped today to become compatible to the new FairMQ.

davidrohr:
errors unrelated, merging

rbx:
@ktf for the channel-config option I have a PR open already, shouldn't be too long until that is tagged.

PR #3568, opened 2021-11-16
xrootd update to 5.3.3

adriansev:
fast bugfix release: https://github.com/xrootd/xrootd/blob/v5.3.3/docs/ReleaseNotes.txt
matters for alien.py meta4 based fallback mechanics


ktf:
@TimoWilken can you please warm the cache with this?

TimoWilken:
@ktf the first two builds succeeded and the others builds are halfway through building O2, so I'd call it good enough to merge.

PR #3574, opened 2021-11-18
Disable built-in XRootD in ROOT

TimoWilken:
If the XRootD package is disabled, this should disable support for it in ROOT as well. Otherwise, ROOT downloads and compiles against an old version of XRootD instead.

Cc: @Barthelemy -- this should fix the issue in QC Mac CI.

### DIFF
--- a/root.sh
+++ b/root.sh
@@ -97,6 +97,15 @@ else
   ROOT_HAS_NO_PYTHON=1
 fi
 
+if [ -n "$XROOTD_ROOT" ]; then
+  ROOT_XROOTD_FLAGS="-Dxrootd=ON -DXROOTD_ROOT_DIR=$XROOTD_ROOT"
+else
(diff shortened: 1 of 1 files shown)


ktf:
LGTM, as usual we should warm up the cache before merging.

root.sh
```diff
@@ -97,6 +97,15 @@ else
   ROOT_HAS_NO_PYTHON=1
 fi
 
+if [ -n "$XROOTD_ROOT" ]; then
```
adriansev, line 100: @TimoWilken @ktf  Why shouldn't we use just a plain `-Dbuiltin_xrootd=0` ? This way, the xrootd capability of root is directly dependent on presence of xrootd itself in the system...

ktf: which we do not control and which could give issues, though...

PR #3586, opened 2021-11-22
xjalienfs:: 1.3.6 hot bugfix

adriansev:
for in-zip files the url used is not the meta filepath! we should take this into account.
this bug affects all in-zip files downloads


adriansev:
@ktf @costing guys, it would really help it this can be fast-tracked! thanks a lot and sorry for the mess!

PR #3592, opened 2021-11-24
Treat warnings as errors in O2Physics CI

jgrosseo:


jgrosseo:
Testing here: https://github.com/AliceO2Group/O2Physics/pull/337

ktf:
In principle now it should be  fixed.

PR #3598, opened 2021-11-24
Introduce VMC recipe

benedikt-voelkel:
* adiabatic approach

  * VMC recipe depends on ROOT, therefore remove ROOT dependency from all
    VMC related recipes and add VMC as the actual dependency

  * for now do not touch anything else and still take ROOT's VMC
    (this keeps everything working as before as for instance O2
     compilation would crash otherwise)

  * following this, everything will move to VMC standalone

### DIFF
--- a/mcsteplogger.sh
+++ b/mcsteplogger.sh
@@ -4,6 +4,7 @@ tag: "v0.3.0"
 source: https://github.com/AliceO2Group/VMCStepLogger.git
 requires:
   - "GCC-Toolchain:(?!osx)"
+  - ROOT
   - VMC
   - boost
 build_requires:
--- a/mcsteplogger.sh
+++ b/mcsteplogger.sh
@@ -1,10 +1,10 @@
 package: MCStepLogger
 version: "%(tag_basename)s"
-tag: "v0.2.0"
+tag: "v0.3.0"
 source: https://github.com/AliceO2Group/VMCStepLogger.git
 requires:
   - "GCC-Toolchain:(?!osx)"
-  - ROOT
+  - VMC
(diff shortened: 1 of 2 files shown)


benedikt-voelkel:
@sawenzel  @TimoWilken is this mergeble?

ktf:
I would say so. I was asking @sawenzel for the review of the functionality, but if you say he is aware we can merge immediately.

mcsteplogger.sh
```diff
@@ -1,10 +1,10 @@
 package: MCStepLogger
 version: "%(tag_basename)s"
-tag: "v0.2.0"
+tag: "v0.3.0"
 source: https://github.com/AliceO2Group/VMCStepLogger.git
 requires:
   - "GCC-Toolchain:(?!osx)"
-  - ROOT
+  - VMC
```
sawenzel, line 7: I would prefer to keep the full list of dependencies. It cannot hurt (for people not knowing that VMC will bring in ROOT).

benedikt-voelkel:
@ktf it is not super urgent, we can wait for @sawenzel 's feedback and then merge it. In the meantime, I will prepare and test the incremental PRs needed to finally move everything to `VMC` standalone

mcsteplogger.sh
benedikt-voelkel: ok, sure, I will make the`ROOT` dep explicit again.

benedikt-voelkel:
@ktf @sawenzel I brought back the ROOT dep.
ok to be merged from my side

sawenzel (approved):

PR #3604, opened 2021-11-25
Bump librdkafka to v1.8.2

awegrzyn:


vascobarroso (approved):

ktf:
Why do you need the setenv?

awegrzyn:
Just for development purposes, but indeed I can keep it in my fork

ktf:
Ok. Keep in mind that the environment provided by the modulefiles is not meant, by design, to include the development variables you need to build. In general my reccomandation is to make sure the cmakecache / compile_commands.json is self contained and does not need extra development information. If you want to have the environment as used by alibuild you should source etc/profile.d/init.sh.

ktf:
Errors unrelated. Relevant part compiled. Merging.

PR #3616, opened 2021-11-30
Install libs to lib instead of lib64

benedikt-voelkel:


benedikt-voelkel:
Done to harmonise with other packages which would install to `lib64` and therefore explicitly told to install to `lib`. 
Also fixes the automatic module generation with `alibuild-generate-module --bin --lib > $MODULEFILE` which does not work with `lib64`

PR #3619, opened 2021-11-30
FST: enlarge testing to include basic analysis checks

sawenzel:


o2-full-system-test.sh
```diff
@@ -24,11 +24,16 @@ $O2_ROOT/prodtests/full_system_test_ci_extra_tests.sh
 popd
 rm -Rf $BUILDDIR/full-system-test-sim
 
+# during build, the Python paths are not setup (in contrast to a full module load env)
+# python is needed by the O2DPG scheduler
+export PYTHONPATH=${PYTHONPATH}:${PYTHON_MODULES_ROOT}/share/python-modules/lib/python/site-packages
+
```
TimoWilken, line 30: Would it be better to use the following in the frontmatter instead? Then the path is set for this recipe and for anything that may end up depending on this recipe, I think.
```yaml
append_path:
  PYTHONPATH: "$PYTHON_MODULES_ROOT/share/python-modules/lib/python/site-packages"
```

davidrohr (dismissed):
fine with me if it works.

o2-full-system-test.sh
sawenzel: I tried to put:
```
build_requires:
  - Python
  - Python-modules
append_path:
  PYTHONPATH: "$PYTHON_MODULES_ROOT/share/python-modules/lib/python/site-packages"
```
but it is not working correctly.

ktf: append_path is for the dependent of the recipe, not for the build itself, so it needs to be there in the `Python-modules` recipe, not in this one.

ktf: The issue is the LHAPDF recipe, which resets the PYTHONPATH.

ktf (changes requested):
We should fix LHAPDF instead.

ktf:
https://github.com/alisw/alidist/pull/3625 should fix it.

o2-full-system-test.sh
```diff
@@ -24,11 +24,16 @@ $O2_ROOT/prodtests/full_system_test_ci_extra_tests.sh
 popd
 rm -Rf $BUILDDIR/full-system-test-sim
 
+# during build, the Python paths are not setup (in contrast to a full module load env)
+# python is needed by the O2DPG scheduler
+export PYTHONPATH=${PYTHONPATH}:${PYTHON_MODULES_ROOT}/share/python-modules/lib/python/site-packages
```
ktf, line 29:
```suggestion
```

ktf:
#3625 is now merged, so if you could remove the PYTHONPATH part, we can merge this.

sawenzel:
recipe updated. Should be ok now.

ktf (approved):

PR #3627, opened 2021-12-02
xjalienfs tag 1.3.7

adriansev:
hot bug-fix release : 
* return the right exitcode of the last executed command
O2 CCDB API use this return exitcode to check for token validity and presence


adriansev:
@ktf @costing could this be fast-tracked? Thanks a lot!

zensanp (approved):

costing (approved):

PR #3634, opened 2021-12-06
ignore .DS_Store files

mario-krueger:


ktf:
Fine, however if you have a .DS_Store it probably means you need to disable spotlight for your build directory...

PR #3639, opened 2021-12-08
Go to head of O2DPG - just for now for testing, will be done automatically eventually

davidrohr:


o2dpg.sh
```diff
@@ -1,6 +1,6 @@
 package: O2DPG
 version: "%(tag_basename)s"
-tag: master
+tag: 9ba38e202024c2d844bbafca191331155a685be1
```
ktf, line 3: I think it's enough to drop "version".

davidrohr: ok, but I thought our conclusion was to retag daily?
So we would not drop the `version`?

davidrohr:
merging, to go ahead with changes in O2.
We can clean this up afterwards.

PR #3640, opened 2021-12-08
Move to VMC standalone

benedikt-voelkel:
introduce necessary recipe updates

* add to simulation recipe but make also explicit in individual VMC
  recipes

* remove ENABLE_VMC

* O2sim built tested with altered O2 and AEGIS locally

* full_system_test runs locally

* disable VMC where ROOT5 is built

* move to foreseen AEGIS branch alice/v1.4

* keep `VMC` as explicit dependency in `simulation` recipe to be more transparent

* provide symlink `libVMC.so` -> `libVMCLibrary.so`, in particular for older scripts that rely on things like `gSystem->Load("libVMC")`

### DIFF
--- a/aegis.sh
+++ b/aegis.sh
@@ -1,6 +1,6 @@
 package: AEGIS
 version: "%(tag_basename)s"
-tag: alice/v1.4
+tag: v1.4-alice1
 requires:
   - ROOT
   - VMC
--- a/aegis.sh
+++ b/aegis.sh
@@ -3,6 +3,7 @@ version: "%(tag_basename)s"
 tag: v1.4
 requires:
   - ROOT
+  - VMC
   - pythia6
 build_requires:
   - CMake
--- a/aegis.sh
+++ b/aegis.sh
@@ -1,6 +1,6 @@
 package: AEGIS
 version: "%(tag_basename)s"
-tag: v1.4
+tag: alice/v1.4
 requires:
   - ROOT
   - VMC
--- a/defaults-o2.sh
+++ b/defaults-o2.sh
@@ -3,6 +3,7 @@ env:
   CMAKE_BUILD_TYPE: RELWITHDEBINFO
   CXXFLAGS: -fPIC -O2 -std=c++17
   CXXSTD: '17'
+  ENABLE_VMC: 'ON'
   GEANT4_BUILD_MULTITHREADED: 'ON'
 overrides:
   AliPhysics:
--- a/defaults-o2.sh
+++ b/defaults-o2.sh
@@ -3,7 +3,6 @@ env:
   CMAKE_BUILD_TYPE: RELWITHDEBINFO
   CXXFLAGS: -fPIC -O2 -std=c++17
   CXXSTD: '17'
-  ENABLE_VMC: 'ON'
   GEANT4_BUILD_MULTITHREADED: 'ON'
 overrides:
   AliPhysics:
--- a/fluka_vmc.sh
+++ b/fluka_vmc.sh
@@ -1,10 +1,11 @@
 package: FLUKA_VMC
 version: "%(tag_basename)s"
-tag: "4-1.1-vmc1"
+tag: "4-1.1-vmc2"
 source: https://gitlab.cern.ch/ALICEPrivateExternals/FLUKA_VMC.git
 requires:
   - "GCC-Toolchain:(?!osx)"
   - ROOT
+  - VMC
 build_requires:
   - CMake
   - FLUKA
@@ -41,7 +42,7 @@ proc ModulesHelp { } {
 set version $PKGVERSION-@@PKGREVISION@$PKGHASH@@
 module-whatis "ALICE Modulefile for $PKGNAME $PKGVERSION-@@PKGREVISION@$PKGHASH@@"
 # Dependencies
-module load BASE/1.0 ${ROOT_REVISION:+ROOT/$ROOT_VERSION-$ROOT_REVISION} ${GCC_TOOLCHAIN_REVISION:+GCC-Toolchain/$GCC_TOOLCHAIN_VERSION-$GCC_TOOLCHAIN_REVISION}
+module load BASE/1.0 ${ROOT_REVISION:+ROOT/$ROOT_VERSION-$ROOT_REVISION} ${VMC_REVISION:+VMC/$VMC_VERSION-$VMC_REVISION} ${GCC_TOOLCHAIN_REVISION:+GCC-Toolchain/$GCC_TOOLCHAIN_VERSION-$GCC_TOOLCHAIN_REVISION}
 # Our environment
 set FLUKA_VMC_ROOT \$::env(BASEDIR)/$PKGNAME/\$version
 setenv FLUKA_VMC_ROOT \$FLUKA_VMC_ROOT
--- a/fluka_vmc.sh
+++ b/fluka_vmc.sh
@@ -4,7 +4,6 @@ tag: "4-1.1-vmc2"
 source: https://gitlab.cern.ch/ALICEPrivateExternals/FLUKA_VMC.git
 requires:
   - "GCC-Toolchain:(?!osx)"
-  - ROOT
   - VMC
 build_requires:
   - CMake
--- a/fluka_vmc.sh
+++ b/fluka_vmc.sh
@@ -4,6 +4,7 @@ tag: "4-1.1-vmc2"
 source: https://gitlab.cern.ch/ALICEPrivateExternals/FLUKA_VMC.git
 requires:
   - "GCC-Toolchain:(?!osx)"
+  - ROOT
   - VMC
 build_requires:
   - CMake
--- a/geant3.sh
+++ b/geant3.sh
@@ -1,8 +1,9 @@
 package: GEANT3
 version: "%(tag_basename)s"
-tag: v3-7
+tag: v3-9
 requires:
   - ROOT
+  - VMC
 build_requires:
   - CMake
   - "Xcode:(osx.*)"
@@ -40,7 +41,7 @@ proc ModulesHelp { } {
 set version $PKGVERSION-@@PKGREVISION@$PKGHASH@@
 module-whatis "ALICE Modulefile for $PKGNAME $PKGVERSION-@@PKGREVISION@$PKGHASH@@"
 # Dependencies
-module load BASE/1.0 ROOT/$ROOT_VERSION-$ROOT_REVISION
+module load BASE/1.0 ROOT/$ROOT_VERSION-$ROOT_REVISION VMC/$VMC_VERSION-$VMC_REVISION
 # Our environment
 set GEANT3_ROOT \$::env(BASEDIR)/$PKGNAME/\$version
 setenv GEANT3_ROOT \$GEANT3_ROOT
--- a/geant3.sh
+++ b/geant3.sh
@@ -2,7 +2,6 @@ package: GEANT3
 version: "%(tag_basename)s"
 tag: v3-9
 requires:
-  - ROOT
   - VMC
 build_requires:
   - CMake
--- a/geant3.sh
+++ b/geant3.sh
@@ -2,6 +2,7 @@ package: GEANT3
 version: "%(tag_basename)s"
 tag: v3-9
 requires:
+  - ROOT
   - VMC
 build_requires:
   - CMake
--- a/geant4_vmc.sh
+++ b/geant4_vmc.sh
@@ -1,9 +1,10 @@
 package: GEANT4_VMC
 version: "%(tag_basename)s"
-tag: "v5-3"
+tag: "v5-4"
 source: https://github.com/vmc-project/geant4_vmc
 requires:
   - ROOT
+  - VMC
   - GEANT4
   - vgm
 build_requires:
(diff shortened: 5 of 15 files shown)


benedikt-voelkel:
@TimoWilken @ktf 

The CI will break at this point. Can we just stop it now already?

TimoWilken:
I've marked the PR as draft, that'l stop the PR from running.

benedikt-voelkel:
@ihrivnac @sawenzel 
Here is a draft PR. I could build `O2sim`

As mentioned by @davidrohr on JIRA (https://alice.its.cern.ch/jira/browse/O2-2689?filter=-3) AEGIS needs a twist as well which I did locally.

I ran the `full_system_test` which works just fine

Let's discuss details here. In the meantime I will open a PR for O2 and AEGIS.

Also, another test with `aliBuild build AliPhysics --defaults release` is running which should build with ROOT5. Let's see, I expect that to not be affected at all

benedikt-voelkel:
As a ROOT5 build:
`aliBuild build AliPhysics --defaults release` runs through, all `AliRoot` and `AliPhysics` tests succeed

sawenzel:
To me, this looks ok.

benedikt-voelkel:
I am working on `AliRoot` at the moment and will check if that works as well. I am going to post the result here. If that goes through as well, I will open the O2 and AliRoot PRs

benedikt-voelkel:
Be backwards compatible and provide a symlink `libVMC.so` to `libVMCLibrary.so` so that `AliPhysics` and `AliRoot` macros can find it when using things like `gSystem->Load("libVMC")`

aegis.sh
```diff
@@ -3,6 +3,7 @@ version: "%(tag_basename)s"
 tag: v1.4
 requires:
   - ROOT
+  - VMC
```
ihrivnac, line 6: This dependence is surprising; I looked in the code and the only usage of VMC is to access mag field in GeneratorCosmics:
 ```c++
 //
  auto fld = TGeoGlobalMagField::Instance()->GetField();
  if (TVirtualMC::GetMC() && TVirtualMC::GetMC()->GetMagField()) {
    fld = TVirtualMC::GetMC()->GetMagField();
  }
```
Does it make sense? I thought, that we always define field via TGeo and then set it to VMC.

File: simulation.sh [hunk 2]
```diff
@@ -1,6 +1,7 @@
 package: simulation
 version: v1.0
 requires:
+  - VMC
```
ihrivnac, line 4: I think that this is not needed, as the dependence on VMC is already declared in GEANT3 and GEANT4_VMC

vmc.sh
```diff
@@ -17,6 +19,7 @@ cmake "$SOURCEDIR"                             \
     ${CXXSTD:+-DCMAKE_CXX_STANDARD=$CXXSTD}
 
 cmake --build . -- ${JOBS:+-j$JOBS} install
+ln -s $INSTALLROOT/lib/libVMCLibrary.so $INSTALLROOT/lib/libVMC.so
```
ihrivnac, line 22: Why is this link needed?

ihrivnac: I see, that you commented this above.

simulation.sh
benedikt-voelkel: It is not needed, just to make it explicit. But we can remove it of course

vmc.sh
benedikt-voelkel: It's when ROOT macros from AliRoot or AliPhysics do something like `gSystem->Load("libVMC")`

simulation.sh
benedikt-voelkel: indeed @ihrivnac should we make it explicit in the `fairroot.sh` recipe rather than here? I would just like to have it explicitly in central recipes so that it is clear what happens.

ihrivnac: I don't see a need for making it explicit here or in FairRoot, as the VMC is anyway nowhere else used (except for the AEGIS case where its usage is disputable)

aegis.sh
benedikt-voelkel: mark as resolved for now, coming back to it if necessary

benedikt-voelkel:
we should merge https://github.com/AliceO2Group/AEGIS/pull/11 and https://github.com/AliceO2Group/AliceO2/pull/7841 before we start the CI chain here

simulation.sh
ihrivnac: ROOT dependency can be dropped in FairRoot. We have also many explicit dependencies on ROOT in O2, and we do not add ROOT in requires in o2.sh as it is already implicitly included. Do you find this misleading as well?

fluka_vmc.sh
```diff
@@ -1,10 +1,11 @@
 package: FLUKA_VMC
 version: "%(tag_basename)s"
-tag: "4-1.1-vmc1"
+tag: "4-1.1-vmc2"
 source: https://gitlab.cern.ch/ALICEPrivateExternals/FLUKA_VMC.git
 requires:
   - "GCC-Toolchain:(?!osx)"
   - ROOT
```
ihrivnac, line 7: We can remove ROOT as VMC already depends on ROOT

geant3.sh
```diff
@@ -1,8 +1,9 @@
 package: GEANT3
 version: "%(tag_basename)s"
-tag: v3-7
+tag: v3-9
 requires:
   - ROOT
```
ihrivnac, line 5: We can remove ROOT as VMC already depends on ROOT

geant4_vmc.sh
```diff
@@ -1,9 +1,10 @@
 package: GEANT4_VMC
 version: "%(tag_basename)s"
-tag: "v5-3"
+tag: "v5-4"
 source: https://github.com/vmc-project/geant4_vmc
 requires:
   - ROOT
```
ihrivnac, line 6: We can remove ROOT as VMC already depends on ROOT

simulation.sh
benedikt-voelkel: ok, true. If we remove ROOT from FairRoot then I agree
* remove ROOT from all VMC recipes
* remove ROOT from FairRoot
* remove VMC from simulation

I will test that and add it to the PR

geant3.sh
benedikt-voelkel: removed

geant4_vmc.sh
benedikt-voelkel: removed

fluka_vmc.sh
benedikt-voelkel: removed

simulation.sh
benedikt-voelkel: I have removed the requirements as listed in my previous comment

ihrivnac: Great, thank you

ktf:
Please remember to remove [Draft] and change it a normal PR when you are ready.

davidrohr:
FairROOT seems to fail to build without ROOT

fluka_vmc.sh
ihrivnac: It turns out that this is required when we build FairRoot without simulation; can you put ROOT back in fairroot.sh?

benedikt-voelkel:
Hello @ktf
It seems the failing mac CI is unrelated to the changes. Seems to be a problem with `autotools` there

ktf:
@MichaelLettrich was the OS updated?

MichaelLettrich:
@benedikt-voelkel  @ktf : the macOS version is Big Sur (11.6.1)

```bash
alibuild@aido2osx3 ~ % xcodebuild -version
Xcode 13.1
Build version 13A1030d

alibuild@aido2osx3 ~ % which autoconf      
/usr/local/bin/autoconf

alibuild@aido2osx3 ~ % autoconf --version
autoconf (GNU Autoconf) 2.71
Copyright (C) 2021 Free Software Foundation, Inc.
License GPLv3+/Autoconf: GNU GPL version 3 or later
<https://gnu.org/licenses/gpl.html>, <https://gnu.org/licenses/exceptions.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

Written by David J. MacKenzie and Akim Demaille.
```

anything else you need?

ktf:
Maybe we should reboot the machine? @MichaelLettrich can you take care of it?

MichaelLettrich:
@ktf done

benedikt-voelkel:
Thanks @MichaelLettrich 
Could you by any chance restart the macOS CI? I don't seem to have the power to do so

MichaelLettrich:
@benedikt-voelkel  I've restarted the PR and it looks like it reproduces the same error.

benedikt-voelkel:
Thanks @MichaelLettrich 
I saw it, looks weird to me as I can't see how this is related to any of the changes I made.
Maybe it is possible to run on another mac machine? Unfortunately, I don't have access to a mac myself, in that case I would try to reproduce/see what the problem is.

MichaelLettrich:
All alidist PRs are built on the same mac and we clearly have PRs that pass the CI, so it should be related to this PR.

benedikt-voelkel:
With Michael's help I was able to build `aliBuild build O2Suite --defaults o2` on the mac machine `aido2osx3` in `/Users/alibuild/bvolkel`
I can run the simulation and the library path to the VMC lib is correctly resolved (see `/Users/alibuild/bvolkel/test_vmc_standalone/dyld.log`)
Is it possible that for this CI the build area is somehow corrupted?

MichaelLettrich:
@ktf @TimoWilken  so now the CI machine was restarted, the build area was cleared and @benedikt-voelkel managed to build it by hand on the CI machine but the build still fails when running on CI. Is there anything specific in the CI environment that can trigger the issue on macOS ?

benedikt-voelkel:
Indeed, it is the same as before. I don't know what I can do from here. As I said, when I did it on the machine, the only thing was checking out this `alidist` branch and running `aliBuild` as described above.
If anyone has a hint, please  let me know.
Again, `autotools` is built way before any of the packages I changed the recipes for, right? I don't see how my changes could have an impact on that.

I could of course rebase the PR on top of the latest `alidist@master` and trigger a new CI run if that would be a possibility.

ktf:
In the past the autotools issues have been issues with XCode on the machine not being properly updated. I will merge this, since the arm version is green. Let's keep an eye on it.

ktf:
Actually, before we merge this, since it will trigger a full rebuild, @TimoWilken could you please warm up the cache?

Alternatively, do we really need to modify so many defaults? Can't we change those where we need to say "NO" or have some logic which is default independent?

ihrivnac:
On my Mac (11.5.2), autotools are taken from the system:
```
[MacBook-Pro-5] ~/work/projects/aliceO2 > aliBuild --defaults o2 -j 4 build O2

==> aliBuild can take the following packages from the system and will not build them:
      GSL, FreeType, CMake, libxml2, zlib, libpng, autotools, lzma, bz2, `lz4`
```

benedikt-voelkel:
Concerning the defaults:
In the `VMC` recipe we could check if ROOT is version 5. In that case we consider the `VMC` build done, not doing anything. You think that would work? I can try it in that case.

That means that the `release` and `prod-latest` defaults need no change **IF** we then also put `ROOT` back to the `Geant3` and `Geant4_VMC` recipes. Because otherwise - in these `ROOT5` modes, they wouldn't find `ROOT`

I would say, all other defaults containing `ENABLE_VMC` should be changed so this logic disappears.

ktf:
To me it sounds like a good idea. Having to rebuild the whole stack for this kind of change is a bit of a pain...

benedikt-voelkel:
Ok, I will give that a shot! Thanks!

vmc.sh
```diff
@@ -8,21 +8,38 @@ build_requires:
...
+  MODULEDIR="$INSTALLROOT/etc/modulefiles"
+  MODULEFILE="$MODULEDIR/$PKGNAME"
+  mkdir -p "$MODULEDIR"
+  alibuild-generate-module --bin --lib > $MODULEFILE
+  cat >> "$MODULEFILE" <<EoF
+  setenv VMC_ROOT \$PKG_ROOT
+  prepend-path ROOT_INCLUDE_PATH \$PKG_ROOT/include/vmc
 EoF
+
+else
```
benedikt-voelkel, line 39: Is this approach acceptable?

ktf: apart from the fact I would do:

```bash
case $ROOT_VERSION in
  6*) ... ;;
  *) .. ;;
esac
```

it seems fine to me. You could also generate the module file once, before actually building the package.

vmc.sh
```diff
@@ -8,21 +8,38 @@ build_requires:
...
+  ROOT_INCLUDE_PATH: "$VMC_ROOT/include/vmc"
 ---
 #!/bin/bash -e
-cmake "$SOURCEDIR"                             \
-  -DCMAKE_CMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE} \
-  -DCMAKE_INSTALL_PREFIX="$INSTALLROOT"        \
-  -DCMAKE_INSTALL_LIBDIR=lib                   \
+
+root_version=$(root-config --version)
+root_version=${root_version:0:1}
```
ktf, line 17: There is already ROOT_VERSION which is defined by alibuild the moment you require root.

defaults-o2.sh
```diff
@@ -3,7 +3,6 @@ env:
   CMAKE_BUILD_TYPE: RELWITHDEBINFO
   CXXFLAGS: -fPIC -O2 -std=c++17
   CXXSTD: '17'
-  ENABLE_VMC: 'ON'
```
ktf, line 6: Does it arm to leave it there for now? this will trigger a rebuild as well...

benedikt-voelkel: ok, will leave that and only remove the corresponding line in `root.sh`

vmc.sh
benedikt-voelkel: I used that as suggested

defaults-o2.sh
benedikt-voelkel: no defaults are changed now

vmc.sh
benedikt-voelkel: Modulfile is created at the beginning and only updated with the corresponding env variables if VMC is actually built

ktf (approved):

benedikt-voelkel:
Concerning failing mac CI this time: https://github.com/AliceO2Group/O2Physics/pull/432#issuecomment-1012408319

ktf:
The error is indeed genuine, albeit unrelated. I would say we can merge this after @TimoWilken has warmed up the cache.

TimoWilken:
Caching is done!

PR #3651, opened 2021-12-13
Bump JAliEn to 1.4.5

zensanp:
Update JAliEn (Java components only). Doesn't affect AliPhysics / O2 in any way.

Also bumps XRootD in defaults-jalien to v5.4.0.

### DIFF
--- a/defaults-jalien.sh
+++ b/defaults-jalien.sh
@@ -17,7 +17,7 @@ overrides:
     tag: v3.4.0_1.045-alice1
   XRootD:
     source: https://github.com/xrootd/xrootd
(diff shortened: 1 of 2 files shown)


adriansev (approved):

ktf:
The multi_node test is failing again. @costing last time this happened I think it was a problem with the machine. Could you please check alissandra02?

costing:
@ktf since the raid controller configuration change the machines are responsive, there is no IOWait on them, builds seem fast (at least from the cpu usage saw-like profile), so the current issue seems to be different.

Thanks for the quick merge!

PR #3655, opened 2021-12-14
Use builtin VMC fallback

benedikt-voelkel:
* use ROOT's 'builtin VMC as fallback for MC (bumped to version v0.4.0)

* update VMC recipe already to most recent

* temporarily VMC recipe is not used

### DIFF
--- a/mcsteplogger.sh
+++ b/mcsteplogger.sh
@@ -1,11 +1,10 @@
 package: MCStepLogger
 version: "%(tag_basename)s"
-tag: "v0.3.0"
+tag: "v0.4.0"
 source: https://github.com/AliceO2Group/VMCStepLogger.git
 requires:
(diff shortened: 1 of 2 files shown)


benedikt-voelkel:
@ktf I decided we should do the transition now as smooth as possible. I added the possibility of falling back to ROOT's builtin VMC for the MCSTepLogger now.
With this PR there is temporarily no recipe that depends on VMC
If that goes through, we can concentrate in the O2 PR, cause in that case we will not anymore have any possible header and library clashes since only ROOT VMC is there

benedikt-voelkel:
Is it possible to merge this and then tackle the O2 PR from there?

ktf:
Yes, sure. I assume you already got greenlight by @sawenzel, so merging this.

PR #3658, opened 2021-12-15
Bump OpenSSL to 1.1.1m

ktf:


pzhristov:
@kft Interesting if this version compiles with all our packages :) When I tested it six months ago, we had many issues.

adriansev:
note that openssl 1.1 needs kerberos and it lacks the option `no-krb5`

adriansev:
also take note of this: https://wiki.openssl.org/index.php/Compilation_and_Installation#PREFIX_and_OPENSSLDIR
```
--prefix and --openssldir control the configuration of installed components. The behavior and interactions of --prefix and --openssldir are slightly different between OpenSSL 1.0.2 and below, and OpenSSL 1.1.0 and above.

The rule of thumb to use when you want something that "just works" for all recent versions of OpenSSL, including OpenSSL 1.0.2 and 1.1.0, is:

specify both --prefix and --openssldir
set --prefix and --openssldir to the same location
```

ktf:
@adriansev as far as I understand it from https://www.mail-archive.com/openssl-users@openssl.org/msg86201.html, kerberos is completely gone, not required.

ktf:
@adriansev ok, adapted openssldir.
@pzhristov let's see, a bunch of externals were updated in the last few months, in particular XRootd. If this works, we can always fallback to 1.0 for AliRoot and co.

adriansev:
> @adriansev as far as I understand it from [mail-archive.com/openssl-users@openssl.org/msg86201.html](https://www.mail-archive.com/openssl-users@openssl.org/msg86201.html), kerberos is completely gone, not required.

oh, thanks for info! I mis-remembered this point, sorry for wrong info

adriansev:
> @adriansev ok, adapted openssldir. @pzhristov let's see, a bunch of externals were updated in the last few months, in particular XRootd. If this works, we can always fallback to 1.0 for AliRoot and co.

@ktf i think that also the step of module generation should be adapted for the new openssldir:
```
++ alibuild-generate-module --lib --bin
/sw/SPECS/slc8_x86-64/OpenSSL/v1.1.1m-local1/OpenSSL.sh: line 43: etc/modulefiles/OpenSSL: No such file or directory
```

adriansev:
@ktf and rsync part :(
```
++ cat
++ mkdir -p /sw/INSTALLROOT/85bffd70bedc0b2ca1c789251764b739fa67d4da/slc7_x86-64/OpenSSL/v1.1.1m-local1/etc/modulefiles
++ rsync -a --delete etc/modulefiles/ /sw/INSTALLROOT/85bffd70bedc0b2ca1c789251764b739fa67d4da/slc7_x86-64/OpenSSL/v1.1.1m-local1/etc/modulefiles
rsync: change_dir "/sw/BUILD/85bffd70bedc0b2ca1c789251764b739fa67d4da/OpenSSL//etc/modulefiles" failed: No such file or directory (2)
rsync error: some files/attrs were not transferred (see previous errors) (code 23) at main.c(1179) [sender=3.1.2]
```

adriansev:
`xrootd v3.3.6` will never work with openssl 1.1 unless someone invest the time for openssl 1.1 support

adriansev:
for build/O2/alidist-cs8 (our) curl should be downloaded before and build just after openssl to avoid:
```
++ mkdir -p /sw/INSTALLROOT/b75b5c6ff988609d11bc3a2503b6a2390ef0e60d/slc8_x86-64/CMake/v3.19.2-local2/etc/modulefiles
++ rsync -a --delete etc/modulefiles/ /sw/INSTALLROOT/b75b5c6ff988609d11bc3a2503b6a2390ef0e60d/slc8_x86-64/CMake/v3.19.2-local2/etc/modulefiles
## End sw/BUILD/CMake-latest/log
## Begin sw/BUILD/FreeType-latest/log
++ URL=http://download.savannah.gnu.org/releases/freetype/freetype-2.10.1.tar.gz
++ curl -L -o freetype.tgz http://download.savannah.gnu.org/releases/freetype/freetype-2.10.1.tar.gz
curl: symbol lookup error: /lib64/libk5crypto.so.3: undefined symbol: EVP_KDF_ctrl, version OPENSSL_1_1_1b
```

ktf:
Even better, we should not fetch freetype via curl.

ktf:
> xrootd v3.3.6 will never work with openssl 1.1 unless someone invest the time for openssl 1.1 support

@adriansev agreed, that's why I am trying to have a recipe which works for both, so that we  do not need to play special tricks for aliroot / aliphysics.

ktf:
Not sure what is wrong, but seems to be consistently wrong.

ktf:
Superseded by #4019.

PR #3660, opened 2021-12-15
Bump AliECS to v0.30.1

teo:


vascobarroso (approved):

ktf:
Mac issue seems to be genuine.

ktf (changes requested):
Can you please check? I think the issue on mac is real.

teo:
@ktf Yep, confirmed. I'm using a linker flag that's unsupported on Mac. Fix incoming.

vascobarroso (approved):

ktf (approved):
Ok, if it compiles on macOS.

PR #3657, opened 2021-12-15
[QC-716] Add correct ROOT_INCLUDE_PATH

Barthelemy:

PR #3662, opened 2021-12-15
Fix module file

benedikt-voelkel:


benedikt-voelkel:
Fixes problem introduced by https://github.com/alisw/alidist/pull/3657

ktf:
this looks wrong to me . what is the error?

ctolon:
I'm also getting problem O2 env.  Logs here:

Loading libInfoLogger/v2.4.1-3
  Module ERROR: can't read "PKG_ROOT": no such variable
    In '/home/batu/alice/sw/MODULES/ubuntu2004_x86-64/libInfoLogger/v2.4.1-3'
    Please contact <root@localhost>

benedikt-voelkel:
Let me check again

ktf:
https://github.com/alisw/alidist/pull/3665 should be the correct fix.

benedikt-voelkel:
Yes, sorry, I just pushed the wrong thing. Thanks!

ktf:
So this was merely removing the broken packages, I think, which is obviously wrong. The actual fixes are in #3663 and #3662 which use the correct variable for the root of the package (PKG_ROOT is only available when using the alibuild-generate-module helper).

Barthelemy:
sorry about that. There is also the MONITORING_ROOT to use. I can prepare a last PR.

Barthelemy:
https://github.com/alisw/alidist/pull/3666

awegrzyn:
@Barthelemy I don't think it is necessary

ktf:
@Barthelemy indeed I posted the recipe for the automatically generated modulefiles, but forgot that you still had non automatically generated ones.

PR #3665, opened 2021-12-16
Fix modulefile for libinfologger

ktf:


ktf:
PKG_ROOT is only available when using alibuild-generate-module.

PR #3666, opened 2021-12-16
Fix the modules file for monitoring

Barthelemy:


ktf:
No, this is correct, IMHO. Do you actually see an error?

awegrzyn (changes requested):
Monitoring uses autogenerated modulefiles therefore I don't think it is necessary

PR #3672, opened 2021-12-17
Force libuv 1.40 as minimum dependency

ktf:


ktf:
All tests passed. Only difference is in the comment. Merging. This is required by the latests changes in O2.

PR #3714, opened 2022-01-10
Bump ReadoutCard to v0.35.0

kostorr:


ktf:
@TimoWilken this has had "build/O2/alidist" in `Expected — Waiting for status to be reported` since yesterday. Could you have a look?

PR #3727, opened 2022-01-13
Add own valgrind

ktf:
Apparently compiling with GCC 10.2 and using system valgrind creates troubles on some systems (e.g. Ubuntu 18.04) which are not able to properly find the  debug symbols. Using valgrind consistently compiled for our stack solves that problem.


ktf:
Not being tested. Merging.

PR #3730, opened 2022-01-13
Alidist - freetype.sh - change system check

maireiphc:
Apparently, the command 'freetype-config' has been deprecated. 
e.g. it has been removed from the package libfreetype6-dev in Ubuntu 20.04. 
The command used here to check the Freetype version potentially available on system (system check) will never work under Ubuntu 20.04.

It is by now recommended to use pkg-config i.e. 'pkg-config freetype2 --cflags'

. Under Ubuntu 18.04 LTS, both option should work (freetype-config redirect silently towards pkg-config, transitional package).
Not tested but expected from http://manpages.ubuntu.com/manpages/bionic/en/man1/freetype-config.1.html
. Under CentOS7, same : both options work (commands tested on lxplus)

I checked the response of the command under Ubuntu 20.04 LTS.
After the change on such an OS, 'aliDoctor AliGenerators' plans now to take FreeType from the system.

### DIFF
--- a/freetype.sh
+++ b/freetype.sh
@@ -7,7 +7,7 @@ build_requires:
   - system-curl
 prefer_system: (?!slc5)
 prefer_system_check: |
-  printf "#include <ft2build.h>\n" | c++ -xc++ - `pkg-config freetype2 --cflags` -c -M 2>&1;
+  printf "#include <ft2build.h>\n" | c++ -xc++ - `freetype-config --cflags 2>/dev/null` `pkg-config freetype2 --cflags 2>/dev/null` -c -M 2>&1;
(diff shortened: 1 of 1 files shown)


freetype.sh
```diff
@@ -7,7 +7,7 @@ build_requires:
   - system-curl
 prefer_system: (?!slc5)
 prefer_system_check: |
-  printf "#include <ft2build.h>\n" | c++ -xc++ - `freetype-config --cflags` -c -M 2>&1;
+  printf "#include <ft2build.h>\n" | c++ -xc++ - `pkg-config freetype2 --cflags` -c -M 2>&1;
```
ktf, line 10: I would suggest we leave both:

```suggestion
  printf "#include <ft2build.h>\n" | c++ -xc++ - `freetype-config --cflags 2>/dev/null` `pkg-config freetype2 --cflags 2>/dev/null` -c -M 2>&1;
```

since there are platforms (most notably mac) where pkgconfig is not (yet?) a dependency of the freetype you get. Having both should not have any significant side effect.

ktf:
prefer_system_recipe is not part of the hash, so we do not need warming the cache.

PR #3740, opened 2022-01-17
Make sure ZeroMQ is added as AliRoot dependency in modulefile if needed

ktf:


ktf:
@pzhristov this is needed to have ZeroMQ being picked up as a runtime dependency from AliRoot (besides adding the `requires: ZeroMQ` at the top, which I understood you already did).

PR #3741, opened 2022-01-17
Revert "Use builtin VMC fallback"

ktf:
Reverts alisw/alidist#3655


benedikt-voelkel:
MCStepLogger `v0.3.0` only work with VMC standalone. Actually `v0.4.0` uses fallback to builtin. This revert should not be necessary.
The changes with the bumped VMC tag only check internally during the VMC build whether or not ROOT was built with its builtin VMC

PR #3749, opened 2022-01-19
Move to VMC standalone

benedikt-voelkel:
* swap logic of ENABLE_VMC: only if set, VMC standalone is built
  -> for all other ROOT6 builds, builtin is still used
     * defaults-next-root6
     * defaults-user-next-root6
     * defaults-jalien
     * defaults-o2-dataflow (see below)
  -> ROOT5 builds not affected

* default-o2-dataflow not changed, no simulation used, so in that case
  ROOT would still be built with buitlin VMC

* defaults-generators not primarily meant for builds including detector
  simulation -> ENABLE_VMC not set

### DIFF
--- a/aegis.sh
+++ b/aegis.sh
@@ -1,8 +1,9 @@
 package: AEGIS
 version: "%(tag_basename)s"
-tag: v1.4
+tag: v1.4-alice1
 requires:
   - ROOT
+  - VMC
   - pythia6
 build_requires:
   - CMake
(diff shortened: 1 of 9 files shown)


davidrohr (approved):
nice, tested it on my local setup with system installation of ROOT 6 without buildit VMC and works nicely

benedikt-voelkel:
Thanks for the x-check David!
The full CI now reports a copyright problem unrelated to this PR it seems.

benedikt-voelkel:
Maybe I should also confirm here that I of course checked and built `AliPhysics@master` with `next-root6` defaults

benedikt-voelkel:
@ktf could you let me know if there is anything to be addressed or if any other x-check is needed before merging this?

ktf:
It would have been good to warm the cache, since this will rebuild ROOT, but I guess it's too late now.

benedikt-voelkel:
I'll keep an eye on AliPhysics and AliRoot this time. Please let me know if you encounter any problems

PR #3750, opened 2022-01-19
Fix check for the year...

ktf:


ktf:
We need to bump the year, otherwise we need to tell people to use 2020 also for new files, and of all the years we should "Greyhound day" ourselves, 2020 is the one I like the least...
I am actually amazed how everyone in 2021 copied the copyright statement from the previous year...
Anyways, it should only affect new files, so it should not be a big deal and actually makes sense that whenever the file is changed we bump the copyright accordingly.

PR #3752, opened 2022-01-19
Update o2checkcode.sh

ktf:


ktf:
Apparently the list of changed files is wrong and we end up with a complain for each copyright statement. :-/ Will look at it tomorrow..

PR #3756, opened 2022-01-20
Introduce standalone TBB

sawenzel:
Build ROOT against standalone TBB (Intel Thread Building Blocks)
and make TBB available for other packages.

### DIFF
--- a/tbb.sh
+++ b/tbb.sh
@@ -1,0 +1,25 @@
+package: TBB
+version: "v2021.5.0"
+tag: v2021.5.0
+source: https://github.com/oneapi-src/oneTBB/
+build_requires:
+ - "GCC-Toolchain:(?!osx)"
+ - CMake
+prefer_system: .*
+prefer_system_check: |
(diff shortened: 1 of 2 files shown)


ktf:
@TimoWilken could you warm up the cache with this?

tbb.sh
```diff
@@ -0,0 +1,35 @@
...
+cmake --build . -- ${JOBS:+-j$JOBS} install
+
+# make soft link of lib64 dir to lib
+pushd "$INSTALLROOT"
+  # Fix ambiguity between lib/lib64
+  if [[ ! -d lib && -d lib64 ]]; then
+    ln -nfs lib64 lib
+  elif [[ -d lib && ! -d lib64 ]]; then
+    ln -nfs lib lib64
+  fi
```
ktf, line 27: Why not simply using `-DCMAKE_INSTALL_LIBDIR=lib`?

sawenzel: no idea. I took the inspiration from different recipes in alidist

ktf: Thank you for your update. Indeed we should probably fix also the other recipes.

ktf (approved):

TimoWilken:
Caching is done.

PR #3762, opened 2022-01-21
Use -O3 and -march=znver2 for builds on the EPN + unrelated but required fixes

davidrohr:


defaults-o2-epn.sh
```diff
@@ -1,7 +1,8 @@
 env:
-  CFLAGS: -fPIC -O2
+  CFLAGS: -fPIC -O3 -march=znver2
```
ktf, line 2: This of course mean that we will do the builds on one of the EPNs, correct?

davidrohr:
Well, that is a question. If we have a cpu with an incompatible instruction set, the build would perhaps fail, since we also run some of the software that we build. On the other hand, usually any avx2 compatible cpu would normally work. Perhaps i would just try with our normal builders, perhaps it just works.

ktf:
Sounds like a plan. Let's see what breaks. Keep in mind that we run not only tests, but also things like cling and build time.

davidrohr:
I ran a build with defaults-epn on my laptop with skylake CPU, and that worked well and is usable on a Rome.
So I guess it will work. If we have much older Intel CPUs in the build farm, perhaps we could pin that particular build to some specific node.

For the deployment: let's wait until some time next week before we deploy this, first we need to update FMQ / ROCm and corresponding O2, once that is stable, we can play with the compiler flags.

PR #3772, opened 2022-01-25
Update ofi to prevent issues with the new version on old platforms (O2-2752)

ktf:


ktf:
@ironMann @davidrohr  @vascobarroso any objections to this? it looks like suddenly ofi does not compile anymore on different platforms (probably due to an update of libatomic on the system).

Barthelemy:
O2-2752

davidrohr (approved):
sure, don't see a problem if it passes the FullCI

awegrzyn:
On FLP production setups we use `libfabric` from CS8 (`1.14.0-1`) so this won't change anything for us.

ktf:
Ok, then even better.

ironMann:
cc @dennisklein

ktf:
Do we actually use the libfabric we ship in alidist anywhere? Maybe we should just drop it, if no one is actually using it.

PR #3776, opened 2022-01-26
Fixed link to libVMCLibrary

pzhristov:


ktf:
i think there is an option to do that at cmake time. Something like CMAKE_LIB_PREFIX.

pzhristov:
@ktf This is a simple fix for already existing and not working link, so we can improve it later.

ktf:
Sorry, I thought this was something else. Your change is fine.

PR #3779, opened 2022-01-26
Use Catalina as deployment target.

pzhristov:
This is the version supported by cling of Root v6-24-06 (MacOSX Monterey)


ktf:
This will trigger a full rebuild on mac. @TimoWilken could you warm up the cache with it?

sawenzel:
It seems to have caused a complete rebuild even on CC7 and CC8.

ktf:
Indeed... Modifying defaults forces a full rebuild everywhere...

PR #3782, opened 2022-01-27
Fix missing LD_LIBRARY_PATH for fmt

ktf:
Now that we have the library version, we must make sure we have the module adapted accordingly.


ktf:
Tested on hyperloop, only modulefile affected. Merging.

PR #3785, opened 2022-01-28
Bump FairMQ to v1.4.47

rbx:
Contains @ktf's contribution with the GUI mode for the built-in controller plugin.
Needed for https://github.com/AliceO2Group/AliceO2/pull/8004.

### DIFF
--- a/fairmq.sh
+++ b/fairmq.sh
@@ -1,6 +1,6 @@
 package: FairMQ
 version: "%(tag_basename)s"
-tag: v1.4.46
+tag: v1.4.47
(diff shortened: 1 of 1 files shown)


ktf (approved):
Thanks. Does it change anything else critical besides my "GUI" plugin?

rbx:
Nothing critical. Fixed a warning and done some changes to how the test suite is executed (now using mostly ipc instead of tcp). And changed from using `FairMQSomething` to ``fair::mq::Something` in many places, but both are still not available and not (yet) deprecated.

PR #3787, opened 2022-01-28
TBB: Add ROOT_INCLUDE_PATH

sawenzel:
Needed by simulation on the GRID

### DIFF
--- a/tbb.sh
+++ b/tbb.sh
@@ -23,4 +23,9 @@ MODULEDIR="$INSTALLROOT/etc/modulefiles"
 MODULEFILE="$MODULEDIR/$PKGNAME"
 mkdir -p "$MODULEDIR"
 alibuild-generate-module --bin --lib > $MODULEFILE
+cat >> "$MODULEFILE" <<EOF
+# extra environment
+set TBB_ROOT \$::env(BASEDIR)/$PKGNAME/\$version
+prepend-path ROOT_INCLUDE_PATH \$TBB_ROOT/include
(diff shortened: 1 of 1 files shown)


sawenzel:
@ktf @TimoWilken : Can we try to have this in today so that GRID nighlies pick it up next morning?

tbb.sh
```diff
@@ -23,4 +23,9 @@ MODULEDIR="$INSTALLROOT/etc/modulefiles"
 MODULEFILE="$MODULEDIR/$PKGNAME"
 mkdir -p "$MODULEDIR"
 alibuild-generate-module --bin --lib > $MODULEFILE
+cat >> "$MODULEFILE" <<EOF
+# extra environment
+set TBB_ROOT \$::env(BASEDIR)/$PKGNAME/\$version
```
TimoWilken, line 28: Do you need a `TBB_ROOT` variable separately or can you use `PKG_ROOT` provided by `alibuild-generate-module`?

sawenzel: probably not needed but it's anyway not exported. I copied the pattern from other recipes.

TimoWilken: Fair enough. I'll merge tonight when the Linux CI is done, but before the nightly tag starts.

ktf: Well, we should probably fix all of them. Anyways, since this is broken let's go ahead with it. @TimoWilken make sure the cache is hot as well as this will rebuild ROOT.

PR #3790, opened 2022-01-29
Update vc.sh

ktf:


ktf:
@davidrohr as you requested. I  was wondering if there was not a better way to accomplish what you need to do which does not involve an environment variable. What are you trying to do?

ktf:
Not needed anymore.

PR #3795, opened 2022-01-30
Fixes for MacOSX Monterey

pzhristov:


ihrivnac:
@pzhristov , would you mind to make a PR with this change in https://github.com/Geant4/geant4 so that we get this change in the next Geant4 patch and can switch back to geant4 distribution in the recipe? Thank you,

PR #3805, opened 2022-02-01
datadist: bump to v1.1.0

ironMann:
This contains support for adoption of existing SHM segments. 
Online run script will need a parameter update (tfbuilder.sh)


davidrohr:
@ironMann : Is this backward-compatible to what is currently running on the FLPs, or do we have to bump in sync?

ironMann:
@davidrohr Yes, this should be compatible with previous version

davidrohr (approved):

PR #3808, opened 2022-02-02
Bump FairMQ to v1.4.48

rbx:
Contains changes for R3C-646 to allow "soft" shmem reset after a crash (previously only possible after an orderly shutdown).

Also adds UnmangedRegion size to shmmonitor debug output.

### DIFF
--- a/fairmq.sh
+++ b/fairmq.sh
@@ -1,6 +1,6 @@
 package: FairMQ
 version: "%(tag_basename)s"
-tag: v1.4.47
+tag: v1.4.48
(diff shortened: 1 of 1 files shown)


davidrohr (approved):
tested in FST, thx for improving this.

vascobarroso:
@davidrohr we are about to make a new FLP Suite release for deployment next Monday. Shall we include this ?

davidrohr:
It should not make a difference anyway, since the change is only for the epn shm management tool. If the CI is green, just include it.

PR #3827, opened 2022-02-08
Bump abseil version

ktf:
We have reports of abseil not working with GCC 11.1 and ubuntu 21.1, so let's see if this is enough to fix it.


ktf:
@teo apparently grpc would need to be updated for this, but then we have our own version and I am not sure it will actually work. Closing for now, but it might bite us in  the future.

PR #3830, opened 2022-02-09
datadist: bump to v1.2.0

ironMann:
This needs synchronized upgrade with FLPs.


davidrohr:
@ironMann : I assume you have already seen that all builds are failing?

davidrohr:
ok, wait, all builds are failing with this, also other PRs.

davidrohr:
this should merged only if we decide to deploy it before the MW2, otherwise the RPMs builds will pick it up automatically.

PR #3836, opened 2022-02-10
We should build ms_gsl with the GCC we also use for O2, not the system one

davidrohr:


sawenzel (approved):

ktf:
AFAICT, ms_gsl is header only. Why would this be needed?

ktf:
Also, depending on CMake brings GCC-Toolchain in the environment indirectly. What is currently broken without this?

ktf:
I would also check the prefer_system check for GCC does not select the system one because it's recent enough, yet it might not have the specific architecture support turned on...If you have a log feel free to point it to me.

ktf (changes requested):
I think this is not actually the issue.

PR #3846, opened 2022-02-15
bump readout.sh

sy-c:
v2.8.2


vascobarroso (approved):

ktf:
There are currently issues with CERN gitlab, which hangs while checking wether some packages need to be rebuilt. I've opened an issue with CERN IT at:

https://cern.service-now.com/service-portal?id=ticket&table=incident&n=INC3059658

I will let you know when the situation improves.

PR #3848, opened 2022-02-16
Bump qualitycontrol to v1.46.0

Barthelemy:


knopers8 (approved):

knopers8:
@TimoWilken @ktf I am a little bit surprised that QC is being automatically updated with nightlies as the default in alidist. I know that this is a requirement for O2DPG and O2sim, but could it be done only for those two, not for the whole alidist?

ktf:
I understood from @sawenzel and previously discussed it with @Barthelemy that this would have been ok. We can of course revisit, but someone will have to give up something because there is clearly conflicting needs. Notice that no one will actually touch the flp branches, so IMHO it makes sense that alidist master is as close as possible to what is actually built in the nightlies.

knopers8:
I had a short chat with @Barthelemy (who is on holidays) and he also had understood it that alidist@master would still use QC releases. Let's clarify this once he's back, we can survive like this for a few days, no worries.

ktf:
Ok. Alternatively, @sawenzel, can't you simply use the daily tags, e.g. https://github.com/alisw/alidist/blob/O2PDPSuite-nightly-20220201 ?

PR #3853, opened 2022-02-17
Increase string buffer and trim install path

mfasDa:
OpenLoops complains at runtime about install path exceeding
the buffersize allocated for strings. There are two reasons
for this:
- The max buffer is allocated to 255 chars, which is on
  modern systems rather conservative, and with the design
  of aliBuild the exceeding of the buffer can be easily
  achieved. 1000 characters is a safer setting.
- The install_path string is not trimmed, therefore any
  buffer will be exceeded (bug in OpenLoops itself)
The buffersize is a config parameter that must be patched
in any case, for the trim function an on-the-fly patch would
allow taking OpenLoops directly from the official repository.

### DIFF
--- a/openloops.sh
+++ b/openloops.sh
@@ -12,6 +12,10 @@ rsync -a --delete --exclude '**/.git' --delete-excluded $SOURCEDIR/ .
 
 unset HTTP_PROXY # unset this to build on slc6 system
 
+# Due to typical long install dir paths used by aliBuikd the string lenghts must be increased
+# In addition a trim statement is missing for the install path
+sed -i 's/max_string_length\ =\ 255/max_string_length\ =\ 1000/g' pyol/config/default.cfg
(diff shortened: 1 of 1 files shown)


mfasDa:
The trim bug is actually fixed in OpenLoops 2.1.2, an upgrade from 2.1.2 would solve the issue as well, still the string buffer size must be set manually. The official repository for OpenLoops is https://gitlab.com/openloops/OpenLoops.git, the current ALICE fork uses a 1-1 clone of the tag.

ktf:
Seems to break AliGenerators, could you check if related? Also, doesn't fortran support PATH_MAX?

mfasDa:
```
/bin/sh: gengetopt: command not found
```
We had this issue a couple of times in aligenerators builds, particularly of generators using ThePEG, however I don't recall how this was resolved. 

Concerning the PATH_MAX: I don't know whether it is used in OpenLoops, I think it is pretty much done all with config parameters :( . Patching the project is beyond scope, sorry.

PR #3856, opened 2022-02-18
Allowed to choose precision in FFTW3

jan-busa:
This pull request allows user to choose FFTW3 precision by setting environment variables FFTW3_DOUBLE or FFTW3_LONG_DOUBLE to ON. If no variable is set, original behavior (build in single precision) is kept.


ktf:
Fine with me. @TimoWilken could you warm up the cache?

TimoWilken:
Caching is done!

PR #3880, opened 2022-02-25
datadist: bump to v1.3.0, add ucx dependency

ironMann:
This adds a new dependency: Unified Communication X Library (UCX). It's required to utilize the IB fabric to the full extent, supporting tag matching and adaptive routing offloads. It should only be required for EPN and FLP builds (slc8).
For compiling ucx features, CentOS builders must have recent rdma-core and rdma-core-devel packages installed, v35 or newer.  UCX tag v1.12.1-rc2 is required, as it fixes a bug with excessive CPU utilization.

DataDistribution v1.3.0 again will need synchronized upgrade on FLP and EPNs.


ironMann:
Forgot to mention, the ucx library available in centos repos cannot be used because they don't enable required compile flags.

datadistribution.sh
```diff
@@ -13,20 +13,19 @@ requires:
   - protobuf
   - O2
   - fmt
+  - "ucx:(slc8)"
```
ironMann, line 16: Will this include the lib for both EPN and FLP builder?

ktf: Unless the FLP default explicitly disables it, yes.

ironMann:
I think the builders are missing rdma-core-devel : https://github.com/alisw/docks/pull/236

ktf:
@TimoWilken could you merge the PR and redeploy?

awegrzyn:
As `rdma-core-devel` is installed by default on CS8, would it be possible to add a check to UCX recipe?

TimoWilken:
All green except for fullCI (which is unrelated); merging.

davidrohr:
@ktf @martenole : FYI: I have disabled building the O2PDPSuite for EPN with the latest DD, but it is now using the version of the FLPSuite. Since this must be updated in sync. We'll have the next regular nightly build with DD 1.3 next Monday, when the next FLP Suite will be installed.
@ironMann : If you need a build for tests before, please let me know or just build yourself using an alidist branch.

PR #3905, opened 2022-03-04
Bump DataDist to v1.3.2

ironMann:
Hi @vascobarroso @davidrohr please include for the next week upgrade. Thanks


ironMann:
@davidrohr Please include in the next EPN build. This should improve startup time.

PR #3907, opened 2022-03-04
Fix setuptools on Ubuntu 21.10

saganatt:
Ciao @ktf ,

this is the solution that enabled me to build full O2Physics stack.
It was enough to specify precise, older setuptools version in both Python-modules and Python-modules-list, no change to versions of other packages.

### DIFF
--- a/python-modules-list.sh
+++ b/python-modules-list.sh
@@ -40,6 +40,7 @@ env:
     pandas==1.2.3
     scikit-learn==0.24.1
   PIP39_REQUIREMENTS: |
+    setuptools==60.8.2
     numpy==1.21.4
     scipy==1.7.3
     Cython==0.29.21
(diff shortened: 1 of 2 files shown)


ktf:
ok, great. @TimoWilken could you warm the cache with this?

TimoWilken:
Hi @saganatt, @ktf, the build on slc7 (CI and caching) fails as `setuptools==60.8.2` is not available there; the highest version pip can find there is 59.6.0 (see the full log linked on the CI results page).

Perhaps changing `setuptools==60.8.2` to `setuptools<=60.8.2` would work?

Also, since you're installing setuptools in `python-modules.sh`, presumably you don't need to list it separately in `python-modules-list.sh` as well, right?

saganatt:
Hi @TimoWilken, listing the version in python-modules.sh only does not help - it keeps installing 60.9. Changing only python-modules-list.sh also does not produce an effect, only combination of them both.

I will check locally with '<='

ktf:
Seems to be fine. @TimoWilken whenever you can cache this we can merge.

TimoWilken:
Caching is done!

PR #3924, opened 2022-03-09
Add missing GCC dependencies

davidrohr:
@ktf : I have no understood what happened last time: When we have a new enough system CMake, but an old system GCC, then we pick up the system CMake, and then we do not pull in GCC-Toolchahin via CMake, and then we can fail when the system GCC is too old for these packages.
And it fails also for header-only stuff, since even there the CMake checks the C compiler with the requested compile flags, i.e. even though it doesn't compile anything, also the header only stuff fails if the system GCC does not understand the requested CMAKE_CXXFLAGS.

### DIFF
--- a/asio.sh
+++ b/asio.sh
@@ -4,6 +4,7 @@ tag: v1.19.1
 source: https://github.com/FairRootGroup/asio
 build_requires:
   - CMake
+  - "GCC-Toolchain:(?!osx)"
 prepend_path:
   ROOT_INCLUDE_PATH: "$ASIO_ROOT/include"
 ---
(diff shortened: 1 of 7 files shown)


ktf:
Ah, ok. I think we said at some point that a mixed system /  non system setup for gcc and cmake is not allowed (because cmake itself is indeed a C++ program), but we do not have any good way to enforce that. So the proper solution is actually  to install a recent gcc on the system, IMHO. That said, this should not harm the recommended "uniform" setup and if it works for the mixed case, I do not see anything against it. I just suspect that in any case the "mixed" setup will hit us later on, though.

PR #3925, opened 2022-03-10
Bump ms-gsl to 4.0.0

davidrohr:


ktf:
is it using C++20?

davidrohr:
why that? You are using std::accumulate. Before, you were getting it from ms_gsl, but no the gsl headers don't load it any more.
So it should be added in the O2 file. I have a PR for it here: https://github.com/AliceO2Group/AliceO2/pull/8308

ktf:
I misread the changes for std::accumulate, nevermind.

PR #3926, opened 2022-03-10
Update o2-full-system-test.sh

sawenzel:
Take out analysis testing. Currently broken and influences by O2Physics (which is under review)


sawenzel:
I hope this stabilizes the full-CI. @davidrohr

davidrohr (approved):
thx, I fully agree to this.
I would only reenable this once the `SIM_CHALLENGE_ANATESTING` runs in O2Physics CI in whatever mode.

PR #3944, opened 2022-03-15
Make sure RPATH location in AEGIS is correct

ktf:


ktf:
I think it's actually CMake to be wrong, but this seems to do the job.

PR #3945, opened 2022-03-15
Make sure FairLogger has a proper version even when doing development

ktf:


dennisklein:
> Maybe the tag should be embedded in the source as part of the release process?

Ya, this is an old question. IMHO, the git repo is a first class quality of our source tree and tools that still model source tree as just a single version tarball need to adapt and learn how to include the git repo. I would hate it a lot to duplicate version numbers in two places. (Also automatic commits are a no go.)

However, what we can do, is create a `dist` build target, that generates such a source tarball for backwards compatibility (and then generate and attach those as release assets). - Would this work?

---

Just for my curiosity, in which environment/use case are you dropping the git history and why?

dennisklein:
btw, the solution you propose in this PR I like even better than a dist tarball. In fact I like it so much, that I will make it a [documented feature](https://github.com/FairRootGroup/FairCMakeModules/issues/26) for [`fair_get_git_version`](https://fairrootgroup.github.io/FairCMakeModules/latest/module/FairProjectConfig.html#fair-get-git-version) which FairLogger does not yet use, but it really should and will in one of the next releases.

ktf:
I just had a spurious tag in fairlogger, which confused the current command. The "tarball" behaviour was just to make a more general example. If you are saying this is the way to go, I am also happy.

ktf:
(also notice github provides you tarballs automatically)

PR #3955, opened 2022-03-16
Use our own FreeType mirror

TimoWilken:


ktf:
I assume you tested if freetype cannot be built out of source already? Something like:

```bash
cd $BUILDDIR
$SOURCEDIR/configure ...
```

ktf (approved):

ktf:
Sigh... Can you cache this and then we merge?

TimoWilken:
Caching is done.

PR #4003, opened 2022-03-25
Bump 

ironMann:
Bump DataDist to v1.3.7
@vascobarroso The memory usage default is decreased to 128MB

Requires ppconsul bump to v0.2.3

### DIFF
--- a/datadistribution.sh
+++ b/datadistribution.sh
@@ -1,6 +1,6 @@
 package: DataDistribution
 version: "%(tag_basename)s"
-tag: v1.3.4
+tag: v1.3.7
 requires:
   - "GCC-Toolchain:(?!osx)"
   - boost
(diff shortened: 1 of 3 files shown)


vascobarroso:
@ironMann great thanks a lot! Let's see if it makes the CI happy :crossed_fingers:

ktf:
@vascobarroso This is working in the PDP CI, shall I merge and you cherry pick it in your branch or shall I hold as well?

vascobarroso (approved):

vascobarroso:
@ktf you can merge, that will actually simplify testing in our nightly CI. If still broken, I will report here. 
For completeness, this is not for next Monday, it is for the week after (as agreed yesterday with the RC).

ktf:
The issue I see is that that will end up in the PDP suite releases, so I would like either @davidrohr or @shahor02 to greenlight it.

vascobarroso:
@ironMann I confirm that the new DD passes our FLP Suite CI. Thanks for the fix!

PR #4012, opened 2022-03-28
[O2-2821] Add the missing ROOT include path

Barthelemy:


davidrohr (approved):

ktf:
Trivial and not actually tested. Merging.

PR #4015, opened 2022-03-29
Bump POWHEG to r3964 and add direct photon process

mfasDa:


mfasDa:
Depends on https://github.com/alisw/POWHEG/pull/6

mfasDa:
Hello @maireiphc,

The fixes in the makefiles are forward-ports from the existing release (r3693), they are in the branch as separate commits but were not part of the master branch. They are needed for compilation on and for compilation with gfortran-10+. (check the branch with ALICE-specific patches of that version) The only new change I added in the makefiles is to set the option `-fallow-type-mismatch` for all gfortran versions starting from 10 instead of specifying version 10 explicitly, in order to add support for gfortran-11. These parts in the makefiles must be adapted by us, as in the repository there are sometimes only placeholders or hard coded paths.

Otherwise this code is  a 1-1 port of v3964. As you see the amount of code changed in the main program in the new POWHEG release is tiny, the main new part is the direct photon process. 

Cheers

Markus

mfasDa:
Depends on which compiler version we are using. As far as I understand we are still using gcc7 for the defaults-generators, then this will not need the patches. In case however one switches to gcc10 (which I hope at some point will be done in order to align all packages under the o2 defaults) then those related to fortran will be needed. Those related to libc++ are needed for macOS.

fjonasALICE:
Any update on this? Would definitley be very useful to have POWHEG direct photon processes available, so following this PR now :-)

ktf:
@alisw/pagmc  please review and eventually merge this.

ktf:
Ping.

maireiphc:
Sorry for my slow reactivity.

Ok, we can give a shot by picking Markus source code of POWHEG with this recipe change. 
We will need to change towards the official ALICE directory with possibly a new release/tag : https://github.com/alisw/POWHEG/tags.

The PR https://github.com/alisw/POWHEG/pull/6 is now pushed to the official ALICE POWHEG repo. To prepare grounds in parallel.

maireiphc (approved):

PR #4019, opened 2022-03-30
OpenSSL 1.1 + patched recipes to use our curl on slc8

TimoWilken:
Based on #3658, but rebased to fix merge conflicts, and I've opened this against master so that the CI runs.

On slc8, we need to use our own custom-built curl (instead of the system one), as the system one conflicts with our OpenSSL.

This also applies to Monitoring and Ppconsul, which originally specified the system curl on slc8 only -- this had to be changed to our own on slc8 as well. @ktf, @ code owners, is this OK?


cgal.sh
```diff
@@ -1,12 +1,13 @@
 package: cgal
+tag: releases/CGAL-4.6.3
 version: "4.6.3"
+source: https://github.com/CGAL/cgal
```
ktf, line 4: Unfortunately we cannot do this for cgal. The repository is simply too big. Has this changed?

monitoring.sh
```diff
@@ -5,8 +5,7 @@ requires:
   - boost
   - "GCC-Toolchain:(?!osx)"
   - "ApMon-CPP:(?!osx)"
-  - "system-curl:(slc8)"
```
ktf, line 8: I think monitoring wants  to use system-curl on slc8. Why changing this?

ktf: @awegrzyn to confirm

ppconsul.sh
```diff
@@ -5,8 +5,7 @@ source: https://github.com/oliora/ppconsul
 requires:
   - boost
   - "GCC-Toolchain:(?!osx)"
-  - "system-curl:(slc8)"
-  - "curl:(?!slc8)"
+  - curl
```
ktf, line 8: Same here.

ktf (changes requested):

cgal.sh
TimoWilken: Hm, good point. I didn't notice it particularly, but the checkout of cgal in `sw/MIRROR/cgal` is 654 MiB. I can change this back to fetching the tarball tomorrow.

monitoring.sh
TimoWilken: On slc8, the system curl fails when our OpenSSL is in `LD_LIBRARY_PATH`. So either we install a wrapper script for curl there, or we compile this with our own curl.

TimoWilken:
@awegrzyn, the Monitoring and Ppconsul packages specifically required `system-curl` on CentOS 8, but our own aliBuild-built `curl` everywhere else. What was the reason for that? And can the same goal be accomplished by disabling the `curl` package in the defaults you're using?

awegrzyn:
On FLPs, as we don't have modules, having our `curl` was breaking `yum` on CC7. I haven't tested what happens on CS8, but I assume this could have impact on `dnf`.

awegrzyn:
As I'm not in the loop which software, running on FLPs, requires new OpenSSL (and curl)?

TimoWilken:
OpenSSL is required by arrow, ROOT, libjalienO2. It's also required for building curl and CMake, so everything built using CMake gets built against it.

Curl is required by Monitoring, Ppconsul and O2 itself. It is also required for building Clang and Python-modules.

awegrzyn:
@TimoWilken my question was more in sense, what requires **non-system** curl, Monitoring and ppconsul can use `curl` from system as it does now.

TimoWilken:
In theory nothing, I think, but can Monitoring and Ppconsul use the system OpenSSL too?

On CentOS 8, you can't use the system curl while the new OpenSSL is in `LD_LIBRARY_PATH` (because the system openssl was built with a bunch of special CentOS patches, and the system curl expects a bunch of symbols that are present in the system openssl but not in ours).

So it's all-or-nothing -- either everything uses our openssl and our curl, or everything uses the system openssl and the system curl (by disabling the curl and OpenSSL packages).

awegrzyn:
Okay, in such case we'd prefer using `curl` in `o2-dataflow` defaults. But of course you can go ahead with changes in other defaults.

TimoWilken:
Hi @teo @vascobarroso, unfortunately it's impossible to compile mesos against the system libsasl2 with our own OpenSSL in the environment at the same time (i.e. when compiling with `o2` defaults). Compiling mesos with `o2-dataflow` defaults still works.

I've removed mesos from the O2Suite package to make the fullCI work (hopefully!). If you need mesos installed somewhere, we could factor it out into another package that depends on O2Suite and mesos? (Or introduce a package e.g. O2Suite-deps which includes everything O2Suite needs except mesos, then change O2Suite to depend on that plus mesos, and make o2-full-system-test/o2fullci depend only on O2Suite-deps rather than O2Suite).

Would that be an acceptable compromise?

ktf:
@teo @vascobarroso could you comment on @TimoWilken question?

vascobarroso:
@TimoWilken @ktf Teo is away until next Tuesday, I will ask him to have a look at this once he is back.

TimoWilken:
I rebased to fix a merge conflict, hence the force push.

I propose as a solution the changes in fcaa92c4f99fed79c1aa44350015265afda40018. For users of O2Suite, this should factor mesos out transparently, while the CI packages (`o2-full-system-test.sh` and `o2fullci.sh`) stop indirectly depending on mesos.

TimoWilken:
OK, reverted to the commit before last, which just removes mesos from O2Suite (and was green previously).

vascobarroso:
@TimoWilken @ktf @awegrzyn this PR seemed to have a side effect on our RPM generation: since this night, the mesos RPM is no longer generated, which then breaks the FLP Suite. Any idea why ?

ktf:
We dropped mesos from o2suite:

https://github.com/alisw/alidist/pull/4019/files#diff-021eaa79cb09b4f86a24b70045467451491eaf0941258204aa3afaf8232facacL16

how about we add it back and we try? In principle there is no need to disable it anymore since we understood from @teo that SSL is not actually used.

ktf:
https://github.com/alisw/alidist/pull/4113 should add it back.

vkucera:
@ktf @TimoWilken 
This commit breaks the installation of Python-modules on Ubuntu 20.04.
```
++ python3 -m pip install -IU pip
WARNING: pip is configured with locations that require TLS/SSL, however the ssl module in Python is not available.
WARNING: Retrying (Retry(total=4, connect=None, read=None, redirect=None, status=None)) after connection broken by 'SSLError("Can't connect to HTTPS URL because the SSL module is not available.")': /simple/pip/
WARNING: Retrying (Retry(total=3, connect=None, read=None, redirect=None, status=None)) after connection broken by 'SSLError("Can't connect to HTTPS URL because the SSL module is not available.")': /simple/pip/
WARNING: Retrying (Retry(total=2, connect=None, read=None, redirect=None, status=None)) after connection broken by 'SSLError("Can't connect to HTTPS URL because the SSL module is not available.")': /simple/pip/
WARNING: Retrying (Retry(total=1, connect=None, read=None, redirect=None, status=None)) after connection broken by 'SSLError("Can't connect to HTTPS URL because the SSL module is not available.")': /simple/pip/
WARNING: Retrying (Retry(total=0, connect=None, read=None, redirect=None, status=None)) after connection broken by 'SSLError("Can't connect to HTTPS URL because the SSL module is not available.")': /simple/pip/
Could not fetch URL https://pypi.org/simple/pip/: There was a problem confirming the ssl certificate: HTTPSConnectionPool(host='pypi.org', port=443): Max retries exceeded with url: /simple/pip/ (Caused by SSLError("Can't connect to HTTPS URL because the SSL module is not available.")) - skipping
ERROR: Could not find a version that satisfies the requirement pip (from versions: none)
ERROR: No matching distribution found for pip
```

ktf:
Could you check if https://github.com/alisw/alidist/pull/4118 fixes the issue? Apparently we cannot pick up python from the system on ubuntu anymore.

vkucera:
> Could you check if #4118 fixes the issue? Apparently we cannot pick up python from the system on ubuntu anymore.

Yes, that seems to compile fine. Thanks

PR #4028, opened 2022-04-01
Bump qualitycontrol to v1.51.2

Barthelemy:


davidrohr:
why do we go from 1.52 to 1.51.2? Is the downgrade intended?

Barthelemy:
This is a patch for v1.51 with FV0 enabled. V1.52 is the new release that brings a number of other changes. I thought it could be safe to have a patch version on the v1.51 branch.

PR #4036, opened 2022-04-04
As it is a requirement, load ONNXRuntime

benedikt-voelkel:


benedikt-voelkel:
For instance when only O2 is loaded, a simple
```bash
o2-sim-serial
```
crashes with 
```bash
o2-sim-serial: error while loading shared libraries: libonnxruntime.so.1.7.2: cannot open shared object file: No such file or directory
```

PR #4037, opened 2022-04-04
Update mesos mirror

TimoWilken:
The previously used mirror is down.


TimoWilken:
I've started caching this PR.

TimoWilken:
Merging as currently broken. CI builders will pick up the cached packages when they're done.

PR #4040, opened 2022-04-05
bump grpc to v1.45.1

ironMann:
This is to explore reverting to upstream grpc versions. 
There are sporadic issues related to grpc observed in DD during runs. I'm not sure if the upgrade will solve anything, but it's worth a try.

The original issue that prompted us to patch and pin the grpc library is hopefully gone now, as FMQ stopped dynamically loading the executable to look for plugins.


davidrohr:
@vascobarroso : Shall we merge this, or does FLP want to do some tests before?

vascobarroso:
@ironMann @davidrohr the timing of this update is a bit tricky since @teo is away. But for me we can go ahead once the CI is green and we will then do some extra tests on our side. If they fail, we can always revert this.

ironMann:
Added a fix for mesos build. 

@vascobarroso @davidrohr We can start with an EPN build first. grpc should be wire compatible. As long as the control plugins work/load properly, we should be ok. I've done a quick test with `-P odc` and `-P OCClite`.

davidrohr:
ok, but perhaps let's do such a test once the CI is green.

ironMann:
@vascobarroso in #4059, I think the OCClite will correctly load with the upstream grpc. DD tasks correctly open the control grpc-sever port and even terminate properly on a signal.

Could you please try re-adding the changes for the next iteration. Thanks

vascobarroso:
@ironMann ok for me. Let's then re-add this once #4059 is merged and see what the CI says about it.

davidrohr:
The new DD is merged, will you test GRPC in the CI without bumping it in alidist, or shall we merge this PR again?

vascobarroso:
@davidrohr the FLP Suite CI seems to like this change, so you can go ahead and merge this PR again in `master`.

vascobarroso:
@davidrohr @ironMann just to confirm: who takes care of re-merging this ?

PR #4056, opened 2022-04-10
Bump DDS to 3.7.9

davidrohr:


ktf:
@TimoWilken could you check why the macOS build is marked as failed? I could not find any particular reason.

TimoWilken:
The Mac build previously had a failing QC unit test, but subsequently succeeded.

Now, the fullCI error is (right at the end of the plain-text log):

```
ERROR: Full system test did generate no raw data for MID
```

davidrohr:
Tested locally, no idea why no MID data was created in here, apparently not related to DDS. Merging

PR #4058, opened 2022-04-11
Using C++17

pzhristov:
Temporary fix, later we should move to defaults-o2 and remove both defaults-next-root6.sh and defaults-user-next-root6.sh.


ktf:
Merging. Notice that I switched the CI to use the o2 defaults in any case. I guess we should now move the releases to o2 as well and call it done. It might actually help some feature which says what is the "default default" to be used for a given package, so that we do not need to tell people to use one, at least in general.

PR #4060, opened 2022-04-12
Bump FairLogger to v1.11.0

ktf:
This introduces the "detail" level which we can use for debug messages which should not be removed when compiling optimised builds and should not end up to info logger.


davidrohr (approved):
Makes total sense. Actually I feel bad I didn't think about it when we added the important level :(

ktf:
We also need to have some associated changes in O2. I will do them once this is merged.

PR #4068, opened 2022-04-14
Revert "Revert "bump grpc to v1.45.1""

ironMann:
Reverts alisw/alidist#4050

The FLP control plugin and DD problem seems to be fixed in #4059


davidrohr:
ok, I'll leave this open and not merge it. But would be good if we could check this soon after Eastern, since this will fix some issues of DD.

ironMann:
picked up in #4089. Closing...

PR #4025, opened 2022-03-31
Bump VMC dependent packages:

ihrivnac:
- VMC: v2-0
- GEANT4_VMC: v6-1
- GEANT3: v4-1
- FLUKA_VMC: 4-1.1-vmc3
- GEANT4: v11.0.1
- MCStepLogger: v0.5.0
- vgm: v5-0
- FairRoot: v18.4.7

PR #4076, opened 2022-04-15
Revert "Bump VMC dependent packages: (#4025)"

sawenzel:
This reverts commit a5e888e797d98058320f4f4c4eb700919ffa44a7.

We see SIGNAL11 with G4 and the simulation hangs indefinitely.
This needs to be understood before we can use the new version.

https://alice.its.cern.ch/jira/browse/O2-2901

### DIFF
--- a/fairroot.sh
+++ b/fairroot.sh
@@ -1,6 +1,6 @@
 package: FairRoot
 version: "%(short_hash)s"
-tag: "v18.4.7"
+tag: "v18.4.2"
 source: https://github.com/FairRootGroup/FairRoot
 requires:
   - generators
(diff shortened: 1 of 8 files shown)


pzhristov:
@ihrivnac Do we have any 10.7.X tag of Geant4 that is compatible with the VMC changes?

sawenzel:
@pzhristov : I would just revert the complete PR (including VMC devs) for a while. I have verified over the last days that it runs stable and does not show the problems of O2-2901. Then, we can study the problem in parallel while offering a stable service for productions.

sawenzel:
I think this is ready to go. Problems in generators + O2/alidist are due to build-issues in other parts of the software stack.
Previously failing checks with real issues (fullCI + macOS) are ok now.

pzhristov (approved):

PR #4084, opened 2022-04-19
onnxruntime: do not build python bindings for M1 Macs

ktf:


ktf:
@pzhristov with this we do not need any custom python, at least not for the general O2 development flow.

PR #4093, opened 2022-04-21
PYTHONPATH for xjalienfs

sawenzel:
Set python path for xjalienfs. This is needed to access alien.py functionality and alien tokens
during aliBuild execution (such as to access CCDB objects for simulation tests done
within aliBuild)


adriansev (approved):
LGTM

PR #4080, opened 2022-04-18
python:: use proper pip and coherent triad of pip,setuptools,wheel

adriansev:
@ktf FYI
* usage of the safe form of `python3 -m pip` for pip operations
* use latest of pip, setuptools, wheel

this was tried on centos7, ubuntu 20.04 and fedora 35
for reference the change logs of involved packages are below
* https://github.com/pypa/pip/blob/main/NEWS.rst
* https://github.com/pypa/setuptools/blob/main/CHANGES.rst
* https://github.com/pypa/wheel/blob/main/docs/news.rst

Hopefully this will solve the problems encountered by others (but which i could not reproduce)

PR #4100, opened 2022-04-22
Revert "python:: use proper pip and coherent triad of pip,setuptools,wheel"

ktf:
Reverts alisw/alidist#4080


ktf:
Reverting this as it breaks M1 macs.

PR #4105, opened 2022-04-22
Bump odc.sh to version 0.68

sawenzel:
Version 0.67 had a compile bug on some platforms (for instance alibit) due to typo std::cout != std::endl. See here: https://github.com/FairRootGroup/ODC/commit/2def9473543e2a1e262de5c76990eb9222e88361.

### DIFF
--- a/odc.sh
+++ b/odc.sh
@@ -1,7 +1,7 @@
 #Online Device Control
 package: ODC
 version: "%(tag_basename)s"
-tag: "0.67"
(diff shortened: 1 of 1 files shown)


davidrohr:
There is actually already #4102, which does the same, but its CI didn't yet start, even though I approved it.

sawenzel:
Very good! Didn't notice this before.

PR #4112, opened 2022-04-26
Use alisw g4

benedikt-voelkel:


benedikt-voelkel:
@pzhristov @ihrivnac @sawenzel

geant4.sh
```diff
@@ -1,7 +1,8 @@
 package: GEANT4
 version: "%(tag_basename)s"
-tag: "v10.7.2-alice1"
+tag: "alice/v11.0.1-patches"
```
pzhristov, line 3: this should be v11.0.1-alice1

geant4.sh
```diff
@@ -1,7 +1,8 @@
...
 source: https://github.com/alisw/geant4.git
+#source: https://gitlab.cern.ch/geant4/geant4.git
```
pzhristov, line 5: Is there any reason to keep the commented source_

pzhristov: We can keep the line.

PR #4113, opened 2022-04-26
Add back mesos in O2Suite

ktf:


vascobarroso (approved):

mesos.sh
```diff
@@ -20,6 +16,10 @@ build_requires:
 - protobuf
 - Python-modules
 - abseil
+- system-apr
```
awegrzyn, line 19: @ktf Why those dependencies were moved from `requires` to `build_requires` ? They are required during runtime and this is why our RPM break.

awegrzyn: https://github.com/alisw/alidist/pull/4122

ktf: no, they are not. They are empty packages, no? The actual libraries are on the system and they are there regardless of the package being deployed. No?

PR #4115, opened 2022-04-26
Add temporary repository for ML model storage

fcatalan92:
@ktf As discussed by mail, I have created a temporary repository to store the ML models for the inference on Hyperloop. It will be deleted when CCDB storage will be available.

I made the module an O2 requirement since there are developments that could need this feature.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -73,3 +73,6 @@ xjalienfs*          @alisw/alien-experts
 datadistribution*   @alisw/epn-experts
 dds*                @alisw/epn-experts
 odc*                @alisw/epn-experts
+
+# ML
+mlmodels*           @fcatalan92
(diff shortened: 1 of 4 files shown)


ktf:
Thanks. 
@awegrzyn I guess we need to disable this for o2-dataflow, as we do for ONNX?

awegrzyn:
Yes, thanks !

ktf:
@fcatalan92 do you actually need this in O2 proper? I.e. will it be used for the sync / async reco / simulation? If this is just O2Physics related we should probably move the dependency there.

fcatalan92:
@ktf Hi, my understanding is that this will be used also for calibration / QC / simulation. So it is needed in O2.

ktf:
ok, so i leave it as it is for now. @awegrzyn please make sure you pick it up in your flp suite branch, so that for next version we can drop the workaround I currently have to do the EPN builds.

awegrzyn:
As it's a bit late for v0.56.0 it will come with following release.

PR #4117, opened 2022-04-26
Disable ODC is defaults O2

ktf:


ktf:
This is needed because apparently ODC links against system SSL (due to gRPC requirement) even if gRPC itself was built with our own OpenSSL.
@dennisklein @rbx I suspect something is wrong in either the gRPC config file or in the ODC cmakelists itself. Could someone have a look?
@davidrohr @ironMann  FYI.

rbx:
gRPC location in ODC looks standard: https://github.com/FairRootGroup/ODC/blob/master/CMakeLists.txt#L70
There is nothing special done with regard to SSL OpenSSL.

ktf:
agreed, but then why does it pick ssl from the system even if grpc was compiled with the alidist one?

dennisklein:
I agree that essentially a `find_package(OpenSSL)` plus adding the appropriate ssl target dependencies to the `gRPC::grpc++` target are missing. https://github.com/FairRootGroup/ODC/issues/62 should fix this.

ktf:
@TimoWilken any idea on why the token is not there?

```
## sw/BUILD/O2-full-system-test-latest/log
--
Detected critical problem in logfile digi.log
digi.log-[76823:FDDDigitizer]: [20:50:37][INFO] FDD: Sending ROMode= 3 to GRPUpdater
digi.log-[76829:CPVDigitizer]: [20:51:18][STATE] Exiting FairMQ state machine
digi.log:[76829:CPVDigitizer]: [20:51:18][FATAL] unknown error while setting up workflow in o2-sim-digitizer-workflow.
digi.log-[ERROR] pid 76829 crashed with 1
digi.log-[INFO] QUIT_REQUESTED
killing child 74827
killing child 74830
--
[76829:CPVDigitizer]: [20:50:16][WARN] CCDB: Did not find an alien token; Cannot serve objects located on alien://
[76829:CPVDigitizer]: [20:50:16][INFO] CCDB: Reading o2::cpv::Pedestals from CPV/Calib/Pedestals
[76829:CPVDigitizer]: [20:50:16][WARN] CCDB: Did not find an alien token; Cannot serve objects located on alien://
[76829:CPVDigitizer]: [20:50:16][FATAL] Got nullptr from CCDB for path CPV/Calib/Pedestals and timestamp 1651005833191
```

PR #4118, opened 2022-04-26
Use own python on ubuntu18

ktf:
We need this because system python uses OpenSSL1.0 and we end up mixing things up.


ktf:
Merging this as it solves the current issues on ubuntu. We probably need to decide a better strategy.

ktf:
Merging this as it solves the current issues on ubuntu. We probably need to decide a better strategy.

PR #4122, opened 2022-04-27
Fix broken Mesos runtime

awegrzyn:
Fixes mesos runtime.


ktf:
Why is this required? Those packages are empty.

awegrzyn:
Yes, but (1) for consistency, (2) they appear in modulefile therefore we know whether it's runtime dependency or not.

PR #4124, opened 2022-04-27
Do not require Python modules on osx_arm64. 

pzhristov:
Check for specific setuptools==60.8.2 that work for XRootD on Mac M1.

### DIFF
--- a/agile.sh
+++ b/agile.sh
@@ -7,7 +7,7 @@ requires:
   - boost
   - lhapdf5
   - HepMC
-  - Python-modules
+  - Python-modules:(?!osx_arm64)
 build_requires:
   - "autotools:(slc6|slc7)"
   - SWIG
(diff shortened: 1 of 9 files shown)


pzhristov:
This PR should be checked on Mac M1 where we do not use miniconda3 for the python packages. Miniconda is useful since it enables the Metal GPUs to tensorflow, etc., but some people may rely on the standard pip3.

pzhristov:
@ktf  The check correctly found unsupported version and failed. I updated the setuptools==60.8.2 on our M1 server, so now the problem is fixed. If you agree, I propose to merge.

ktf:
This is fine with me, did you update all the servers, though? I would have expected the test to be redone and become green.

ktf:
I have restarted the test just to make sure.

pzhristov:
The compilation is OK, but something is wrong with the tests.

ktf:
That test usually fails when there is some CCDB hiccup, so it should be safe.

PR #4136, opened 2022-04-30
Pick up the provided OpenSSL

ktf:
Without this it picks up the system version on linux.

### DIFF
--- a/qualitycontrol.sh
+++ b/qualitycontrol.sh
@@ -43,7 +43,7 @@ incremental_recipe: |
 case $ARCHITECTURE in
   osx*) 
       [[ ! $BOOST_ROOT ]] && BOOST_ROOT=$(brew --prefix boost)
-      [[ ! $OPENSSL_ROOT ]] && OPENSSL_ROOT_DIR=$(brew --prefix openssl@1.1)
+      [[ ! $OPENSSL_ROOT ]] && OPENSSL_ROOT=$(brew --prefix openssl@1.1)
       [[ ! $LIBUV_ROOT ]] && LIBUV_ROOT=$(brew --prefix libuv)
       SONAME=dylib
   ;;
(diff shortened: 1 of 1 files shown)


chiarazampolli:
Hello @ktf , 

Are these errors genuine, or could this be merged?

Thanks!

ktf:
I think they are actually genuine.

@Barthelemy @knopers8 could you investigate a bit? It seems to me that the cmake from QC actually picks up OpenSSL from the system. This was meant to force it to pick it up from our own build, where specified, but still it does not work. I remember you had to do something weird to pick up openssl, can you remind me what that was?

ktf:
Ok, I am now forcing the OpenSSL detection also in O2. In principle the actual issue could actually be fixed at QC level by dropping the dependency on `CURL::libcurl` since `O2::CCDB` should already provide a consistent set of dependencies. @Barthelemy is looking into it. However, since that would require a new QC release, let's see if the changes proposed in the PR can converge faster.

o2.sh
```diff
@@ -241,7 +241,7 @@ module load BASE/1.0 \\
             ${FASTJET_REVISION:+fastjet/$FASTJET_VERSION-$FASTJET_REVISION}                         \\
             ${CGAL_REVISION:+cgal/$CGAL_VERSION-$CGAL_REVISION}                                     \\
             ${GLFW_REVISION:+GLFW/$GLFW_VERSION-$GLFW_REVISION}                                     \\
-            ${FMT_REVISION:+fmt/$FMT_VERSION-$FMT_REVISION}                                         \\
```
Barthelemy, line 244: Are you removing fmt ?

ktf: Nope.

chiarazampolli:
Hello, 

Are the errors still genuine?

Chiara

ktf:
Yes. It looks like gRPC is built with system openssl on cc8.

ktf:
It should be fixed in the last commit.

chiarazampolli:
Hello, 
I am sorry to bother about this, but the tests are all red now...

chiarazampolli:
Sigh, still not ok... @ktf

singiamtel:
Setting as draft to save CI tester cycles

PR #4164, opened 2022-05-06
Revert "Disable ODC in O2 defaults "

davidrohr:
Reverts alisw/alidist#4117

@ktf @lkrcal @rbx : Sorry, but I don't agree to disabling ODC for the O2 defaults.
We need to test it in the alidist CI. Yesterday we bumped ODC in alidist, and the new version was simply not compiled, so we didn't spot a compile error, so we cannot deploy it today.

@dennisklein @rbx : What is the status on the OpenSSL issue that @ktf reported with ODC before? Can this be fixed?

### DIFF
--- a/defaults-o2.sh
+++ b/defaults-o2.sh
@@ -8,7 +8,6 @@ env:
   MACOSX_DEPLOYMENT_TARGET: '10.15'
 disable:
   - mesos
-  - ODC
 overrides:
   AliPhysics:
     version: '%(commit_hash)s_O2'


dennisklein:
I also fixed a few minor bugs in and modernized the `odc.sh` recipe: https://github.com/dennisklein/alfadist/blob/odc_ci/odc.sh, if you are interested.

rbx:
The fix from Dennis is now merged and released with ODC 0.70.

This version also includes a fix for missing includes in the infologger module. Although I have no tested the infoLogger fix, as I cannot reproduce the original problem. The fix is based on the error log that was posted in Mattermost. Let me know if there is still another issue.

PR #4168, opened 2022-05-06
update python to v3.9.12

adriansev:
FYI @ktf @TimoWilken @sawenzel
tested on centos 7 and osx_x86-64 (and the lowest denominator that worked on both).

### DIFF
--- a/python.sh
+++ b/python.sh
@@ -1,6 +1,6 @@
 package: Python
 version: "%(tag_basename)s"
-tag: v3.6.10
+tag: v3.9.12
 source: https://github.com/python/cpython
 requires:
  - AliEn-Runtime:(?!.*ppc64)
@@ -60,6 +60,7 @@ EOF
 fi
 
 ./configure --prefix="$INSTALLROOT"  \
+            ${OPENSSL_ROOT:+--with-openssl=$OPENSSL_ROOT} ${OPENSSL_ROOT:+--with-openssl-rpath=no} \
             --enable-shared          \
             --with-system-expat      \
             --with-ensurepip=install
--- a/rivet.sh
+++ b/rivet.sh
@@ -1,6 +1,6 @@
 package: Rivet
 version: "%(tag_basename)s"
-tag: "3.1.1-alice1"
+tag: "3.1.6-alice1"
 source: https://github.com/alisw/rivet
 requires:
   - GSL
--- a/root.sh
+++ b/root.sh
@@ -93,7 +93,7 @@ if [[ -d $SOURCEDIR/interpreter/llvm ]]; then
   PYTHON_EXECUTABLE=$(python3 -c 'import distutils.sysconfig; print(distutils.sysconfig.get_config_var("exec_prefix"));')/bin/python3
 else
   # Non-ROOT 6 builds: disable Python
-  ROOT_PYTHON_FLAGS="-Dpyroot=OFF"
+  ROOT_PYTHON_FLAGS="-Dpython=OFF -Dpyroot=OFF"
   ROOT_PYTHON_FEATURES=
   ROOT_HAS_NO_PYTHON=1
 fi


adriansev:
this solves https://alice.its.cern.ch/jira/browse/O2-2920

adriansev:
AliRoot/alidist test is irrelevant as it tries to build ROOT v5.34.30 very old python bindings .. some decision must be made to disable python bindings for ancient ROOT

adriansev:
on the yoda part, it seems that it's not python >= 3.8 ready as tp_print was changed to tp_vectorcall_offset, see
https://docs.python.org/3/c-api/typeobj.html#c.PyTypeObject.tp_vectorcall_offset
and
https://github.com/cython/cython/issues/2976

sawenzel:
Is there any news here? I think we should go on for defaults-o2.sh.

adriansev:
on a second look at AliRoot/alidist (and root.sh recipe) output it seems that the recipe code to prevent python bindings to be built does not work... i will take a look at this, but on the yoda part i cannot act because of lack of knowledge on it..

adriansev:
so, i'm a little git-challenged so i'm grateful for any tips to fix the inclusion of some commits from master (beside the clean approach to close this and open a new clean PR)

adriansev:
> Hi @adriansev, I managed to clean your branch [https://github.com/fcatalan92/alidist/tree/python_new](https://github.com/fcatalan92/alidist/tree/python_new?rgh-link-date=2022-05-17T12%3A41%3A40Z) with the side effect of adding my name in a commit. You can avoid that doing it yourself with
> 
> ```shell
> git pull --rebase upstream master
> git rebase -i HEAD^^^^^
> ```
> 
> `upstream` is the name of the alisw remote

Hi @fcatalan92 !! thanks a lot!!

ktf:
@adriansev what's the status here? the generator issue is actually a genuine one.

adriansev:
> @adriansev what's the status here? the generator issue is actually a genuine one.

@ktf Yes, it is a genuine one, but outside of my possibility/knowledge of action.
i see that the version of yoda used is ancient (see https://yoda.hepforge.org/)
moreover starting with 1.9.0 i see:
```
The 1.9 YODA release series is an increment of 1.8 with breaking changes to the yodamerge script, which is now much faster in particular for many large files, but as a consequence no longer has a 'simple stacking' mode. For this the new yodastack script has been introduced. This release also removes support for ROOT version 5, introduces the generic Fillable, Binned, and Scatter base classes and the fillDim() and rmPoint[s]() methods on fillables and scatters respectively, and fixes numerous small bugs and annoyances.
```

looking into https://gitlab.com/hepcedar/yoda/-/blob/release-1-9-x/ChangeLog
i see some python fixes but nothing explicitly on our error.
grepping in the release-1-9-x branch does not find any instance of `tp_print`
so the problem could be fixed already.

The decision to be made is on using ROOT6 only latest version of yoda (only then we can see if we still have the problem and make an issue report to developers for fixing.)

ktf:
@alisw/pagmc could you comment on this? What are your plans for YODA?

adriansev:
@ktf also i just noticed this in yoda.sh:
```
prepend_path:
  PYTHONPATH: $YODA_ROOT/lib64/python2.7/site-packages:$YODA_ROOT/lib/python2.7/site-packages
```

maybe this should be changed to something more flexible like:
```
python$( ${PYTHON_EXECUTABLE} -c 'import sys; print(f"{sys.version_info.major}.{sys.version_info.minor}")' )
```

maireiphc:
Hello Adrian (all),

  sorry for the lack of reactivity here.
Let me exposed what I know, what I suspect and what I do not know (sorry in advance).

Yoda is the histogram package independent from Rivet but used by the later to store data out of MCgenerator+Rivet processing.
It is C++ framework but in part steered with Python.

Currently in the AliGenerator bundle, the yoda recipe uses v1.8.2. See https://github.com/alisw/alidist/blob/master/yoda.sh
If need be, we have alternatively [v1.9.0-alice1](https://github.com/alisw/yoda/tree/1.9.0-alice1) available under our own alisw repo.
The official release is currently at the version 1.9.5, https://yoda.hepforge.org/downloads

As far as I understand, the configure file of YODA tries to figure out which Python(+Cython) version can be picked.
https://github.com/alisw/yoda/blob/v1.8.2/configure
https://github.com/alisw/yoda/blob/1.9.0-alice1/configure
Look for variable : YODA_PYTHONPATH

A priori, YODA should be compatible with Python3 but I did not figured out where it flips between Python2.7 to Python 3
likley here : https://github.com/alisw/yoda/blob/2d2ff436b561a4c6391c37bf423c0a6daefd3e15/configure#L17709 for v1.8.2 in order of appearance, ti may prefer Python 3.8 over Python 2.7 (?)

What Adrian propose as upgrade of yoda.sh is maybe going anyway in a proper direction, with adapting to what is available on the systems. What I do not know is how it has to be dealt wrt to Python modules (python modules remain rather obscure to my level of knowledge).

Note : 
AliGenerators currently compile against gcc-toolchain (https://github.com/alisw/alidist/blob/master/defaults-generators.sh), which is defaulted currently v7.3.0-alice2 (https://github.com/alisw/alidist/blob/master/gcc-toolchain.sh).
We have in mind to move towards v10.2.0-alice2 like done for O² environements (https://github.com/alisw/alidist/blob/master/defaults-o2.sh)

adriansev:
@maireiphc @ktf Hi everyone! So, i would say that before going forward in this PR, the yoda would have to be updated to increase the chances of successful build with newer python (at least the offending line `__pyx_type_4yoda_4util_Base.tp_print = 0;` from `YODA/pyext/yoda/util.cpp:4898` is no longer found in 1.9.x versions) 
So, this will wait first for yoda update (by a yoda user/expert) then this PR should be tested again.
One comment that i would have looking over yoda configure and alidist recipe is that
yoda have a default requirement for ROOT6(PyROOT) but the yoda recipe does not specify ROOT at requires/build_requires.. also the apriori definition of prepend_path (with hardcoded python2.7 usage) maybe should be removed and if needed just use/export the `PYTHON_VERSION` and `PYTHON`

maarten-litmaath:
Hi all,
there also are ugly warnings we want to get rid of:

$ /cvmfs/alice.cern.ch/bin/alienv enter JAliEn
[JAliEn] ~ > alien.py token-info > /dev/null 
/cvmfs/alice.cern.ch/el6-x86_64/Packages/xjalienfs/1.4.2-1/lib/python/site-packages/OpenSSL/crypto.py:8: CryptographyDeprecationWarning: Python 3.6 is no longer supported by the Python core team. Therefore, support for it is deprecated in cryptography and will be removed in a future release.
  from cryptography import utils, x509

miranov25:
Hello @ktf and all
Is there an update? We would like to use  new Python as soon as possible.
Marian

ktf:
I just merged #4359 and restarted the test for aligenerators. If that works, I suggest we merge and continue from there.

ktf:
Ok, the next issue is rivet. @maireiphc could you bump that version as well? It looks like a similar issue with modern pythons.

antoniocosilva:
Hi @ktf, I'm not so familiar with all the details as Antonin, but I'm replying because Antonin may take a while to answer since he's enjoying a very deserved vacation. My understanding is that Rivet should be bumped to tag: "3.1.6-alice1" in rivet.sh, that's the newest version and is compatible with YODA 1.9.5.

ktf:
Ok, I have done both here and in https://github.com/alisw/alidist/pull/4392.

sawenzel:
@ktf: The CI succeeds fully for this bump. Not sure #4392 is actually needed or blocking this PR.

sawenzel:
Ah ok. I see that it's part of this commit as well. Then I think we can go ahead with the merge.

sawenzel (approved):

awegrzyn:
@ktf On FLPs we use python from system. As `prefer_system_check` was not been updated in this PR, it means we gonna keep using 3.6.8. Is that okay?

ktf:
i suppose so. things which require the new python are not part of the dataflow default.

PR #4179, opened 2022-05-10
Use OpenSSL from the system if > 1.0

ktf:


ktf:
The old check (force OpenSSL 1.0.x) does not make sense anymore since everyone should be using ROOT6 and the new alien / xrootd.

martenole:
Hi, unfortunately this does not work for me on Ubuntu 20.04 with aliBuild v1.11.3.
When I do `aliDoctor O2 --defaults o2` I get: SUCCESS: Package OpenSSL will be picked up from the system.
But then when I do `aliBuild build O2 --defaults o2` it still builds OpenSSL@OpenSSL_1_1_1m. I do have the dev package installed though: libssl-dev/focal-updates,focal-security,now 1.1.1f-1ubuntu2.13

So I still have problems when I load the O2 environment after the build that I cannot use `ssh` or `git push` etc

openssl.sh
```diff
@@ -4,7 +4,7 @@ tag: OpenSSL_1_1_1m
 source: https://github.com/openssl/openssl
 prefer_system: (?!slc5|slc6)
 prefer_system_check: |
-  if [ `uname` = Darwin ]; then test -d `brew --prefix openssl@1.1 || echo /dev/nope` || exit 1; fi; echo '#include <openssl/bio.h>' | c++ -x c++ - -I`brew --prefix openssl@1.1`/include -c -o /dev/null || exit 1; echo -e "#include <openssl/opensslv.h>\n#if OPENSSL_VERSION_NUMBER >= 0x10100000L\n#error \"System's GCC cannot be used: we need OpenSSL 1.0.x to build XrootD. We are going to compile our own version.\"\n#endif\nint main() { }" | cc -x c++ - -I`brew --prefix openssl@1.1`/include -c -o /dev/null || exit 1
+  if [ `uname` = Darwin ]; then test -d `brew --prefix openssl@1.1 || echo /dev/nope` || exit 1; fi; echo '#include <openssl/bio.h>' | c++ -x c++ - -I`brew --prefix openssl@1.1`/include -c -o /dev/null || exit 1; echo -e "#include <openssl/opensslv.h>\n#if OPENSSL_VERSION_NUMBER < 0x1000000L\n#error \"System's GCC cannot be used: we need OpenSSL >1 to build our own XrootD. We are going to compile our own version.\"\n#endif\nint main() { }" | cc -x c++ - -I`brew --prefix openssl@1.1`/include -c -o /dev/null || exit 1
```
martenole, line 7: There's also a small typo here: GCC should be replace by OpenSSL

ktf:
Could you try with the updated version?

martenole:
It is still building OpenSSL unfortunately

ktf:
Ok, the issue is of course that on platforms where we provide the prebuilt  packages the prefer_system will be enabled only if `--always-prefer-system` is specified on the commandline. Otherwise it will either download it or try to build our own version.

PR #4184, opened 2022-05-11
Better error detection for sim_challenge in fullCI

sawenzel:
sim_challenge can fail in multiple ways:
a) by directly exiting with code!=0
   (not leaving a textual marker)

b) if one of the sub-tests (analysis) failed
   - leaving a textual marker - but the script exited normally

This commit adjusts the error detection to cover both cases.


sawenzel:
@shahor02

shahor02:
thanks!

ktf (approved):

PR #4187, opened 2022-05-12
Enable ONNXruntime on FLP and EPN

fcatalan92:
@ktf @ihrivnac As discussed at the last WP3 meeting, this PR enables the installation of the onnxruntime library for FLPs and EPNs. The use case is the QC (FLP) and calibration (EPN) for the EMCAL that will need to run the inference of machine learning models @mfasDa. 
Of course, let me know if there is any concern.

### DIFF
--- a/defaults-o2-dataflow.sh
+++ b/defaults-o2-dataflow.sh
@@ -29,8 +29,6 @@ disable:
   - O2sim
   - O2-full-system-test
   - O2Physics
-  - ONNXRuntime
-  - MLModels
   # Fall back to the system OpenSSL and curl.
   - OpenSSL
   - curl
--- a/defaults-o2-epn.sh
+++ b/defaults-o2-epn.sh
@@ -8,7 +8,6 @@ env:
   GEANT4_BUILD_MULTITHREADED: 'ON'
 disable:
   - O2Physics
-  - ONNXRuntime
   - OpenSSL
   - curl
   - mesos


ktf:
@awegrzyn @vascobarroso @ironMann if this is ok for you could you please approve?

ironMann:
@ktf @fcatalan92 I'm not the right person to approve this. As it looks to me, this is a completely new processing components being added to the online chain. 
For the EPN part, before rollout, it should go through PDP and EPN validations (best contact the SRCs), and its impact on the CPU and GPU performance, and data flow latency evaluated with the FST.

pzhristov:
@fcatalan92 Do we really want to rely on ONNX? I remember we discussed possible use of SOFIE from Root that was supposed to be much faster and better integrated in our software stack. I haven't used it myself, so please let me know if it is not suitable and what is missing.

ktf:
@ironMann I think PDP already gave the green light at the last WP3, modulo the fact that indeed we might want to use SOFIE rather than ONNX when it becomes available officially (@pzhristov we should move to the new ROOT to be able to use it, no?).

sawenzel:
My 2 cents: I think it is reasonable to go ahead here to unblock developments and testing of ML algorithms on FLP/EPN. ONXX is the de-facto industry standard and people are using it already elsewhere (O2Physics / simulation) in our software stack. The library itself is small and I don't think it will harm the software installation.

I have not heard of SOFIE, but it appears to be in experimental state. It seems to be a more risky path to take at this moment.

fcatalan92:
@ironMann This PR enables the possibility of using onnxruntime but no actual task on EPN is using it at the moment (so there isn't any new processing component yet). I agree with you that the performance and the impact on the data flow latency must be evaluated but this is not a property of the library itself. These quantities will depend heavily on the task (e.g. how big are the models, how many times are you using them), so this validation should be addressed by the experts which would like to use ML in their specific task.

@pzhristov TMVA SOFIE is a very interesting option, but there are some caveats at the moment: requires ROOT v6.26 (any timeline for its deployment?), is currently labelled as experimental by the ROOT team, is less flexible than onnxruntime, no BDT support yet (heavily used in analysis). Of course, it has the pros of the integration with ROOT and possibly with the arrow data format. On the performance point, I think they are on par https://indico.cern.ch/event/855454/contributions/4605010/attachments/2352902/4014369/732_pdf.pdf .
IMHO further discussion and some tests are needed before taking a decision on which software to use. 

The benefit of having onnxruntime now is that people could start to develop, test, and integrate their ML models. Switching from onnxruntime to TMVA SOFIE in a later stage should not be problematic (at least for neural networks).

davidrohr:
I don't see any problem with this. With this PR it becomes only available as a library.
And once it is use it must go through the same tests as all the other software.

ktf:
@awegrzyn can you greenlight this?

pzhristov (approved):

vascobarroso (approved):

PR #4195, opened 2022-05-15
Add recipe for Upcgen MC event generator

nburmaso:
This is a recipe for Upcgen, a MC event generator dedicated to simulation of ultra-peripheral collisions. The code is available at the [repository](https://github.com/nburmaso/upcgen).

### DIFF
--- a/upcgen.sh
+++ b/upcgen.sh
@@ -1,7 +1,7 @@
 package: Upcgen
 version: "%(tag_basename)s"
 tag: tag-15-05-22-3
-source: https://github.com/nburmaso/upcgen
+source: https://github.com/alisw/upcgen
 requires:
   - ROOT
   - HepMC3
--- a/upcgen.sh
+++ b/upcgen.sh
@@ -1,0 +1,40 @@
+package: Upcgen
+version: "%(tag_basename)s"
+tag: tag-15-05-22-3
+source: https://github.com/nburmaso/upcgen
+requires:
+  - ROOT
+  - HepMC3
+  - pythia
+build_requires:
+  - CMake
+  - GCC-Toolchain:(?!osx.*)
+---
+#!/bin/bash -e
+
+cmake $SOURCEDIR                          \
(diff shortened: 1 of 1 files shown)


upcgen.sh
```diff
@@ -0,0 +1,40 @@
+package: Upcgen
+version: "%(tag_basename)s"
+tag: tag-15-05-22-3
+source: https://github.com/nburmaso/upcgen
```
sawenzel, line 4: is this your private fork of Upcgen ?

I believe that if a fork is necessary, we should have an official ALICE version under alisw. But @ktf @TimoWilken may further clarify the rules.

nburmaso: Hi Sandro. This is the main repository of Upcgen, there is no official ALICE version of it. Should I prepare one?

TimoWilken: Hi @nburmaso. Are you OK with transferring the nburmaso/upcgen repo to the alisw organisation? This will update its name to alisw/upcgen. I will give you full admin access to the alisw/upcgen repo in that case.

I've sent you an invitation to the alisw GitHub organisation -- please check your emails for the link to accept the invitation or go to <https://github.com/alisw>. When you've accepted it, you should have the rights to transfer the repo over -- go to your repo's settings and at the bottom, click the "transfer" button.

If transferring is not an option, please let me know and we can work out an alternative solution.

nburmaso: Hi @TimoWilken. Thank you for the suggestion, but I would like to keep the main repository as a personal one. Can I make a mirror of it inside `alisw` for the recipe?

TimoWilken: OK, you should be able to fork it yourself. If this case, though, any tags that you refer to in the alidist recipe must exist in the alisw/upcgen repo (whether they exist in your personal repo or not).

nburmaso: Ok, sounds good to me. I have created a fork and switched to it in the recipe.

PR #4214, opened 2022-05-20
Bump MLModels

fcatalan92:


fcatalan92:
@kft @TimoWilken Hi, I don't have permission to merge PR related to this recipe. I'm in the CODEOWNERS but probably some further step is needed.
https://github.com/alisw/alidist/blob/d6acad2d5f4cb9cd3488dc74a65e9a115b513d10/CODEOWNERS#L78
I see the message "Unknown owner: make sure @fcatalan92 exists and has write access to the repository"

TimoWilken:
Hi @fcatalan92, I've given you write access.

fcatalan92:
Hi @TimoWilken, thanks! I see no differences although, should I try to close the PR and open it again?
BTW, this update can skip CI I think. The recipe just copies some files, no code is built.

TimoWilken:
Ah, I expect this is because the security settings for this repo are set such that you need an approving review in order to merge. Even though you are the sole code owner, GitHub doesn't let you review your own PR, I think. @ktf, is this rule still useful even with the separate "require code owner reviews" rule enabled?

I also don't know if you can override the required CI status checks. In case you can't, you should at least be able to enable auto-merge, which will merge this PR once all the required status checks are green (and you have the required review).

O2 depends on MLModels; does it (or something depending on it) use the models for CI such that an unexpected file structure would cause CI errors elsewhere? If these files aren't actually used by any CI, then it's fine, otherwise I'd be hesitant to merge without at least the O2 checks passing.

fcatalan92:
> O2 depends on MLModels; does it (or something depending on it) use the models for CI such that an unexpected file structure would cause CI errors elsewhere? If these files aren't actually used by any CI, then it's fine, otherwise I'd be hesitant to merge without at least the O2 checks passing.

There is no CI using the models currently and I don't think there will be in the future.

TimoWilken (approved):

TimoWilken:
OK. Force-merging in this case; we might have to find a better way to handle this in future.

PR #4226, opened 2022-05-24
Dynamically set up env variables to speed up startup of ROOT

sawenzel:
See also https://github.com/alisw/root/commit/d6ea57cb6226f37e517bba241a546ee576ed236e.

Thanks to @TimoWilken.


sawenzel:
CI passes. Can we merge this?

sawenzel:
Well, this means we need another annotation or injected information to detect an o2-epn default.
Let's revert it for now to stabilize the situation.

PR #4229, opened 2022-05-25
Annotate FLP builds with special env variable

sawenzel:
This is in order to be able to know when we are
buiding for FLP and to take build decisions based
in this flag

See for instance https://alice.its.cern.ch/jira/browse/O2-2975

### DIFF
--- a/defaults-o2-dataflow.sh
+++ b/defaults-o2-dataflow.sh
@@ -5,7 +5,6 @@ env:
   CFLAGS: "-fPIC -O2"
   CMAKE_BUILD_TYPE: "RELWITHDEBINFO"
   CXXSTD: "17"
-  O2_BUILD_FOR_FLP: "ON"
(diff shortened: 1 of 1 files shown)


sawenzel:
@davidrohr

sawenzel:
@ktf : I think once alibuild provides this internally, we can again get rid of this annotation. (If you tag and release new alibuild to the FLP build machines also fine with me.)

ktf:
I would do it only for the O2-customization package, so that we do not rebuild the whole stack.

sawenzel:
I am not familiar with the O2-customization package. Would the builds for FLP pick this up?

ktf:
Yes, it should, the idea is to be able to customize O2 only, without having to rebuild every single package because of the choice.

sawenzel:
Fine for me

PR #4241, opened 2022-05-30
Bump FairMQ to v1.4.51

rbx:
To fix/prevent https://alice.its.cern.ch/jira/browse/EPN-143.


davidrohr:
@rbx : Do you need a PR to AliceO2 afterwards, or is this all that is needed?

rbx:
This is all, no changes in O2 necessary.

PR #4245, opened 2022-05-30
Fix ROOT hack on macOS

ktf:


ktf:
Currently broken. Merging.

PR #4254, opened 2022-06-01
Revert changes to O2 modulefile (ROOT env)

sawenzel:
The recent changes with dynamic Tcl execution in the o2 modulefile seem to be unstable ("race condition") and need more testing/refinement.

See https://alice.its.cern.ch/jira/browse/O2-2981


sawenzel:
@davidrohr @martenole

PR #4255, opened 2022-06-01
xjalienfs::Hot bugfix release to 1.4.0

adriansev:
Fix typo that prevents copy operations on systems where uid != gid


costing (approved):

sawenzel:
Tested locally and it fixes the problem spotted on alibi. This should be merged today before 18:00.

PR #4260, opened 2022-06-02
xjalienfs 1.4.2 critical bugfix

adriansev:
@ktf 
critical bugfix for downloads in directories with parents not writable by user.
whilst checking for writability of destination hierarchy, the actual destination was not taken into account, so the download would fail because the parent of destination is not writable.
affects MC productions/DPG

### DIFF
--- a/xjalienfs.sh
+++ b/xjalienfs.sh
@@ -1,6 +1,6 @@
 package: xjalienfs
 version: "%(tag_basename)s"
-tag: "1.4.0"
+tag: "1.4.1"
 source: https://gitlab.cern.ch/jalien/xjalienfs.git
 requires:
  - "OpenSSL:(?!osx)"
--- a/xjalienfs.sh
+++ b/xjalienfs.sh
@@ -1,6 +1,6 @@
 package: xjalienfs
 version: "%(tag_basename)s"
-tag: "1.4.1"
+tag: "1.4.2"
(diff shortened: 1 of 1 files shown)


costing (dismissed):

davidrohr:
All CIs passed except FullCI which failed due to not finding an alien token. The check that fails is
```
  alien-token-info >& /dev/null
  if [[ $? != 0 ]]; then
    echo "No alien token present" 1>&2
    exit 1
  fi
  ```
@adriansev : Has anything changed wrt to the alien token in this jalienfs version that could explain that? (The digitization before ran through actually, but it also doesn't check for the presense of the token).
Or @TimoWilken : Do we have again problems with alien tokens in the CI builders?

adriansev:
@davidrohr actually this is on my side, as i could reproduce the problem.
yes, there were a lots of moving code blocks to function, and it seem that my CI test for environment token usage is flawed and did not caught this .. working on it, sorry for the inconvenience!

costing (approved):

adriansev:
@davidrohr so, i fixed the problem and i could have access and correct exitcode for tokens content as env vars. could you please also check if this fix the behavior that you seen?
Thanks a lot!

davidrohr:
We'll see in the FullCI. If it passes, we can merge.

PR #4292, opened 2022-06-13
Provide HEPMC3 ROOT_INCLUDE_PATH during alibuild stage

sawenzel:
This fixes `Missing FileEntry for HepMC3/...` that appear during
full-CI FST or O2Physics compilation.

In this case, the module files are not used and hence the ROOT_INCLUDE_PATH was not correctly set.


sawenzel:
@davidrohr

PR #4312, opened 2022-06-20
Make rapidjson usable from ROOT

ktf:


awegrzyn:
It looks like running `alibuild-generate-module` without `--lib` does not populate `$PKG_ROOT`

PR #4323, opened 2022-06-23
Bump qualitycontrol to v1.63.1

Barthelemy:


sy-c (approved):

davidrohr:
why do we try to bump to 1.63.1 now when another PR to 1.64 is already open?

Barthelemy:
Because we don't know when v1.64 will be installed... it is part of the new flp suite but I don't know when, and if, it will be installed. Thus I prepared v1.63.1 as a back up solution that we can install instead in case we cannot install v1.64.

vascobarroso:
Closing in favor of #4319. If we need to use QC 1.63.1 at P2 I will cherry pick this into branch `flp-suite-v0.62.0`.

PR #4356, opened 2022-07-06
Make CXXSTD and -std=c++X agree for next-root6 defaults

TimoWilken:
We specify `CXXSTD=17`, but `-std=c++11`, which confuses ROOT. ROOT v5-25-02 only supports C++14 and up, so its build fails with the wrong `-std=`.

Cc: @chiarazampolli

### DIFF
--- a/defaults-next-root6.sh
+++ b/defaults-next-root6.sh
@@ -3,7 +3,7 @@ version: v1
 disable:
   - arrow
 env:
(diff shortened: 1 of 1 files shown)


TimoWilken:
AliRoot Mac CI is broken, and all checks using `next-root6` defaults on this PR are green, so merging this.

PR #4359, opened 2022-07-07
Bump YODA to 1.9.5

pzhristov:


ktf:
This now compiles. Shall we go ahead with this?

maireiphc:
AliGenerators compiles indeed, but it is a surprise for me...
How YODA 1.9.5 can be compiled against a Python2.7 version ?! The settings may be overwritten somewhere, and it is not somewhere directly in the defaults-generators.sh.
@adriansev was facing an Python2 issue in YODA exactly, blocking compilation while giving a try with compatibility with newer python3 (3.9.12) in PR https://github.com/alisw/alidist/pull/4168

maireiphc:
Hello Giulio,

  to my knowledge, we do not require explicitly python 3.6 in https://github.com/alisw/alidist/blob/master/defaults-generators.sh.
(i.e. no explicit requirement of Python-modules)
If it happens, it would be behind the scene via some requirements made for some dependencies.
But for instance, so far YODA has no explicit dependence on ROOT.sh (which itself requires Python-modules).

Maybe a graph from "alibuild deps --defaults generators --outgraph Yoda-Dep.pdf" would clarify where it commes from.

ktf:
@maireiphc So, shall  we merge this? is anyone relying on the old version of YODA?

pzhristov:
@maireiphc @ktf I think we should not require any support for Python 2.7 - this version is obsolete and not supported. The minimal version that we already use is 3.6, and we have the request to move to 3.9.12

ktf:
I fully agree, since we also dropped building ROOT5 (which YODA 1.9 does not support) I see no reason  why we should stick with the old version.

maireiphc:
I am not objecting the merging, which, yes, goes into the right direction.
We can move on as needed (and possibly break AliGenerators deployment - if ever it breaks anything. No problem, we have enough CVMFS deployment at hand).

I am just surprised that it works.
So we should come at some point to that and either fix or at least clean the recipe.
I actually suspect this will not work everywhere, but maybe just on old distribution where python3 is not default on the system while python2 is.

I mean: currently in https://github.com/alisw/alidist/blob/master/yoda.sh l.11 +12, we have:
```
prepend_path:
  PYTHONPATH: $YODA_ROOT/lib64/python2.7/site-packages:$YODA_ROOT/lib/python2.7/site-packages
```
i.e. while bumping to Yoda 1.9.5, there is a hard path pointing to python2.7 here; for some reason, python3.6 gets picked up under slc7 but this does not work at least on Ubuntu.
Likely because of current:
```
requires:
  - boost
  - "Python:slc.*"
  - "Python-system:(?!slc.*)"
```

For visualisation of dependencies, I have run aliBuild deps with the current recipe Vs the one under construction on my local computer (= mods in diff listed above in the thread)
`aliBuild deps YODA --defaults generators --architecture slc7_x86-64 --outgraph zYODA-deps-Current-v2022-07-09.pdf`
= [zYODA-deps-Current-v2022-07-09.pdf](https://github.com/alisw/alidist/files/9094726/zYODA-deps-Current-v2022-07-09.pdf)

`aliBuild deps YODA --defaults generators --architecture ubuntu2004_x86-64 --outgraph zYODA-deps-AfterLocalModsOnRecipe-v2022-07-09.pdf)`
= [zYODA-deps-AfterLocalModsOnRecipe-v2022-07-09.pdf](https://github.com/alisw/alidist/files/9094729/zYODA-deps-AfterLocalModsOnRecipe-v2022-07-09.pdf)

PR #4378, opened 2022-07-12
Fix sed invocations for MacOS

TimoWilken:
`sed -i` on Mac needs an argument specifying the backup suffix.

### DIFF
--- a/evtgen.sh
+++ b/evtgen.sh
@@ -17,8 +17,8 @@ env:
 rsync -a --delete --exclude '**/.git' $SOURCEDIR/ ./
 
 # adjust the configure scripts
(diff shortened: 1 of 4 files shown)


ktf (approved):

pzhristov:
Hi @TimoWilken, why do we have to create *.bak? sed -i replaces the original file, while in your change you will create a new file with extension bak and remove it, so no change will be used. Am I missing anything? The problem on Mac was that we had -i "commands", which does not work, but -i -e "commands" should be OK.

TimoWilken:
Ah, if `-i -e` is fine, that's great! This was a problem reported by a user, but I didn't read closely and it seems they have a `-i "commands"`, without the `-e`.

PR #4386, opened 2022-07-18
Bump DebugGUI

ktf:


ktf:
Only affects macOS. Merging.

PR #4390, opened 2022-07-19
Update starlight.sh

preghenella:


preghenella:
https://github.com/alisw/STARlight/pull/4

PR #4408, opened 2022-07-27
Fixed compilation in several AliGenerator packages

pzhristov:
Some packages still do not compile on MacOS, so they are switched off in aligenerators.sh.


pzhristov:
The build of ALiGenerators in the CI is using defaults-release, which relies on obsolete gfortran and does not accept -fallow-argument-mismatch, so the updated EPOS Makefile does not work. I will put support for such obsolete gfortran versions.

PR #4413, opened 2022-07-29
bump arrow

aalkin:


ktf:
did you miss xsimd?

aalkin:
xsim was added in #4411

aalkin:
@ktf @pzhristov 
The error `undefined reference to `typeinfo for llvm::ObjectCache'` seems to point to disabled LLVM's RTTI. I assume we need to enable it in for clang we build on linux. We can also try to use `-fno-rtti` for arrow, but I would expect it is actually used by gandiva.

aalkin:
Update: on ubuntu 22.04 I cannot reproduce the error, everything builds normally.

pzhristov:
I reproduced the problem on Ubuntu 18.04, but it is not clear to me how in the past the RTTI was OK, and now it is not.

TimoWilken:
Hi @aalkin @pzhristov, I've just pushed a commit that enables building RTTI in LLVM. Let's see if that fixes it.

aalkin:
Looks like enabling RTTI solved the issue, the remaining error is unrelated.

pzhristov:
Yes, I am merging this PR.

PR #4418, opened 2022-08-01
xrootd:: build only client components

adriansev:


adriansev:
Hi @TimoWilken @ktf ! as discussed, the proposal to build only client components for xrootd went to another (this) PR. On the GRID services, the proposal was welcomed, it remain to see if there any users that use xrootd server components from alibuild/cvmfs

PR #4421, opened 2022-08-02
Build LLVM with run-time type info

TimoWilken:
Required by arrow v9 (#4413).

This change is also present in #4413.


TimoWilken:
Cc: @aalkin

PR #4427, opened 2022-08-03
Fix gdb script relocation

ktf:


ktf:
Merging since currently broken in any case.

vascobarroso:
@ktf we now see the following error when installing our RPMs: 

```
 Running scriptlet: o2-O2-20220902-1.slc8.x86_64                          1/20
/usr/sbin/ldconfig: /opt/o2/lib/libarrow.so.900.0.0-gdb.py.unrelocated is not an ELF file - it has the wrong magic bytes at the start.
```

Is this something to be worried about ?

ktf:
No, it should not be a real issue. We should probably nevertheless move it somewhere else when we bump the arrow version..

PR #4428, opened 2022-08-03
Fix RapidJSON modulefile

awegrzyn:
https://github.com/alisw/alidist/pull/4312


ktf:
Merging since currently broken. The issue is most likely unrelated and actually fixed by #4427.

PR #4430, opened 2022-08-03
Readout: require mysql only on production platform

ktf:


ktf:
@sy-c would you be ok with this? So that we do not have to require mqsql installations for users e.g. on mac or ubuntu.

ktf:
Actually maybe we could drop it completely for defaults-o2 and keep it only for defaults-dataflow? You use that for the FLP deployment, no?

sy-c:
Hello @ktf, both solutions are fine, I have no preference.

sy-c (dismissed):

ktf:
I will actually do both...

PR #4431, opened 2022-08-03
golang is responsibility of alisw/dataflow-experts

ktf:


ktf:
@vascobarroso @teo @Barthelemy @awegrzyn FYI.

PR #4440, opened 2022-08-07
datadist: v1.4.10

ironMann:
Hi @davidrohr, this contains only the change to accept fractional raw tf percentage. Please include it in the next EPN bump, it is compatible with existing FLP-DD versions.

### DIFF
--- a/datadistribution.sh
+++ b/datadistribution.sh
@@ -1,6 +1,6 @@
 package: DataDistribution
 version: "%(tag_basename)s"
-tag: v1.4.8
+tag: v1.4.9
 requires:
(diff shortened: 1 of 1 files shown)


davidrohr (approved):

ktf:
@Barthelemy the error in QC seems to be genuine and a new one. Could you take a look?

ktf:
That said, it's most likely unrelated.

ironMann:
Hi @davidrohr @martenole @vascobarroso , 
Please take DD v1.4.10 for the next release. Only bugfixes compared to v1.4.9.
Thanks

PR #4442, opened 2022-08-08
Bump qualitycontrol to v1.68.1

Barthelemy:


kostorr (approved):

Barthelemy:
@TimoWilken @ktf There seems to be a problem in cs8 build

awegrzyn:
https://github.com/alisw/alidist/pull/4445

Barthelemy:
@awegrzyn I am not sure whether we should close this one. We won't install v1.69.0 immediately at P2, it is to be included in the next FLP Suite. V1.68.1 is to be installed at P2 tomorrow. 

I am not against installing QC v1.69 at P2 tomorrow (outside FLP Suite) but I don't know what @pippohub think

Barthelemy:
ignore my previous message, thank you

PR #4457, opened 2022-08-12
defaults-generators.sh - gcc-toolchain standards

maireiphc:
In order to be inline with what is used for defaults-o2.sh (https://github.com/alisw/alidist/blob/master/defaults-o2.sh), we try to move AliGenerators to our most recent standards for gcc-toolchain.
By default, gcc-toolchain is still set to : v7.3.0-alice2 (https://github.com/alisw/alidist/blob/master/gcc-toolchain.sh).
Here we intend to define the defaults-generators behaviour set to : v10.2.0-alice2

### DIFF
--- a/defaults-generators.sh
+++ b/defaults-generators.sh
@@ -5,6 +5,9 @@ env:
   CMAKE_BUILD_TYPE: RELWITHDEBINFO
   CXXFLAGS: -fPIC -g -O2 -std=c++14
 overrides:
+  GCC-Toolchain:
+    tag: v10.2.0-alice2
+    version: v10.2.0-alice2
   boost:
     requires:
     - GCC-Toolchain:(?!osx)


maireiphc:
Hello there, the AliGenerators build check failed Mon 3:44 CET, but I did not manage to locate in the log the place where it fails.
If someone can tell me what are the key words to look for, I could have a second look.
(Error, fails, failure, exit, ... did not bring me to a conclusive place)

pzhristov:
PR #4468 should solve the problems in the compilation of aligenerators.

maireiphc:
Ok, now I see in the log issued from the build check of AliGenerators that a paragraph "Error messages" is present. Before it only mentioned "Compiler Warnings".
And indeed as far as the errors are concerned, it points to errors in EPOS v3.111, a repetitive type mismatch between a real and an integer in the various calls of subroutine getbran().
The same issue has been met on my Kubuntu machine with gcctoolchain `v10.2.0-alice2` and the fix brought with the EPOS tag `v3.111-alice5` should be curing the issue also on (slc7 + gcctoolchain `v10.2.0-alice2`), as pointed by Peter.

See:
- version bump on alidist/epos.sh https://github.com/alisw/alidist/pull/4468, 
- after merge request on our EPOS GitLab in v3.111 https://gitlab.cern.ch/ALICEPrivateExternals/epos/-/merge_requests/2

PR #4473, opened 2022-08-20
alidist/YODA - introduce (py)ROOT dependency

maireiphc:
(py)ROOT can be introduced as dependency to YODA.
The feature is at least needed in order get a feature allowing the user to move from one format to the next between YODA and ROOT.
A corresponding (python+cython) package is to be built within YODA + see enduser python scripts YODA/bin/root2yoda + YODA/bin/yoda2root

Mods in yoda.sh recipe:
1. The recipe now _requires_ ROOT (pyROOT as a matter of fact).
2. The _configure_ needs to enable the (py)ROOT compatibility : ./configure --disable-silent-rules --enable-root (for any platform : macOS, slc, Ubuntu...)
3. The YODA _module_ finally needs to load ROOT along.

- Extra -

The commit also content notes and 2 questions about Python management in the recipe.
This is for my own education (PAG-MC).
The recipe is used here as a playground; expert views are needed for clarification.
Comments put in the recipe will likely be removed in a next commit,
while information got here from exchange with experts may be moved to aliBuild documentation :
https://github.com/alisw/alibuild/tree/master/docs

A very similar python interrogation runs in parallel for the lhapdf.sh recipe, where python is also needed as recipe requirement.

### DIFF
--- a/yoda.sh
+++ b/yoda.sh
@@ -11,15 +11,23 @@ requires:
 build_requires:
   - "autotools:(slc6|slc7)"
 prepend_path:
-  PYTHONPATH: $YODA_ROOT/lib/python3.9/site-packages # _Question 1_ : 3.9 hardcoded here ?
-    # should/could we move to sthg more adaptative with sthg like $PYTHON_VERSION ?
-    # or use $YODA_ROOT/bin/yoda-config --pythonpath to get the pythonpath ? (~like done below here for the module file)
+  PYTHONPATH: $YODA_ROOT/lib/python/site-packages # See below at built time the management towards generic path .../lib/python/site-packages
 ---
+#!/bin/bash
+
 rsync -a --exclude='**/.git' --delete --delete-excluded $SOURCEDIR/ ./
 
 [[ -e .missing_timestamps ]] && ./missing-timestamps.sh --apply || autoreconf -ivf
 
 (
+
+PYTHON_EXECUTABLE=$(/usr/bin/env python3 -c 'import sys; print(sys.executable)') 
+    # python3 here should come from our own Python package ("requires"), see alidist/python.sh
+    # The command will return sthg like "python3.9"
+PYTHON_VER=$( ${PYTHON_EXECUTABLE} -c 'import sys; print(f"{sys.version_info.major}.{sys.version_info.minor}")' )
+    # The command will return sthg like "3.9"
+
+
 unset PYTHON_VERSION
     # _Question 2_ :
     # YODA configure needs at least 2 python-related environment variables :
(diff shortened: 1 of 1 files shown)


yoda.sh
```diff
@@ -7,23 +7,69 @@ requires:
   - "Python:(?!osx)"
   - "Python-modules:(?!osx)"
   - "Python-system:(osx.*)"
+  - ROOT
 build_requires:
   - "autotools:(slc6|slc7)"
 prepend_path:
-  PYTHONPATH: $YODA_ROOT/lib/python3.9/site-packages
+  PYTHONPATH: $YODA_ROOT/lib/python3.9/site-packages # _Question 1_ : 3.9 hardcoded here ?
```
adriansev, line 14: See xrootd recipe https://github.com/alisw/alidist/blob/master/xrootd.sh#L84-L92
here just use a simple `python` and later (after the compilation and install), detect the actual python directory and create a `python` symlink to it.

maireiphc: Hello Adrian, 
thanks for the pointer and the advice.
I have implemented the changes in the spirit of what is done for XRootD (a simplified version here).
2 questions:

1. Any issue to anticipate with this solution on the front of the relocation issues (moving from build machine to execution ones) ?
2. Any opinion on some alternative prepand_path command below, making use of the yoda-config command ?
```
   > echo `exec yoda-config --pythonpath`
/scratch/AliceSuite/sw/ubuntu2004_x86-64/YODA/v1.9.5-local4/lib/python3.9/site-packages
   > echo $(exec yoda-config --pythonpath) | gawk -F '/' '{print $(NF-2)"/"$(NF-1)"/"$(NF)}'
lib/python3.9/site-packages
```
i.e. in the recipe, sthg like:
```
prepend_path:
  PYTHONPATH: "$YODA_ROOT/`echo $(exec yoda-config --pythonpath) | gawk -F '/' '{print $(NF-2)"/"$(NF-1)"/"$(NF)}'`"
   #  Getting the path towards python site-packages from YODA, e.g. $YODA_ROOT/lib/python3.9/site-packages
```
or simpler, but likely creating issues with relocation (?) :
```
  PYTHONPATH: `echo `exec yoda-config --pythonpath`
```


In other words, I am wondering:

1. whether it is possible to invoke bash command within the instruction 
2. and what is immediately discoverable on-the-fly after the build via some commands (is yoda-config already accessible to be launched ?).

adriansev: @maireiphc Hi! Sorry, i did not seen your reply! so:
1. you do not want absolute paths (as not to have relocation problems)
you just need a `python` --> `python3.9` symlink (where `3.9` you get from the PYTHON_VER, see https://github.com/alisw/alidist/blob/master/xrootd.sh#L24)
so, https://github.com/alisw/alidist/blob/master/xrootd.sh#L95-L100 is all you need

2. prepend_path: dictionary whose key-value pairs are an environment variable name and a path to be prepended to it, as it happens in LD_LIBRARY_PATH. This happens only after the package declaring the prepend_path in question is built, so it is not available in the same recipe (just like variables declared using env).
from http://alisw.github.io/alibuild/reference.html#the-header

so, IMHO i would not use (or rely) on shell commands run in prepend_path

Basically, the problem is: given that in a similar context, the XROOTD recipe works for us, what is the problem with doing the same for yoda?

maireiphc: Hello @adriansev, 
thanks for the reply and consideration on question 1 of the initial commit.
I consider the item resolved (what I implemented with commit https://github.com/alisw/alidist/pull/4473/commits/bfd895195bbf3a00d1b668e4493075525ed19998, second one here in this PR thread, does the work via symlink).
Do you confirm ?

maireiphc:
@ktf, @adriansev @pzhristov : question1 seems cleared by now.

We are still left with question 2 : why the sole instruction "unset PYTHON_VERSION" is enough for the body of the recipe to work ? why it is not necessary to define PYTHON_VERSION ourselves in the recipe ? - some feature somewhere seems to discover the version 3.9 for us but I did not spot who and where.
See comments in the recipe body, commit https://github.com/alisw/alidist/pull/4473/commits/2e106b70ebe0135520d7c943bc7230e3604ee10c, l.24 and later

ktf:
As described at http://alisw.github.io/alibuild/reference.html#the-body, the PYTHON_VERSION is injected automatically by alibuild.

maireiphc:
Good morning @ktf,
thanks for the hint. 
rereading the documentation about the recipe you pointed me to, I have now got an extra point that I miss:
We require a package called "python", aliBuild will then define an env variable at build time that will be <PACKAGE>_VERSION, so "PYTHON_VERSION"... In our present case, the version tag for our python package is "v3.9.12".
So, ok, doing nothing further, the requirement of the "python" package for YODA will inject that "v3.9.12" into the game, letting the YODA configure look for a command python$PYTHON_VERSION=pythonv3.9.12 at some check level. This is consistent with what I observed and reported in the commits for this PR.

What I still do not get is why, at the beginning of the recipe body, just unsetting the PYTHON_VERSION put in place by aliBuild is enough in terms of instruction.
Unsetting should let PYTHON_VERSION be an empty/undefined variable. 
How PYTHON_VERSION manages to be set to "3.9" ?
My question is: is there something specific done in aliBuild behind the scene ? or is it something done by the YODA configure itself to recover the situation ? (finding our python3 binary likely)

yoda.sh
maireiphc: @adriansev : ping for a potential confirmation on this go-ahead (symlink towards python/site-package)

maireiphc:
> My question is: is there something specific done in aliBuild behind the scene ? or is it something done by the YODA configure itself to recover the situation ? (finding our python3 binary likely)

@ktf : any feeling about the alternative above ?
My key question is if I miss sthg on the aliBuild side that you can educate me about; if the "magic" is on the YODA side, then ok it is rather on me to discover where is the thing.

In practice, the recipe should be operational as it is, it can be merged. On the other hand, you may want me to clean the comments and reduce them to their bare minimum before merging. You may tell.

yoda.sh
adriansev: @maireiphc yes, i confirm that this is the way to go (it's the only way to have a static PYTHONPATH declaration when the python bindings are local to the package)

maireiphc: Great. Thanks for the feedback ! - I will declare now this "conversation as resolved".

pzhristov:
I propose to merge this PR.

PR #4491, opened 2022-08-27
Fix ONNXRuntime build on older architectures

fcatalan92:
This PR fixes the build on some older versions of mac os and ubuntu, probably where the default python is 2.7.
The recipe silently required python in the build, now this is made explicit.


fcatalan92:
@ktf Could we merge this PR? It solves an issue reported in the mailing list and I checked that it solves the problem with a few affected people.

PR #4512, opened 2022-09-05
Adjust lib dir

benedikt-voelkel:


benedikt-voelkel:
it seems the new version installs to `lib64` by default

sawenzel (approved):

benedikt-voelkel:
failing fullCI seems to be unrelated / failed in general

benedikt-voelkel:
@pzhristov is that something we can merge?
As mentioned the exported lib dir is currently not aligned with the actual lib dir (`lib` vs ` lib64`)

ktf (approved):

ktf:
evtgen was not in the list of CODEOWNERS, so it needed some extra approval. I changed this in #4521.

PR #4517, opened 2022-09-06
Use our own EVTGEN repository

ktf:
hepforge not reliable


ktf:
Do we really need EVTGEN for our O2PDPSuite?????

ktf:
@shahor02 @pzhristov FYI

shahor02:
@ktf thanks! As for your question: O2PDPSuite covers also the simulation and since O2Sim depends on the EVTGEN, I guess also O2PDPSuite should do.

PR #4521, opened 2022-09-07
Make evtgen simulation-experts responsibility

ktf:


ktf:
@sawenzel FYI. This way if you approve changes in evtgen, it should merge directly.

PR #4524, opened 2022-09-07
Bump monitoring to v3.13.1

awegrzyn:
Auto-generated PR for the following release https://github.com/AliceO2Group/Monitoring/releases/tag/v3.13.0

### DIFF
--- a/monitoring.sh
+++ b/monitoring.sh
@@ -1,6 +1,6 @@
 package: Monitoring
 version: "%(tag_basename)s"
-tag: v3.12.0
+tag: v3.13.0
 requires:
   - boost
   - "GCC-Toolchain:(?!osx)"
--- a/monitoring.sh
+++ b/monitoring.sh
@@ -1,6 +1,6 @@
 package: Monitoring
 version: "%(tag_basename)s"
-tag: v3.13.0
+tag: v3.13.1
 requires:
   - boost
   - "GCC-Toolchain:(?!osx)"


awegrzyn:
@ktf After merging this you probably want to adapt Analysis: `enableProcessMonitoring(15, {Monitor::Cpu, Monitor::Mem, Monitor::Smaps})`

awegrzyn:
Okay, it looks like there's naming issue, lemme change that...

vascobarroso (approved):

Barthelemy:
@ktf @TimoWilken Could you have a look at the failing tests ?

ktf:
They look like CCDB issues. @costing could you take a look?

costing:
Can it be that two entities write and read back from the same path? It seems that the assertion that what was written is what is read back doesn't work, likely due to a concurrent operation. As`Test/CachingA` and `Test/CachingB` grew to 230k and 117k respectively, this seems plausible. Could these tests pick unique, random paths to test against? And wipe everything at the end of the execution?

Barthelemy:
I see that Sandro has copied the code I had in testCcdb to prevent issues when running test concurrently: https://github.com/AliceO2Group/AliceO2/pull/9798

ktf:
I suspect that's also CCDB related. The test runs slow and then it hits the timeout. @sawenzel do you have any idea of what is going on there?

PR #4525, opened 2022-09-07
datadist v1.5.0 and ucx v1.13.1

ironMann:
Hi @martenole @vascobarroso @davidrohr,

Please bundle these DD and ucx versions in the next FLP/PDP suites. I'm not sure if the compatibility will be preserved as the RDMA library is bumped. It might be ok.

- Improved cancellation of Initializing states, and exit on control system signals before reaching READY states.
- Option to drop a percentage of TimeFrames on FLPs, before building them to EPNs, i.e., not sending them over network
- Comes with bump of ucx to v1.13.1 (from v1.12.1)

### DIFF
--- a/datadistribution.sh
+++ b/datadistribution.sh
@@ -1,6 +1,6 @@
 package: DataDistribution
 version: "%(tag_basename)s"
-tag: v1.4.12
+tag: v1.5.0
 requires:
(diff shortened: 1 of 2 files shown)


davidrohr:
errors unrelated, merging

PR #4543, opened 2022-09-15
datadist v1.5.1

ironMann:
Hi @martenole @vascobarroso, this contains adaptation to monitoring changes and improvements for relocation of tf scheduling to an infra node. Please add to the next upgrade. 
I think it will need the latest monitoring library, which is already in alidist.


vascobarroso:
@ironMann @lkrcal if we want this in the next release it should be merged by the end of tomorrow

ktf:
Merged as discussed in the SRC/GPU meeting.

PR #4548, opened 2022-09-20
Allow releasing AliPhysics built with o2 defaults

TimoWilken:
The current release procedure replaces the "current" AliPhysics tag in overrides, but this didn't work before when building with o2 defaults, as there was no override.

With this change, the current AliRoot/AliPhysics release script works with o2 defaults too.

### DIFF
--- a/defaults-o2.sh
+++ b/defaults-o2.sh
@@ -12,7 +12,10 @@ disable:
 overrides:
   AliPhysics:
     version: '%(commit_hash)s_O2'
+    tag: v5-09-57h-01
   AliRoot:
+    version: '%(commit_hash)s_O2'
+    tag: v5-09-57h
     requires:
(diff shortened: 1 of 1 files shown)


ktf:
Does this actually trigger  full rebuild or not?

TimoWilken:
It shouldn't, but I've started a cache build anyway. I'll write when it's done.

TimoWilken:
Caching is done.

fcatalan92:
@TimoWilken I and others encountered problems building Aliphysics with the O2 defaults after this commit.
Setting the Aliroot tag to a newer one `v5-09-59e` solved the issue. It looks to me that the changes of this commit force the AliRoot version even if AliPhysics is in development mode using the master branch, thus the conflict.

PR #4558, opened 2022-09-26
Avoid building MbsAPI for FairRoot

ktf:


ktf:
@dennisklein do you understand what is going on here? I am trying to get rid of some dependencies, and libMbsAPI is one of the top offenders due to the fact it brings in ROOT. I tried passing BUILD_MBS=OFF but this causes the errors  you see in https://ali-ci.cern.ch/alice-build-logs/alisw/alidist/4558/f6f944115f5420d88114f16d517719116d306c28/build_alidist_O2Suite_o2_macOS/pretty.html do you understand what might be the  culprit?

dennisklein:
hm, yes, this option is broken. you need https://github.com/FairRootGroup/FairRoot/commit/695275eaea4c7eba9a9a8ecabace838fd85da5f9. I can do a v18.4.9 point release.

ktf:
thanks!

Ciao,
Giulio

dennisklein:
Candidate is in https://github.com/FairRootGroup/FairRoot/tree/v18.4_patches, however, it is not validated yet (we don't have good test coverage for that part). It would be very much appreciated, if you could test this branch, if it works for you.

ktf:
Trying now. BTW, what do we lose if we drop the boost dependency in FairRoot? We of course use boost elsewhere, but I have the impression in FairRoot you bring in way more stuff than (say) FairMQ.

dennisklein:
> what do we lose if we drop the boost dependency in FairRoot?

AFAICS, it is mostly used in examples and Boost.Serialization is included in some places in libBase, but I doubt it is essential.

(I even wish, we could drop boost from FairMQ, but the program_options stuff is not so easy to lose now, I am afraid.)

dennisklein:
FairRoot also has the Cling dimension. So, some dependencies are not even needed at build time really, only for the `source-me.sh` script, that FairRoot installs, so that the user macros can find the non-system installations of certain deps. (which is totally the wrong place to solve this :( ..., and which is one of the pain points that stalls the FairRoot CMake modernization now for years ...)

ktf:
Is there an option to drop boost, even if not found on the system?

dennisklein:
> Is there an option to drop boost, even if found on the system?

Nope, not at the moment. But I think it should be possible, I have put it on the agenda: https://github.com/FairRootGroup/FairRoot/issues/1228.

dennisklein:
https://github.com/FairRootGroup/FairRoot/releases/tag/v18.4.9

ktf:
@TimoWilken new error...

TimoWilken:
@ktf Thanks for the ping. That path refers to the previous Nomad allocation, but we moved everything to the new, now-running allocation because we use "sticky" ephemeral disks in Nomad to preserve the build cache across job restarts.

I don't know if MCStepLogger is doing something funny and hardcoding paths that shouldn't be hardcoded in its exported CMake files, or if this is a general problem where the only solution is to stop reusing the build area across restarts.

ktf:
Rest of the tests were already done. Merging.

PR #4561, opened 2022-09-27
Install xjalienfs correctly if running aliBuild inside venv

TimoWilken:
When we run aliBuild from a virtual environment, the old xjalienfs recipe installs xjalienfs inside that virtualenv, not inside its $INSTALLROOT where it should go. Fix that and force it to install inside $INSTALLROOT. We always add the $INSTALLROOT to our $PYTHONPATH anyway.

@ktf @adriansev Does this look alright to you?

I've tested this on slc7 when running aliBuild inside a virtualenv. In that case, alien.py is installed properly and I can run it without ImportErrors.

### DIFF
--- a/xjalienfs.sh
+++ b/xjalienfs.sh
@@ -3,25 +3,29 @@ version: "%(tag_basename)s"
 tag: "1.4.2"
 source: https://gitlab.cern.ch/jalien/xjalienfs.git
 requires:
- - "OpenSSL:(?!osx)"
- - "osx-system-openssl:(osx.*)"
- - XRootD
- - AliEn-Runtime
- - Python-modules:(?!osx_arm64)
+  - "OpenSSL:(?!osx)"
+  - "osx-system-openssl:(osx.*)"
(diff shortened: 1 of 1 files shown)


TimoWilken:
Hm, that's a good point. The Python-modules recipe manually deactivates any venv it finds, but seems to do some special handling of `system-requirements.txt`. Not sure if anything actually uses that file.

For xjalienfs, I've got an issue with the CI system at the moment, where an update of xjalienfs messed up the CI's venv (outside of `sw/`), which I want to avoid.

ktf:
This still fails no?

TimoWilken:
Yes, because the current (alidist master) recipe affects the host venv, rather than anything inside `sw/`. So other builds running breaks the host, which makes this PR fail too. This PR doesn't *fix* the host's venv once it's broken, it just prevents it from getting broken if every build has this change.

ktf:
As discussed, fine to merge then.

PR #4575, opened 2022-10-03
Bump JAliEn-ROOT 0.6.7

ktf:
This drops the dependency of JAliEn-ROOT on RDataframe and Graf3d.


ktf:
@TimoWilken I have the impression there is something wrong with thew fullCI checker. Could you have a look?

PR #4581, opened 2022-10-06
Bump qualitycontrol to v1.77.0

Barthelemy:


ktf:
For the record, this is still breaking because of gitlab:

```
RuntimeError: Error 128 from git ls-remote --heads --tags https://gitlab.cern.ch/jalien/alice-grid-utils.git: fatal: unable to access 'https://gitlab.cern.ch/jalien/alice-grid-utils.git/': The requested URL returned error: 503
```

I have reopened the ticket. Apparently we are not the only ones having troubles.

PR #4583, opened 2022-10-06
bump jalien-ROOT to 0.6.8

sawenzel:
same reasoning as #4582


ktf:
Not sure what's up with one of the alibuildmac02, it pings, but I cannot login. @costing this is hosted downstairs, no? Do you have better insight of what it's happening with it?

costing:
Yes, it's hosted downstairs but I don't have the credentials for it. We can look at it together on Mon.

PR #4585, opened 2022-10-07
Drop ApMon from monitoring

ktf:


ktf:
@costing @awegrzyn @davidrohr as discussed.

PR #4592, opened 2022-10-10
Revert "Drop ApMon from monitoring"

ktf:
Reverts alisw/alidist#4585

### DIFF
--- a/monitoring.sh
+++ b/monitoring.sh
@@ -4,6 +4,7 @@ tag: v3.14.0
 requires:
   - boost
   - "GCC-Toolchain:(?!osx)"
+  - "ApMon-CPP:(?!osx)"
   - curl
   - libInfoLogger
 build_requires:


ktf:
This apparently does not work on the grid. @awegrzyn is there a way to disable explicitly ApMon  support? As far as  I can tell it is being loaded by Monitoring and then it fails because it's not in path anymore.

awegrzyn:
Okay, I'm not aware what's being used on the grid. On FLPs we disable it in `defaults-o2-dataflow.sh`. If you want to disable it during runtime, I don't think it's possible atm.

ktf:
I meant I wanted to disable it at build time.  In principle, no one is using it, so I am not sure how we can end up with Monitoring linking against it. Maybe it's actually somehow found in the docker image? Do you have some cmake flag which does not even look for it on the system?

awegrzyn:
True, you're right, it does look in `/usr/local`. Lemme disable it.

ktf:
Indeed I think this is the issue...

PR #4594, opened 2022-10-10
ROOT Python scripts: don't hardcode build-time path to Python

TimoWilken:
ROOT is currently built using an absolute path specified as `$PYTHON_EXECUTABLE`. This is then hardcoded into Python scripts installed by ROOT, but the build-time Python executable is not necessarily the same as the runtime one (e.g. for packages built using Jenkins and installed on CVMFS).

This was initially done (in commit 226160ae138e8d72822895c8916df4eb42c3e2f9) in order to handle Python executables installed by environment managers like asdf and pyenv. See the message of the above commit for details.

Instead, when we aren't using a Python executable from an env manager, use just "python3" as the executable, so that it is looked up in `$PATH` at runtime. At runtime, we should be inside an alienv environment. If we built our own Python, it will be picked up from `$PATH`; if not, we will fall back to the system Python as expected.

### DIFF
--- a/root.sh
+++ b/root.sh
@@ -87,14 +87,25 @@ fi
 if [[ -d $SOURCEDIR/interpreter/llvm ]]; then
   # ROOT 6+: enable Python
   ROOT_PYTHON_FLAGS="-Dpyroot=ON"
-  ROOT_PYTHON_FEATURES="pyroot"
   ROOT_HAS_PYTHON=1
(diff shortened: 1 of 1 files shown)


davidrohr:
@TimoWilken : can we merge this?

TimoWilken:
@davidrohr I'll just make sure it's cached for the CI, then I'll merge it!

PR #4595, opened 2022-10-10
Remove ApMon from Monitoring

awegrzyn:
https://github.com/alisw/alidist/pull/4592

### DIFF
--- a/monitoring.sh
+++ b/monitoring.sh
@@ -27,6 +27,7 @@ fi
 cmake $SOURCEDIR                                              \
       -DCMAKE_INSTALL_PREFIX=$INSTALLROOT                     \
       ${BOOST_REVISION:+-DBOOST_ROOT=$BOOST_ROOT}                 \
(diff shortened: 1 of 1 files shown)


monitoring.sh
```diff
@@ -28,7 +27,6 @@ fi
 cmake $SOURCEDIR                                              \
       -DCMAKE_INSTALL_PREFIX=$INSTALLROOT                     \
       ${BOOST_REVISION:+-DBOOST_ROOT=$BOOST_ROOT}                 \
-      ${APMON_CPP_REVISION:+-DAPMON_ROOT=$APMON_CPP_ROOT}         \
```
ktf, line 31: This one I guess we can actually leave, no? Or you removed the option completely?

awegrzyn: Option is still there, I don't have preference so I brought it back

vascobarroso (approved):

PR #4596, opened 2022-10-10
Adding faster zlib implementation from Cloudflare

sawenzel:
Known to be a faster replacement for zlib.
Has the potential to speedup ROOT IO heavy
workflows by some ~10%.

A free lunch since drop-in replacement.

### DIFF
--- a/zlib-cloudflare.sh
+++ b/zlib-cloudflare.sh
@@ -1,0 +1,24 @@
+package: zlib-cloudflare
+version: "%(tag_basename)s"
+tag: gcc.amd64
+source: https://github.com/cloudflare/zlib/
+build_requires:
+ - "GCC-Toolchain:(?!osx)"
+ - CMake
+ - alibuild-recipe-tools
+---
+#!/bin/sh
+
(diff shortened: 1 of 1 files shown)


sawenzel:
I came across this recently. It appears that this is a faster zlib implementation that is essentially free lunch for our processing. The AOD conversion speeds up by 10% using this lib as a drop in replacement (just change the LD_LIBRARY_PATH) and I expect gains in every place where write large ROOT files (TPC clusterization, digitization). This comparison was done against the standard version zlib that we also build ourselves.

I'd like to start a discussion whether we can move to this optimized version, or at least build it in the software stack so that MC can use it on the GRID.

@ktf, @pzhristov , @costing

ktf:
I am all for it. We should also redo all the benchmarks we did in the past with other algorithms, IMHO. Now that we have I/O CPU budget to spare, maybe we can afford higher compression rates.

sawenzel:
So I think we could go on here. This will not change in production for the moment, just make this available for testing.

PR #4597, opened 2022-10-10
Bump qualitycontrol to v1.78.2

Barthelemy:


Barthelemy:
Closing because it is not an official version, only for Giulio

PR #4608, opened 2022-10-14
Adjustments to compile on ARM/aarch64

sawenzel:
* O2 does not depend on KFParticle anymore (KFParticle does not compile currently on ARM)
* adjust Clang and VecGeom to compile on ARM

see https://alice.its.cern.ch/jira/browse/O2-3278

### DIFF
--- a/clang.sh
+++ b/clang.sh
@@ -28,6 +28,7 @@ esac
 case $ARCHITECTURE in
   *_x86-64) LLVM_TARGETS_TO_BUILD=X86 ;;
   *_arm64) LLVM_TARGETS_TO_BUILD=AArch64 ;;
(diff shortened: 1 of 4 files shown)


TimoWilken:
Does O2 not use KFParticle at all any more or should it still be built on x86? If so, you can use `"KFParticle:(?!_aarch64)"` as a `requires` instead of removing it completely.

pzhristov:
@TimoWilken I think KFParticle dependency is a remnant of the time when O2Physics was not yet created. O2 itself probably will never depend on KFParticle, but O2Physics most probably will use it. So we can move it there and protect it for the moment using "KFParticle:(?!_aarch64)"

PR #4645, opened 2022-11-02
Install pkgconfig file for UUID

TimoWilken:
XRootD requires this, as it searches for libuuid using pkg-config in its CMake recipes. This leads to build failures on Ubuntu. While users can install libuuid on their system to provide the `*.pc` file, this seems silly if we're building the library anyway.

I've also fixed some linter errors related to quoting issues in the script, as this seemed like a good opportunity.


ktf:
Can you warm up the cache with this?

PR #4650, opened 2022-11-04
Bump arrow to 10.0.0

aalkin:


pzhristov:
@aalkin Should we start tagging the arrow branch instead of using hashes?

ktf:
Definitely. Although it  apparently does not build as it seems to be lacking the option for C++17.

aalkin:
@ktf that option supposed to come from `defaults`, is it not?

ktf:
I think in the default we simply have the env variable, then we need to expose it to cmake.

PR #4653, opened 2022-11-04
root.sh :: fix python shebangs

adriansev:
@TimoWilken I noticed that the 10 root commands based on python use invalid python path.
With this PR i copied from xjalienfs the fixing of shebangs.


TimoWilken (approved):

TimoWilken:
Looks good to me; I'll cache this (as merging this will cause lots of rebuilds and slowdowns in CI) and then merge it.

PR #4670, opened 2022-11-17
Fix igprof recipe to work on hyperloop

ktf:


ktf:
error unrelqted. tested on hyperloop and not part of the test suite.

PR #4672, opened 2022-11-17
Bump ROOT to v6-26-10-alice4

ktf:


pzhristov:
We need a fix in ROOT/cmake/modules/FindArrow.cmake, otherwise the cmake configuration fails with Arrow 9.0.0.

ktf:
@pzhristov as discussed I reverted the arrow 10 cmake fixes in root so that this will hopefully work with arrow 9.

pzhristov (dismissed):

PR #4689, opened 2022-11-28
Bump debug gui

ktf:


davidrohr:
Will this fix it only for the remote GUI, or also for the imgui DebugGUI when I run locally?

ktf:
For the OpenGL backend, as I said, I will have to bump the full ImGUI, so I will do it as time permits.

PR #4690, opened 2022-11-28
fix the version of setuptools

Barthelemy:


sy-c (dismissed):

Barthelemy:
what about `setuptools==59.6.0` in `defaults-o2-dataflow` and we change it asap we move to python 3.9 ?

awegrzyn:
Yes, we also have distinction between pythons, so we could have `setuptools==59.6.0` for 3.6 and `setuptools==65.5.1` for 3.9.

Barthelemy:
see my commit

awegrzyn:
@ktf Is that okay with you?

Barthelemy:
Could someone approve and merge ?

PR #4692, opened 2022-11-29
Bump DebugGUI version to v0.6.7

ktf:


ktf:
Changes only some javascript file. Merging.

PR #4694, opened 2022-11-29
Added EPOS4 recipe

jackal1-66:
The build was tested in a Ubuntu machine


epos4.sh
```diff
@@ -0,0 +1,77 @@
...
+requires:
+  - ROOT
+  - fastjet
+env:
+  EPOVSN: "4.0.0"
+  EPO: "${EPOS_ROOT}/epos4/"
+  LIBDIR: "${EPO}bin"
+  OPT: "./"
+  HTO: "./"
+  CHK: "./"
```
ktf, line 14: These will be set at build time for the dependencies of EPOS4. Is that what you really need?

epos4.sh
```diff
@@ -0,0 +1,77 @@
...
+set EPOS_ROOT \$::env(BASEDIR)/$PKGNAME/\$version
+setenv EPOS_ROOT \$EPOS_ROOT
+setenv EPOVSN 4.0.0
+# Final slash is required by EPOS, please leave it be
+setenv EPO \$::env(EPOS_ROOT)/epos4/
+prepend-path PATH \$::env(EPO)bin
+setenv LIBDIR \$::env(EPO)bin
+setenv OPT ./
+setenv HTO ./
+setenv CHK ./
```
ktf, line 76: Is this what you really need? These will be set at runtime when you load the module.

jackal1-66: It's in the tutorial also of the installation of EPOS4 standalone. They need to be set and that should be the default one, but the user can change it afterwards obviously.

epos4.sh
jackal1-66: I removed them cause they aren't needed (for now).

maireiphc (approved):

PR #4697, opened 2022-11-30
New attempt at fixing the check in qc repo

Barthelemy:


Barthelemy:
@awegrzyn @ktf @knopers8

PR #4701, opened 2022-12-04
Add ITSResponse package

mconcas:
Hi @shahor02 @ktf, @TimoWilken 

As briefly discussed I moved the numerous files related to ITS ALPIDE chip responses to a dedicated repository: https://github.com/AliceO2Group/ITSChipResponseData
Here I am adding a recipe to get them as a prerequisite during O2 installation.
In pull-request https://github.com/AliceO2Group/AliceO2/pull/10352 the files are used to create the .root file.
Can this work? Let me know if something should be done differently.

Giulio, does this approach still require not to be run during the CI?

### DIFF
--- a/itsresponse.sh
+++ b/itsresponse.sh
@@ -7,8 +7,8 @@ build_requires:
 ---
 #!/bin/bash -e
 
-mkdir -p ${INSTALLROOT}/response
-cp -r ${SOURCEDIR}/* $INSTALLROOT/response
+mkdir -p "$INSTALLROOT/response"
+cp -r "$SOURCEDIR"/* "$INSTALLROOT/response"
 
 # Modulefile
 MODULEDIR="$INSTALLROOT/etc/modulefiles"
--- a/itsresponse.sh
+++ b/itsresponse.sh
@@ -1,0 +1,23 @@
+package: ITSResponse
+version: "%(tag_basename)s"
+tag: v1.0.0
+source: https://github.com/AliceO2Group/ITSChipResponseData.git
+build_requires:
+  - alibuild-recipe-tools
+---
+#!/bin/bash -e
+
+mkdir -p ${INSTALLROOT}/response
+cp -r ${SOURCEDIR}/* $INSTALLROOT/response
+
+# Modulefile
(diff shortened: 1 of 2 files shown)


mconcas:
Hi, any feedback on this?

mconcas:
Hi @shahor02  
> In general looks correct to me but I will not be able to test/go in details in nearest time.
Ok, not a problem. The package "installation" itself should be harmless if I did not overlook something.
Then I guess one could have already some feedack from CI.
>I did not understand your comment about not running this in the CI, why it should not run?
In first place I thought that we could have stored these files on a remote repo and just make the converter exe to fetch them from there at O2 installation stage. When we briefly chatted about this with Giulio he mentioned that probably this should have been avoided during CI.
Now, whit this "module" approach, the executable looks in module-related path for the files and then runs on local repo.
And probably will not be a worry for any automated thing,

TimoWilken:
Hi @mconcas, looks good to me in general (and the fullCI error is unrelated; it's the O2Physics build running out of memory).

I've pushed a small commit fixing a few shell quoting issues in the script, I hope that's OK.

mconcas:
> Hi @mconcas, looks good to me in general (and the fullCI error is unrelated; it's the O2Physics build running out of memory).
> 
> I've pushed a small commit fixing a few shell quoting issues in the script, I hope that's OK.

Sure, thanks a lot!

awegrzyn:
This is inconsistent with `O2` modulefile...
Depending on usecase, either O2 modulefile needs to modified or `ITSResponse` moved under `build_requires`
@mconcas Could you clarify if it's runtime or build time dependency?

btw. @ktf Would it be possible to move it to `alibuild-generate-module`?

mconcas:
> This is inconsistent with `O2` modulefile... Depending on usecase, either O2 modulefile needs to modified or `ITSResponse` moved under `build_requires` @mconcas Could you clarify if it's runtime or build time dependency?
> 
> btw. @ktf Would it be possible to move `O2` modulefile to `alibuild-generate-module`?

Hi, it is required during the build time. My bad there, see: #4725 .

PR #4703, opened 2022-12-05
[QC-469] remove python deps for repocleaner (moved to pip)

Barthelemy:


awegrzyn (approved):
:star2:

ktf:
Check it actually works having the variable empty.

PR #4704, opened 2022-12-05
Added EPOS4 compatibility to AliGenerators

jackal1-66:


ktf:
I am a bit confused about the renaming of the variables. Can you make sure that if they are not needed they are not exported but just set as local shell variables? Thanks.

jackal1-66:
Ciao @ktf, I had to rename that way those variables in order to make EPOS4 and EPOS coexist in AliGenerators. I tested it in my local installation and it worked like a charm. I also edited the EPOS4 source accordingly.

PR #4712, opened 2022-12-07
Bump JAliEn-ROOT and libjalienws

ktf:


ktf:
Trusting @costing and merging, since it compiles fine on ARM mac.

PR #4723, opened 2022-12-11
Removed edit committed on 23/09/2020

jackal1-66:
After this edit, it was not possible anymore to launch Rivet on the GRID, including the loading of the AliGenerators package. 
So a test could be valuable to regain the usage of the framework, whose last version runnable is the 22-09-2020 on the GRID.


jackal1-66:
@ktf Ciao Giulio, I see that the macOs build fails for an error that is completely unrelated to my pull request, can you please check what's the issue?

PR #4725, opened 2022-12-13
Move ITSresponse in O2 build_requires

mconcas:
@awegrzyn: FYI

### DIFF
--- a/o2.sh
+++ b/o2.sh
@@ -26,7 +26,6 @@ requires:
   - FFTW3
   - ONNXRuntime
   - MLModels
-  - ITSResponse
 build_requires:
   - GMP
   - MPFR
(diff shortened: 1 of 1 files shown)


martenole:
@mconcas is this package already required for the O2 nightly tag of today?

ktf:
@martenole notice I had to add it to the "override" packages in the Jenkins job description.

martenole:
I was just wondering, because the WeeklyO2Release build uses the most recent FLP suite branch in alidist. But then it takes the o2 recipe from alidist@master. Here then the ITSresponse package is listed, but its not yet included in the alidist branch from the FLP suite. This makes the build fail with `ERROR: [Errno 2] No such file or directory: 'alidist/itsresponse.sh'`
So for building I am now not overwriting the o2 recipe anymore. Of course this would not work if O2 already required it. But it doesn't yet, this I asked Matteo and he confirmed to me on mattermost

ktf:
@martenole You need to add ITSResponse to the ALIDIST_OVERRIDE_PKGS and you should be fine. I did it now for WeeklyO2Release too.

awegrzyn:
@ktf Could you please merge? This breaks our RPM generation.

PR #4726, opened 2022-12-13
[QC-901] downgrade ROOT for o2-dataflow until we have a proper fix fr…

Barthelemy:
…om ROOT


vascobarroso (approved):

pzhristov:
@Barthelemy Does reverting the PR  https://github.com/root-project/root/pull/11334/files solve the problem of TList? In such case instead of going back to the old Root we can patch our fork and provide a new tag.

PR #4727, opened 2022-12-13
Update AliEn-CAs to latest commit

TimoWilken:
This commit of [alisw/alien-cas](https://github.com/alisw/alien-cas) includes updated Grid certificates.

In addition, I've taken the opportunity to fix a few formatting and quoting issues.


TimoWilken:
I've cached this and the error is unrelated; merging.

PR #4730, opened 2022-12-14
Check alidist recipes using a more thorough linter

TimoWilken:
I've been working on [a linter for alidist recipes](https://github.com/TimoWilken/alidistlint) occasionally (when I've had time between other things). It takes a recipe, feeds the header through yamllint, any scripts through shellcheck, and applies a few checks of its own to make sure the YAML header is valid.

This PR adds a workflow that checks any changed recipes (for PRs) or all recipes (on push) using alidistlint.

It includes all the checks done by the old `scripts/lint-recipes`, so I'm removing that in this PR. However, I've kept the previous check for circular dependencies, as that's not covered by alidistlint.


TimoWilken:
Hi @ktf, after applying #4859, the following errors remain. However, fixing these would change the respective recipes' hashes and cause rebuilds.

What do you think the best course of action would be here?

```
alice-grid-utils.sh:16:23: error: Couldn't find end token `EoF' in the here document. [SC1044]
autotools.sh:41:9: error: -d doesn't work with globs. Use a for loop. [SC2144]
defaults-o2-dataflow.sh:47:7: error: overrides.O2-customization.env.ENABLE_UPGRADES: must be of string type [ali:schema]
defaults-o2-dataflow.sh:48:7: error: overrides.O2-customization.env.BUILD_ANALYSIS: must be of string type [ali:schema]
defaults-o2-dataflow.sh:49:7: error: overrides.O2-customization.env.BUILD_EXAMPLES: must be of string type [ali:schema]
defaults-o2-dataflow.sh:50:7: error: overrides.O2-customization.env.O2_BUILD_FOR_FLP: must be of string type [ali:schema]
fairmq.sh:21:3: error: duplication of key "ROOT_INCLUDE_PATH" in mapping [yl:key-duplicates]
glog.sh:1:0: error: metadata not found or empty (is the '\n---\n' separator present?) [ali:empty]
looptools.sh:28:59: error: Modules 4 does not allow colons in prepend-path [ali:colons-prepend-path]
mesos-workqueue.sh:1:0: error: metadata not found or empty (is the '\n---\n' separator present?) [ali:empty]
openloops.sh:27:13: error: Double quote array expansions to avoid re-splitting elements. [SC2068]
openloops.sh:32:13: error: Double quote array expansions to avoid re-splitting elements. [SC2068]
openmp.sh:2:1: error: version: must be of string type [ali:schema]
parquet-cpp.sh:9:1: error: duplication of key "build_requires" in mapping [yl:key-duplicates]
pda.sh:1:0: error: metadata not found or empty (is the '\n---\n' separator present?) [ali:empty]
rivettask.sh:2:1: error: version: %-format error: unsupported format character '{' (0x7b) at index 1 [ali:schema]
rivettask.sh:2:1: error: version: invalid substitution type '%{'; use only %(...)s (or use '%%' if you want a literal '%') [ali:schema]
rivettask.sh:2:1: error: version: invalid substitution type '%{'; use only %(...)s (or use '%%' if you want a literal '%') [ali:schema]
rivettask.sh:2:1: error: version: invalid substitution type '%{'; use only %(...)s (or use '%%' if you want a literal '%') [ali:schema]
zlib-cloudflare.sh:24:24: error: Couldn't find end token `EOF' in the here document. [SC1044]
```

ktf:
Apart from autotools.sh, defaults-o2-dataflow.sh, zlib-cloudflare.sh, the others we can fix. For the remaining ones I would add some exclusion list and fix them once we really rebuild everything (unfortunately we missed the GCC bump)..

TimoWilken:
Hi @ktf, thanks, see #4860 for those.

TimoWilken:
@ktf, on the `master` branch of alidist, there are now zero linter errors from alidistlint; only warnings and hints. I'd say that makes this ready to merge.

PR #4749, opened 2023-01-09
Attempt at recipe for bookkeeping

Barthelemy:


sy-c (dismissed):

bookkeeping-api.sh
```diff
@@ -0,0 +1,39 @@
...
+    SONAME=dylib
+  ;;
+  *) SONAME=so ;;
+esac
+
+cmake $SOURCEDIR/cxx-client                  \
+      ${CMAKE_GENERATOR:+-G "$CMAKE_GENERATOR"}                 \
+      ${CMAKE_BUILD_TYPE:+-DCMAKE_BUILD_TYPE=$CMAKE_BUILD_TYPE} \
+      -DCMAKE_INSTALL_LIBDIR=lib          \
+      -DOPENSSL_ROOT_DIR=$OPENSSL_ROOT_DIR \
```
TimoWilken, line 31: Please use `${OPENSSL_ROOT_DIR:+-DOPENSSL_ROOT_DIR="$OPENSSL_ROOT_DIR"}` instead, to handle the case where the variable is not set (e.g. if OpenSSL is disabled).

bookkeeping-api.sh
```diff
@@ -0,0 +1,39 @@
...
+incremental_recipe: |
+  cmake --build . -- ${JOBS+-j $JOBS} install
+  mkdir -p $INSTALLROOT/etc/modulefiles && rsync -a --delete etc/modulefiles/ $INSTALLROOT/etc/modulefiles
+---
+
+case $ARCHITECTURE in
+  osx*)
+    # If we preferred system tools, we need to make sure we can pick them up.
+    [[ ! $GRPC_ROOT ]] && GRPC_ROOT=`brew --prefix grpc`
+    [[ ! $OPENSSL_ROOT ]] && OPENSSL_ROOT_DIR=$(brew --prefix openssl@1.1)
```
TimoWilken, line 20: There seems to be some confusion between `OPENSSL_ROOT` and `OPENSSL_ROOT_DIR` here. Is this intentional?

bookkeeping-api.sh
```diff
@@ -0,0 +1,39 @@
...
+
+    SONAME=dylib
```
TimoWilken, line 22: The `SONAME` variable doesn't seem to be used anywhere, and it's not exported. Can you remove it?

TimoWilken (commented):
Hi @Barthelemy, looks good to me in general, but I'm a bit confused about how some variables are used (see comments below).

bookkeeping-api.sh
Barthelemy: I think that I copied it blindly from `o2.sh`. I'll remove it.

bookkeeping-api.sh
Barthelemy: agreed

bookkeeping-api.sh
Barthelemy: yes, indeed, I will fix that. 

Still, there is something wrong with the grpc recipe because we should not have to explicitely specify the dependency. We load grpcs that should be enough to get openssl.

Barthelemy:
@TimoWilken @ktf Any idea why I am getting `/sw/SOURCES/grpc/v1.50.1/v1.50.1/src/core/tsi/ssl_transport_security.cc:865:20: error: 'EC_KEY_new_by_curve_name' was not declared in this scope` ? 

I don't see a reason the changes in the grpc recipe could cause this. Moreover I have tested with success on CS8 and mac.

sy-c (approved):

Barthelemy:
@TimoWilken could you approve/merge please ? thank you

TimoWilken (approved):

PR #4751, opened 2023-01-09
Add IgProf as a dependency of O2PDPSuite

ktf:


ktf:
I would like to have IgProf available in the environment for the memory investigations which I am currently doing with @chiarazampolli & co.

PR #4758, opened 2023-01-12
Bump qualitycontrol to v1.86.0

Barthelemy:


sy-c (approved):

ktf:
@sawenzel can we increase the timeout for `o2sim_hepmc`? It seems to go in timeout very often in the last couple of months.

PR #4761, opened 2023-01-12
Revert "Add IgProf as a dependency of O2PDPSuite (#4751)"

martenole:
This reverts commit 42e1b53588852990bad6f8ce619a22b0897b9fd2.

@ktf the build failed again, so I would revert your commit for now: https://alijenkins.cern.ch/job/DailyBuilds/job/WeeklyO2Release/241


ktf:
Indeed. I am looking into it. Not sure how come the CI actually worked with it.

PR #4762, opened 2023-01-12
Make sure we can build igprof on the EPNs

ktf:


ktf:
Tested by hand on the EPN. Error unrelated.

PR #4767, opened 2023-01-16
Fix alidist recipe linter errors

TimoWilken:
This fixes errors in alidist recipes that would be found by the linter in https://github.com/alisw/alidist/pull/4730.

I expect these changes won't be controversial, as it's mostly just whitespace and formatting fixes in YAML headers. However, there are a few formatting and syntax fixes for recipes, too.

See individual commits for details and rationale of individual changes.

### DIFF
--- a/abseil.sh
+++ b/abseil.sh
@@ -1,6 +1,6 @@
 package: abseil
 version: "%(tag_basename)s"
-tag:  "20220623.1"
+tag: "20220623.1"
 requires:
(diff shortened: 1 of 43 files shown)


ktf:
Please mark this as draft

ktf:
This will trigger a huge rebuild storm. I would do a PR for the things which affect leaves / almost leaves and then we check what is left.

PR #4776, opened 2023-01-20
Fix linter errors that don't cause rebuilds

TimoWilken:
These are whitespace changes or changes to recipes that are force-rebuilt every time anyway.

See individual commits for details.

Needed for https://github.com/alisw/alidist/pull/4730.


TimoWilken:
Looking at the CI builders, I can confirm that these changes don't cause any extra rebuilds.

PR #4781, opened 2023-01-25
Add O2DPG sim test recipe

benedikt-voelkel:


benedikt-voelkel:
A few refinements for the underlying tests to be run is here: https://github.com/AliceO2Group/O2DPG/pull/834

PR #4799, opened 2023-02-06
Keep going when reporting QualityControl build errors

ktf:


davidrohr:
why do we want that? to see all build errors?
I am not sure if this is always the right thing to do.

ktf:
@Barthelemy @knopers8 I suggest we add the "-k 0" flag when building QC, as it will report all build errors before exiting with a non-zero exit code. This may slow down testing PRs as it won't stop at the first error, but it allows seeing / fixing all issues at once. For small projects like QC, it might actually be an overall win in terms of time spent checking PR.

ktf:
I can make it optional for the CI, but does it really matter? In general QC builds very fast, no?

ktf:
To be clear, -k0 does not prevent incremental builds from happening.

Barthelemy:
Hi, 
I don't mind the addition of the flag. Let's see if it gets annoying locally or not and then we can decide to have it only for the CI.

Barthelemy (dismissed):

ktf:
Looks like we are using make to build QC. Is that actually expected?

Barthelemy:
on my machine it uses `ninja`. Could it be caused by the piece of code you removed in the recipe ?

ktf:
doh

Barthelemy (approved):

ktf:
macOS errors unrelated. Merging.

PR #4809, opened 2023-02-09
Bump geant3 to v4.2

ihrivnac:
Includes fix of a segmentation violation observed on MacOSX with M1/M2 processors with AliRoot (https://github.com/vmc-project/geant3/pull/40)


ihrivnac:
@pzhristov , can you try ?

PR #4813, opened 2023-02-10
Do not fetch clang-format out of band

ktf:


ktf:
Avoid fetching things out of band, so that we do not need to worry about the build environment having a working OpenSSL.

ktf:
Given we need to bump llvm to compile arrow 11, let's do it in the same PR.

davidrohr:
this is obsolete, since git-clang-format is now installed from clang15 anyway

PR #4827, opened 2023-02-15
Point test to source dir

benedikt-voelkel:
* in the source dir there is the entire git repo (.git)

* from there it is possible to collect based on
  * ALIBUILD_HEAD_HASH
  * ALIBUILD_BASE_HASH

* all tests will be executed in the build dir while the source dir will stay untouched

### DIFF
--- a/o2dpg-sim-tests.sh
+++ b/o2dpg-sim-tests.sh
@@ -17,7 +17,7 @@ pushd $BUILDDIR/o2dpg-sim_tests
 LOGFILE="o2dpg-sim-tests.log"
 
 O2DPG_TEST_EXITCODE=0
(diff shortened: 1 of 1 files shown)


benedikt-voelkel:
Hej @TimoWilken 
The change is just to give the path so that the source dir is known as discussed.
It will change of course nothing right now.

Only when https://github.com/AliceO2Group/O2DPG/pull/875 is merged, we will start to dynamically pick up changes and run corresponding tests in the build dir of this recipe.

TimoWilken:
Merged since this is a harmless change and it only affects the O2DPG CI. Cheers @benedikt-voelkel!

PR #4833, opened 2023-02-17
Keep logs after build

benedikt-voelkel:


benedikt-voelkel:
@TimoWilken this is hopefully the last adjustment for now

PR #4834, opened 2023-02-17
ROCm: add at 5.3

ktf:


davidrohr:
So first, this is not ROCm but this is the amd app sdk which we use for the OpenCL 1 backend.
Second, this lacks the renaming of the libamdocl64.so file, which will otherwise clash with a file of the same name that comes from ROCm.
Third, I am not sure I like overriding the OCL ICD paths. Doesn't that mean we break the OpenCL installations that come from the system?

davidrohr:
I think if we do this, we should set only `OCL_ICD_FILENAMES`, which adds additional ICD files, but doesn't break the OpenCL implementations from the system.
But still, this will not help with the fact that ROCm wants to be installed to /opt. I agree we can relocate the APP SDK.

ktf:
Ah, ok, apologies. I missed the fact there were RPMs installed as well, I just stopped parsing the provision.sh at CUDA.

I can do  the renaming.

For the actual rocm, can't we use cpio to unpack the RPMs in our preferred location? Alternatively, don't they provide a similar self contained script for the installation?

davidrohr:
They have an installer, e.g. https://repo.radeon.com/amdgpu-install/22.10/rhel/8.4/amdgpu-install-22.10.50100-1.el8.noarch.rpm
But that is an RPM which you install, and then you can run it and it will download and install the required RPMs for the ROCm roles you want to use.

Again, I am not sure if the destination directory can be changed. I think it is hardcoded in several places.
E.g., I can install ROCm 5.1 and 5.3 on the same server, goes to /opt/rocm-[version], and then I can compile software with both of them, and when running them they'll find the right libraries automatically. Didn't check in detail, but could be rpath or whatever.

PR #4839, opened 2023-02-21
Macro name change for clarity, keep workerlog

benedikt-voelkel:


benedikt-voelkel:
@sawenzel I will wait for this being merged before updating https://github.com/AliceO2Group/O2DPG/pull/875.

@TimoWilken This is another trivial change, I would say

PR #4844, opened 2023-02-22
modulefile for libunwind

sawenzel:
This is needed so that

`alienv enter IgProf`

correctly works (loading also the necessary libunwind dependency).


TimoWilken (approved):

ktf:
Thanks!

PR #4849, opened 2023-02-23
Add workflow creation

benedikt-voelkel:


benedikt-voelkel:
Hello @TimoWilken 
Since everything works fine for a single test run, I added another quick test for workflow creation.
When we have a better solution for the `diff`, I will adjust everything in the O2DPG tests themselves.
Would that be okay?

This additional test only adds a few seconds to the current testing.

benedikt-voelkel:
Thanks a lot and absolutely no worries concerning the diff. In principle everything works as expected.
You are helping quite a lot with this thing!

PR #4860, opened 2023-02-28
Fix linter errors in higher-level packages

TimoWilken:
Fix the remaining linter errors in packages that won't cause a rebuild of the entire software stack.

Also fix linter warnings in the same packages, as the package hashes will change anyway, and this way the warnings won't suddenly crop up the next time anyone sends a PR including these packages.

Also delete parquet-cpp recipe; see 0bacb346878420e9f8d6a4a8e57c640f60a71efa for details.

Draft to avoid too many rebuilds in CI.

Needed for #4730.


Barthelemy:
fine with me

ktf (dismissed):

vascobarroso (dismissed):

TimoWilken:
Thanks for the reviews!

I'll merge when the CI succeeds.

vascobarroso (approved):

PR #4869, opened 2023-03-02
Drop asiofi / asio from FairMQ

ktf:
They are now deprecated and will be removed in 1.5.0.

### DIFF
--- a/fairmq.sh
+++ b/fairmq.sh
@@ -6,8 +6,6 @@ requires:
   - boost
   - FairLogger
   - ZeroMQ
-  - asiofi
-  - asio
 build_requires:
   - flatbuffers
   - CMake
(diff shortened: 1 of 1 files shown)


TimoWilken:
Is it worth removing the asiofi and asio recipes from alidist entirely, since they won't be tested by CI after this change and will presumably just become stale? They don't seem to be used by anything other than FairMQ.

davidrohr:
I think so, and also libfabric should be removed.
I asked 2 weeks ago (on 16th), and said we'd remove it in 2 weeks since noone complains. So I think we cann remove the 3 now.

dennisklein:
> Is it worth removing the asiofi and asio recipes from alidist entirely, since they won't be tested by CI after this change and will presumably just become stale? They don't seem to be used by anything other than FairMQ.

From my point of view, they can go away, perhaps also `ofi.sh`. Originally, they were all introduced by me, but I have not tracked if anybody else started to depend on them.

ktf:
No objections to dropping the recipes completely, but let's do that in a separate PR, since this one is already undergoing testing....

PR #4870, opened 2023-03-02
Remove packages orphaned by FairMQ change

TimoWilken:
This removes asiofi, asio and ofi since they were only used by FairMQ prior to #4869.

This PR also cleans up some leftovers from #4869, including one now-unused variable. @ktf, There was a comment in the FairMQ recipe warning not to disable the `-DBUILD_EXAMPLES`, but since you changed that, can the comment go too?

In addition, port the change to FairMQ from #4860 to batch the O2 rebuild with the one from this PR and/or #4869.


dennisklein:
> There was a comment in the FairMQ recipe warning not to disable the -DBUILD_EXAMPLES, but since you changed that, can the comment go too?

Just a headsup to avoid double efforts: I will make a PR within the next days with a more general cleanup of the `fairmq.sh` recipe.

TimoWilken:
Hi @dennisklein, great, thanks!

If you're already changing the FairMQ recipe, perhaps you'll also find https://github.com/TimoWilken/alidistlint useful -- we're in the process of introducing this as a code checker for alidist recipes. It finds common issues in shell scripts and alidist recipe metadata.

dennisklein:
@TimoWilken @ktf Maybe a quick style question:

A CMake configure option of this form `${ZEROMQ_ROOT:+-DZeroMQ_ROOT=$ZEROMQ_ROOT}`, do you prefer to have it explicitely written out in the cmake configure command line? alibuild started out by using capitalized `$<PACKAGE>_ROOT` env variables while CMake then later introduced support for case-sensitive `<Package>_ROOT` env and CMake variables. In principle, one could drop explicit CMake config lines in the form above since FairMQ supports the capitalized env variables set by alibuild - I would appreciate the reduced boilerplate, but not sure about your latest (de facto) policies here?

TimoWilken:
@dennisklein, I'm not sure that "our" CMake actually uses the `$<PACKAGE>_ROOT` env vars by default -- at least the way I understand `cmake --help-policy CMP0074` is that the env vars are ignored by default, and we don't configure the policy otherwise as far as I know.

TimoWilken:
Ah, fair enough! In that case, I don't mind either way. I don't know of an ALICE policy on this, I don't think it's come up before...

TimoWilken:
Rebased and removed cleanups obsoleted by #4882. Now, this PR just removes the unused asio, asiofi and ofi recipes.

PR #4873, opened 2023-03-03
Fix FairMQ Modulefile

ktf:
Extra \ breaks module.


ktf:
Not currently working and apparently not tested. Merging.

PR #4880, opened 2023-03-06
Fix typo in readout recipe

martinboulais:
Fix `bookkeeping` typo in readout recipe


sy-c (approved):
the change is ok for me, although I don't understand how it could have worked so far on the dev system.

sy-c (approved):

awegrzyn:
IMO it works because CMake cache stores path to CONFIG under `..._DIR`

PR #4843, opened 2023-02-22
Check new FairMQ version

ktf:

PR #4881, opened 2023-03-06
Check FairMQ `v1.5.0-rc.1`

dennisklein:
2nd round before release with modernized recipe (1st round in #4843)


dennisklein:
FairMQ build seems to succeed, but `-Werror` renders the checks kinda useless for the purpose of testing the rc :(

PR #4882, opened 2023-03-07
Modernize FairMQ recipe

dennisklein:
@TimoWilken https://github.com/TimoWilken/alidistlint worked like a charm for me (on Fedora 36) thx!

PR #4883, opened 2023-03-08
Add explicit flatbuffers dependency to ODC

TimoWilken:
#4882 removed flatbuffers as a FairMQ dependency, but it is apparently needed for ODC. Not sure why the CI on #4882 didn't catch this.


davidrohr:
currently broken + trivial, merging

PR #4885, opened 2023-03-08
daily tags: auto-update dist definitions

alibuild:
The daily-tags.sh script has created a tag release with these changes.


vascobarroso:
@alisw/pdp-experts can this be merged ? The latest QC needs a commit in O2 that is in this tag (or newer).

PR #4887, opened 2023-03-08
Bump qualitycontrol to v1.93.0

Barthelemy:


vascobarroso (approved):

awegrzyn:
...and second questions is which O2 does it use? The version set in recipe?

TimoWilken:
Hi @awegrzyn, yes, it compiles QC; you can search for `## Begin sw/BUILD/QualityControl` in the full text log to see it.

The check uses the `dev` branch of O2.

awegrzyn:
Okay, thanks, would it be possible to move this check to use actual version set in the recipe?

davidrohr:
I don't fully understand what you mean? You want to test against an older o2 version but not dev?
I think at some point we said we only support o2/dev and not the previous tags. Thus it could be that older o2 veraions do not compile with current alidist etc.. Hence i think we should not test against old O2 but always against o2/dev.

davidrohr:
What version of the recipe do you mean? The o2 version defined in o2.sh? How is that bumped? Daily to the latest tag?

But it principle, as this is an alidist check, it would not only affect qc and dd, but it would imply that all package changes are tested against o2/de in tze normal ci and against the tagged o2 in the dataflow ci. Do we want that?

awegrzyn:
Yes, I referred to version set in `o2.sh`. I don't know how it is bumped but it is the one we use for FLP Suite (not `dev`), therefore I'd strongly push to move dataflow check to use that version.

davidrohr:
And for when is that relevant? For the time when creating the flpsuite branche? Or for the whole lifetime of an flpsuite version?

I think in any case we should not block alidist updates for colliding with older o2 versions of the current flp suite, since flp suite lifetime is usually larger than a week.

awegrzyn:
Nah, I think we're not at the same line.
Image that there are changes in O2@dev that new QC needs. You bump QC in alidist (but not O2), build passes because it uses O2@dev.
Then, any user that wants to compile QC get an error, because `aliBuild` uses what's in `o2.sh` not `dev`...

davidrohr:
But I think that is exactly the expected behavior, we support only o2/dev.
For the nightly builds, the current state of alidist/master and o2/dev get tagged together, and they are expected to match.

Users building O2 themselves are anyway expected tp build as a development package using O2/dev.

If for some reason you want to build an o2 tag, not dev, then you should not use alidist/master but the corresponding tag in alidist (that is why we tag it together such that it matches).

@ktf : correct me if I said something wrong.

davidrohr:
O2 is bumped in alidist to have a somewhat up to date version, and to have the right version in the alidist tag. In my opinion, we could also have the dev branch in there by default, and the tagged version only in the alidist tag of the same name.

Checking the tagging history, o2.sh is bumped only at successful nightly builds. There are sometimes longer gaps when we had jenkins issues.
If we test alidist PRs against the tagged O2, whenever we want to bump something in alidist that needs an O2 change, one would need to wait for that O2 change to become tagged. I am not sure this is what we want.
But we can rediscuss in WP3 then.

PR #4892, opened 2023-03-09
Bookkeeping api requires protobuf

davidrohr:
doesn't compile without for me, no idea why it works in the CI.
But the bookkeeping API Cmake clearly needs protobuf


TimoWilken:
Hi David, in CI it is pulled in transitively via grpc. Maybe you have that package disabled locally? Anyway, with your fix here, it makes more sense. Thanks!

PR #4895, opened 2023-03-10
GenTopo fix typo

martenole:
@davidrohr is it OK to change here, or should I rather change the filename in O2DPG?


davidrohr:
thx, fine for me this way

PR #4899, opened 2023-03-13
Bump bookkeeping version and fix grpc+openssl

martinboulais:


Barthelemy (approved):

davidrohr:
@Barthelemy : The O2/alidist error can be ignored I think, but what about the maxOS-arm error?

Barthelemy:
@davidrohr good point. Let me check and fix it.

Barthelemy:
@davidrohr I have fixed the issue with the macos-arm check. I don't think that the errors about GPU are related.

PR #4906, opened 2023-03-14
Test creation of test analyses workflow

benedikt-voelkel:
* test if any changes break the principal working of the test analyses


benedikt-voelkel:
This additional test takes less than a second.
Related O2DPG PR is https://github.com/AliceO2Group/O2DPG/pull/944

PR #4908, opened 2023-03-15
Drop unneeded copyright check for O2

ktf:
We have a separate rule in Github for that.


ktf:
@TimoWilken is this what we need?

TimoWilken (approved):

ktf:
@TimoWilken Can you disable any linting which is not fully fixed already? We should never enable cosmetic checks for things which are not already fully under control.

ktf:
Currently broken. Merging.

PR #4913, opened 2023-03-16
Update geant4 to next minor patch version 11.0.4

sawenzel:


ihrivnac:
> @ihrivnac : Are you ok with doing this bump or do you see reasons to do a full release validation study?

As this is just patch release, I think, there is no need for full release validation study.

sawenzel:
@ktf @TimoWilken : Can this be merged please? There is a  test failure as documented here https://alice.its.cern.ch/jira/browse/O2-3653 which seems systematic and due to missing token or broken CCDB connection.

PR #4915, opened 2023-03-18
Add KFParticle to O2 dependencies

mpuccio:
Hi @pzhristov, @ktf,

This is the modification I mentioned on Thursday.
It does not create havoc, as we already build the KFParticle package for O2Physics.

Cheers,
Max

### DIFF
--- a/defaults-o2-epn.sh
+++ b/defaults-o2-epn.sh
@@ -10,6 +10,7 @@ env:
   GEANT4_BUILD_MULTITHREADED: 'ON'
 disable:
   - O2Physics
+  - KFParticle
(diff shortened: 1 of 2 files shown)


mpuccio:
Hi @ktf,

it is already disabled for dataflow, see https://github.com/alisw/alidist/blob/master/defaults-o2-dataflow.sh#L29 

Cheers,
Max

mpuccio:
Hi @ktf,
can we merge this or do you think I still need to implement something?
Cheers,
Max

davidrohr:
fine for me to merge it, won't create much overhead

PR #4916, opened 2023-03-20
Bump jalien-root to 0.7.0

costing:
Memory leaks fixed by Sandro


sawenzel:
I think this can be merged.

PR #4920, opened 2023-03-22
Bump Root to v6-28-02

pzhristov:


root.sh
```diff
@@ -1,7 +1,7 @@
 package: ROOT
 version: "%(tag_basename)s"
-tag: "v6-26-10-alice5"
-source: https://github.com/alisw/root.git
```
sawenzel, line 4: I think that we will loose a couple of commits in our fork which have not been backported to root. This should not be done.

sawenzel (changes requested):
I think that we will loose a couple of commits in our fork which have not been backported to root. This should not be done.

ktf (approved):

ktf:
Merging this to see if we spot other issues in production. We can have a alice1 tag with the missing treatment of ROOT_LDSYSPATH. Are we sure this is still relevant, though?

sawenzel:
> Merging this to see if we spot other issues in production. We can have a alice1 tag with the missing treatment of ROOT_LDSYSPATH. Are we sure this is still relevant, though?

Yes, I have no reason to believe otherwise.

PR #4924, opened 2023-03-24
Bump qualitycontrol to v1.95.1

Barthelemy:


sy-c (approved):

vascobarroso:
@alisw/alidist-maitainers can this be merged ?

awegrzyn:
We again end up in situation where `alidist@master` does not compile because QC is bumped but O2 not...

ktf:
Well, the set of tests which we run is decided by you (FLP), so if you want to prevent this from happening you should switch, e.g.:

https://github.com/alisw/ali-bot/blob/master/ci/repo-config/mesosci/cs8/o2-alidist-dataflow.env

to not use AliceO2 dev. This of course will have the problem that QC will then need to implement any changes in a backward compatible way (at least for a few days), but again, I think that's a good idea in any case.

davidrohr:
@ktf : Can we then bump O2 automatically to the latest nightly tag in alidist every day, so have to wait only till the next morning, not few days?

PR #4925, opened 2023-03-24
add missing `--libs` to bookkeeping-api recipe

Barthelemy:


Barthelemy:
@vascobarroso @martenole

PR #4941, opened 2023-04-02
[O2-3661] Switch off a warning treated as error

pzhristov:


sawenzel (dismissed):

sawenzel:
Can this be merged please? Many people seem to be stuck on the Mac due to this.

TimoWilken:
Check failures unrelated.

martenole:
Somehow with this change my manual builds on the EPNs with Alma linux fail because of an CMake error:
```
CMake Error at /home/marten/alice/sw/slc8_x86-64/Clang/v15.0.7-6/lib/cmake/llvm/LLVMExports.cmake:59 (set_target_properties):
  The link interface of target "LLVMSupport" contains:

    zstd::libzstd_shared

  but the target was not found.  Possible reasons include:

    * There is a typo in the target name.
    * A find_package call is missing for an IMPORTED target.
    * An ALIAS target is missing.
```
Full log is attached. Does someone have an idea what might be wrong?
[arrowLog.txt](https://github.com/alisw/alidist/files/11168567/arrowLog.txt)

sawenzel:
> Somehow with this change my manual builds on the EPNs with Alma linux fail because of an CMake error:
> 
> ```
> CMake Error at /home/marten/alice/sw/slc8_x86-64/Clang/v15.0.7-6/lib/cmake/llvm/LLVMExports.cmake:59 (set_target_properties):
>   The link interface of target "LLVMSupport" contains:
> 
>     zstd::libzstd_shared
> 
>   but the target was not found.  Possible reasons include:
> 
>     * There is a typo in the target name.
>     * A find_package call is missing for an IMPORTED target.
>     * An ALIAS target is missing.
> ```
> 
> Full log is attached. Does someone have an idea what might be wrong? [arrowLog.txt](https://github.com/alisw/alidist/files/11168567/arrowLog.txt)

We probably need to integrate Alma Linux into the official CI... (I guess we can revert this for now as MacOS builds are anyway broken with XCode 13.4)

pzhristov:
@sawenzel How is this change related top the issue on Alma Linux? I do not see any direct reference to the modified warning options.

sawenzel:
> @sawenzel How is this change related top the issue on Alma Linux? I do not see any direct reference to the modified warning options.

No idea. Better to ask Ole.

martenole:
I also don't know. But when I revert this commit here then the build works fine for me

martenole:
I have removed everything in `$HOME/alice/sw` except for MIRROR, so I was doing a clean build with everything up to date.
On my ubuntu PC I don't have any problems with this change.

TimoWilken:
libzstd should come from the host system, I don't think it's in alidist.

There are no precompiled CentOS/Alma 8 tarballs for aliBuild yet with this change. Perhaps before, you were using arrow tarballs compiled on a system with libzstd-devel installed, and you don't have this package on the EPNs, so the manual compilation fails?

TimoWilken:
> We probably need to integrate Alma Linux into the official CI...

This is already done. The fullCI runs under Alma 8, and has been for a couple of weeks.

martenole:
Hi @TimoWilken 
thanks a lot, indeed libzstd-devel was not installed on the EPNs. Now it works. I'll ask for that package to be installed by default.
Cheers,
Ole

PR #4953, opened 2023-04-12
Add Python 3.11 clause to Python modules list

TimoWilken:
On MacOS Ventura, our CI machines use python@3.11 from brew.

Under Python 3.11, we can install psutil==5.9.4 as a wheel, but not 5.9.0, leading to a pip failure.

Nothing else is changed from `PIP310_REQUIREMENTS`.


TimoWilken:
Merging since it's currently broken on Mac and no CI Linux platform uses Python 3.11.

PR #4958, opened 2023-04-18
ACTS: build analysis apps, correct version

plariono:
This PR allows to build the analysis apps and corrects the version to "main" as the AliceO2Group fork doesn't have any version.
Tagging @mconcas, @sawenzel for approval.

### DIFF
--- a/acts.sh
+++ b/acts.sh
@@ -1,5 +1,5 @@
 package: ACTS
-version: "main"
+version: "v23.5.0-alice"
 requires:
   - ROOT
   - pythia
--- a/acts.sh
+++ b/acts.sh
@@ -1,5 +1,5 @@
 package: ACTS
-version: v23.4.0
+version: "main"
 requires:
   - ROOT
   - pythia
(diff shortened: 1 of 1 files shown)


mconcas (dismissed):

acts.sh
```diff
@@ -1,5 +1,5 @@
 package: ACTS
-version: v23.4.0
+version: "main"
```
sawenzel, line 2: main is a "moving" version and it is easy to break the CI when you commit to ACTS main. In general, it would be preferred to use fixed tags (even if they are in your development branch). @plariono : Would you be able to tag your own development in ACTS (with some ALICE specific tag)?

plariono: Hi @sawenzel, here the "main" branch refers to the AliceO2Group fork, which we use as a temporary solution. 
When the necessary PR gets merged in the main ACTS repository, we'll come back to a fixed version. 
> @plariono : Would you be able to tag your own development in ACTS (with some ALICE specific tag)?
I'm not sure. Where exactly such tag should be?

mconcas: > main is a "moving" version and it is easy to break the CI when you commit to ACTS main. In general, it would be preferred to use fixed tags (even if they are in your development branch). @plariono : Would you be able to tag your own development in ACTS (with some ALICE specific tag)?

Isn't "main" a moving version only if we update the repository? Just to be sure there is not some automatic synchronisation mechanism I am not aware of.

Concerning the tagging, this we can surely do on AliceO2Group/ACTS side.

sawenzel: yes sure. but update of the repo might happen without testing. I am not strictly against using "main" ... just wanted to raise the issue.

plariono: > We would create a tag, by releasing the ACTS code, in the fork of ACTS that you maintain

I see, so it will be a tag in our fork of ACTS, indeed that would be nice.

plariono:
@mconcas @sawenzel can this get approved please? thank you!

PR #4964, opened 2023-04-25
alien-cas.sh :: always use latest version of alien-cas

adriansev:
@ktf @TimoWilken until we have other solution, let's make sure we _always_ use the latest version of our CAs repository

### DIFF
--- a/alien-cas.sh
+++ b/alien-cas.sh
@@ -1,6 +1,6 @@
 package: AliEn-CAs
 version: v1
-tag: f62625ede780d455b3b7878064bcfee6bd9a4f53
+tag: master
 source: https://github.com/alisw/alien-cas.git
 build_requires:
   - alibuild-recipe-tools


TimoWilken (approved):
Looks good to me, I'll merge when it's cached.

ktf:
-1 no, we should not have master as a tag, because it will always check upstream the version.

adriansev:
@ktf 
> -1 no, we should not have master as a tag, because it will always check upstream the version.

well, that's the point.. why wouldn't we want to always check the alien-cas latest content?

ktf:
Because it means that every build will have to go to the server, slowing down things and potentially breaking if disconnected / the remote has a glitch. Given this thing changes once every 2 / 3 years, I see no point in creating potential issues for no real benefit.

adriansev:
@ktf the remote is github so, is there is a glitch anything else have a problem. Moreover as you saw, the changes can be ignored with the cost of biting us when is too late.. also, i'm not sure if you are aware, all the CAs must change their certificates to update the hash from sha1 to sha256 (as default crypto policies on el9 no longer accepts sha1), so in the future there will be changes, so our usage of alien-cas must use the latest version (until we settle on a solution to have the certificate management out of the packages). but, well, you know best the side effects, let me know if i should close this. Thanks!

ktf:
> @ktf the remote is github so, is there is a glitch anything else have a problem.

Not true. As long as you have everything compiled locally, which most of the people do when developing locally, it will not have to go to the server. Moreover, we do not want to have people recompiling everything just because someone added a CA they do not care about to a package they did not request to update.

>  CAs must change their certificates to update the hash from sha1 to sha256 (as default crypto policies on el9 no longer accepts sha1), so in the future there will be changes, so our usage of alien-cas must use the latest version (until we settle on a solution to have the certificate management out of the packages). but, well, you know best the side effects, let me know if i should close this. 

Yes, I think we should close this. We update tags daily without any problems, I do not see why we should do an exception for a trivial package which gets updated once every few months at best.

That said, I think even the in package behaviour is actually fine. What we need to fix is:

* In case a user certificate can be authenticated without any extra steps, we do not set X509, because we assume that the sys admin is taking care of the CAs.
* In case /cvmfs/ is available, the module file should point to a common installation, possibly doing the same check as above.
* Failing that, we do things as now, since it means that the user is taking care of the builds locally.

We should also make sure that only toplevel packages depend on alien-cas, because it does not make sense that we currently rebuild the whole software stack if a CA changes.

PR #4966, opened 2023-04-25
Prefer using system or CVMFS CA certs

TimoWilken:
Try to detect when we can use system certificates or those from CVMFS, and prefer them if possible. This should avoid situations where old O2 tags have expired certs pinned.

Still fallback to the old behaviour, which is fine for CI and local development.

Cc: @ktf @adriansev

### DIFF
--- a/alien-cas.sh
+++ b/alien-cas.sh
@@ -2,9 +2,17 @@ package: AliEn-CAs
 version: v1
 tag: f62625ede780d455b3b7878064bcfee6bd9a4f53
 source: https://github.com/alisw/alien-cas.git
+prefer_system: .*
+prefer_system_check: |
+  # If we can verify the user's cert, the CERN CA must be installed
+  # system-wide. In that case, don't install our certs.
+  openssl verify ~/.globus/usercert.pem > /dev/null 2> /dev/null
 build_requires:
   - alibuild-recipe-tools
 env:
+  # This is only used for O2/O2Physics unit tests. These are run
+  # locally (in which case these CAs are fine and assumed to be
+  # up-to-date) or in CI (ditto).
   X509_CERT_DIR: "$ALIEN_CAS_ROOT/globus/share/certificates"
 ---
 #!/bin/bash -e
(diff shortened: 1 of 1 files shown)


TimoWilken:
One weirdness I've run into is that old OpenSSL versions (1.0.2k, on CentOS 7) don't exit with an error status when the CA cert is expired, they only print an error message. This might have to be handled specially...

alien-cas.sh
```diff
@@ -18,5 +26,19 @@ MODULEFILE="$MODULEDIR/$PKGNAME"
...
+try {
+  exec openssl verify \$::env(HOME)/.globus/usercert.pem > /dev/null 2> /dev/null
+} trap CHILDSTATUS {} {
+  # If we couldn't validate the user's cert, either it doesn't exist
+  # or the CERN CA isn't installed as a system cert. In the latter
+  # case, we should redirect to CVMFS or to our certs.
+  if [file isdirectory /cvmfs/alice.cern.ch/etc/grid-security/certificates] {
+    # Prefer current certs from CVMFS, if available. This avoids
+    # failures due to expired certs.
+    setenv X509_CERT_DIR /cvmfs/alice.cern.ch/etc/grid-security/certificates
```
ktf, line 38: Rather than an hardcoded single place, I would use the latest version of this package, so that we can still update the grid by simply bumping something in alidist and no "out of band" procedures are involved.

ktf: I.e. something like `ls \$::env(BASEDIR)/AliEn-CAs/ | tail -1`...

adriansev:
Hi @TimoWilken Would be possible to take into account and honor X509_USER_CERT, X509_USER_KEY, X509_CERT_FILE, X509_CERT_DIR? Maybe, even check if X509_CERT_DIR, X509_CERT_FILE are set and if so, keep them?
Also, AFAIK, on macos cvmfs is mounted under /Users/Shared, but i'm not sure if macos users are a targer for this..
Also, for the moment the actual package that export X509_CERT_DIR is alien-runtime

adriansev:
also, unrelated directly to this but could be worth mentioning, we do a lot of find rsync stuff .. wouldn't be easier to do this directly in alien-cas repository and then just point X509_CERT_DIR to that location? (and also as Giulio already suggested, by adding tags to alien-cas with the same name/ver as IGTF would be easier to manage versioning)

PR #4967, opened 2023-04-25
Delete obsolete defaults definitions

TimoWilken:
These defaults are old, unsupported and shouldn't be used any more -- "o2" is the correct defaults file now.

This PR doesn't change anything for old AliPhysics tag series, as they are taken from `AliPhysics-v5-09-*-01-patches` branches.

@ktf, is this OK?

I'm not sure about the "coverage", "jalien" and "generators" defaults -- are these still supported?


TimoWilken:
Changed as you suggested @ktf.

I'm still not sure about the "coverage", "jalien" and "generators" defaults -- should any of those also be deleted?

ktf:
- defaults-generators.sh I think we need. Pinging @jackal1-66 and @maireiphc  
- jalien and coverage we probably do not need.

jackal1-66:
> * defaults-generators.sh I think we need. Pinging @jackal1-66 and @maireiphc
> * jalien and coverage we probably do not need.

defaults-generators.sh is still used in some instances when building AliGenerators, so it would be better to keep it.

PR #4970, opened 2023-04-26
Remove setuptools pinning in XRootD

TimoWilken:
This pinning was added in fa73bfc806f3e4012b68b2c1e018a5ae79f05f59 due to an issue when setting up the M1 Macs, but it works with the latest setuptools on the new M2 Macs.

Cc: @pzhristov @adriansev @ktf

### DIFF
--- a/python-modules.sh
+++ b/python-modules.sh
@@ -60,10 +60,8 @@ python3 -m venv $PYTHON_MODULES_INSTALLROOT
 . $PYTHON_MODULES_INSTALLROOT/bin/activate
 
 # Upgrade pip
-python3 -m pip install -IU pip
 # Install setuptools upfront, since this seems to create issues now...
-python3 -m pip install -IU "setuptools<=60.8.2"
-python3 -m pip install -IU wheel
+python3 -m pip install -IU pip setuptools wheel
 
 # FIXME: required because of the newly introduced dependency on scikit-garden requires
 # a numpy to be installed separately
--- a/xrootd.sh
+++ b/xrootd.sh
@@ -43,10 +43,6 @@ case $ARCHITECTURE in
     # This fix is needed only on MacOS when building XRootD Python bindings.
     export CFLAGS="${CFLAGS} -isysroot $(xcrun --show-sdk-path)"
     unset UUID_ROOT
-    if [ "$(python3 -c 'import setuptools; print(setuptools.__version__)')" != "60.8.2" ]; then
-      echo 'Please install setuptools==60.8.2'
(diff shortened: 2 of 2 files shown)


adriansev:
@TimoWilken i see some setuptools pinning also in python-modules-list.sh, so those pinnings should be also removed..

TimoWilken:
@adriansev in python-modules-list, we pin the versions of *all* packages to a known-working set on each platform/python version, so that's intentional as far as I know. I'm not sure we want to always use the latest setuptools automatically there.

The pinning I'm trying to get rid of here is independent of that, and is causing the occasional user confusion on MacOS-arm64, since we don't use python-modules/python-modules-list there afaik.

adriansev:
> @adriansev in python-modules-list, we pin the versions of _all_ packages to a known-working set on each platform/python version, so that's intentional as far as I know. I'm not sure we want to always use the latest setuptools automatically there.
> 
> The pinning I'm trying to get rid of here is independent of that, and is causing the occasional user confusion on MacOS-arm64, since we don't use python-modules/python-modules-list there afaik.

@TimoWilken well, it's the same venv so anything that comes after that will update anyway what is installed .. 
also python-modules are used on any python related requirement, and python-modules use Python-modules-list, so, any python related package will use Python-modules-list
so, as example, if you look in
/cvmfs/alice.cern.ch/el7-x86_64/Packages/Python-modules/1.0-368/share/python-modules/lib/python/site-packages  (path from O2Physics::nightly-20230425-1)
you will see that certifi is at latest version even if pinned in python.sh
moreover you can observer the multiple versions of setuptools installed (58.1.0, 60.8.2 and 65.5.1)
so, IMHO the pinning from python-modules-list is useless as the un-pinned installation of packages will update all the requirements anyway (and the various packages will have different minimum requirements)... python is not made with version freezing in mind...

adriansev (approved):

adriansev:
and macos-latest runner is described here:
https://github.com/actions/runner-images#available-images

adriansev:
@TimoWilken i just observed an pip installation problem with xrootd in my workflow, and it was fixed only after i pinned pip and setuptools: https://github.com/adriansev/jalien_py/blob/master/.github/workflows/codeql-analysis.yml#L72
so, it seems that python install infrastructure of xrootd is outdated w.r.t. latest pip/setuptools so we should pin both pip and setuptools both in python.sh and python-modules-list

TimoWilken:
This change is also present in #4972; closing in favour of that PR.

PR #4972, opened 2023-04-27
python:: unique pip,setuptools pinned version:: tryout fixing xrootd problems

adriansev:
Hi @TimoWilken @ktf This is a tryout to use a common basic pinned requirements
over all platforms/arch, with the purpose to make xrootd python bindings work without
special customization that would/could change the common to all framework venv.
Please take a look, and let's see if it works :)

PR #4974, opened 2023-04-27
Integrate Conda env into Python-modules

TimoWilken:
Conda is required for GPU hardware support on ARM (M1/M2) Macs, in order to use Tensorflow.

Previously, this was expected to be installed manually, as documented at https://alice.its.cern.ch/jira/browse/O2-2072, but that way packages are not automatically kept compatible with alidist packages.

This PR installs a Conda environment as the Python-modules package instead. This integration also means all packages that previously disabled Python-modules on osx_arm64 can now depend on the package normally.

Tested on an M2 Mac; I can run the tensorflow test script listed at https://developer.apple.com/metal/tensorflow-plugin/ successfully in the resulting `alienv enter Python-modules/latest` environment.

@pzhristov, if you could test this with some O2-related script that uses the GPU, that would be great!

### DIFF
--- a/python-modules-list.sh
+++ b/python-modules-list.sh
@@ -76,6 +76,26 @@ env:
     dryable==1.0.5
     responses==0.10.6
     pandas==1.1.5
+  # Keep the PIPxy version in sync with the Conda env we install in Python-modules!
+  # Everything but the first two lines copied from PIP39_REQUIREMENTS, but with versions
+  # adjusted such that wheels are available and for compatibility with tensorflow.
+  PIP39_REQUIREMENTS_osx_arm64: |
+    tensorflow-macos==2.12.0
+    tensorflow-metal==0.8.0
+    PyYAML==5.4.1
+    psutil==5.9.5
+    uproot==4.1.0
+    numpy==1.23.5
+    scipy==1.10.1
+    Cython==0.29.21
+    seaborn==0.11.0
+    scikit-learn==1.2.2
+    sklearn-evaluation==0.12.0
+    Keras==2.12.0
+    xgboost==1.7.5
+    dryable==1.0.5
+    responses==0.10.6
+    pandas==1.5.3
   PIP310_REQUIREMENTS: |
     PyYAML==5.4
     psutil==5.9.0
@@ -106,20 +126,9 @@ env:
     dryable==1.0.5
     responses==0.10.6
     pandas==1.1.5
+build_requires:
+  - alibuild-recipe-tools
 ---
-# Modulefile
-MODULEDIR="$INSTALLROOT/etc/modulefiles"
-MODULEFILE="$MODULEDIR/$PKGNAME"
-mkdir -p "$MODULEDIR"
-cat > "$MODULEFILE" <<EoF
-#%Module1.0
-proc ModulesHelp { } {
-  global version
-  puts stderr "ALICE Modulefile for $PKGNAME $PKGVERSION-@@PKGREVISION@$PKGHASH@@"
-}
-set version $PKGVERSION-@@PKGREVISION@$PKGHASH@@
-module-whatis "ALICE Modulefile for $PKGNAME $PKGVERSION-@@PKGREVISION@$PKGHASH@@"
-# Dependencies
-module load BASE/1.0
-# Our environment
-EoF
+#!/bin/bash -e
+mkdir -p "$INSTALLROOT/etc/modulefiles"
+alibuild-generate-module > "$INSTALLROOT/etc/modulefiles/$PKGNAME"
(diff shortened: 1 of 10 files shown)


TimoWilken:
Sorry, forgot to mention: I also changed the XRootD recipe as in https://github.com/alisw/alidist/pull/4970 (but I didn't submit this change here to avoid conflicts with the other PR). If you want to try this, please delete the check for `setuptools==60.8.2` from xrootd.sh.

python-modules-list.sh
```diff
@@ -76,6 +76,27 @@ env:
...
+    Cython==0.29.21
+    seaborn==0.11.0
+    scikit-learn==1.2.2
+    sklearn-evaluation==0.12.0
+    Keras==2.12.0
+    xgboost==1.7.5
+    dryable==1.0.5
+    responses==0.10.6
+    pandas==1.5.3
+    setuptools==65.5.1
```
adriansev, line 99: Hi @TimoWilken given that setuptools is installed as common dependency in #4972 maybe here is redundant?

TimoWilken: Good point @adriansev. I'll rebase this PR on top of yours once yours is merged and fix stuff like this!

pzhristov:
@TimoWilken I confirm that the Metal GPU is detected and used by tensorflow. Merging...

pzhristov:
Just one additional comment: the python installation on my M1 needed setuptools, wheel, six, pyyaml and requests to start successfully the build.

TimoWilken:
> @TimoWilken I confirm that the Metal GPU is detected and used by tensorflow. Merging...

Thanks for checking @pzhristov!

> Just one additional comment: the python installation on my M1 needed setuptools, wheel, six, pyyaml and requests to start successfully the build.

Did you install alibuild through pip or did you run it straight from the git folder? At least pyyaml and requests are required by alibuild (to parse alidist recipes and to download tarballs, respectively). If you install alibuild using pip, those dependencies should have been installed automatically...

TimoWilken:
If you haven't merged yet, could you wait until after chep? Just in case it causes problems on one of the CI Macs.

pzhristov:
Sorry, your comment came just after I merged it :( Yes, I am using everything from github, no pip or brew for the dependencies.

TimoWilken:
Alright, no problem at all!

mfasDa:
@TimoWilken @adriansev  After updating alidist and removing conda which I installed manually accroding to https://alice.its.cern.ch/jira/browse/O2-2072 - which seems to be in conflict with the conda installed now by aliBuild - I was able to build again, but when doing alien.py commands I get the following warning:
```
XRootD module could not be imported! Not fatal, but XRootD transfers will not work (or any kind of file access)
 Make sure you can do:
python3 -c 'from XRootD import client as xrd_client'
```
Looking at the python bindings it looks the problem is that the bindings are built for python 3.11, which comes now from brew, while at runtime we load python 3.9.16 from conda. This probably needs a fix in the XRootD build recipe forcing the python bindings to be built with the python from conda as well.

adriansev:
maybe `. "$PYTHON_MODULES_INSTALLROOT/bin/activate"` line should be common to all python installing recipes? i'm not sure if PYTHON_MODULES_INSTALLROOT is defined externally, but anyway if we set a common place to install python stuff then all python recipes should using uniformly i would think .. but only @TimoWilken know the overall view and what could/should be done or not ..

mfasDa:
@TimoWilken I have tried adding the lines to the python-modules.sh build recipe and with this the build of xrootd crashes with a well-known error reported recently quite frequently:
```
2023-06-12@11:33:48:DEBUG:QualityControl:XRootD:aliceo2:   ld: library not found for -ldl
2023-06-12@11:33:48:DEBUG:QualityControl:XRootD:aliceo2:   clang: error: linker command failed with exit code 1 (use -v to see invocation)
2023-06-12@11:33:48:DEBUG:QualityControl:XRootD:aliceo2:   error: command '/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/c++' failed with exit code 1
2023-06-12@11:33:48:DEBUG:QualityControl:XRootD:aliceo2:   error: subprocess-exited-with-error
```

TimoWilken:
Hi @mfasDa, this error usually indicates that:

1. aliBuild built some packages against python@3.X from brew
2. brew updated its python and/or installed a new python@3.Y
3. now aliBuild is picking up python@3.Y and trying to build stuff against it, combined with stuff built against the older python@3.X

This is a recurring problem and we're working on a proper solution to this. Unfortunately, for now, you'd have to delete anything depending on python (even indirectly) from your `sw` directory, or start a fresh `sw` directory. Could you try that?

TimoWilken:
Hi @mfasDa, sorry, if that didn't work, could you try https://github.com/alisw/alidist/pull/5035?

PR #4980, opened 2023-05-02
Cefp stable 2022 v3

mpuccio:
@TimoWilken would it be possible to create a new CEFP2022v3 tag starting from this?


TimoWilken:
Hi @mpuccio, I'm building it now!

PR #4985, opened 2023-05-04
Drop ApMon-CPP

ktf:


ktf:
@costing do we still use this anywhere?

ktf:
@costing I was trying with the new ROOT and apparently ApMon.h confuses things because of the various `_POSIX_*` variables  defined at the top.

pzhristov (approved):

PR #4987, opened 2023-05-11
Revert "Disable ApMon-CPP in O2 (#4985)"

TimoWilken:
This reverts commit 1f4e3e39f4bb23d9e23188b152f29a18221705b6.

Hopefully, this will fix the problems reported in O2-3760 for now.

### DIFF
--- a/defaults-o2.sh
+++ b/defaults-o2.sh
@@ -11,7 +11,6 @@ env:
 disable:
   - mesos
   - MySQL
-  - ApMon-CPP
 overrides:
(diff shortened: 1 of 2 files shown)


pzhristov:
@TimoWilken I would be surprised if this helps, but we will see :)

TimoWilken:
@pzhristov I'm not convinced this is the problem either, but it's the only idea I've got for now... Of course, we can revert this reversion if the problem persists!

PR #4990, opened 2023-05-15
Add test for RelVal

benedikt-voelkel:
test itself only runs if files in RelVal/ changed

### DIFF
--- a/o2dpg-sim-tests.sh
+++ b/o2dpg-sim-tests.sh
@@ -13,27 +13,26 @@ fi
 TEST_DIR=$BUILDDIR/o2dpg-sim_tests
 LOGFILE="${TEST_DIR}/o2dpg-sim-tests.log"
 
+# remove everything for a fresh start
 rm -Rf ${TEST_DIR}
 mkdir ${TEST_DIR}
 pushd ${TEST_DIR}
 
 O2DPG_TEST_GENERATOR_EXITCODE=0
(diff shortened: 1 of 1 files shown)


o2dpg-sim-tests.sh
```diff
@@ -31,9 +31,15 @@ O2DPG_TEST_ANALYSISQC_EXITCODE=0
 { O2DPG_TEST_REPO_DIR=${SOURCEDIR}/../../../O2DPG/${O2DPG_VERSION}/0 "${O2DPG_ROOT}/test/run_analysisqc_tests.sh" >> ${LOGFILE} 2>&1 ; O2DPG_TEST_ANALYSISQC_EXITCODE=$?; } || true  # don't quit immediately on error
 # only log/test files created, nothing to delete here
 
+O2DPG_TEST_RELVAL_EXITCODE=0
+{ O2DPG_TEST_REPO_DIR=${SOURCEDIR}/../../../O2DPG/${O2DPG_VERSION}/0 "${O2DPG_ROOT}/test/run_relval_tests.sh" >> ${LOGFILE} 2>&1 ; O2DPG_TEST_RELVAL_EXITCODE=$?; } || true  # don't quit immediately on error
+# keep only logs, remove everything else for now
+# if we needed/possible in the future, we might keep some of the other files
+find . -type f ! -name '*.log' -delete
```
ktf, line 38: If you are running in BUILD, I think there is no need to remove the logs, given they do not get packaged in any case. Just make sure you delete them at the beginning, so that we start from a clean slate.

PR #4999, opened 2023-05-21
Added Rivet in AliPhysics

jackal1-66:
I added Rivet as a requirement in AliPhysics and my tests were successful. Having it could be useful to run analysis on pre-existing hepmc files and is a prerequisite for running on LEGO trains (previous AliRivetTask developed by Christian Holm Christensen on generated AODs)


jackal1-66:
A different approach is now used for the RIVET compatibility issue: an exception catcher prevents environment from not opening in all operating systems.

PR #5008, opened 2023-05-26
Link Python to distro-independent libcrypt.so

TimoWilken:
We use el7-built Python binaries on el9 as well, but there, only libcrypt.so.2 is available (as opposed to libcrypt.so.1 on el7). However, loading .so.2 runtime still works, so patchelf the python executable to refer to the generic unversioned library.

@zensanp Is there some way you could test this? You can build the package as we do in the nightly builds using `aliBuild build -a slc7_x86-64 --docker-image registry.cern.ch/alisw/slc7-builder:patchelf Python` (the regular build container doesn't yet have patchelf).

### DIFF
--- a/python.sh
+++ b/python.sh
@@ -83,7 +83,7 @@ pushd "$INSTALLROOT/bin"
 popd
 
 if [ "$(uname)" = Linux ]; then
-  # We want to use el7-built Pythons on el7, e.g. on the Grid. On el7,
+  # We want to use el7-built Pythons on el9, e.g. on the Grid. On el7,
   # we compile against libcrypto.so.1, which is not available on el9
   # (it's libcrypt.so.2 instead). However, Python still works when
   # loaded with libcrypt.so.2, so just replace the library reference.
--- a/python.sh
+++ b/python.sh
@@ -82,6 +82,14 @@ pushd "$INSTALLROOT/bin"
   [[ -x python3-config ]] || ln -nfs "$PYTHON_CONFIG_BIN" python3-config
 popd
 
+if [ "$(uname)" = Linux ]; then
+  # We want to use el7-built Pythons on el7, e.g. on the Grid. On el7,
+  # we compile against libcrypto.so.1, which is not available on el9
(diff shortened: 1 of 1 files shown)


python.sh
```diff
@@ -82,6 +82,14 @@ pushd "$INSTALLROOT/bin"
   [[ -x python3-config ]] || ln -nfs "$PYTHON_CONFIG_BIN" python3-config
 popd
 
+if [ "$(uname)" = Linux ]; then
+  # We want to use el7-built Pythons on el9, e.g. on the Grid. On el7,
+  # we compile against libcrypto.so.1, which is not available on el9
+  # (it's libcrypt.so.2 instead). However, Python still works when
+  # loaded with libcrypt.so.2, so just replace the library reference.
+  patchelf --replace-needed libcrypt.so.1 libcrypt.so "$INSTALLROOT/bin/python"
```
ktf, line 90: I do  not  understand this. Shouldn't we simply be compiling against our own openssl?

TimoWilken:
It's libcrypt (from glibc), not libcrypto (from openssl) in this case. I'm not sure compiling our own glibc is a good idea, but we might be able to use [libxcrypt](https://github.com/besser82/libxcrypt) instead (which [apparently Gentoo use as well](https://wiki.gentoo.org/wiki/Project:Toolchain/libcrypt_implementation)). I don't know if this would cause any compatibility/interop problems with the host system (e.g. if we link against system libs that were linked against glibc's libcrypt), but [the libxcrypt compatibility notes](https://github.com/besser82/libxcrypt#compatibility-notes) sound promising:

> On Linux-based systems, by default libxcrypt will be binary backward compatible with the libcrypt.so.1 shipped as part of the GNU C Library. This means that all existing binary executables linked against glibc’s libcrypt should work unmodified with this library’s libcrypt.so.1.
>
> However, the converse is not true: programs linked against libxcrypt will not work with glibc’s libcrypt.  Also, programs that use certain legacy APIs supplied by glibc’s libcrypt (encrypt, encrypt_r, setkey, setkey_r, and fcrypt) cannot be compiled against libxcrypt.

PR #5024, opened 2023-06-09
Cleanup ROOT recipe

ktf:
* Drop support for ROOT5
* Add support for C++20
* Make sure the plugins / paths are set up correctly in the system.rootrc also in the case the incremental recipe is used.

### DIFF
--- a/root.sh
+++ b/root.sh
@@ -35,14 +35,14 @@ incremental_recipe: |
          "$INSTALLROOT/etc/plugins/TSystem/P030_TAlienSystem.C" \
          "$INSTALLROOT/etc/plugins/TFile/P070_TAlienFile.C"
   # Add support for ROOT_PLUGIN_PATH envvar for specifying additional plugin search paths
-  grep -v '^Unix.*.Root.PluginPath' $INSTALLROOT/etc/system.rootrc > system.rootrc.0
+  grep -v '^Unix.*.Root.PluginPath' "$INSTALLROOT/etc/system.rootrc" > system.rootrc.0
   cat >> system.rootrc.0 <<\EOF
   # Specify additional plugin search paths via the environment variable ROOT_PLUGIN_PATH.
   # Plugins in \$ROOT_PLUGIN_PATH have priority.
   Unix.*.Root.PluginPath: $(ROOT_PLUGIN_PATH):$(ROOTSYS)/etc/plugins:
   Unix.*.Root.DynamicPath: .:$(ROOT_DYN_PATH):
   EOF
-  mv system.rootrc.0 $INSTALLROOT/etc/system.rootrc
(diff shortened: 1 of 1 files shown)


root.sh
```diff
@@ -130,6 +135,7 @@ cmake $SOURCEDIR
       ${CXX11:+-Dcxx11=ON}                                                             \
       ${CXX14:+-Dcxx14=ON}                                                             \
       ${CXX17:+-Dcxx17=ON}                                                             \
+      ${CXX17:+-Dcxx20=ON}                                                             \
```
TimoWilken, line 138: Why keep the `CXXvv` variables? You removed the code that sets them.

If you keep them, shouldn't the last one be `${CXX20:+-Dcxx20=ON}` instead?

root.sh
```diff
@@ -34,26 +34,31 @@ incremental_recipe: |
...
          "$INSTALLROOT/etc/plugins/TFile/P070_TAlienFile.C"
+  # Add support for ROOT_PLUGIN_PATH envvar for specifying additional plugin search paths
+  grep -v '^Unix.*.Root.PluginPath' $INSTALLROOT/etc/system.rootrc > system.rootrc.0
+  cat >> system.rootrc.0 <<\EOF
+  # Specify additional plugin search paths via the environment variable ROOT_PLUGIN_PATH.
+  # Plugins in \$ROOT_PLUGIN_PATH have priority.
+  Unix.*.Root.PluginPath: $(ROOT_PLUGIN_PATH):$(ROOTSYS)/etc/plugins:
+  Unix.*.Root.DynamicPath: .:$(ROOT_DYN_PATH):
+  EOF
+  mv system.rootrc.0 $INSTALLROOT/etc/system.rootrc
```
TimoWilken, line 45: Why not quote `$INSTALLROOT`?

ktf:
```suggestion
  # Add support for ROOT_PLUGIN_PATH envvar for specifying additional plugin search paths
  grep -v '^Unix.*.Root.PluginPath' "$INSTALLROOT/etc/system.rootrc" > system.rootrc.0
  cat >> system.rootrc.0 <<\EOF
  # Specify additional plugin search paths via the environment variable ROOT_PLUGIN_PATH.
  # Plugins in \$ROOT_PLUGIN_PATH have priority.
  Unix.*.Root.PluginPath: $(ROOT_PLUGIN_PATH):$(ROOTSYS)/etc/plugins:
  Unix.*.Root.DynamicPath: .:$(ROOT_DYN_PATH):
  EOF
  mv system.rootrc.0 "$INSTALLROOT/etc/system.rootrc"
```

root.sh
ktf:
```suggestion
```

TimoWilken:
@ktf caching is done (for both PRs together). Do you want to merge them today or on Monday?

PR #5030, opened 2023-06-12
Fix use of Python-modules binaries during subsequent builds

TimoWilken:
On osx_arm64, we need to use the python binary from Python-modules (i.e. from Conda) in subsequent builds like ROOT.

These paths are in a non-standard place intentionally (for updatable RPM support), so aliBuild does not pick them up as it would usually.

### DIFF
--- a/python-modules.sh
+++ b/python-modules.sh
@@ -10,6 +10,8 @@ build_requires:
   - Python-modules-list
   - alibuild-recipe-tools
 prepend_path:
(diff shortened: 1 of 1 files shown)


mfasDa:
@TimoWilken see comment on https://github.com/alisw/alidist/pull/4974 I ran into trouble when building xrootd.

TimoWilken:
With https://github.com/alisw/alidist/pull/5035 it should work, but let's see...

PR #5032, opened 2023-06-13
Next attempt to enable GDML for Geant4

sawenzel:
* build G4 with GDML support
* update recipe for XercesC with correct tag
* use tools to generate modulefiles for XercesC + Geant4 (code reduction)

Note that in the past we had problems in relocating XercesC + Geant4. Let's hope that this has been fixed meanwhile.

### DIFF
--- a/geant4.sh
+++ b/geant4.sh
@@ -5,9 +5,11 @@ tag: "v11.0.4"
 source: https://gitlab.cern.ch/geant4/geant4.git
 requires:
   - "GCC-Toolchain:(?!osx)"
+  - xercesc
 build_requires:
   - CMake
   - "Xcode:(osx.*)"
+  - alibuild-recipe-tools
 prepend_path:
   ROOT_INCLUDE_PATH: "$GEANT4_ROOT/include:$GEANT4_ROOT/include/Geant4"
 incremental_recipe: |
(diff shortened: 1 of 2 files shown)


sawenzel:
@TimoWilken : We need to cross-check that it still builds whenever XercesC is coming from the remote store. In past, this wasn't the case.

geant4.sh
```diff
@@ -56,22 +60,12 @@ make install
...
-set version $PKGVERSION-@@PKGREVISION@$PKGHASH@@
-module-whatis "ALICE Modulefile for $PKGNAME $PKGVERSION-@@PKGREVISION@$PKGHASH@@"
-# Dependencies
-module load BASE/1.0 ${XERCESC_REVISION:+xercesc/$XERCESC_REVISION-$XERCESC_REVISION}
-# Our environment
+alibuild-generate-module --bin --lib > $MODULEFILE
+cat >> "$MODULEFILE" <<EOF
+# extra environment
 set GEANT4_ROOT \$::env(BASEDIR)/$PKGNAME/\$version
 setenv GEANT4_ROOT \$GEANT4_ROOT
```
ktf, line 67: Do we really need GEANT4_ROOT?

sawenzel: It was there before so I kept it. It cannot hurt and is independent of the intention of this PR.

xercesc.sh
```diff
@@ -27,17 +28,8 @@ make install
...
-}
-set version $PKGVERSION-@@PKGREVISION@$PKGHASH@@
-module-whatis "ALICE Modulefile for $PKGNAME $PKGVERSION-@@PKGREVISION@$PKGHASH@@"
-# Dependencies
-module load BASE/1.0
-# Our environment
+alibuild-generate-module --lib > $MODULEFILE
+cat >> "$MODULEFILE" <<EOF
+# extra environment
 set XERCESC_ROOT \$::env(BASEDIR)/$PKGNAME/\$version
```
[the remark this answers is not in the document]
TimoWilken: This can definitely be removed -- it only sets a Tcl variable, so it'll be set and thrown away immediately.

TimoWilken:
> @TimoWilken : We need to cross-check that it still builds whenever XercesC is coming from the remote store. In past, this wasn't the case.

I can build this and upload it to the remote store, then try reusing it. What's a simple command I can run in the alienv environment to make sure it's working?

PR #5038, opened 2023-06-15
Bump tag/commit for jq

sawenzel:
jq - official tag 1.6 - does not compile on the latest MacOS.

In absence of a newer release, we use commit hash 52d5988, which was tested to compile fine with alibuild on linux + Mac.

@co-author: Peter Hristov

### DIFF
--- a/jq.sh
+++ b/jq.sh
@@ -1,5 +1,5 @@
 package: jq
-version: v1.6
+version: v1.6-alice1
 tag: 52d5988
 source: https://github.com/stedolan/jq.git
 build_requires:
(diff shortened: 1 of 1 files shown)


jq.sh
```diff
@@ -1,6 +1,6 @@
 package: jq
 version: v1.6
```
sawenzel, line 2: unsure ... what we need to put as "version"?

ktf: `v1.6-alice1`, I would say.

ktf (approved):

PR #5039, opened 2023-06-15
Drop GEANT4_VMC examples

ktf:


sawenzel:
Do you have any particular motivation for this change? I have nothing against it, in principle.

ktf:
Not really, I just ended up recompiling it a few times due to changes in ROOT (which were quickly recompiled thank to the incremental build) and this saves 10 / 20 seconds (because it does not have an incremental build). We could also simply add an incremental recipe actually, if you prefer.

sawenzel:
@ihrivnac

ihrivnac:
> @ihrivnac

It's fine with me; I use the examples to test geant4_vmc updates only in stand-alone build

ktf:
Thanks. Dropping it from the main build.

PR #5048, opened 2023-06-21
Update acts.sh: compile with GEANT4 flags=ON

altsybee:


sawenzel (dismissed):

altsybee:
Hi, I removed the "trailing spaces after \" in L26, hopefully now it should pass the checks..

ktf:
I would be good to actually pass the GEANT4 path explicitly to CMake, to avoid picking up an eventual system installation.

altsybee:
> I would be good to actually pass the GEANT4 path explicitly to CMake, to avoid picking up an eventual system installation.
Could you implement this change, please?.. I'm not an expert in this business at all

sawenzel (approved):

sawenzel:
I updated this PR with @ktf suggestion and made Geant4 a runtime requirement. For me it builds. I guess we can safely merge given that this is used just for R&D and not production.

altsybee:
Thanks a lot, sure!

altsybee:
So, now it fails at something unrelated to the acts.sh change, I guess..

PR #5051, opened 2023-06-22
Bump Data Distribution to v1.6.0

cuveland:
This bumps DataDistribution to v1.6.0, which addresses a number of pending requests. TfBuilder will now terminate in case of a gRPC failure or if it cannot connect to one of the FLPs.


davidrohr (approved):

davidrohr:
Closing, since bump is included in #4843 together with FMQ

PR #5052, opened 2023-06-23
Re-enable simulation in O2 builds; find geant4 + xercesc

sawenzel:
Since the introduction of GDML support in Geant4,
some builds of O2 failed to include simulation. This lead to failure of the fullCI and absence of o2-sim on the GRID.

The problem is that finding Geant4 during O2-cmake triggers again a search for XercesC, which seems to fail.

This commit fixes the issue by asking cmake to prefer CONFIG package searches which guides it better to pickup XercesC from the alibuild software stack.

This appears to be the shortest possible (hot)fix for the issue.


sawenzel:
@TimoWilken : This is an urgent fix. We should merge it as soon as fullCI is green. Currently we have fullCI failures on all other builds due to this ... and missing simulation in CVFMS releases.

TimoWilken:
Thanks @sawenzel, I'll merge it as soon as the fullCI is done!

TimoWilken:
FullCI is green; other failures are known and unrelated; merging.

PR #5059, opened 2023-06-27
Bump FairMQ to 1.7.0

ktf:
Needed to have autoBind=false as default behaviour.

### DIFF
--- a/fairmq.sh
+++ b/fairmq.sh
@@ -38,6 +38,7 @@ cmake "${SOURCEDIR}"                                              \
       ${CXX_COMPILER:+-DCMAKE_CXX_COMPILER=${CXX_COMPILER}}       \
       ${CMAKE_BUILD_TYPE:+-DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE}} \
       -DCMAKE_INSTALL_LIBDIR=lib                                  \
+      -DFAIRMQ_CHANNEL_DEFAULT_AUTOBIND=OFF                       \
       -DDISABLE_COLOR=ON                                          \
       -DBUILD_EXAMPLES=ON                                         \
       -DBUILD_TESTING=ON
--- a/fairmq.sh
+++ b/fairmq.sh
@@ -1,5 +1,5 @@
 package: FairMQ
-version: "v1.6.0"
+version: "v1.7.0"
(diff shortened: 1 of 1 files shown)


rbx:
Note that with this change any FairMQ topology that uses this build (with `-DFAIRMQ_CHANNEL_DEFAULT_AUTOBIND=OFF`) and uses ODC/DDS plugin to dynamically configure TCP ports, will have to set `autoBind=false` in the channel config of the dynamically configured channel. Otherwise dynamic port configuration will fail.

ktf:
Ok, nevermind then. Can we have autoBind on by default when DDS / ODC is detected?

rbx:
If autoBind is off by default, I could turn it on in the ODC plugin constructor (I think the lifetime allows it, have to double check). But then wouldn't you be back to square one? Does https://alice.its.cern.ch/jira/browse/O2-2818 happen in a setup without ODC? And turning it off again, if the plugin sets it on, will be much trickier - you have to think about the lifetime and who sets config last.

But I'm not sure if changing the defaults without much thinking, based on a single issue (easily fixable with config), is a great idea.
Can you change the channel config on your side? That's what it is for. You could even hardcode it in the framework by calling SetProperty after device creation.

ktf:
Actually I do not understand anymore. Indeed the issue happens with ODC, @knopers8 could you confirm? If yes, then we do not have any case where the autoBind is useful, no? No one uses DDS standalone in ALICE, AFAICT.

rbx:
ODC always uses DDS. But whether or not autoBind is used depends on the contents of the topology file and channel configuration.
If the topology contains properties prefixed with `fmqchan_`, like this:
```<property name="fmqchan_aggregator-proxy-barrel_sp" scope="global"/>```, then the channel addresses will be exchanged. 

Addresses can still be hardcoded in the channel config, then the hardcoded values will be exchanged.
Or the address values can be left out, like this:
` --channel-config name=aggregator-proxy-barrel_sp,method=bind,type=sub,transport=zeromq,rateLogging=10 `
(taken from the linked file). Then autoBind will be used.

These example values are taken out of one of your EPN topologies few weeks ago. So, I assume that the feature is used to some degree. Thus, if you turn autoBind off by default, those channel configs will have to be adjusted.

PR #5061, opened 2023-06-28
Use ROOT @ v6-28-04-alice1

ktf:
This has a special patch which will prevent some dictionaries to be loaded just because the ROOT::Experimental namespace is referenced.

### DIFF
--- a/root.sh
+++ b/root.sh
@@ -1,7 +1,7 @@
 package: ROOT
 version: "%(tag_basename)s"
-tag: "v6-28-04"
-source: https://github.com/root-project/root.git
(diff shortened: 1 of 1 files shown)


ktf:
@sawenzel I see more errors with XercesC. Could you take a look?

sawenzel:
@ktf: We bumped XercesC last week. At that moment all CIs of alidist where green https://github.com/alisw/alidist/pull/5057. Weird that it doesn't build now!? On CC7 it builds fine locally for me. Maybe we have some interference, caching effect on the build machines?

PR #5068, opened 2023-07-02
alienpy 1.5.2

adriansev:
* Fast release for increasing timeout for cp operations. (this is the target of release)

* Improve mechanics for `cp -fastcheck` (now also aliased to `-skipmd5`) to check also for the age of destination file. If destination present and older than lfn then i will be re-downloaded
* ccdb command improvements (and helpers in example directory for QC shifter duties)


adriansev:
ping @TimoWilken

PR #5069, opened 2023-07-03
Wipe GEANT4's dependency path cache

TimoWilken:
When we build GEANT4, it caches its dependencies' locations on disk, but this information will be wrong if we create a tarball and unpack it elsewhere.

Delete this cache, so that GEANT4 searches anew on each build, and respects the *_ROOT paths we give it.

Discussing with @ktf, this is what we came up with. @sawenzel this should fix the CI errors when searching for xercesc.


ktf (dismissed):

sawenzel (dismissed):

TimoWilken:
Hm, vgm still fails because now `XercesC_LIBRARY` and `XercesC_INCLUDE_DIR` are unset. I guess that means we still need to set these in `alidist/vgm.sh`? Or maybe `-DXercesC_ROOT="$XERCESC_ROOT"` would be enough?

ktf:
I suppose so. @ihrivnac I remember you fixed something about this kind of relocations. Was VMC left out?

TimoWilken:
Hm, this is very weird. This now seems to work, but not on Mac, where vgm generates a bad Makefile(???).

Also, weirdly, build/O2/alidist-dataflow-cs8 doesn't have any logs...

TimoWilken:
Oh never mind the missing logs, o2-dataflow disables GEANT4 so nothing was rebuilt.

For the vgm problem on Mac, I'll try to investigate tomorrow.

@ktf do you want to merge now anyway?

ktf:
I do not think breaking macOS on ARM is a good idea. Let's have a look at it tomorrow.

TimoWilken:
I don't know how I missed this earlier... only setting `XercesC_ROOT` works for me on Mac (I can build GEANT4_VMC). I haven't yet tested it on Linux, tbc...

TimoWilken:
Closing since #5072 was merged.

PR #5072, opened 2023-07-05
Find XercesC via CMAKE_PREFIX_PATH

sawenzel:
* do not use cached information about packages in Geant4 installation
* help finding XercesC via CMAKE_PREFIX_PATH

### DIFF
--- a/xercesc.sh
+++ b/xercesc.sh
@@ -11,6 +11,8 @@ prefer_system_check: |
 prepend_path:
   ROOT_INCLUDE_PATH: "$XERCESC_ROOT/include"
   LD_LIBRARY_PATH: "$XERCESC_ROOT/lib"
+env:
+  CMAKE_PREFIX_PATH: "$CMAKE_PREFIX_PATH:$XERCESC_ROOT"
 ---
 
 cmake $SOURCEDIR                                         \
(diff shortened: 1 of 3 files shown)


xercesc.sh
```diff
@@ -11,6 +11,8 @@ prefer_system_check: |
 prepend_path:
   ROOT_INCLUDE_PATH: "$XERCESC_ROOT/include"
   LD_LIBRARY_PATH: "$XERCESC_ROOT/lib"
+env:
```
ktf, line 14: I think this would need to be prepend_path:

```suggestion
prepend_path:
```

because otherwise Xerces will overwrite everything else.

TimoWilken: It should be in `prepend_path` to handle the case where `CMAKE_PREFIX_PATH` is empty in the outside environment.

There's already a `prepend_path` block above, so merge it into that one (as `CMAKE_PREFIX_PATH: "$XERCESC_ROOT"`) to avoid wiping out the statements already there. Incidentally, aliBuild should already handle `LD_LIBRARY_PATH`, so we probably don't need that one...

TimoWilken:
Merging since CI is broken and 081370e worked.

PR #5078, opened 2023-07-10
Bump qualitycontrol to v1.107.2

Barthelemy:


vascobarroso (approved):

davidrohr:
> I guess that it won't build against the latest O2

I have seen that, but if we bump QC in alidist, we have to make sure it also builds with latest O2, otherwise we break the O2 CI, since it will pick up the new QC :(.
Can we fix this in QC, to be compatible with latest O2/dev as well? Do we need to do anything on O2 side?

Barthelemy:
@davidrohr checking...

Barthelemy:
@davidrohr it is probably missing https://github.com/AliceO2Group/QualityControl/commit/42981ae72d5392102253d54ca6c50061b2492bef
I can add it to v1.107 --> v1.107.3

Barthelemy:
As Vasco just pointed out to me, we don't need this version in alidist, it is just to redo the RPMs for the FLPs in prod

PR #5080, opened 2023-07-11
Bump bookkeeping to v0.59.0

martinboulais:


TimoWilken (approved):

vascobarroso:
@TimoWilken any idea why this PR went to you for approval instead of @alisw/dataflow-experts ?

TimoWilken:
@vascobarroso `bookkeeping-api.sh` isn't listed in [the CODEOWNERS file](https://github.com/alisw/alidist/blob/master/CODEOWNERS), so it fell back to the default entry. I'll add it! Are there any other unlisted recipes that should be owned by @alisw/dataflow-experts as well?

vascobarroso:
@TimoWilken thanks! Looking at the list, I cannot see any other that should be owned by @alisw/dataflow-experts.

TimoWilken:
Great, thanks for checking @vascobarroso!

PR #5100, opened 2023-08-07
Pin Python package versions using constraint files

TimoWilken:
Constraint files are supported in pip 7.1+. The oldest we have pre-installed (on slc7 and slc8) is pip 9.0.3, so this is supported everywhere.

### DIFF
--- a/python-modules.sh
+++ b/python-modules.sh
@@ -64,14 +64,16 @@ case $ARCHITECTURE in
 esac | tr -s '[:space:]' '\n' > requirements.txt
 
 # Install pinned basic requirements for python infrastructure
-python3 -m pip install -IU -r base-requirements.txt
+# Use the requirements also as a constraints file to pin versions of dependencies
+# (https://pip.pypa.io/en/stable/user_guide/#constraints-files).
+python3 -m pip install -IU -c base-requirements.txt -r base-requirements.txt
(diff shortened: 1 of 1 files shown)


TimoWilken:
Hi @ktf @adriansev, what do you think about this?

It seems that with the previous approach of installing Python packages with a plain `-r requirements.txt`, the versions specified in that file aren't necessarily always installed (e.g. we specify Cython==0.x, but things still broke when the latest version became 3.x).

With constaint files, pip should always install the exact version that we specify, and not let anything else override it, even if something else would otherwise pull in a different version.

TimoWilken:
You're right @adriansev, xjalienfs and other packages that install Python packages on top of Python-modules might end up causing version updates, since they won't use the constraint file. This wouldn't completely prevent newer package versions from being installed accidentally then...

adriansev:
@TimoWilken if there is a need for it, i can add my requirements to Python-modules and constrain them there .. but note the the whole dependency chain would also need to be constrained and we need to establish if this does not make some packages incompatible with other ..

TimoWilken:
Yeah, it's probably a bigger job. It might be better to install the constraint files in the Python-modules-list package, so that other Python-based packages can use them when they install their own stuff.

PR #5106, opened 2023-08-16
Bump qualitycontrol to v1.114.0

Barthelemy:


martenole:
Can this be merged? I guess the macOS builder is still not working in any case, right?

martenole:
Ah, then maybe @davidrohr or @ktf can help?

PR #5113, opened 2023-08-23
GenTopo: install also python scripts

davidrohr:


martenole:
@davidrohr can we revert this commit until we actually have python scripts in the folder?

With this commit my build fails with 
```
09:11:06 2023-08-23@09:11:06:DEBUG:GenTopo:GenTopo:nightly-20230823.1: cp: cannot stat '/jenkins/workspace/build-any-ib/sw/20156530/1/SOURCES/GenTopo/nightly-20230823.1/nightly-20230823.1/DATA/tools/epn/*.py': No such file or directory
```

martenole:
Actually I could work around it by selecting not alidist@master but the latest FLP suite branch @davidrohr

davidrohr:
argh, actually I tried the `cp` command locally with a missing `.py` file, and it worked, and did the copy correctly.
However, the return code is non-zero, and aliBuild checks that. Sorry for that...

You can either revert, or just use an older alidist slug for the build.

PR #5117, opened 2023-08-25
Add abseil as o2 build dependency

MichaelLettrich:
librans needs abseil as a build dependency as it relies on its fast hash-map implementation.

### DIFF
--- a/o2.sh
+++ b/o2.sh
@@ -29,6 +29,7 @@ requires:
   - KFParticle
   - RapidJSON
 build_requires:
+  - abseil
   - GMP
   - MPFR
   - googlebenchmark


ktf:
Can we avoid this?

MichaelLettrich:
I want to use `absl::flat_hash_map`  as a fast Lookup table in the rANS entropy coder. This is pretty much a drop-in-replacement for `std::unordered_map`, so no new functionality is gained, and actually `std::unordered_map` is easier to work with because of pointer stability.

But there is a penalty of 3x-5x for plain lookups in my benchmarks. The problem is that these lookups are one of the hottest parts of the compression code. I have unsuccessfully tried for weeks to use different types of data-structures to close the performance  gap, but could not find anything better that worked and was readily available in alidist. 
So I cannot really avoid this unless we are willing to suffer a significant (roughly 25x) slowdown of the new compression code or stick to the old code and miss out on up to 2.8x decrease in CTF file size for P-P and up to 15% in Pb-Pb.

Are there any technical reasons apart from not wanting to introduce yet another dependency to O2 ?

shahor02:
Hi @ktf @davidrohr , since we need to validate new compression code asap, could we converge quickly on this PR?

ktf:
@shahor02 Fair enough, hopefully we will have all the usages of abseil always synchronised.

MichaelLettrich:
@ktf   OK, I see. So it is not about absl in general or yet another dependency, but other components using absl? (have no idea about that)
@davidrohr writing a fast hash-table is a very difficult task I would not go there. But the way I integrated it, it is very easily replaceable  - literally a one-liner,  if interfaces are STL compatible. (+ validation ofc).

For now its probably best if we could keep absl in, but I can perpare something during YETS that does not use absl.

MichaelLettrich:
@ktf I cannot see the logs currently, but is there a glitch with alidist and macos or is that an actual, real error? 
if its a glitch, can we reset those two tests? It would be really good to merge this PR ASAP to move forward with the validation of my O2 pr.

shahor02:
@ktf the macOS macro loading check fails in all PRs. Can one merge this?

davidrohr:
I see macOS CI failling generally also for other PRs, so seems unrelated. Merging.

PR #5130, opened 2023-09-07
Bump libwebsockets to 4.3.2

davidrohr:


davidrohr:
@ktf : could you have a look? It fails on MacOS with `error: field has incomplete type 'HMAC_CTX' (aka 'struct hmac_ctx_st')`.
I removed a patch we had in the old version for MacOS, but that patch was not compatible wtih >= 4.0, and the comment said it should be removed with >= 4.0.

ktf:
I will check in a bit.

ktf:
I think the issue is that libwebsocket uses a new OpenSSL API, while we still have an old v1.1.1m version which we build on macOS. I think this goes back to some AliEn / JAlien / XRootD requirement. @costing do you know if we can update from the grid side?

costing:
If we are not building the old alien any more there should be no issue to use a new OpenSSL.

davidrohr:
So is someone creating a PR to bump OpenSSL? Which version do we want to bump to?

ktf:
I just did. It needed to be done in this PR, actually.

davidrohr:
@ktf : this still fails on MacOS due to `HMAC_CTX` having incomplete type, as before.

ktf:
So the issue is that the test for HMAC_CTX_new fails with:
 
```
558 : && /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/cc -Wno-deprecated-declarations -Wno-deprecated -Wa    ll -Wextra -Wno-unused-parameter -Wconversion -Wsign-compare -Wstrict-aliasing -fvisibility=hidden -Wundef  -Wuninitialized -Wtype-limits -Wign    ored-qualifiers -fPIC -O2  -Werror -Wno-deprecated-declarations -pthread -Wno-error=unused-command-line-argument -DCHECK_FUNCTION_EXISTS=HMAC_C    TX_new -arch arm64 -isysroot /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX14.0.sdk -mmacosx-versio    n-min=10.15 -Wl,-search_paths_first -Wl,-headerpad_max_install_names  CMakeFiles/cmTC_09b8d.dir/CheckFunctionExists.c.o -o cmTC_09b8d  /opt/homebrew/opt/openssl@3/lib/libssl.dylib && :
559 ld: Undefined symbols:
560   _HMAC_CTX_new, referenced from:
561       _main in CheckFunctionExists.c.o
562 clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

this is because apparently CHECK_FUNCTION only knows about libssl, but not libcrypto. This is probably an issue also elsewhere, actually, e.g. grpc.  @Barthelemy @knopers8 @teo @awegrzyn you might be interested as well. https://github.com/alisw/alidist/pull/5149 seems to work on my laptop.

@costing which version of JAlien should we use with libwebsockets 4.3.2?

davidrohr:
we need jalien-root 0.7.3 for the new libwebsockets, but we just downgraded to 0.7.1: https://github.com/alisw/alidist/pull/5146

ktf:
Superseded by #5149.

PR #5136, opened 2023-09-13
O2 requires abseil not only for building

martenole:


sawenzel (approved):

MichaelLettrich:
I am wondering about the requires part? Abseil is compiled as a static libary and therefore linked as such throughout the build. There is no need to load that module if I understand correctly?

MichaelLettrich:
Ok, that part I did not take into account. Then changing it to from `build-requires` to `requires` totally makes sense.

ktf:
This breaks macOS.

PR #5140, opened 2023-09-19
Bump boost to v1.82.0

ktf:


ktf:
Needed to work with the updated XCode.

TimoWilken:
Hi @ktf, from the log, it looks like it's running out of space in `/dev/shm`, i.e. running out of memory!

The machine seems to have enough disk space available (the 1TB disk is half empty).

ktf:
You are right. @dennisklein @rbx is there any known issue in using boost 1.83.0 with FairMQ?

rbx:
Nothing known. But we also don't have a test environment with 1.83 afaik.

davidrohr:
I have been using boost 1.82 for few weeks now on my laptop without problems.
Haven't tried 1.83 yet.

ktf:
Let's try to use 1.82, then...

ktf:
Still broken. I need to check in details.

davidrohr:
OK, meanwhile I tried 1.83 locally, and that also works for me.
But I am also not running multiple workflows in parallel.

PR #5141, opened 2023-09-19
Bump FairRoot to v18.4.9-alice2

ktf:


ktf:
Needed for the new XCode.

ktf:
The only difference is https://github.com/alisw/FairRoot/commit/2465f4c8a951eb3a1cd081b06c688e9c5eb111f1 which should be safe and it's basically what is done upstream. @dennisklein do you see any problem with it?

Merging to get things compiling again.

PR #5144, opened 2023-09-21
Disable Geant4 MT mode

sawenzel:
Rationale:

*) ALICE is not using MT functionality of Geant4
*) recent perf benchmarks have shown that (at least on some linux kernels),
   the MT mode has a ~2% overhead spent in function _dl_tlsdesc_dynamic.

This overhead is best avoided by switching off MT and it will lead to a speedup of Geant4 in many cases.


TimoWilken:
Hi @sawenzel, no objections from my side, but please let me know before merging! Since this PR changes an env variable in a defaults file, I expect this will lead to a full rebuild of the entire software stack, so I'd like to cache it before it's merged.

sawenzel:
A quick benchmark on alibi (running CC7) shows that the runtime for 150pp events goes down from ~680s to ~640s with this PR. Output is identical. So this is even a substantially larger effect than %2 (for whatever reason) and it is free lunch.

TimoWilken:
I've started the caching, so we can merge on Monday!

ktf:
Is this done?

ktf:
Why do we have those variables in the defaults? That will mean that whenever we change them, we rebuild everything. Is that really needed? Why not simply having a "simulation-config.sh" package so that only the simulation dependencies get rebuilt?

PR #5146, opened 2023-09-23
revert jalien-root to 0.7.1

costing:


adriansev (approved):

ddobrigk:
Hi everyone, I see a CI error - but it looks unrelated? 

I am a bit curious to merge this for testing, but OK, I leave it up to you. Even with reverting, of course, it would also be very important to understand *why* version 0.7.3. is causing a large slowdown and address that as well.

davidrohr:
let's see if it helps...

PR #5149, opened 2023-09-25
Move to JAliEn-ROOT 0.7.4

ktf:
Move to JAliEn-ROOT 0.7.4

* Fix static cast
* Improve detection of OpenSSL to use both libcrypto and libssl

### DIFF
--- a/jalien-root.sh
+++ b/jalien-root.sh
@@ -1,6 +1,6 @@
 package: JAliEn-ROOT
 version: "%(tag_basename)s"
-tag: "0.7.1"
+tag: "0.7.4"
 source: https://gitlab.cern.ch/jalien/jalien-root.git
 requires:
   - ROOT
--- a/libwebsockets.sh
+++ b/libwebsockets.sh
@@ -6,6 +6,7 @@ build_requires:
   - CMake
   - "GCC-Toolchain:(?!osx)"
   - "OpenSSL:(?!osx)"
+  - ninja
 prefer_system: "osx"
 prefer_system_check: |
   printf '#if !__has_include(<lws_config.h>)\n#error \"Cannot find libwebsocket\"\n#endif\n' | cc -I$(brew --prefix libwebsockets)/include -c -xc - -o /dev/null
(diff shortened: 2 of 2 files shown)


davidrohr:
well, the problem will be, now that we downgraded jalien-root again, to check where the performance problem in analysis comes from, this will not be compatible anymore to libwebsockets.

ktf:
Ok, but the detection of OPENSSL needs to be corrected in any case. It currently works by accident (here and in other places, I fear).

costing (dismissed):

ktf:
The issue on mac is a genuine error. Probably it's now caught by new new XCode. I am preparing a fix which will require bumping to 0.7.4.

ktf:
Tested on hyperloop as well...

ktf:
@TimoWilken the error with git is again there. Any idea of what this might be?

```
Git command for package 'Clang' failed.
Command: git clone --bare https://github.com/llvm/llvm-project /build/nomad/alloc/3b592022-e81d-4247-9f4f-2e37864d8242/ci/local/o2-alidist/sw/MIRROR/clang
In directory: .
Exit code: -15
```

ktf:
@davidrohr could you check on Arch linux?

davidrohr:
I tried this and it works for me.
@costing : shall we merge and then check if it is ok in the Analysis?
Could we actually confirm that after reverting JAliEn-ROOT to 0.7.1, the Analysis issue was gone?

ktf:
@ddobrigk @jgrosseo can you confirm that going back to 0.7.1 the issue disappeared?

costing:
I've asked David (Chinellato) about analysis jobs, waiting for his reply. Ruben also had it on his machine and as far as I understood reverting fixed it. The same issue with 0.7.3 was observed in reco jobs too and Chiara has prepared a new async package with 0.7.1, we'll see if it helps there.

Before bumping libwebsockets by default, could we have a build with a suffix, to be tried out as a batch of jobs first? I'll be away next week and I'm reluctant to disturb the production setup during this time.

davidrohr:
Not sure how we can create such a tag. We can merge and then use the override file option in jenkins, to take the files from alidist/master.
Or we could use a different alidist slug, which has Giulio's PR in.

ktf:
For the record, I am building a daily tag with my branch https://alijenkins.cern.ch/job/DailyBuilds/job/AdHocO2Physics/282/. @ddobrigk can you validate it once it's done so that we can merge it?

ktf:
daily-20231023-2201-1 uses this and it's now deployed.

ktf:
For the record, I am now comparing a train run on the grid (kudos to @jgrosseo) with a patched version:

https://alimonitor.cern.ch/hyperloop/train-run/129599

with the vanilla one from the same day:

https://alimonitor.cern.ch/hyperloop/train-run/129692

With 90% of the jobs done and 70% of the files, the runtimes seem compatible. I will merge this once the test terminates for good.

ktf:
Test completed, timings  are compatible. Merging.

PR #5150, opened 2023-09-27
Have ZeroMQ compile with new XCode

ktf:


ktf:
Dropping the tests. We do not need them and they do not compile anymore on newer XCode.

PR #5156, opened 2023-10-02
Update Python packages for py3.11 to working versions

TimoWilken:


TimoWilken:
Caching #5156, #5165, #5198: <https://alijenkins.cern.ch/job/CacheO2Package/85/>

PR #5158, opened 2023-10-03
Use openssl3 on macOS

ktf:


ktf:
This seems to fix macOS.

PR #5159, opened 2023-10-03
Use python modules compatible to python 3.6

davidrohr:
@ktf : This is what we did to fix compilation on EPNs.

### DIFF
--- a/python-modules-list.sh
+++ b/python-modules-list.sh
@@ -2,9 +2,9 @@ package: Python-modules-list
 version: "1.0"
 env:
   PIP_BASE_REQUIREMENTS: |
-    pip==23.0.1
-    setuptools==65.5.1
-    wheel==0.40.0
+    pip==21.3.1
+    setuptools==59.6.0
+    wheel==0.37.1
   PIP_REQUIREMENTS: |
     requests==2.27.1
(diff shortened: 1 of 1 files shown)


ktf:
@adriansev any objections to this?

adriansev:
> @adriansev any objections to this?

not really .. actually i do not get why is needed since we have our own python 3.9, but if this make thinks work then it's ok .. will the CI will build xrootd? (to make sure that xrootd python bindings are installed without problems)

davidrohr:
The point is that our minimum python requirement is 3.6, which is the default python of CC8, so we use the system python on the FLPs / EPNs. The python modules versions which I downgraded require 3.7, so currently our setup is broken if you have system python 3.6. It uses that, and then fails to build the python modules.
So we have to either downgrade the modules, or bump the minimum python version to 3.7.

adriansev:
indeed in EL7 the last available versions are the ones mentioned in this commit .. and i absolutely had no idea that in el8 there is still python3.6 when in el9 the python is 3.9

ktf:
Ok, let's merge this, I think there is some benefit in having a configuration which is compatible with CC7 / CC8.

PR #5166, opened 2023-10-05
Use openssl@3 on macOS

ktf:
Use openssl@3 on macOS


ktf:
Superseeded by #5175.

PR #5167, opened 2023-10-05
Cleanup abseil

ktf:
While we do not need to expose abseil to ROOT (if anything because it does not
actually work) we can still benefit from some of the clean-ups I did. In particular
this will:

* Build with ninja
* Generate the modulefile on the fly


abseil.sh
```diff
@@ -5,28 +5,29 @@ requires:
   - "GCC-Toolchain:(?!osx)"
 build_requires:
   - CMake
+  - ninja
   - alibuild-recipe-tools
 source: https://github.com/abseil/abseil-cpp
+prefer_system: "osx"
+prefer_system_check: |
+  printf '#include <absl/container/flat_hash_map.h>' | c++ -std=c++17 -I`brew --prefix abseil`/include -c -xc++ - >/dev/null
```
ktf, line 13:
```suggestion
  printf '#include <absl/container/flat_hash_map.h>' | c++ -std=c++17 -I"$(brew --prefix abseil)/include" -c -xc++ - >/dev/null
```

singiamtel:
Merged by #5785

PR #5176, opened 2023-10-08
Allow shortcutting builds

ktf:
This shows how the alibuild-variant-support package could be used to e.g. speed up builds of O2Physics tutorials by only building the Tutorials/install target, rather than the whole package.

### DIFF
--- a/o2physics.sh
+++ b/o2physics.sh
@@ -12,11 +12,20 @@ build_requires:
   - CMake
   - ninja
   - alibuild-recipe-tools
+  - alibuild-variant-support
 source: https://github.com/AliceO2Group/O2Physics
 incremental_recipe: |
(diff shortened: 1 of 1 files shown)


ktf:
@aalkin @TimoWilken this is what I think I was mentioning last week / in the past. 

Right now it would only work with:

    ALIBUILD_BUILD_VARIANT=tutorial aliBuild build O2Physics --defaults o2 --debug

but it would be cool to have support in aliBuild so that one could do:

    aliBuild build O2Physics#tutorials --defaults o2 --debug

with the "#tutorial" being stripped to determine the package name, yet inserted in the develHash, so that if one drops it, it rebuilds the development package without need to cleanup.

awegrzyn:
Are there any docs on how to use `ALIBUILD_BUILD_VARIANT`? It seems like interesting feature.

ktf:
not yet, we are still developing the idea. we could brainstorm things a bit together though, since indeed you / FLP was one of the customers I had in mind for such a feature.

PR #5177, opened 2023-10-09
GPU: add target architecture override hip,cuda

mconcas:
After any of the GPU settings we reserve the possibility to override the target architectures for HIP.
@davidrohr


davidrohr:
thx, could you add a similar override for CUDA?

PR #5189, opened 2023-10-26
Revert removing llvm-spirv build from clang

davidrohr:
Just found out by chance that the internal clang spirv backend is not working. (Not at all in LLVM15, and in LLVM17 it still produces broken code.) Thus, for now we have to keep bundling the external tool with clang. So far, O2 was building with partial OpenCL compilation, due to a bug in the O2 CMake, which silently disabled the SPIRV translation in this case.

### DIFF
--- a/clang.sh
+++ b/clang.sh
@@ -52,6 +52,19 @@ cmake "$SOURCEDIR/llvm" \
 
 cmake --build . -- ${JOBS:+-j$JOBS} install
 
+git clone -b "${PKGVERSION%%.*}.0.0" https://github.com/KhronosGroup/SPIRV-LLVM-Translator
+mkdir SPIRV-LLVM-Translator/build
+pushd SPIRV-LLVM-Translator/build
+cmake ../ \
+  -G Ninja \
+  -DLLVM_DIR="$INSTALLROOT/lib/cmake/llvm" \
+  -DLLVM_BUILD_TOOLS=ON \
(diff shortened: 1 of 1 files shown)


davidrohr:
Hi @TimoWilken : I know. If the CI gets green, I'd merge it tomorrow evening, so everything should be up to date Monday morning.

davidrohr:
And it is still not as bad as changing GCC :)

TimoWilken:
Perfect, thank you @davidrohr! And you're right, haha, GCC would be much worse :)

TimoWilken:
Hi @davidrohr, during caching there was an error:

```
++ git clone -b v15.0.7 https://github.com/KhronosGroup/SPIRV-LLVM-Translator
Cloning into 'SPIRV-LLVM-Translator'...
Remote branch v15.0.7 not found in upstream origin
```

The upstream repo only seems to have vX.0.0 available in all recent vX.* series: https://github.com/KhronosGroup/SPIRV-LLVM-Translator/tags

Perhaps something like the following would work instead?

```bash
git clone -b "${PKGVERSION%%.*}.0.0" https://github.com/KhronosGroup/SPIRV-LLVM-Translator
```

davidrohr:
@TimoWilken : This seems to work now. FullCI will require a change to O2, which I'll merge after this one.
The other two errors seem unrelated. Could you cache it now?

PR #5190, opened 2023-10-26
Fix ALIBUILD_O2_OVERRIDE_*_ARCHS

mconcas:
This simply was not working. @davidrohr

### DIFF
--- a/o2.sh
+++ b/o2.sh
@@ -191,8 +191,8 @@ cmake $SOURCEDIR -DCMAKE_INSTALL_PREFIX=$INSTALLROOT                            
       ${DISABLE_GPU:+-DENABLE_CUDA=OFF -DENABLE_HIP=OFF -DENABLE_OPENCL=OFF -DENABLE_OPENCL2=OFF}         \
       ${ALIBUILD_ENABLE_CUDA:+-DENABLE_CUDA=ON}                                                           \
       ${ALIBUILD_ENABLE_HIP:+-DENABLE_HIP=ON}                                                             \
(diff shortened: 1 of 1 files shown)


davidrohr:
I don't understand. Your change in this PR shouldn't make any difference, and I don't understand why it was not working before. For me such syntax works:
```
qon@qon ~ $ FOO=foo
qon@qon ~ $ echo ${FOO:+--test ${FOO}}
--test foo
qon@qon ~ $ FOO=
qon@qon ~ $ echo ${FOO:+--test ${FOO}}
```

mconcas:
> I don't understand. Your change in this PR shouldn't make any difference, and I don't understand why it was not working before. For me such syntax works:
> 
> ```
> qon@qon ~ $ FOO=foo
> qon@qon ~ $ echo ${FOO:+--test ${FOO}}
> --test foo
> qon@qon ~ $ FOO=
> qon@qon ~ $ echo ${FOO:+--test ${FOO}}
> ```

I don't understand either, I was using it on EPN and it seemed to ignore the `HIP_AMDGPUTARGET` override.
Then I copied the same format used in other lines in o2.sh and it started working.

Maybe there is no correlation.
Indeed your very same test works on EPN shell.

davidrohr:
I think you tried an incremental build? It might not work with that.
When you changed the o2.sh recipe, that has triggered a rebuild

mconcas:
> I think you tried an incremental build? It might not work with that. When you changed the o2.sh recipe, that has triggered a rebuild

yes, it was incremental. I did some weak tentatives to re-trigger it by changing lines in FindO2GPU.cmake and CMakeListst of ITSTrackingGPU.

PR #5191, opened 2023-10-27
Changing AEGIS recipe for Sonoma build

ChSonnabend:
Modifying dependencies of
- boost
- dds
- aegis
for a successful build on macOS Sonoma

### DIFF
--- a/aegis.sh
+++ b/aegis.sh
@@ -21,7 +21,8 @@ if [ $FVERSION -ge 10 ]; then
    echo "Fortran version $FVERSION"
    SPECIALFFLAGS=1
 fi
-cmake $SOURCEDIR -DCMAKE_INSTALL_PREFIX=$INSTALLROOT       \
+cmake $SOURCEDIR -DCMAKE_CXX_FLAGS=-isystem\ "/opt/homebrew/include" \
+                 -DCMAKE_INSTALL_PREFIX=$INSTALLROOT       \
                  ${CMAKE_GENERATOR:+-G "$CMAKE_GENERATOR"} \
                  -DCMAKE_BUILD_TYPE=$CMAKE_BUILD_TYPE      \
                  -DCMAKE_SKIP_RPATH=TRUE \


ChSonnabend:
Tagging @TimoWilken

TimoWilken:
Thank you very much @ChSonnabend! I'll mark the PR as "ready for review" so that the CI will run.

TimoWilken (approved):
Approving for CI

aegis.sh
```diff
@@ -21,7 +21,8 @@ if [ $FVERSION -ge 10 ]; then
    echo "Fortran version $FVERSION"
    SPECIALFFLAGS=1
 fi
-cmake $SOURCEDIR -DCMAKE_INSTALL_PREFIX=$INSTALLROOT       \
+cmake $SOURCEDIR -DCMAKE_CXX_FLAGS=-isystem\ "/opt/homebrew/include" \
```
ktf, line 24: What are you trying to achieve here?

ChSonnabend: This was needed to find the system dependencies for nlohmann/json. Not sure if this is needed in general, but I added it here for now.

dds.sh
```diff
@@ -1,6 +1,6 @@
 package: DDS
 version: "%(tag_basename)s"
-tag: "3.7.22"
+tag: "3.7.23"
```
ktf, line 3: This is good to go already, so I split it in a different PR (https://github.com/alisw/alidist/pull/5192).

aegis.sh
ktf: If this is some "out-of-band" external, we should probably have a separate recipe for it.

boost.sh
```diff
@@ -1,6 +1,6 @@
 package: boost
-version: v1.75.0
-tag: v1.75.0
+version: v1.82.0
+tag: v1.82.0
```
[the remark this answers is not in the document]
ChSonnabend: Ah ok, this I didn't know. For me it works on Sonoma (and it works only like this, not with the older version...).

ktf (changes requested):
Thank you for your contribution. I have a few comments inline. I would use the orginal PR for the boost changes. Moreover, the aegis changes cannot be merged as they are, since they hardcode a global system path in the search. If you really do not want to package json for now, you should at least use `$(brew --prefix json)/include` and have it as part of a macOS specific flag.

boost.sh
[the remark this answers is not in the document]
ktf: Yes, it compiles and runs fine with 1.82.0 as long as you run one workflow at the time. When you run more than one, sometimes it fails. You can find more details in the error report of the other PR.

ChSonnabend:
After some exchanges with @ktf only the AEGIS changes were integrated in this PR.

PR #5193, opened 2023-10-27
Make sure builds find our protobuf headers

TimoWilken:
The header version must match the version of the protoc binary, which we put in PATH already. Protobuf's installed pkgconfig file contains the right information.

This should solve a build failure observed by a user that had a mismatching version of protobuf installed system-wide.

@ktf, if you prefer only the first commit, I can force-push to remove the cleanup one.

### DIFF
--- a/protobuf.sh
+++ b/protobuf.sh
@@ -4,6 +4,10 @@ source: https://github.com/protocolbuffers/protobuf
 build_requires:
   - CMake
   - "GCC-Toolchain:(?!osx)"
+prepend_path:
+  # The protobuf headers must match the protoc binary version, so prevent the
+  # use of system headers by putting ours first in the path.
+  PKG_CONFIG_PATH: "$PROTOBUF_ROOT/lib/pkgconfig"
 ---
 
 cmake $SOURCEDIR/cmake                  \
--- a/protobuf.sh
+++ b/protobuf.sh
@@ -4,38 +4,24 @@ source: https://github.com/protocolbuffers/protobuf
 build_requires:
   - CMake
   - "GCC-Toolchain:(?!osx)"
+  - alibuild-recipe-tools
 prepend_path:
   # The protobuf headers must match the protoc binary version, so prevent the
   # use of system headers by putting ours first in the path.
   PKG_CONFIG_PATH: "$PROTOBUF_ROOT/lib/pkgconfig"
 ---
-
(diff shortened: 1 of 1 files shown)


protobuf.sh
```diff
@@ -4,34 +4,24 @@ source: https://github.com/protocolbuffers/protobuf
...
-# Our environment
-set PROTOBUF_ROOT \$::env(BASEDIR)/$PKGNAME/\$version
-setenv PROTOBUF_ROOT \$PROTOBUF_ROOT
-prepend-path LD_LIBRARY_PATH \$PROTOBUF_ROOT/lib
-prepend-path PATH \$PROTOBUF_ROOT/bin
+mkdir -p "$INSTALLROOT/etc/modulefiles"
+alibuild-generate-module --bin --lib > "$INSTALLROOT/etc/modulefiles/$PKGNAME"
+cat >> "$INSTALLROOT/etc/modulefiles/$PKGNAME" <<EoF
+setenv PROTOBUF_ROOT \$PKG_ROOT
 EoF
```
ktf, line 27: Do we really need such a variable? Why?

```suggestion
```

TimoWilken: I only kept it because it was there before. But at runtime, we shouldn't need it, right? Removed.

protobuf.sh
```diff
@@ -4,34 +4,21 @@ source: https://github.com/protocolbuffers/protobuf
...
+  # use of system headers by putting ours first in the path.
+  PKG_CONFIG_PATH: "$PROTOBUF_ROOT/lib/pkgconfig"
 ---
-
-cmake $SOURCEDIR/cmake                  \
-    -DCMAKE_INSTALL_PREFIX=$INSTALLROOT \
-    -Dprotobuf_BUILD_TESTS=NO           \
-    -Dprotobuf_MODULE_COMPATIBLE=YES    \
-    -Dprotobuf_BUILD_SHARED_LIBS=OFF    \
+#!/bin/bash -e
```
ktf, line 13: For the record, this is not actually needed nor used...

TimoWilken: You're right, we run the script with `bash -e <script>` explicitly, so the shebang line doesn't matter. I only find it helps to make the run conditions obvious, and it helps tools like shellcheck figure out that this is how the script is invoked.

PR #5194, opened 2023-10-27
fix typo, system detection of jq was broken

davidrohr:
@ktf : I assume this was a typo?


ktf:
Yes, I was probably testing the recipe without having to rebuild everything.

PR #5195, opened 2023-10-27
Allow using libwebsockets from system also on Linux

davidrohr:
@ktf : Any reason why this is limited to osx?


ktf:
I guess not.

PR #5200, opened 2023-11-02
Add libuv dependency to libwebsockets

yuw726:
Enable libuv support in JAliEn-ROOT by adding libuv as a dependency in libwebsockets recipe.

This change doesn't turn on libuv in jalien-root immediately, but prepares the transition. It is controlled by a dedicated lws option in the code whether libwebsockets use libuv loop or the default one.

### DIFF
--- a/libwebsockets.sh
+++ b/libwebsockets.sh
@@ -8,6 +8,7 @@ build_requires:
   - "OpenSSL:(?!osx)"
   - ninja
   - alibuild-recipe-tools
+  - libuv
 # On Mac, Brew's libwebsockets loads Brew's Python, which confuses ROOT.
 prefer_system: "(?!osx_)"
 prefer_system_check: |
@@ -47,6 +48,9 @@ cmake $SOURCEDIR                                                    \
       ${OPENSSL_ROOT:+-DOPENSSL_LIBRARIES=$OPENSSL_ROOT/lib/libssl.$SONAME;$OPENSSL_ROOT/lib/libcrypto.$SONAME}     \
       ${OPENSSL_ROOT:+-DLWS_OPENSSL_INCLUDE_DIRS=$OPENSSL_ROOT/include}                                             \
       ${OPENSSL_ROOT:+-DLWS_OPENSSL_LIBRARIES=$OPENSSL_ROOT/lib/libssl.$SONAME;$OPENSSL_ROOT/lib/libcrypto.$SONAME} \
+      -DLWS_WITH_LIBUV=ON                                           \
+      ${LIBUV_ROOT:+-DLIBUV_INCLUDE_DIRS=$LIBUV_ROOT/include}       \
+      ${LIBUV_ROOT:+-DLIBUV_LIBRARIES=$LIBUV_ROOT/lib}              \
       -DLWS_HAVE_OPENSSL_ECDH_H=OFF                                 \
       -DLWS_WITHOUT_TESTAPPS=ON
 cmake --build . --target install ${JOBS+-j $JOBS}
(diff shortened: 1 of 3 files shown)


libwebsockets.sh
```diff
@@ -36,6 +37,9 @@ cmake $SOURCEDIR                                                    \
       ${OPENSSL_ROOT:+-DOPENSSL_LIBRARIES=$OPENSSL_ROOT/lib/libssl.$SONAME;$OPENSSL_ROOT/lib/libcrypto.$SONAME}     \
       ${OPENSSL_ROOT:+-DLWS_OPENSSL_INCLUDE_DIRS=$OPENSSL_ROOT/include}                                             \
       ${OPENSSL_ROOT:+-DLWS_OPENSSL_LIBRARIES=$OPENSSL_ROOT/lib/libssl.$SONAME;$OPENSSL_ROOT/lib/libcrypto.$SONAME} \
+      -DLWS_WITH_LIBUV=ON                                           \
+      ${LIBUV_ROOT:+-DLIBUV_INCLUDE_DIRS=$LIBUV_ROOT/include}       \
```
ktf, line 41: Please, use `${LIBUV_REVISION:...}` rather than `${LIBUV_ROOT:...}`

ktf (changes requested):
See inline.

ktf (approved):

yuw726:
build/alidist/O2Suite/o2/macOS check fails because libwebsockets cannot link to libuv:
```
Searching for OpenSSL executable and dlls
OpenSSL executable: /opt/homebrew/opt/openssl@3/bin/openssl
 GENCERTS = 0
libuv include dir: /Volumes/build/alice-ci-workdir/alidist-o2/sw/osx_arm64/libuv/v1.40.0-local1/include
libuv libraries: /Volumes/build/alice-ci-workdir/alidist-o2/sw/osx_arm64/libuv/v1.40.0-local1/lib
-- Looking for uv-version.h
-- Looking for uv-version.h - not found
-- Looking for uv/version.h
-- Looking for uv/version.h - not found
DIR /Volumes/build/alice-ci-workdir/alidist-o2/sw/BUILD/85465dbd00be206b84a2f6b337cfe01c4a99036a/libwebsockets CMP /Volumes/build/alice-ci-workdir/alidist-o2/sw/SOURCES/libwebsockets/v4.3.2/v4.3.2/cmake
-- Configuring done (7.5s)
CMake Warning at lib/CMakeLists.txt:268 (target_link_libraries):
  Target "websockets" requests linking to directory
  "/Volumes/build/alice-ci-workdir/alidist-o2/sw/osx_arm64/libuv/v1.40.0-local1/lib".
  Targets may link only to libraries.  CMake is dropping the item.


-- Generating done (0.0s)
CMake Warning:
  Manually-specified variables were not used by the project:

    OPENSSL_ROOT_DIR

```

The error comes from https://github.com/warmcat/libwebsockets/blob/v4.3.2/lib/CMakeLists.txt#L268, where `target_link_libraries` tries to link to the libuv lib path instead of the library.

It is strange that it does work on other OSs.

ktf:
As discussed, the problem is that libwebsocket is an archive (.a) library, so it does not bring in automatically the transitive closure of its dependencies. Given in any case you use it, I would explicitly find LibUV and add a dependency for `LibUV::LibUV` in your `CMakeList.txt`.

yuw726:
Hi @ktf, will you have an opportunity to test these recipes together with `add-libuv-support` branch of https://gitlab.cern.ch/jalien/jalien-root/ on Hyperloop?

costing (approved):

zensanp (approved):

PR #5201, opened 2023-11-03
Update ThePEG tag

jackal1-66:
ThePEG modifications required for latest developments on Rivet in AliGenerators


maireiphc (approved):
Needed to improve better compatibility with "older" compilers.
const_cast set up by Christian Holm in 2 files .cc about rivet->analyse(HEPmc GenEvent)

PR #5202, opened 2023-11-03
Enable certain dependencies on RHEL/Alma 9

awegrzyn:


datadistribution.sh
```diff
@@ -13,7 +13,7 @@ requires:
   - protobuf
   - O2
   - fmt
-  - "ucx:(slc8)"
+  - "ucx:(slc8|slc9)"
```
ktf, line 16:
```suggestion
  - "ucx:(slc.*)"
```

mesos.sh
```diff
@@ -16,7 +16,7 @@ requires:
   # - curl
   # - OpenSSL
 build_requires:
-  - "autotools:(slc6|slc7|slc8)"
+  - "autotools:(slc6|slc7|slc8|slc9)"
```
ktf, line 19:
```suggestion
  - "autotools:(slc.*)"
```

o2suite.sh
```diff
@@ -12,7 +12,7 @@ requires:
   - QualityControl
   - "DataDistribution:(?!osx)"
   - "ALF:(?!osx)"
-  - "mesos:(slc8)"
+  - "mesos:(slc8|slc9)"
```
ktf, line 15:
```suggestion
  - "mesos:(slc.*)"
```

ktf (commented):
Any reason not to use a wildcard? In the end we do not support slc6 / slc7 anymore.

awegrzyn:
> Any reason not to use a wildcard? In the end we do not support slc6 / slc7 anymore.

Yes, of course, it makes more sense.

ktf (dismissed):

awegrzyn:
@TimoWilken It seems that `build/O2/alidist` check running at `alientest02`  is missing a system dependency which it: `rdma-core-devel`

TimoWilken:
Hi @awegrzyn, that check runs on slc7, so changing `slc8` to `slc.*` in `datadistribution.sh` means it also tries to build ucx on slc7, which it didn't before. Maybe for that one, it would be worth saying `slc[89]_` instead?

awegrzyn:
Ah ok, thanks for feedback, I thought all checks are already at `slc8`.

PR #5204, opened 2023-11-05
Attempt at compiling with latest ROOT

ktf:
Attempt at compiling with latest ROOT

### DIFF
--- a/geant3.sh
+++ b/geant3.sh
@@ -19,9 +19,16 @@ if [ $FVERSION -ge 10 ]; then
    echo "Fortran version $FVERSION"
    SPECIALFFLAGS=1
 fi
+case $ARCHITECTURE in
+  osx*) SONAME=dylib ;;
+  *) SONAME=so ;;
+esac
+
 cmake $SOURCEDIR -DCMAKE_INSTALL_PREFIX=$INSTALLROOT      \
                  -DCMAKE_BUILD_TYPE=$CMAKE_BUILD_TYPE     \
                  ${CXXSTD:+-DCMAKE_CXX_STANDARD=$CXXSTD}  \
+                 -DVDT_INCLUDE_DIR="$ROOT_ROOT/include"       \
+                 -DVDT_LIBRARY="$ROOT_ROOT/lib/libvdt.$SONAME"      \
                  -DCMAKE_SKIP_RPATH=TRUE \
                  ${SPECIALFFLAGS:+-DCMAKE_Fortran_FLAGS="-fallow-argument-mismatch -fallow-invalid-boz -fno-tree-loop-distribute-patterns"}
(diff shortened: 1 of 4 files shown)


geant3.sh
```diff
@@ -19,9 +19,16 @@ if [ $FVERSION -ge 10 ]; then
...
+case $ARCHITECTURE in
+  osx*) SONAME=dylib ;;
+  *) SONAME=so ;;
+esac
+
 cmake $SOURCEDIR -DCMAKE_INSTALL_PREFIX=$INSTALLROOT      \
                  -DCMAKE_BUILD_TYPE=$CMAKE_BUILD_TYPE     \
                  ${CXXSTD:+-DCMAKE_CXX_STANDARD=$CXXSTD}  \
+                 -DVDT_INCLUDE_DIR="$ROOT_ROOT/include"       \
+                 -DVDT_LIBRARY="$ROOT_ROOT/lib/libvdt.$SONAME"      \
```
ihrivnac, line 31: I don't see how this could help, as there is no usage of `vdt` inside VMC packages; if a Root library has this dependence, it should be defined in its target dependencies in the CMake Config generated by ROOT.

ktf: Apparently the new ROOT does not do that. Could you follow up with them?

ihrivnac: I can, but as I posted already in JIRA, I was not able to reproduce this `vdt` error on my side (outside aliBuild), ans asked you to provide me a log where this happens. In this PR, all tests passed ok.

ktf: To reproduce, the best is probably if you use alibuild. Given this PR is green, I would propose we merge it and then we clean it up later, if needed.

ihrivnac: I can reproduce the error with alibuild; but still I do not see what makes the difference between the standalone build, where the error does not happen. I will investigate more next week. 
It's ok to merge, as you say, we can clean-up later.

ihrivnac: The difference is now understood; in the standalone build, I usually build with ROOT and Geant4 environment set; and this added ROOT in the searched patch for `Vdt`. When building strictly with ROOT  given by CMake `ROOT_DIR` variable, I get the same error.

This can be fixed by the attached patch to root; with this patch the setting of the VDT variables is not needed. 
[patch_cmake_vdt.txt](https://github.com/alisw/alidist/files/13336344/patch_cmake_vdt.txt)

ktf:
I guess we could simply drop VDT if we are not using it, as @sawenzel suggested.

ihrivnac:
> I guess we could simply drop VDT if we are not using it, as @sawenzel suggested.

Yes, I tried it out locally, and O2 builds ok without VDT

ktf:
Superseded by #5214.

PR #5205, opened 2023-11-05
Use Ninja generator for ROOT

ktf:
Use Ninja generator for ROOT


ktf:
This is fine and it will be included by #5214.

PR #5211, opened 2023-11-09
Modernize gRPC recipe

ktf:
Modernize gRPC recipe

* Properly use alibuild-generate-recipe.
* Do not download packages out of band, since in any
  case we are not using them.

### DIFF
--- a/grpc.sh
+++ b/grpc.sh
@@ -17,12 +17,6 @@ incremental_recipe: |
   mkdir -p $INSTALLROOT/etc/modulefiles && rsync -a --delete etc/modulefiles/ $INSTALLROOT/etc/modulefiles
 ---
 #!/bin/bash -e
-
-pushd $SOURCEDIR
-git checkout "$GIT_TAG"
-git submodule update --init
-popd
-
 SONAME=so
 case $ARCHITECTURE in
   osx*)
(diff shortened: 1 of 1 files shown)


ktf:
@teo @awegrzyn any reason for the submodule init? I just tried and it seems to compile even without on my mac.

awegrzyn:
LGTM, submodules are only needed if dependency provider is set to `module` (which is not the case).

ktf:
Yes, that's what I meant. Let's see the build then and if @teo has no objections, I think we can merge this when green (and save us a bit of bandwidth and build time.

teo:
IIRC the submodule stuff was needed at some point in the distant past for one of the bundled dependencies. I'm ok with dropping it if it works now.

ktf:
Ok, let's see. Seems good so far.

ktf:
Mmm... they systematically fail with:
```
 1/28 Test  #1: o2-qc-test-core .........................***Timeout  30.03 sec
 7/28 Test #10: testTriggers ............................***Timeout  30.02 sec
11/28 Test #15: testCheckWorkflow .......................***Timeout  50.02 sec
```

ktf:
any idea?

awegrzyn:
@Barthelemy @knopers8 Could you have a look at: https://ali-ci.cern.ch/alice-build-logs/alisw/alidist/5211/d702dd883972a84fee66d2aa9be9b44397d9b0d5/build_O2_alidist-cs8/pretty.html ?

TimoWilken:
Hm, the checks are all running on bare-metal machines with local disks, but could loading the QC libraries still be a problem?

Barthelemy:
having a look

ktf:
Issue understood, it's the DNS config on some of the test machines which was not updated correctly, apparently.

ktf:
Ubuntu issue unrelated and irrelevant.

PR #5222, opened 2023-11-11
Bump FairMQ version to v1.8.1

ktf:
Bump FairMQ version to v1.8.1

### DIFF
--- a/fairmq.sh
+++ b/fairmq.sh
@@ -1,5 +1,5 @@
 package: FairMQ
-version: "v1.8.0"
+version: "v1.8.1"
 source: https://github.com/FairRootGroup/FairMQ
 requires:
   - boost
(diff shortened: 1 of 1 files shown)


ktf:
@davidrohr @teo This should fix the race condition at shutdown we see on the FLP.

davidrohr:
@Barthelemy @knopers8 : I see many QC tests timing out regularly, not only in this PR, but also in O2 PRs. Could you have a look what is the problem?

davidrohr:
ok, tested this at P2, seems OK, global state reaches READY

davidrohr:
ignore my last comment... commented on wrong PR...

TimoWilken:
Done @davidrohr

davidrohr:
errors unrelated, merging

PR #5171, opened 2023-10-06
Update Rivet (and YODA) builds with an eye toward Rivet integration into O2 pipeline 

cholmcc:
This will do the following updates 
- YODA and Rivet taken directly from upstream repositories.   The older `alisw` repositories are terribly out of date, and the local fixes are no longer needed. 
- YODA depends on HepMC3 to make sure we pull in the right ROOT version (HepMC3 depends on ROOT) when making Rivet 
- Rivet built against HepMC3 as needed for Rivet integration into O2 (see also https://gitlab.com/cholmcc/O2Rivet)

The changes have been checked and validated on `lxplus` albeit in a slightly funny way (because `aliBuild` doesn't play nice with CVMFS).  However, the tests are entirely equivalent to how `aliBuild` would do it from scratch. 

Yours,
_Christian_

PR #5223, opened 2023-11-13
Bump C++ standard to C++17

ktf:


ktf:
@jackal1-66 @maireiphc I need to bump the compilation standard for AliGenerators, since the new ROOT does not work anymore with C++14. Let me know if you have issues with it.

maireiphc:
Helllo @ktf (cc @jackal1-66, @cholmcc ), I suspect some (older) generators may not comply out of the box with the std c++17.
This is to be tried though.

To that end, we may need to first try out the compilation outcome and running time of the pull request
https://github.com/alisw/alidist/pull/5171 = Update Rivet (and YODA) builds with an eye toward Rivet integration into O2 pipeline

Just to avoid to pile troubles on troubles.
Giulio, can you wait for our green light before moving further with this ?

ktf:
Indeed, I merged it after I checked that the compilation actually works correctly.

cholmcc:
Hi, 

Off-hand I would think most EGs work with C++17 if they work with C++14.  Most of the changes between the two (see [here](https://en.cppreference.com/w/cpp/17)) that _are_ incompatible are unlikely to be used by EG code.   In fact, most EGs, as is true of most Physics code, uses antiquated constructs and ideoms and are more likely to conform to C++98 than any more modern standard like C++14.  

Yours,
_Christian_

PR #5227, opened 2023-11-14
Bump arrow to 14.0.1

ktf:
Bump arrow to 14.0.1


ktf:
Let's see if this fixes the issue on macOS...

PR #5228, opened 2023-11-14
Bump madgraph to v3.5.2 and install internal packages

mfasDa:
Install internal packages which are required  for
madgraph running at NLO
- Oneloop
- Collier
- Ninja
- QCDLoop
- RunningCoupling and for analysis
- MadAnalysis5

### DIFF
--- a/madgraph.sh
+++ b/madgraph.sh
@@ -1,16 +1,46 @@
 package: madgraph
 version: "%(tag_basename)s"
-tag: "v2.8.1"
+tag: "v3.5.2"
 source: https://github.com/alisw/MadGraph
 requires:
-  - "Python:slc.*"
-  - "Python-system:(?!slc.*)"
+  - "Python-modules:(?!osx)"
 build_requires:
   - alibuild-recipe-tools
 ---
 #!/bin/bash -e
 
-rsync -a --exclude='**/.git' --delete --delete-excluded $SOURCEDIR/ $INSTALLROOT/
+rsync -a --exclude='**/.git' --delete --delete-excluded "$SOURCEDIR/" "$BUILDDIR/"
+
+# install internal packages 
+cd "$BUILDDIR"
+cat << EOF >> install.dat
+install oneloop
+install ninja
+install collier
+install RunningCoupling
+install QCDLoop
+install MadAnalysis5
+EOF
+
+./bin/mg5_aMC install.dat
+
+# cleanup after build
+rm install.dat
+rm HEPTools/ninja/ninja_install.log 
+rm HEPTools/oneloop/oneloop_install.log
+rm HEPTools/collier/collier_install.log
+rm HEPTools/madanalysis5/madanalysis5_install.log
+rm -rf HEPTools/ninja/Ninja 
+rm -rf HEPTools/oneloop/OneLOop*
+rm -rf HEPTools/collier/COLLIER*
+find QCDLoop -mindepth 1 -maxdepth 1 -not -name include -not -name lib -not -name share -exec rm -rf {} \;
+rm *.tgz
+rm vendor/*tar.gz
+
+# change paths in conifguration file for package relocation
+sed -i.deleteme -e "s|$BUILDDIR|$INSTALLROOT|" input/mg5_configuration.txt
+rm -f input/mg5_configuration.deleteme
+rsync -a "$BUILDDIR/" "$INSTALLROOT/"
 
 #ModuleFile
 mkdir -p $INSTALLROOT/etc/modulefiles


mfasDa:
Bump needed as MadGraph5 v2.8.1 is not compatible with python 3.9

madgraph.sh
```diff
@@ -10,7 +10,41 @@ build_requires:
...
 #!/bin/bash -e
 
-rsync -a --exclude='**/.git' --delete --delete-excluded $SOURCEDIR/ $INSTALLROOT/
+rsync -a --exclude='**/.git' --delete --delete-excluded $SOURCEDIR/ $BUILDDIR/
+
+# install internal packages 
+cd $BUILDDIR
+cat << EOF >> $PWD/install.dat
+install oneloop
+install ninja
```
ktf, line 19: can't you reuse our own ninja? Or is this a separate ninja?

mfasDa: This is a different "ninja", it is not the build tool, it is a tool used in one-loop integration in NLO processes: https://ninja.hepforge.org. Without this OneLoop as loop tool (and consequently madgraph) does not work.

madgraph.sh
```diff
@@ -10,7 +10,41 @@ build_requires:
...
+rm HEPTools/ninja/ninja_install.log 
+rm HEPTools/oneloop/oneloop_install.log
+rm HEPTools/collier/collier_install.log
+rm HEPTools/madanalysis5/madanalysis5_install.log
+rm -rf HEPTools/ninja/Ninja 
+rm -rf HEPTools/oneloop/OneLOop*
+rm -rf HEPTools/collier/COLLIER*
+cd QCDLoop
+todelete=($(ls -1 | grep -v include | grep -v lib | grep -v share))
+for d in ${todelete[@]}; do rm -rf $d; done
```
TimoWilken, line 39: Instead of these three lines, I'd suggest:
```bash
find QCDLoop -mindepth 1 -maxdepth 1 -not -name include -not -name lib -not -name share -exec rm -rf {} +
```
That avoids parsing the output of `ls` and all the string-splitting hairness that comes with that.

madgraph.sh
```diff
@@ -10,7 +10,41 @@ build_requires:
 ---
 #!/bin/bash -e
 
-rsync -a --exclude='**/.git' --delete --delete-excluded $SOURCEDIR/ $INSTALLROOT/
+rsync -a --exclude='**/.git' --delete --delete-excluded $SOURCEDIR/ $BUILDDIR/
```
TimoWilken, line 13: Could you quote any variables you don't want to split? I.e., in this case, use `"$SOURCEDIR/"` and `"$BUILDDIR/"`. Same for other uses of variables below, but I won't mark every one.

TimoWilken (commented):
Thanks @mfasDa! A few comments on the Bash code below, but looks fine to me in principle.

madgraph.sh
mfasDa: OK, this I can change

madgraph.sh
```diff
@@ -10,7 +10,41 @@ build_requires:
...
+
+# cleanup after build
+rm  $PWD/install.dat
+rm HEPTools/ninja/ninja_install.log 
+rm HEPTools/oneloop/oneloop_install.log
+rm HEPTools/collier/collier_install.log
+rm HEPTools/madanalysis5/madanalysis5_install.log
+rm -rf HEPTools/ninja/Ninja 
+rm -rf HEPTools/oneloop/OneLOop*
+rm -rf HEPTools/collier/COLLIER*
```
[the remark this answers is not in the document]
mfasDa: These are the source directories of the dependencies which are downloaded and upacked in the install dir by the install tool.  I though in is not necessary to trash the install dir with sources - we also don't do that for other packages. For OneLOop and COLLIER it contains a version, which I would like to wildcard (therefore the *)

madgraph.sh
```diff
@@ -10,7 +10,41 @@ build_requires:
 ---
 #!/bin/bash -e
 
-rsync -a --exclude='**/.git' --delete --delete-excluded $SOURCEDIR/ $INSTALLROOT/
+rsync -a --exclude='**/.git' --delete --delete-excluded $SOURCEDIR/ $BUILDDIR/
+
+# install internal packages 
+cd $BUILDDIR
+cat << EOF >> $PWD/install.dat
```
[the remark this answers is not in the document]
mfasDa: Just wanted to be safe. Can change that.

madgraph.sh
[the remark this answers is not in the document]
TimoWilken: Makes sense, thank you!

madgraph.sh
```diff
@@ -10,7 +10,39 @@ build_requires:
 ---
 #!/bin/bash -e
 
-rsync -a --exclude='**/.git' --delete --delete-excluded $SOURCEDIR/ $INSTALLROOT/
+rsync -a --exclude='**/.git' --delete --delete-excluded "$SOURCEDIR/" "$BUILDDIR/"
+
+# install internal packages 
+cd $BUILDDIR
```
TimoWilken, line 16:
```bash
cd "$BUILDDIR"
```

madgraph.sh
```diff
@@ -10,7 +10,39 @@ build_requires:
...
+rm  $PWD/install.dat
+rm HEPTools/ninja/ninja_install.log 
+rm HEPTools/oneloop/oneloop_install.log
+rm HEPTools/collier/collier_install.log
+rm HEPTools/madanalysis5/madanalysis5_install.log
+rm -rf HEPTools/ninja/Ninja 
+rm -rf HEPTools/oneloop/OneLOop*
+rm -rf HEPTools/collier/COLLIER*
+find QCDLoop -mindepth 1 -maxdepth 1 -not -name include -not -name lib -not -name share -exec rm -rf {} \;
+cd $BUILDDIR
```
TimoWilken, line 38:
```bash
cd "$BUILDDIR"
```

madgraph.sh
```diff
@@ -10,7 +10,39 @@ build_requires:
...
+rm *.tgz
+rm vendor/*tar.gz
+
+# change paths in conifguration file for package relocation
+sed -i.deleteme -e "s|$BUILDDIR|$INSTALLROOT|" input/mg5_configuration.txt
+rm -f input/mg5_configuration.deleteme
+rsync -a $BUILDDIR/ $INSTALLROOT/
```
TimoWilken, line 45:
```bash
rsync -a "$BUILDDIR/" "$INSTALLROOT/"
```

TimoWilken (commented):
Hi @mfasDa, thank you very much. Sorry, a few more quoting issues that I didn't highlight before. That's it then, though.

madgraph.sh
mfasDa: Not needed any more, removed

madgraph.sh
mfasDa: Done

TimoWilken (approved):
Thank you @mfasDa, looks good to me now! Enabling auto-merge so that it's merged once CI is green (since the AliGenerators check should compile this).

TimoWilken:
Hm, the build fails with:

```
++ ./bin/mg5_aMC install.dat
madgraph requires the six module. The easiest way to install it is to run "python -m pip install six --user"
```

You should probably depend on `Python-modules` instead of `Python`/`Python-system`. We might also have to add six to `Python-modules-list` if it's not installed already...

mfasDa:
> Hm, the build fails with:
> 
> ```
> ++ ./bin/mg5_aMC install.dat
> madgraph requires the six module. The easiest way to install it is to run "python -m pip install six --user"
> ```
> 
> You should probably depend on `Python-modules` instead of `Python`/`Python-system`. We might also have to add six to `Python-modules-list` if it's not installed already...

Hi @TimoWilken, I change the build recipe as you suggested. Hope it works now.

mfasDa:
CI errors unrelated

mfasDa:
@TimoWilken  Do you think we can merge this now?

TimoWilken:
Merged, thanks @mfasDa!

PR #5229, opened 2023-11-14
Bump VecGeom to v1.2.6

benedikt-voelkel:
* update build options

* remove ROOT dependency

### DIFF
--- a/vecgeom.sh
+++ b/vecgeom.sh
@@ -1,52 +1,49 @@
 package: VecGeom
 version: "%(tag_basename)s"
-tag: 89a05d148cc708d4efc2e7b0eb6e2118d2610057
+tag: v1.2.6
 source: https://gitlab.cern.ch/VecGeom/VecGeom.git
 requires:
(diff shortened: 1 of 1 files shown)


vecgeom.sh
```diff
@@ -64,7 +60,7 @@ proc ModulesHelp { } {
 set version $PKGVERSION-@@PKGREVISION@$PKGHASH@@
 module-whatis "ALICE Modulefile for $PKGNAME $PKGVERSION-@@PKGREVISION@$PKGHASH@@"
 # Dependencies
-module load BASE/1.0 Vc/$VC_VERSION-$VC_REVISION ${ROOT_REVISION:+ROOT/$ROOT_VERSION-$ROOT_REVISION}
```
ktf, line 67: since you are cleaning up stuff, any reason for not using `alibuild-generate-module`?

benedikt-voelkel: good point, will do it

PR #5234, opened 2023-11-15
Enable HepMC3 support in SHERPA

mfasDa:


mfasDa:
I would maybe not yet drop the HepMC2 support entirely, for use cases which still rely on HepMC2.

mfasDa:
Hi @TimoWilken , could you please merge this as well?

PR #5235, opened 2023-11-15
Openloops: Use python3 in scons file

mfasDa:
Making compilation work on systems which
no longer provide python2


TimoWilken:
Hi @mfasDa, I think you're right, it's not worth worrying about Python 2 support. On the platforms where it's even installed, we compile our own version (3.9) anyway.

TimoWilken (approved):

mfasDa:
Fix is needed as compilation breaks on systems which no longer provide python2.  I would rather patch this one file during the build process than the openloops repository, as patching in the build recipe makes import of newer versions easier (i.e. would allow us to go to the repo of the vendor if there).

PR #5237, opened 2023-11-16
Cache Go artifacts in build directory

TimoWilken:
This avoids leaving the cache in the user's home directory, or trying to access potentially nonexistent paths in a build container.


ktf:
Currently broken. Merging.

PR #5238, opened 2023-11-16
Bump JETSCAPE to 3.1.1-alice3

ktf:
Bump JETSCAPE to 3.1.1-alice3

Needed to detect the new ROOT

### DIFF
--- a/jetscape.sh
+++ b/jetscape.sh
@@ -1,6 +1,6 @@
 package: JETSCAPE
 version: "%(tag_basename)s"
-tag: "v3.1.1-alice2"
+tag: "v3.1.1-alice3"
 source: https://github.com/alisw/JETSCAPE
 requires:
   - boost


maireiphc:
Hello @ktf,
 are you sure about this bump ?
I may miss sthg, but I cannot find currently the tag 3.1.1-alice3 under https://github.com/alisw/JETSCAPE ?

ktf:
I made a mistake and created a branch, rather than a tag. Unless strictly needed, I guess we can wait for the next bump to create a proper tag.

maireiphc:
OK...
Naïve question:
1. Will the compilation even work, i.e. aliBuild will fetch the proper git branch with the proper existing tag name: 
> tag: "v3.1.1-alice3"
> source: https://github.com/alisw/JETSCAPE

Spontaneously, I would expect that this will not work and thus will break AliGenerators from today on (Jetscape is entitled in the generator bundle), no ?

2. So a fix in the recipe would be needed anyhow to square the situation (?)
No problem to have AliGenerators unavailable for today (we have plenty of them with similar package versions) but I am wondering whether one can live in practice with Jetscape under such recipe.

ktf:
Tags and branches are treated almost the same. The only difference is that by using a branch, you will follow the tip of that branch, if additional commit are added.

PR #5239, opened 2023-11-16
Bump LHAPDF6 to v6.5.2

mfasDa:


mfasDa:
This is needed in order to run with more modern pdfsets (i.e. CT18, NNPDF4.0) which are not supported by LHAPDF v6.2.1

mfasDa:
v6.2.1 was patched in the download script. I had a look whether patches need to be back ported and it seems the download script is reworked significantly so the patches are no longer needed. Consequently there is currently no need for a patch branch and also not for dedicated ALICE tags.

mfasDa:
@ktf @TimoWilken please review!

TimoWilken:
Hi @mfasDa! If you're changing to an upstream tag, should we also change back to the upstream repo as `source:`?

maireiphc:
v6.5.2 is indeed available in our alisw/lhapdf. I have never push the update to the end with the update of the recipe (~1 year old), because at that time, compilation was calling for a rework of the recipe alidist/lhapdf.sh on the Python side.
Sthg in the spirit of what was needed for alidist/yoda.sh:

> requires:
>   - "Python:(?!osx)"
>   - "Python-modules:(?!osx)"
>   - "Python-system:(osx.*)"
> ...
> build_requires:
> - ..
>   - Python
>   - ...

1. @mfasDa : can you tell/check whether, indeed we better update the Python management part of the lhapdf.sh recipe and its sideways (https://github.com/alisw/alidist/blob/master/lhapdf5.sh, https://github.com/alisw/alidist/blob/master/lhapdf-pdfsets.sh)
2. No objection to point to the upstream source if one believes that it is possible without scratch.

mfasDa:
@maireiphc Concerning the python @TimoWilken might better answer here. To my understanding the changes in the lhapdf5 build recipe are related to macOS used from conda, and the need for Python-modules might be due to missing python module dependencies. Seems like for LHAPDF6 this is not the case, otherwise the CI would have complained.

ktf:
@maireiphc @jackal1-66 shall we merge this?

jackal1-66:
Considering the CI didn't complain I would try to merge it yes. LHAPDF is required for many old generators and interfaces, so hopefully this won't break them. In case a patch will be required.

jackal1-66 (approved):

PR #5240, opened 2023-11-16
Move everything to C++20

ktf:
Move everything to C++20

PR #5242, opened 2023-11-17
Bump boost to 1.83.0-alice1

ktf:
Bump boost to 1.83.0-alice1

Custom tag needed due to change in behaviour of boost::interprocess.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/alisw/alidist/pull/5242).
* #5240 (2 commits)
* __->__ #5242 (2 commits)

### DIFF
--- a/boost.sh
+++ b/boost.sh
@@ -1,6 +1,6 @@
 package: boost
-version: v1.75.0
-tag: v1.75.0
+version: v1.83.0-alice1
+tag: v1.83.0-alice1
 source: https://github.com/alisw/boost.git
 requires:
   - "GCC-Toolchain:(?!osx)"
@@ -54,7 +54,7 @@ fi
 
 TMPB2=$BUILDDIR/tmp-boost-build
 case $ARCHITECTURE in
-  osx*) TOOLSET=clang-darwin ;;
+  osx*) TOOLSET=clang ;;
   *) TOOLSET=gcc ;;
 esac
 
@@ -84,10 +84,8 @@ b2 -q                                            \
    --without-graph                               \
    --without-graph_parallel                      \
    --without-locale                              \
-   --without-math                                \
    --without-mpi                                 \
    ${BOOST_NO_PYTHON:+--without-python}          \
-   --without-wave                                \
    --debug-configuration                         \
    -sNO_ZSTD=1                                   \
    ${BZ2_ROOT:+-sBZIP2_INCLUDE="$BZ2_ROOT/include"}  \
--- a/jetscape.sh
+++ b/jetscape.sh
@@ -1,6 +1,6 @@
 package: JETSCAPE
 version: "%(tag_basename)s"
-tag: "v3.1.1-alice4"
+tag: "v3.1.1-alice5"
 source: https://github.com/alisw/JETSCAPE
 requires:
   - boost


ktf:
@dennisklein @TimoWilken I finally managed to reproduce the issue and I found a workaround in a reproducible way. Apparently this is due to the fact the builds run in docker container and `/dev/shm` inside behaves differently from the one outside the container. I can fix the issue by passing `-v /dev/shm:/dev/shm`.

@TimoWilken: did we change anything there?
@dennisklein: did you manage to understand what might be different? I see they changed a few things for FreeBSD "jails". Maybe that's related?

TimoWilken:
Hi @ktf, it hasn't changed recently as far as I know. For fullCI, we pass `--tmpfs=/dev/shm:rw,size=8g,mode=1777` to the docker invocation, but we don't for the other CI checks. I can add something similar for the other checks, since the fullCI one seems to work. Any idea how much SHM we might need? 8G seems like quite a bit.

ktf:
Reproducer for the issue:

```
#include <fairmq/Tools.h>
#include <fairmq/ProgOptions.h>
#include <fairmq/TransportFactory.h>

int main(int argc, char** argv) {
        size_t session{(size_t)getpid() * 1000 + 0};
        fair::mq::ProgOptions config;
        config.SetProperty<std::string>("session", std::to_string(session));

        auto factoryZMQ = fair::mq::TransportFactory::CreateTransportFactory("zeromq");
        auto factorySHM = fair::mq::TransportFactory::CreateTransportFactory("shmem");
}
```

dennisklein:
```console
❯ podman run -it --ipc=none fedora:latest df -h | grep shm
❯ podman run -it --ipc=private fedora:latest df -h | grep shm
shm              63M     0   63M   0% /dev/shm
❯ podman run -it --ipc=host fedora:latest df -h | grep shm
tmpfs            20G   84K   20G   1% /dev/shm
❯ podman run -it --shm-size=4G fedora:latest df -h | grep shm
shm             4.0G     0  4.0G   0% /dev/shm
```

It's not docker, but assuming podman behaves similarly (it shares also cli options with docker), looks like one can containerize processes without seeing any `/dev/shm`, which - to me - looks like whats happening?

ktf:
Yes, indeed, that's what I was about to say: old boost works just fine with the same setup, so the container options should not really matter.

ktf:
I think I found the issue. If I revert https://github.com/boostorg/interprocess/commit/87379be15fc01e5c8e7a54ada237f994ff9c1d56 running the test above in the container works fine.

ktf:
Reading a bit, I understand what is going on. They changed from `ftruncate` to `posix_fallocate`. The difference being that `posix_fallocate` will reserve memory in advance, while `ftruncate` will not. As a side effect this now kills all the checks for which we do not have the `--tmpfs` option that @TimoWilken mention.

For now I will patch boost to re-establish the old behavior, so that we can continue with the C++20 / Sonoma / ROOT 6.30.01 ports, however we should discuss what behaviour we actually need. At the moment we can be sloppy with the default shared memory allocation, because in the end if we do not use it, it will be fine. I guess that could be cleaned up, but I worry of the implications on the reported memory usage e.g. on the Grid.

PR #5243, opened 2023-11-17
Update clean-pr-checks.yml

ktf:


ktf:
@TimoWilken I think we were missing this.

PR #5246, opened 2023-11-21
Require lhapdf as direct dependency

mfasDa:


mfasDa:
@ktf @TimoWilken Necessary on the grid, as lhapdf-pdfsets only has lhapdf as build_requirement. Consequently I get a missing symbol complaint concerning LHAPDF in case I don't specify lhapdf manually in the jdl, which is of course not the desired behaviour.

TimoWilken:
Thanks @mfasDa, looks good to me! If you ping me when the checks are done, I'll merge it.

PR #5247, opened 2023-11-21
Update jetscape.sh to work with new boost

ktf:


ktf:
@maireiphc this should be a real tag.

maireiphc:
Héhé, ok noted. For reference, I just add a comment here to loop back : to be compared to previous hybrid situation in #5238 last week with former "tag" `v3.1.1-alice3`.

PR #5250, opened 2023-11-23
Drop unneeded packages on macOS

ktf:
Drop unneeded packages on macOS

* DataDistribution does not actually work / is needed
* ODC is not needed


ktf:
Trivial, merging.

PR #5253, opened 2023-11-24
alienpy 1.5.4 bugfix

adriansev:
* Due to wrong import ordering, stdout print log messages destined to logfile
	* It could confuse tools that do a scrapping of alienpy stdout

Secondary improvements: 
* Make X509_CERT_{DIR,FILE} requirements without fallback: if they are specified then it should be the only option to be used
* Relax the validity checking of CApath location, it only requires to have at least one pem file into it
* add `-report` option to `ccdb` command

### DIFF
--- a/xjalienfs.sh
+++ b/xjalienfs.sh
@@ -1,6 +1,6 @@
 package: xjalienfs
 version: "%(tag_basename)s"
-tag: "1.5.3"
+tag: "1.5.4"
 source: https://gitlab.cern.ch/jalien/xjalienfs.git
 requires:
   - "OpenSSL:(?!osx)"


costing (approved):

adriansev:
@TimoWilken Hi Timo! is there any problem with O2 test `o2sim_G3` ?
i see in alidist-ubuntu2204 the following:
```
## sw/BUILD/O2-latest/log
--
Running 1 test case...
matBudGeom24268_geometry.root does not exist. Will create it on the fly
[INFO] Loading geometry from matBudGeom24268_geometry-aligned.root
Error in <TFile::TFile>: file /sw/BUILD/f4169c16f147d28e15dee8a42bfba1e810b45a3f/O2/Detectors/Base/matBudGeom24268_geometry-aligned.root does not exist
[FATAL] Failed to open file matBudGeom24268_geometry-aligned.root
For later analysis we write a core dump to core_dump_24268
unknown location(0): [4;31;49mfatal error: in "MatBudLUT": unknown type[0;39;49m
/sw/SOURCES/O2/build_O2_alidist-ubuntu2204-ubuntu2204_x86-64/0/Detectors/Base/test/testMatBudLUT.cxx(31): [1;36;49mlast checkpoint[0;39;49m

515/518 Test #474: o2sim_G3 .............................................................................   Passed   89.78 sec
--

```

TimoWilken:
I'll have a look @adriansev! But ignore that check for now, it's not required to merge.

adriansev:
Good morning @TimoWilken ! Can this be merged?

adriansev:
@TimoWilken just so you know, i updated the tag with a minimal protection
see https://github.com/adriansev/jalien_py/commit/921195a5ca79ece00fbbabb848b05a32d1f89fc3

TimoWilken:
> @TimoWilken just so you know, i updated the tag with a minimal protection see [adriansev/jalien_py@921195a](https://github.com/adriansev/jalien_py/commit/921195a5ca79ece00fbbabb848b05a32d1f89fc3)

Hi @adriansev, are you sure that line does what you want? If `content` is non-empty, it returns `((type, content), None)` and I'm not sure that's really what you intend. If it's not, you should say `else (None, None)` at the end of the line instead of `else None, None`.

adriansev:
@TimoWilken yes you are right! the test for a file missing the md5 (the root cause of the fix) worked because i check the second argument for existence, so while the commit is not correct it worked as expected for a file missing the md5 value ... let me know if you want me to do a retag with the proper parentheses in place

TimoWilken:
I'd suggest making a new tag and sending another alidist PR. Alibuild doesn't like it much when existing tags change what commit they point to, I think. If you'd like, I can merge the new PR straight away, since it's a very small fix!

PR #5255, opened 2023-11-27
alienpy 1.5.5:: quick fix for files missing md5 hash in catalogue

adriansev:
@TimoWilken done :)


adriansev:
for reference: https://github.com/adriansev/jalien_py/compare/1.5.4...1.5.5

TimoWilken:
Thanks @adriansev!

For the record: the only change is a short fix for a problem in 1.5.4, (which was merged previously) so merging straight away.

PR #5262, opened 2023-11-29
Bump to monitoring 3.18.1

ktf:
Bump to monitoring 3.18.1

Required for C++20

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/alisw/alidist/pull/5262).
* #5240
* __->__ #5262

### DIFF
--- a/monitoring.sh
+++ b/monitoring.sh
@@ -1,6 +1,6 @@
 package: Monitoring
 version: "%(tag_basename)s"
-tag: v3.17.5
+tag: v3.18.1
 requires:
   - boost
   - "GCC-Toolchain:(?!osx)"


awegrzyn (approved):

ktf:
@awegrzyn apparently 3.18.0 forces all the dependencies to have -std=gnu++20, due to the change:

```
-target_compile_features(Monitoring PUBLIC cxx_std_17)
+target_compile_features(Monitoring PUBLIC cxx_std_20)
```

IMHO it would be better not to force a standard on dependencies, but simply pickup what the system provides (and complain if you do not support it). @Barthelemy I think this might explain your woes with C++20 as well.

awegrzyn:
Indeed, you are correct, I missed it that it's `PUBLIC`... I reverted it back to `cxx_std_17` as I'm using C++17 features in the interface...

PR #5271, opened 2023-11-30
Install OpenSSL's pkgconfig files

TimoWilken:
This should avoid interference from the system OpenSSL, since most distros install pkgconfig files for OpenSSL, and CMake may find those if we don't put ours in the appropriate path.

Draft for now to avoid slowing down the CI. If this is OK'd, I'll mark it as ready for review to start CI.

### DIFF
--- a/openssl.sh
+++ b/openssl.sh
@@ -21,15 +21,17 @@ build_requires:
   - zlib
   - alibuild-recipe-tools
   - "GCC-Toolchain:(?!osx)"
+prepend_path:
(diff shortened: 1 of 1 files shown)


TimoWilken:
@ktf: This is the OpenSSL pkg-config PR. The files were deleted in commit f1375ac710cda05ffc77c6c60c84693c71dd8704, but I can't find the reasoning for that.

TimoWilken:
Caching succeeded and CI error is unrelated; merging.

PR #5214, opened 2023-11-10
Bump ROOT to v6.30.01-alice2

ktf:
Bump ROOT to v6.30.01-alice2

PR #5276, opened 2023-12-05
Bump ROOT to v6-28-08-alice1

ktf:


sawenzel:
This doesn't do it. We need v6-28-08 according to ROOT team. Testing right now.

sawenzel:
I confirm that with v6-28-08 the issue with the StreamerInfos is fixed.

ktf:
PR updated.

sawenzel:
It appears that all builds fail now. Something was changed in the CMake (related to VDT?? https://github.com/root-project/root/issues/11797)

ktf:
It's probably the same issue as in https://github.com/alisw/alidist/pull/5214/files. Let me use that one, instead.

ktf:
Closing in favour of https://github.com/alisw/alidist/pull/5278.

PR #5277, opened 2023-12-05
bump aegis.sh

sawenzel:
fixes a segfault in QED event generation


sawenzel:
Tested locally. Needed for nightlies.

PR #5294, opened 2023-12-15
Strip libonnxruntime by default

ktf:
Strip libonnxruntime by default

From:

-rwxr-xr-x  1 root root  602M Nov 23 02:08 libonnxruntime.so.1.12.1

to:

-rwxr-xr-x  1 root root   12M Dec 15 09:07 libonnxruntime.so.1.12.1

### DIFF
--- a/fairlogger.sh
+++ b/fairlogger.sh
@@ -22,8 +22,8 @@ case $ARCHITECTURE in
   ;;
   *) ;;
 esac
-
 mkdir -p $INSTALLROOT
+
 
 cmake $SOURCEDIR                                                 \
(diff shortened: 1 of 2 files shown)


ktf:
@TimoWilken how difficult do you think it would be to add automatic debug symbol stripping from aliBuild? I.e. we build the normal tarball and one where we put the symbols removed via objdump. The two tarballs could be uploaded in parallel, but the symbol one is only downloaded upon request. This would also allow us to avoid deploying symbols, unless requested.

TimoWilken:
@ktf Tracked in <https://alice.its.cern.ch/jira/browse/O2-4512>!

PR #5295, opened 2023-12-16
Add MachineLearningHEP

qgp:


machinelearninghep.sh
```diff
@@ -0,0 +1,44 @@
...
+env -u VIRTUAL_ENV ALIBUILD=1 \
+    python3 -m pip install --force-reinstall \
+    --target="$INSTALLROOT/lib/python/site-packages" \
+    ${PIPOPTIONS} "file://$SOURCEDIR"
+
+# rm -rf "${INSTALLROOT:?}/bin"
+mv "$INSTALLROOT/lib/python/site-packages/bin" "$INSTALLROOT/bin"
+
+# Modulefile
+MODULEDIR="$INSTALLROOT/etc/modulefiles"
```
ktf, line 26: You can probably use `alibuild-generate-module --bin` (needs `alibuild-recipe-tools` as `build_requires`) to generate the modulefile (and then append to it the missing parts like PYTHON_PATH).

ktf:
Nice. If it works, LGTM.

qgp:
Thanks @ktf, updated with the recipe for the module file. 

Now, I still see the following points to be addressed:
- Dependencies get installed as `site-packages` of `MachineLearningHEP` and, thus, separately for each installation/build of `MachineLearningHEP` (NB: 2.6 GB!). Is this really what we want? Or should dependencies be installed to some common place?
- With a development package, rerunning `aliBuild build MachineLearningHEP` creates a new build every time. I assume this can be avoided with an incremental recipe.
- `aliBuild init MachineLearningHEP` checks out the tag listed in the recipe, is there a way to make it use the current HEAD of a given branch (other than specifying `MachineLearningHEP@run3` on the command line)?

NB: In parallel, I am transitioning `MachineLearningHEP` from `setup.py` to `pyproject.toml` (alisw/MachineLearningHEP#844).

ktf:
> Dependencies get installed as site-packages of MachineLearningHEP and, thus, separately for each installation/build of MachineLearningHEP (NB: 2.6 GB!). Is this really what we want? Or should dependencies be installed to some common place?

No, we do not want that, but at the moment that's the only viable solution. PIP does not support incremental layers last time I checked.

> With a development package, rerunning aliBuild build MachineLearningHEP creates a new build every time. I assume this can be avoided with an incremental recipe.

Yes, I think that's what does the trick. Also make sure you get avoid byproducts in the source folder or you have a .gitignore which ignores them. Any untracked file in the sources will trigger a rebuild.

>  aliBuild init MachineLearningHEP checks out the tag listed in the recipe, is there a way to make it use the current HEAD of a given branch (other than specifying MachineLearningHEP@run3 on the command line)?

Can't you simply use "run3" as a "tag" in the recipe?

qgp:
> No, we do not want that, but at the moment that's the only viable solution. PIP does not support incremental layers last time I checked.
Ok, then we just have to avoid too many separate builds.
 
> Yes, I think that's what does the trick. Also make sure you get avoid byproducts in the source folder or you have a .gitignore which ignores them. Any untracked file in the sources will trigger a rebuild.
Ok, I have added the incremental recipe, that at least avoids the proliferation of separate builds.

> Can't you simply use "run3" as a "tag" in the recipe?
If that does not violate any policy of not using non-fixed versions, that's fine by me - at least as long as we do not have meaningful tags that we'd want people to use.

ktf:
> If that does not violate any policy of not using non-fixed versions, that's fine by me - at least as long as we do not have meaningful tags that we'd want people to use.

As long as this is not a dependency of O2, I have no objections.

qgp:
> As long as this is not a dependency of O2, I have no objections.

It is not.

ktf:
Sorry, it got lost in the Christmas holidays...

PR #5298, opened 2023-12-18
Enable Python support again

ktf:
Enable Python support again

The folder for llvm moved in ROOT v6.30.XX and therefore aliBuild
things we are dealing with ROOT5. Dropping support for ROOT5 should bring
back the PyROOT support.


pzhristov (approved):
I was about to commit the same fix for the PyROOT issue. All my tests of AliRoot/AliPhysics were OK.

PR #5299, opened 2023-12-19
Revert to v6-28-08-alice1

ktf:


ktf:
Not needed anymore since we managed to fix JAlien-ROOT (needs 0.7.4.1).

PR #5307, opened 2024-01-11
Use O2DPG dev directory

benedikt-voelkel:


benedikt-voelkel:
@TimoWilken This will probably fail here since I assume that O2DPG will not be there as a dev package?

TimoWilken:
The o2dpg-sim-tests isn't built at all for PRs in alidist, so merging!

PR #5309, opened 2024-01-12
Bump GEANT4 to v11.2.0

martenole:
Compiling GEANT4 v11.0.4 with gcc 13.2 fails because of missing `cstdint` includes which is fixed in v11.2.0. 

Not sure if anything speaks agains this bump?

### DIFF
--- a/geant4.sh
+++ b/geant4.sh
@@ -1,6 +1,6 @@
 package: GEANT4
 version: "%(tag_basename)s"
-tag: "v11.0.4"
+tag: "v11.2.0"
 # source: https://github.com/alisw/geant4.git
 source: https://gitlab.cern.ch/geant4/geant4.git
 requires:
--- a/geant4_vmc.sh
+++ b/geant4_vmc.sh
@@ -1,6 +1,6 @@
 package: GEANT4_VMC
 version: "%(tag_basename)s"
-tag: "v6-1-update1-p9"
+tag: "v6-5"
 source: https://github.com/vmc-project/geant4_vmc
 requires:
   - ROOT
--- a/vgm.sh
+++ b/vgm.sh
@@ -1,6 +1,6 @@
 package: vgm
 version: "%(tag_basename)s"
-tag: "v5-0"
+tag: "v5-2"
 source: https://github.com/vmc-project/vgm
 requires:
   - ROOT


benedikt-voelkel:
We are currently validating a bump in general. It would come with a few more changes in fact.
It will be discussed in next week's WP12/13 meeting.

ktf:
It also looks in needs a news VMC...

martenole:
Hm, ok. I only tried manually fixing 11.0.4 and that worked fine. Indeed for v11.2.0 at least vgm and VMC need to be patched :/

ihrivnac:
You can fins the co-working version of VMC packages at this page:
https://vmc-project.github.io/download/tar-files/

Geant4 11.2 requires VGM 5.2 and Geant4 VMC 6.5.

benedikt-voelkel:
The errors in FST do not seem to be related to this PR.
After the discussion we had in the WP1§2/13, in principle, it should be fine to merge, @sawenzel @ihrivnac ?

sawenzel (approved):

pzhristov:
@martenole  @benedikt-voelkel @ihrivnac @sawenzel @ktf This PR breaks the simulation in AliRoot. The medium BODY_Vacuum is not found in GEANT4_VMC/source/run/src/TG4RegionsManager2.cxx:164 (even if it was correctly created), and then the unprotected null pointer is used on line 166 which leads to segmentation violation. I propose to revert these changes until the underlying issue is fixed.

benedikt-voelkel:
@pzhristov thanks for checking.
We are using that version and now rely a lot on it since in particular we use the much more precise handling of transport and production cuts.
In addition it allows to use the requested feature of scaling cross sections.
So I am in favour of keeping it since it will affect current Run3 productions and would lead to a degradation.

sawenzel:
@pzhristov : Couldn't we just revert Geant4 for AliRoot, using the defaults mechanism (we could fix the version of G4 for Run2)? I think we shouldn't let Run2 problems affect Run3 developments.

pzhristov:
I suspect this is not only Run2 problem and the fix has to be done in GEANT4_VMC

benedikt-voelkel:
That is likely true. However, since we don't have the problem yet in Run3 productions, I agree with @sawenzel to keep this in and add the fix and after that, **bump** Geant4_VMC instead of moving back now.
If it is a protection agains a `nullptr`, could there be a quick fix @ihrivnac ?

pzhristov:
Ok, if we rely on the current version, we need a fix. I do not want to have different packages for AliRoot/AliPhysics and O2. It is not only the protection, but probably the treatment of the names that has to be checked.

pzhristov:
It seems that the call to BuildPhysics() is done in a wrong moment. I suppose in O2 this is managed in FairRoot, would you confirm?

benedikt-voelkel:
Yes, in O2, the call to `TVirtualMC::BuildPhysics()` is done in FairRoot's `FairMCApplication::InitMC()`.
For completeness: We are also not overriding `FairMCApplication::InitMC()`

ihrivnac:
@pzhristov ,

The new method of applying cuts can be disabled by `TG4RunConfiguration::SetSpecialCutsOld()`. This function should be called in `g4Config.C`, eg.:
```c++
  // Create Geant4 VMC
  //  
  TGeant4 *geant4 = 0;
  if ( ! gMC ) {
    TG4RunConfiguration* runConfiguration 
      = new TG4RunConfiguration("geomRoot", 
                                "FTFP_BERT_EMV+optical", 
                                "specialCuts+stackPopper+stepLimiter",
                                 true, false);
    runConfiguration->SetSpecialCutsOld();  // ADDED

    geant4 = new TGeant4("TGeant4", 
                         "The Geant4 Monte Carlo : FTFP_BERT_EMV+optical", 
                         runConfiguration);
             // Repeat physics selection in the title; to be removed
             // with new geant4_vmc tag (1.13)            

    cout << "Geant4 has been created." << endl;
  } 
```

Can you try this if it helps ?

pzhristov:
@ihrivnac Hi Ivana, thanks a lot, this additional call solved the problem!

PR #5310, opened 2024-01-12
Bump flatbuffers to v23.5.26

martenole:
To allow for compilation with gcc 13.2 in case nothing speaks against this?

### DIFF
--- a/flatbuffers.sh
+++ b/flatbuffers.sh
@@ -1,5 +1,5 @@
 package: flatbuffers
-version: v1.12.0
+version: v23.5.26
 source: https://github.com/google/flatbuffers
 requires:
   - zlib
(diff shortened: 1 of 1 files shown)


martenole:
Hm, I had built only what I needed to make O2 work. Trying with ONNX seems not to work, even if I try to bump ONNX to its latest version as well. In `ONNXRuntime/cmake/deps.txt` its still specified that flatbuffers v1.12 is used for compilation. But that doesn't compile with gcc 13. So not sure if its possible to compile our software stack with gcc 13 at the moment including ONNX

davidrohr:
For reference, I had issues with flatbuffers in ONNXruntime on my laptop, since its version 1.12 collided somehow with my system flatbuffer version. I tried if I could bump ONNXruntime's flatbuffer, but it is not foreseen.
Searching a bit in the web, some other people complained, but Microsoft decided after internal discussion not to bump flatbuffer C++ for now: https://github.com/microsoft/onnxruntime/pull/10866

Though that was 2022, but apparently still didn't change...
For myself, I have not found a solution. I could not get ONNXruntime working, so I disabled it for me in alidist...

davidrohr:
superseded by other PRs

PR #5311, opened 2024-01-12
Update apmon-cpp.sh

costing:
Bump version to include the compilation fix from Ole

### DIFF
--- a/apmon-cpp.sh
+++ b/apmon-cpp.sh
@@ -1,6 +1,6 @@
 package: ApMon-CPP
 version: "%(tag_basename)s"
-tag: v2.2.8-alice5
+tag: v2.2.8-alice6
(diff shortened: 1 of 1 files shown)


TimoWilken:
Fine by me in principle, but I expect that once merged, this will need manual patching again on CVMFS, so that it works on @jgrosseo's Ubuntu hosts.

davidrohr:
@costing @jgrosseo : Can we merge this? It only adds a fix to include a missing header, so I don't understand why this would require some manual patching, but if @jgrosseo can comment?

jgrosseo:
I think it is needed for each new release. In the past @TimoWilken was doing this but I guess @ktf can do it as well. He (and me) are away this week, though...

PR #5325, opened 2024-01-24
bump infologger

sy-c:


vascobarroso (dismissed):

TimoWilken:
Hi, the `v2.6.0` tag for InfoLogger does not seem to exist, which is why the CI fails.

sy-c:
Hello @TimoWilken, it seems all tests passed but the "build/O2/alidist" which failed apparently without any related message. Can you please check and merge if it is ok ?

PR #5344, opened 2024-02-02
Do not check for member names

ktf:
Do not check for member names

Somehow this check was not fully working before and it's now complaining
about actual violation.

It is deemed however too complicated to fix them at this point in time.


ktf:
@shahor02 as discussed.

ktf:
Currently broken, merging.

PR #5346, opened 2024-02-05
Add GMP requirement to fastjet

martenole:
Not sure if this is really correct, or if instead the `module load GMP` part should be removed in the fastjet recipe. But we should find a fix quickly so opening this now for discussion @ktf

### DIFF
--- a/fastjet.sh
+++ b/fastjet.sh
@@ -4,6 +4,7 @@ tag: "v3.2.1_1.024-alice3"
 source: https://github.com/alisw/fastjet
 requires:
   - cgal
+  - GMP
 env:
   FASTJET: "$FASTJET_ROOT"
 ---


ktf:
I guess this is the correct solution under the assumption that now GMP is needed at runtime.

@cholmcc can you remind me why you had to add the dependency in the first place?

cholmcc:
Sorry for the late answer - vacation time here in DK. 

The dependency was added because `GMP` is needed - both run- and build-time - by `fastjet`.  If it is _not_ set as a dependency, then the generated `fastjet-config` script will have the wrong path to the `GMP` shared library, and this will then be picked up by subsequent packages (e.g., Rivet, O2, and so on) which can lead to run-time errors because `libGMP.so` cannot be found.    The crux is ofcourse that `fastjet` is dynamically linked against `libGMP.so`. 

Yours,
_Christian_

ktf:
we build GMP statically on linux, probably as an historical trick to avoid
worrying about the system version. why would that require a dynamic
dependency?

Ciao,
Giulio

PR #5352, opened 2024-02-13
Don't check namespace names

TimoWilken:
Analogously to https://github.com/alisw/alidist/pull/5344. Spurious `aliceO2-namespace-naming` errors seen e.g. here: https://github.com/AliceO2Group/AliceO2/pull/12682.

Assuming the same reasoning applies as for #5344!


TimoWilken:
Errors unrelated; merging. @ktf please let me know if you prefer to have the errors fixed and reenable the check!

PR #5353, opened 2024-02-13
Fix build with o2 defaults on slc9

TimoWilken:
If we're going to build analysis tags on slc9, we can't assume that building on slc9 implies o2-dataflow defaults (or an FLP/EPN environment).

Move FLP/EPN-specific requirement into defaults-o2-dataflow and defaults-o2-epn.

### DIFF
--- a/datadistribution.sh
+++ b/datadistribution.sh
@@ -13,7 +13,6 @@ requires:
   - protobuf
   - O2
   - fmt
-  - "ucx:(slc8|slc9)"
 build_requires:
   - CMake
 source: https://github.com/AliceO2Group/DataDistribution
--- a/defaults-o2-dataflow.sh
+++ b/defaults-o2-dataflow.sh
@@ -51,6 +51,20 @@ overrides:
       BUILD_ANALYSIS: "OFF"   # Disable analysis in O2
       BUILD_EXAMPLES: "OFF"   # Disable examples in O2
       O2_BUILD_FOR_FLP: "ON"
+  DataDistribution:
+    requires:
+      - "GCC-Toolchain:(?!osx)"
+      - boost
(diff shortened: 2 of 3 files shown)


TimoWilken:
Hi @alisw/epn-experts @alisw/dataflow-experts, any objections?

TimoWilken:
CI error unrelated; merged.

Barthelemy:
@TimoWilken @ktf Our nightly build and test is failing due to : 
```
[root@ansible-ci-cs8-6 ~]# ldd /opt/o2/bin/StfSender
	linux-vdso.so.1 (0x00007ffeaa8ba000)
	libdl.so.2 => /lib64/libdl.so.2 (0x00007ff04206c000)
	libucp.so.0 => not found
	libuct.so.0 => not found
	libucm.so.0 => not found
	libucs.so.0 => not found
```
 Any idea ?

TimoWilken:
Which aliBuild --defaults are you using @Barthelemy?

Barthelemy:
`o2-dataflow`
I am now wondering if the problem is not when generating the RPM. Not sure.

TimoWilken:
If you built locally, does the binary from the build area still work? (`~/alice/sw/slc8_x86-64/O2/latest/bin/StfSender` or something similar)

I'll try to reproduce locally!

Barthelemy:
I am building

Barthelemy:
I think that the RPM generation does not take into account the defaults...

TimoWilken:
I see! Perhaps you could build and install the RPM for the ucx package separately as a workaround?

Barthelemy:
We removed this commit from the release branch. I am trying to update the RPM generation tool. Thank you for your help

PR #5354, opened 2024-02-13
Don't copy everything coming from python in the modules

Barthelemy:


Barthelemy:
Hi Timo, 
we are having issues to build RPMs. 
1. there is a symlink from lib64 to lib which causes difficulties to fpm (but ok, this we could handle)
2. there are files in the python-modules that are copies of the files provided by system rpms: `file /usr/lib/.build-id/d9/306cdb856f29810c5f6262b96d7fa5b807e3bb from install of o2-Python-modules-1.0-6.slc8.x86_64 conflicts with file from package python39-3.9.18-1.module+el8.9.0+20024+793d7211.x86_64`

And when I look at what is there I see for example : 
```
[alibuild@AIDO2QC10 Python-modules]$ ll 1.0-local7
total 12
drwxrwxr-x. 2 alibuild alibuild 4096 Feb 13 08:30 bin
drwxrwxr-x. 4 alibuild alibuild   42 Feb 13 08:30 etc
drwxrwxr-x. 2 alibuild alibuild    6 Feb 13 08:30 include
drwxrwxr-x. 3 alibuild alibuild   37 Feb 13 08:30 lib
lrwxrwxrwx. 1 alibuild alibuild    3 Feb 13 08:30 lib64 -> lib
-rw-rw-r--. 1 alibuild alibuild   70 Feb 13 08:30 pyvenv.cfg
-rw-rw-r--. 1 alibuild alibuild  866 Feb 13 08:30 relocate-me.sh
[alibuild@AIDO2QC10 Python-modules]$ ll 1.0-local7/bin
total 84
-rw-r--r--. 1 alibuild alibuild 1936 Feb 13 08:30 activate
-rw-r--r--. 1 alibuild alibuild  885 Feb 13 08:30 activate.csh
-rw-r--r--. 1 alibuild alibuild  938 Feb 13 08:30 activate.csh.unrelocated
-rw-r--r--. 1 alibuild alibuild 2025 Feb 13 08:30 activate.fish
-rw-r--r--. 1 alibuild alibuild 2078 Feb 13 08:30 activate.fish.unrelocated
-rw-r--r--. 1 alibuild alibuild 8834 Feb 13 08:30 Activate.ps1
-rw-r--r--. 1 alibuild alibuild 1989 Feb 13 08:30 activate.unrelocated
-rwxrwxr-x. 1 alibuild alibuild  234 Feb 13 08:30 easy_install
-rwxrwxr-x. 1 alibuild alibuild  234 Feb 13 08:30 easy_install-3.9
-rwxrwxr-x. 1 alibuild alibuild  225 Feb 13 08:30 pip
-rwxrwxr-x. 1 alibuild alibuild  225 Feb 13 08:30 pip3
-rwxrwxr-x. 1 alibuild alibuild  225 Feb 13 08:30 pip3.9
-rwxr-xr-x. 1 alibuild alibuild 7768 Feb 13 08:30 python
-rwxr-xr-x. 1 alibuild alibuild 7768 Feb 13 08:30 python3
-rwxr-xr-x. 1 alibuild alibuild 7768 Feb 13 08:30 python3.9
-rwxrwxr-x. 1 alibuild alibuild  212 Feb 13 08:30 wheel
```
all these python are real binaries, not symlinks. 

@ktf advised to remove the `--copies` and indeed it seems to help.

ktf:
@TimoWilken and myself discussed a bit. Maybe you can avoid the issue with 

https://stackoverflow.com/questions/60673229/fpm-v-1-11-0-vs-inclusion-of-usr-lib-build-id-files

? The while it's true that the two python are the same, they are in different paths, so there is no risk of overlapping things, no?

Barthelemy:
FYI, using this branch I am able to build the RPMs. 

I will try with the options you mention, however I am a bit fearful of the consequences. Adam is not here for a bit of time and I have little experience. 
For me to understand, is using `--copies` necessary ? what is the drawback of removing it ?

TimoWilken:
The drawback of removing `--copies` is that, sometimes, randomly, pip commands will start failing or silently doing the wrong thing.

The issue is that, if the venv is built with symlinks, then you have an absolute symlink to whatever build path your `python3` binary happens to currently be in. If you delete that build path, or if the final Python-modules package is installed on a different system to where it was built, you'll have a broken link instead.

This might happen to work out fine some of the time, but it'll silently break venv semantics -- so if we deploy this in alidist and users keep installing packages using pip (intending them to go into the venv), those packages will end up somewhere else on their system.

Barthelemy:
I see. 

I will try out what you proposed. 

Otherwise, could we put something in the defaults for `o2-dataflow` ? so that we remove `--copies` just for us  ?

Barthelemy:
I think that we have a workaround on our side. I will revisit it when Adam is back.

PR #5359, opened 2024-02-15
xjalienfs 1.5.8

adriansev:
[Diff 1.5.7..1.5.8](https://github.com/adriansev/jalien_py/compare/1.5.7...1.5.8)

Main changes
* act on ALIEN_SITE env var
* XRootD:: cp : always send md5 when asking for write(upload) access
* XRootD:: cp : for writing(upload) use the used md5 from access request to inform xrootd to use it as pre-computed checksum and verify the uploaded file

### DIFF
--- a/xjalienfs.sh
+++ b/xjalienfs.sh
@@ -1,6 +1,6 @@
 package: xjalienfs
 version: "%(tag_basename)s"
-tag: "1.5.7"
+tag: "1.5.8"
 source: https://gitlab.cern.ch/jalien/xjalienfs.git
 requires:
   - "OpenSSL:(?!osx)"


costing (approved):

adriansev:
@TimoWilken 1 test failed from build/O2/alidist-ubuntu2204
```
18/28 Test #23: batch_test ..............................***Failed   19.24 sec
....
The histogram of the QC Task has less than 225 (75%) of expected samples
```

the test execution would seem fine only the output it would seem invalid but for certain this have no relation to alienpy

Is there any chance that at some point to reduce the alienpy tests to it's own tests? Is really pointless to run all O2 tests for it...

PR #5361, opened 2024-02-16
Run xjalienfs's self-tests after installation

TimoWilken:
In CI, we have the `JALIEN_TOKEN_{CERT,KEY}` variables, so we could use them to run `alien.py`'s self-tests. These are quick and would only be run when xjalienfs is rebuit, i.e. when it is updated.

We should probably add another CI variable (analogous to `ALIBUILD_O2_TESTS` to enable this, instead of detecting the `JALIEN_TOKEN_*` variables, since users might set those. We don't want to accidentally run these tests with users' tokens, since the self-tests create and delete various test files and directories in AliEn.

### DIFF
--- a/xjalienfs.sh
+++ b/xjalienfs.sh
@@ -43,9 +43,11 @@ rm -fv "$INSTALLROOT"/bin/*.bak
 
 # Now that alien.py is installed, we can run its tests. They need a JAliEn
 # token though, so skip them if we have none.
+set +x   # avoid echoing tokens
 if [ -n "$JALIEN_TOKEN_CERT" ] && [ -n "$JALIEN_TOKEN_KEY" ]; then
   PATH="$INSTALLROOT/bin:$PATH" "$SOURCEDIR/tests/run_tests" ci-tests
 fi
+set -x
 
 # Modulefile
 mkdir -p "$INSTALLROOT/etc/modulefiles"
--- a/xjalienfs.sh
+++ b/xjalienfs.sh
@@ -44,7 +44,7 @@ rm -fv "$INSTALLROOT"/bin/*.bak
 # Now that alien.py is installed, we can run its tests. They need a JAliEn
 # token though, so skip them if we have none.
 if [ -n "$JALIEN_TOKEN_CERT" ] && [ -n "$JALIEN_TOKEN_KEY" ]; then
-  PATH="$INSTALLROOT/bin:$PATH" "$SOURCEDIR/tests/run_tests"
+  PATH="$INSTALLROOT/bin:$PATH" "$SOURCEDIR/tests/run_tests" ci-tests
 fi
 
 # Modulefile
--- a/xjalienfs.sh
+++ b/xjalienfs.sh
@@ -1,6 +1,6 @@
 package: xjalienfs
 version: "%(tag_basename)s"
-tag: "1.5.8"
+tag: "1.5.9"
 source: https://gitlab.cern.ch/jalien/xjalienfs.git
 requires:
   - "OpenSSL:(?!osx)"
--- a/xjalienfs.sh
+++ b/xjalienfs.sh
@@ -41,10 +41,13 @@ for binfile in "$INSTALLROOT"/bin/*; do
 done
 rm -fv "$INSTALLROOT"/bin/*.bak
 
-# Now that alien.py is installed, we can run its tests. They need a JAliEn
-# token though, so skip them if we have none.
+# Now that alien.py is installed, we can run its tests.
 set +x   # avoid echoing tokens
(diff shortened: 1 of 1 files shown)


adriansev:
Hi @TimoWilken !!! Thanks a lot for this! So, the copy tests are done in the home directory of user that is running the tests... the only problem that could happen would be if the test is interrupted and the workdir of the tests is not cleaned up .. moreover i realize now that multiple parallel tests by the same user could be problematic

adriansev:
I will work a little bit more on the tests to adapt them to CI usage ... 
For the copy tests would be ok to "hide" the output in logfiles that would/could be inspected if something fail?

xjalienfs.sh
```diff
@@ -41,6 +41,12 @@ for binfile in "$INSTALLROOT"/bin/*; do
 done
 rm -fv "$INSTALLROOT"/bin/*.bak
 
+# Now that alien.py is installed, we can run its tests. They need a JAliEn
+# token though, so skip them if we have none.
+if [ -n "$JALIEN_TOKEN_CERT" ] && [ -n "$JALIEN_TOKEN_KEY" ]; then
+  PATH="$INSTALLROOT/bin:$PATH" "$SOURCEDIR/tests/run_tests"
```
adriansev, line 47: Could you please use `"$SOURCEDIR/tests/run_tests ci-tests"` ?
I'm adding now this argument for running in CI environment ..

TimoWilken: Done, thanks! Though the CI won't pick it up yet because it's still using v1.5.8.

TimoWilken:
> I will work a little bit more on the tests to adapt them to CI usage ... For the copy tests would be ok to "hide" the output in logfiles that would/could be inspected if something fail?

If the output isn't sensitive, it would be better to show it on stderr/stdout -- then people won't have to go digging on the build machines manually any times something fails. And the CI builders run aggressive cleanup routines, so the logs would likely be lost immediately anyway.

xjalienfs.sh
adriansev: yes, i will tag a new version with only the tests when i'm done

adriansev:
> > I will work a little bit more on the tests to adapt them to CI usage ... For the copy tests would be ok to "hide" the output in logfiles that would/could be inspected if something fail?
> 
> If the output isn't sensitive, it would be better to show it on stderr/stdout -- then people won't have to go digging on the build machines manually any time something fails. And the CI builders run aggressive cleanup routines, so the logs would likely be lost immediately anyway.

well, it's not sensitive but could be exhaustive (especially that DEBUG is enabled) .. it's fine with me either way ... so, should i keep the messages on un-redirected?

TimoWilken:
> well, it's not sensitive but could be exhaustive (especially that DEBUG is enabled) .. it's fine with me either way ... so, should i keep the messages on un-redirected?

As long as it's not many megabytes of text, yes, let's keep it in stderr/stdout!

Alternatively, if you want, you could write to a log file and only `cat` that log file if the test fails -- sort of like what `ctest` does for the O2 and O2Physics unit tests.

adriansev (approved):

adriansev:
@TimoWilken I rewrite the tests a little bit in the spirit of what we talked .. could you please take a look (of course when you have time next week) and if everything looks ok with you i will tag a release.
Thanks a lot!!

TimoWilken:
Hi @adriansev, thank you! The only remaining issue I think is that `tests/run_tests` still exits with code 0, even if a test failed -- so the CI would be green, even if tests fail!

It also aborts early if a test fails instead of running the rest of them, but I suppose that's a matter of taste -- up to you :)

adriansev:
@TimoWilken Fixed :) can you please take a look again at the tests? The final exit code will be the number of failed tests, and for `ci-tests` argument all tests will be tried even if one fails ..

TimoWilken:
Looks good to me now, thanks @adriansev!

TimoWilken:
Let me know when you create a tag. I propose to update this PR with the new tag, so we directly see the tests run. Then I'll add an extra `ALIBUILD_XJALIENFS_TESTS` variable for CI, so that the tests aren't accidentally run on user machines with their tokens.

adriansev:
Hi @TimoWilken ! `1.5.9` was tagged! Thanks a lot!!

TimoWilken:
Applied, thank you!

adriansev (approved):

TimoWilken:
Sorry, I forgot one thing: https://github.com/alisw/ali-bot/commit/cff5959320cea6451b10427b447b62eb5e663aa1. I'll restart the affected checks, but then it's fine to merge for me!

TimoWilken:
I see the tests ran successfully for the slc7 build. The error in the Ubuntu build is a known problem unrelated to this PR. Merging!

PR #5362, opened 2024-02-21
Make ROOT use FFTW3 from alidist

TimoWilken:
Install the right version of the FFTW3 library -- ROOT requires the "double", not "float" version.

Also, make sure the ROOT build can find all its dependencies for the features we enable, and fail if it doesn't.


TimoWilken:
Apparently O2 requires the float version of fftw3 while ROOT requires the double one. Also, `-Dfail-on-missing` is going to be more work than I thought to get working on all platforms. Closing in favour of https://github.com/alisw/alidist/pull/5363.

PR #5363, opened 2024-02-21
Split FFTW3 into double and float versions for ROOT and O2

TimoWilken:
ROOT requires libfftw3, but O2 requires libfftw3f.

Also, make sure that ROOT actually finds libfftw3 and doesn't just disable the feature we request.


martenole (approved):
Thanks a lot for your work Timo

davidrohr:
I don't fully understand why we have to split this? My system libfftw compilation of gentoo just builds both libs libfftw and libfftwf. Can't we do this in the recipe? But ok, I also don't care much if there are 2 packages. But it can make it a bit confusing.

TimoWilken:
I agree that doing both in the same package would be cleaner, but I didn't see an obvious way to do this in fftw3's CMake files. As far as I can tell, only one "variant" can be enabled at a time.

I don't know if it would work if we configure as variant 1, build and install it, then wipe out the build dir, re-configure as variant 2, build that and install it in the same prefix as variant 1... I could try that tomorrow.

ktf:
How do we actually know the two variants don't overlap with each other?

TimoWilken:
The only thing that overlaps are the include files, but those are bit-for-bit identical between both variants. https://github.com/alisw/alidist/pull/5363/commits/147166ff8609a0407257228bd5f4dc90c74e2899 works locally for me (I can compile both ROOT and O2 in the slc8-gpu container).

martenole:
Does anything speak against merging this? I have tried to build on the EPNs with this change and then had fftw3 as feature in ROOT as desired.

davidrohr:
well, CI is green except the usual random timeout in the FullCI.
From my side we can merge it to fix it now on the EPNs, but would be good if this could be followed up and cleaned up.
@TimoWilken : could you check why we cannot easily build the f and the non-f variant in one recipe and deliver them together?
For AVX: Wasn't this explicitly disabled since some old grid sites couldn't run it?

TimoWilken:
> @TimoWilken : could you check why we cannot easily build the f and the non-f variant in one recipe and deliver them together?

That's how it works now, since https://github.com/alisw/alidist/pull/5363/commits/147166ff8609a0407257228bd5f4dc90c74e2899. It won't work in a single CMake invocation because fftw3's CMake files only let you enable one variant at a time, but at least it's in a single recipe.

> For AVX: Wasn't this explicitly disabled since some old grid sites couldn't run it?

Ah OK, I suppose that makes sense. The previous recipe was a bit weird -- the variable was declared/exported but never set to a non-empty value, so I couldn't tell whether this was a mistake.

davidrohr:
ah, sorry, didn''t check that you already updated the PR :(.
The title was misleading though :).

TimoWilken:
Oh sorry, I forgot to update the title! :)

PR #5368, opened 2024-02-27
Load grid-base-packages/default in toolchain module

TimoWilken:
For use on the Grid and e.g. in Hyperloop, some packages need to be available (currently just jq). Package these as "grid-base-packages", which has a "default" symlink available.

Then, we can add more packages in future that should be loaded as part of the toolchain, without incurring costly rebuilds of the full software stack.

Draft PR to avoid long-running rebuilds in CI for now.

### DIFF
--- a/gcc-toolchain.sh
+++ b/gcc-toolchain.sh
@@ -181,6 +181,13 @@ if { "\$mod_name" == "GCC-Toolchain" } {
   set base_path [string map "/Modules/modulefiles/ /" \$base_path]
   regexp -- "^(.*)/.*/.*\$" \$base_path dummy base_path
   set base_path \$base_path/Packages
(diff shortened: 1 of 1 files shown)


TimoWilken:
Caching here: https://alijenkins.cern.ch/job/CacheO2Package/103/

TimoWilken:
Caching succeeded, so marking this as ready for review.

ktf:
LGTM; is it safe to merge now?

TimoWilken:
Yes, it's all cached and the `grid-base-packages/default` packages exists for all archs on CVMFS. Merging!

TimoWilken:
I've adjusted the existing the existing Toolchain modulefiles manually, so we don't need to relink them. This just means we won't have to manually edit future GCC-Toolchain modulefiles!

PR #5369, opened 2024-02-28
Package pytorch

TimoWilken:
This is a very large package (~5 GiB), so I didn't put it in Python-modules-list (though it could also go there, in theory, if the `CMAKE_PREFIX_PATH` mangling is ported over).

Binaries work for me (tested with `torchrun --help`), and I can compile a simple example project against the installed C++ library if I use `-DTORCH_CUDA_ARCH_LIST='8.0 8.6 8.9 9.0'` (cf. https://github.com/pytorch/pytorch/issues/113948).

### DIFF
--- a/pytorch.sh
+++ b/pytorch.sh
@@ -1,0 +1,47 @@
+package: pytorch
+version: "%(tag_basename)s"
+tag: "2.2.1"
+requires:
+  - Python
+  - Python-modules
+build_requires:
+  - alibuild-recipe-tools
+prepend_path:
+  PYTHONPATH: "$PYTORCH_ROOT/lib/python/site-packages"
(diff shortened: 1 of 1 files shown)


pytorch.sh
```diff
@@ -1,47 +1,17 @@
-package: pytorch
+package: PyTorch
```
ktf, line 1: do we really need to change the name?

TimoWilken: This is a completely new package. The name is up to you. But anyway, this was for testing by Christian S, so it doesn't need to be merged yet.

ktf: ah, ok, I thought it was already there on CVMFS. Fine with me.

PR #5380, opened 2024-03-11
Get ROOT to compile with new macOS / XCode

ktf:
Get ROOT to compile with new macOS / XCode

Notice you will need both macOS 14.4 and XCode 15.3.

### DIFF
--- a/boost.sh
+++ b/boost.sh
@@ -1,6 +1,6 @@
 package: boost
-version: v1.83.0-alice1
-tag: v1.83.0-alice1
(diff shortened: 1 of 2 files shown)


TimoWilken:
Caching here: https://alijenkins.cern.ch/job/CacheO2Package/105/

ktf:
error in fullCi happens everywhere. tested on my box as well, merging

ktf:
caching was also done

PR #5385, opened 2024-03-15
xjalienfs :: tag 1.6.0 and non-failure trial of gnureadline installation [JAL-178]

adriansev:
* allow to work without loadable readline
* on macos use _only_ gnureadline but work without it if not found
* examples/ add ccdb_mirror, ccdb_time2run, test_grid_conn.C
* fix jobInfo infromation of wn name running the job

### DIFF
--- a/xjalienfs.sh
+++ b/xjalienfs.sh
@@ -24,7 +24,7 @@ prepend_path:
 [[ "$ARCHITECTURE" ==  osx_* ]] && { env -u VIRTUAL_ENV ALIBUILD=1 \
     python3 -m pip install --force-reinstall \
     --target="$INSTALLROOT/lib/python/site-packages" \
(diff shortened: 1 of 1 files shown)


adriansev:
@ktf There are some really weird errors in this 2 particular tests, and i'm fairly certain that have nothing to do with xjalienfs changes.. could you please take a look?
This release mainly fix the macos python 3.12 problems where there is no gnureadline module (<=3.11 works fine)

adriansev:
@ktf This is ready for merging or for another round of tests if you can fix the problems visible in the failed tests. Thanks a lot!

adriansev:
ping @ktf

ktf:
I am not sure this is the best solution, but let's get rid of the problem.

PR #5393, opened 2024-03-22
Fix CGAL rpath

ktf:
Fix CGAL rpath


ktf:
CGAL seems to explicitly disable the RPATH support in CMake, so we need to monkey patch the produced libraries to properly load under the new XCode which seems to be more strict about it.

ktf:
@pzhristov

pzhristov (approved):

PR #5398, opened 2024-03-26
Update Pythia8 alidist recipe

jackal1-66:
Pythia version bumped to 8.311
Important Angantyr fixes implemented in this version.


jackal1-66:
> https://github.com/alisw/alidist/blob/master/defaults-generators.sh#L20 https://github.com/alisw/alidist/blob/master/defaults-o2-epn.sh#L39

Ciao Antonin, I'm aware of those, but I wanted to keep first the edit only to pythia.sh so that I can properly check that everything in the build works fine even on the GRID (it has been tested only in my local machine right now).

jackal1-66:
> Hello @jackal1-66
> 
> thanks for taking this upgrade of version of Pythia from v8304 to v8311 on your shoulder.
> 
> 1. Any reason for the commit ([7fd2c4f](https://github.com/alisw/alidist/commit/7fd2c4faf2e7d061ac65b202500cea00fd2b0160)) to appear here in the list ? Seems like an older fix which is now in place right ?
>    https://github.com/alisw/alidist/blob/master/rivet.sh#L243
> 2. We may "wait" anyhow to fix first the compilation of AGile to work within AliGenerators, to get if the compilation works in the central deployment...
> 3. In the meantime, should we also update (or even remove ?) the default parameter in place for pythia in alidist/defaults-generators.sh ?
> 
> This default choice in `defaults-generators` was introduced by @preghenella (~2022), the exact reason escaped me at the time. Retrospectively, I somewhat guess that it is maybe motivated to align the version of Pythia used among different ALICE environments and avoid version clashes, namely AliGenerators + O² if loaded together in a same environment See also the other `defaults-o2` and `defaults-o2-epn`
> 
> * https://github.com/alisw/alidist/blob/master/defaults-generators.sh#L20
> * https://github.com/alisw/alidist/blob/master/defaults-o2-epn.sh#L39
> * https://github.com/alisw/alidist/blob/master/defaults-o2.sh#L36
> 
> Maybe @ktf, @sawenzel or @pzhristov: we may have a take on such considerations : should we still enforce AliGenerators and 0² to move in sync via `defaults` regarding Pythia8 ? If yes, should we move along from v8304-alice1 to v8311 ? (Possible impact for DPG, as far as I could imagine:
> 
> * central MC production starting from standalone package,
> * for 0²rivet ? maybe transparent, mm not sure
> * ... ?

Ciao Antonin, 
Thanks for taking the time to review the PR. 
1) To be honest I'm not sure why this still appear. The commit has been already merged a long time ago, 
and as I can see only the last commit changes are there (the change to pythia version). I don't know of an easy way to 
edit it here to be honest, that doesn't require me to create another branch... Let me know if it would be good anyway like 
this or if I should do this last procedure.
2 and 3) that's also why I didn't change the defaults, so that the Pythia package could be built as standalone (hopefully) and appear as standalone package on the GRID, but the requirements for O2/AliPhysics and AliGenerators wouldn't change. 
I was able to test the compilation of the new P8 only on my local machines, so I wanted to make sure it worked also on Lxplus and on the GRID.

maireiphc:
For 1. OK, leave it as it is. Seems harmless here if I do not miss anything
For 3. But the thing is that if we do not change the default-generators.sh, the new pythia8 version v8311 can only deployed by user on their machine, there is simply no path towards the CVMFS deployment and grid with standalone package of a custom version tag like v8311... it can only be via AliGenerators where defaults will supersede v8311 by v8034 or via O² with the same superseeding logic... Do I miss sthg ?

jackal1-66:
> For 1. OK, leave it as it is. Seems harmless here if I do not miss anything
> 
> For 3. But the thing is that if we do not change the default-generators.sh, the new pythia8 version v8311 can only be "deployed" by singla user on their own machine, there is simply no path towards the CVMFS deployment and grid with standalone package of a custom version tag like v8311...i.e. it can only be via AliGenerators ... where defaults-generators will anyhow supersede v8311 by v8034 or via O² with the same superseeding logic... Do I miss sthg ?

Since I see also single packages built and available on alimonitor, I was supposing that Pythia8 would be built anyway, without considering the defaults... I guess @ktf can specify if this is the case or not.

jackal1-66:
Ciao @sawenzel, I'm not sure I got your comment right to be honest. Right now I think all the default instances considered (o2, generators and o2-epn) use the same pythia version. What is exactly the thing you're proposing?

sawenzel:
I was proposing **not** to mention pythia in any default anymore ... which will just use the tag mentioned in the pythia.sh recipe. This would make life easier.

jackal1-66:
Ah ok, I understand now. I actually just checked and indeed we're using the last version on each default, so it would improve the situation. For now @maireiphc this can be approved I would say. I will work on editing the other defaults when I have a moment.

maireiphc (approved):
The policy is clear now. Approving.

PR #5412, opened 2024-03-29
GPU: Add RTC compilation test to FullCI

davidrohr:


ktf:
No objections on my side, however i cannot tell what is wrong with the test

PR #5413, opened 2024-04-01
Bump flatbuffers and use internal flatbuffers in ONNXRuntime

davidrohr:


ktf (changes requested):
Why can't we simply move flatbuffers to a version which everyone likes? I would prefer not having two versions.

davidrohr:
ONNXRuntime does not support any newer version, and will not in the foreseeable future as they say they can anyway do this internal build.

ktf:
Fair enough, but why bumping it, then? Is it required to get the new GCC working? If yes, how come the internal version does work?

davidrohr:
That is a good question, I was wondering the same. For me our flatbuffers did not compile with gcc13, while the 9nnx-internal version compiled. I didn't check if they somehow fixed it. Anyhow, I don't like using a many years old version due to onnx decising they do not want to bump...

ktf:
@davidrohr do you understand the  issue on cs8?

davidrohr:
@ktf : which issue on cs8?

ktf:
yesterday it was failing. maybe it was simply a glitch in the wrong package.

davidrohr:
Now Ubuntu fails, but it was green before. So I think it is more random ctests failing.
@pzhristov : What is the status of the tests with analysis? Can we merge this?

davidrohr:
no longer needed with onnx bumped to 1.18

PR #5414, opened 2024-04-04
[Geant4_VMC] Bump to v6-6-p1

benedikt-voelkel:


sawenzel:
How do we proceed here? Do we have some release validation results?

benedikt-voelkel:
We do have a validation: https://indico.cern.ch/event/1399212/contributions/5883872/attachments/2828111/4941146/20240327_slides_geant4_vmc.pdf

mfasDa:
Both the reports of the validation of the GEANT4 itself and the and the VMC here only address tracking and not calorimetry, even though we provide a task in the release validation. This makes it hard to track where the issue with the pi0 mass peak could have been introduced.

Future reports of the release validation please also include calorimetry.

PR #5419, opened 2024-04-05
xjalienfs :: 1.6.1 prevent potential abort in pthread_cancel

adriansev:
There are cases when the following message appear at the exit of alien.py:
```
libgcc_s.so.1 must be installed for pthread_cancel to work
Aborted
```

I encountered this when two versions of openssl were present in system and when trying
`/cvmfs/alice.cern.ch/bin/alienv enter O2sim::async-2023-pp-apass3-20240320.1-1,xjalienfs`

Let's try to prevent this and see if macos is happy with this.
If it works then i will tag 1.6.1

### DIFF
--- a/xjalienfs.sh
+++ b/xjalienfs.sh
@@ -1,6 +1,6 @@
 package: xjalienfs
 version: "%(tag_basename)s"
-tag: "preload_gcc"
+tag: "1.6.1"
 source: https://gitlab.cern.ch/jalien/xjalienfs.git
 requires:
   - "OpenSSL:(?!osx)"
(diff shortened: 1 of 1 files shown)


ktf:
Could you elaborate on what was the actual fix, possibly with a link to the diff?

adriansev:
@ktf sure!
https://github.com/adriansev/jalien_py/compare/1.6.0...1.6.1
the only commit from 1.6.0 to 1.6.1 is 
https://github.com/adriansev/jalien_py/commit/74df2706db896ebd1094029b1db06dad57a85716

adriansev:
Good morning @ktf ! Any resolution on this?

PR #5429, opened 2024-04-14
Fix libuv load path

ktf:
libwebsockets is linked statically, however it requires libuv, which is not, therefore one needs to actually have it in the modulefile. This works just fine on O2, because other stuff brings in libuv, however it breaks when using it for AliROOT, which does not. Hot patched on CVMFS as well.

### DIFF
--- a/jalien-root.sh
+++ b/jalien-root.sh
@@ -7,6 +7,7 @@ requires:
   - xjalienfs
   - XRootD
   - libwebsockets
+  - libuv
 build_requires:
   - json-c
   - CMake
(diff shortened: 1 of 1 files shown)


yuw726:
Hi @ktf, could you clarify this for me? In my local build, `aliBuild` decides that both libwebsockets and libuv will be built and not taken from the system. Therefore the local JAliEn-ROOT has both of them.

But before 0.7.9 JAliEn-ROOT was actually linking against the system libuv because of wrong cmake instruction. I think it messed up the cvmfs version so that it couldn't locate the library on the remote machine.

0.7.9 is using HINTS instead of PATHS [here](https://gitlab.cern.ch/jalien/jalien-root/-/blob/master/cmake/modules/FindLibUV.cmake?ref_type=heads#L53).

It seems to work without the libuv env variable. Do we need it?

ktf:
Yes, you do. libwebsockets is an archive library however libuv is not, therefore it needs to be brought in nevertheless. Moreover libwebsocket is declared as a build_requires, so it's not included in the automatic modulefile generation. I suspect it works for you because you worked in the O2 environment, where libuv is brought in correctly regardless.

PR #5432, opened 2024-04-18
ndmspc v0.20240422.0 release

mvala:


ndmspc.sh
```diff
@@ -40,7 +43,8 @@ MODULEFILE="etc/modulefiles/$PKGNAME"
 alibuild-generate-module --bin --lib > "$MODULEFILE"
 cat >> "$MODULEFILE" <<EoF
 # Our environment
-set NDMSPC_ROOT \$::env(BASEDIR)/$PKGNAME/\$version
-setenv NDMSPC_ROOT \$NDMSPC_ROOT
+setenv NDMSPC_ROOT \$PKG_ROOT
```
ktf, line 46: if you can avoid the environment variable, i would prefer.

ktf (changes requested):

ndmspc.sh
mvala: Do you mean `NDMSPC_ROOT` or `NDMSPC_MACRO_DIR`?

- i assumed that `$NDMSPC_ROOT` should it be there by design
- I can remove `$NDMSPC_MACRO_DIR`, but i would prefer to have it there for users

ktf: Any `$*_ROOT` is not required to be there at runtime, in general, especially as environment variable rather than a shell variable (`setenv` vs `set`).

If you do not need it, I would suggest you drop it. For `$NDMSPC_MACRO_DIR`, I would suggest you find a way to detect it automatically (e.g. rely on it being in `../macros`), but I have no strong opinion.

I think the bottomline i that you should try to avoid environment variables as much as possible.

mvala: > Any $*_ROOT is not required to be there at runtime, in general, especially as environment variable rather than a shell variable (setenv vs set).

I can avoid this
> If you do not need it, I would suggest you drop it. For `$NDMSPC_MACRO_DIR`, I would suggest you find a way to detect it automatically (e.g. rely on it being in ../macros), but I have no strong opinion.

Detection of `macros` directory is exactly via `$NDMSPC_MACRO_DIR` variable. 

I could also use `Unix.*.Root.MacroPath: .:$(ROOTSYS)/macros` and add entry there, but for that we (as alice) would have to have some common `.rootrc` where i can add via `aliBuild` recepie similar as it is 

```
alibuild-generate-module --bin --lib --macros > "$MODULEFILE"
```
But i think it would be overkill. Defining `$NDMSPC_MACRO_DIR` is just fine.

What do you think?

ktf: > I can avoid this
...
> Detection of macros directory is exactly via $NDMSPC_MACRO_DIR variable.

Ok, then let's just drop `setenv NDMSPC_ROOT` in the modulefile. I guess it's fine.

ndmspc.sh
```diff
@@ -19,12 +18,16 @@ if [[ $ALIBUILD_NDMSPC_TESTS ]]; then
...
   CXXFLAGS="${CXXFLAGS} -Werror -Wno-error=deprecated-declarations"
 fi
-cmake --version
 
 # When O2 is built against Gandiva (from Arrow), then we need to use
 # -DLLVM_ROOT=$CLANG_ROOT, since O2's CMake calls into Gandiva's
 # -CMake, which requires it.
-cmake "$SOURCEDIR" "-DCMAKE_INSTALL_PREFIX=$INSTALLROOT"
+cmake "$SOURCEDIR" "-DCMAKE_INSTALL_PREFIX=$INSTALLROOT"          \
+      -G Ninja                                                    \
```
ktf, line 26: You probably need a build_requires for ninja.

ktf (changes requested):
If you fix the build_requires to include Ninja, then I think we are ready to go.

ndmspc.sh
mvala: Done

PR #5441, opened 2024-04-26
bumped aliecs to v1.8.3 and golang to v1.22.2

justonedev1:


knopers8 (approved):

knopers8:
The CI checks might fail if they were started before 13:42, since the v1.8.3 release appeared later than this PR.

PR #5442, opened 2024-04-26
Bump ODC to 0.83.0 and DDS to 3.10

lkrcal:


AnarManafov:
@lkrcal , there is a hot fix for DDS. There was a bug in topology update code. Please, change the version to 3.10

PR #5446, opened 2024-04-29
Improve GCC builds

ktf:
- Move to GCC 13.2.0
- Move to GDB 14.2
- Do not use GCC 7.3.0 as default, since now AliRoot it's built with its own default.
- Add support for CCACHE. While disabled by default it can be enabled by passing CCACHE_DISABLE=false as environment variable.

### DIFF
--- a/defaults-ali.sh
+++ b/defaults-ali.sh
@@ -25,9 +25,6 @@ overrides:
       - Vc
       - ZeroMQ
       - JAliEn-ROOT
-  GCC-Toolchain:
-    version: v12.2.0-alice1
-    tag: v12.2.0-alice1
   cgal:
     version: 4.12.2
   fastjet:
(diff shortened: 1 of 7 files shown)


awegrzyn:
@ktf Very nice, do you have any numbers/experience on speed up when using Ccache with our codebase?

ktf:
No... In the past (15+ years ago) I frankly spent more time chasing hash conflicts than saved build time, but maybe it is worth giving another chance since a lot of time passed. Moreover they now provide distributed caches via redis as well, which might be very interesting if it works as advertised.

ktf:
The other issue is root dictionaries generation which cannot be cached. i propose we simply try and then decide how to proceed.

awegrzyn:
Ok thanks. Yeah, I'm keen on experiment with it as well.

ktf:
Closed in favor of #5449 which has the minimal set required.

PR #5459, opened 2024-05-07
Bump python to 3.11.10

ktf:
Bump python to 3.12.3

ALMA 9 on ARM seems to have troubles with it


pzhristov:
I think python 3.12 still has issues with some packages, my proposal would be to use 3.11.10.

PR #5466, opened 2024-05-13
STARlight introduction in AliGenerators

jackal1-66:
Added STARlight dependency to AliGenerators. STARlight recipe has been edited to pull directly the main branch from the official STARlight repository. In addition the HepMC3 package has been set as a dependency and STARlight is now built with it, so that it could also be useful for RIVET analyses. Tests performed locally are encouraging.

### DIFF
--- a/starlight.sh
+++ b/starlight.sh
@@ -1,6 +1,6 @@
 package: STARlight
-version: "%(tag_basename)s"
-tag: main
+version: "20240714"
+tag: 9980f5d
 requires:
   - HepMC3
 build_requires:
--- a/starlight.sh
+++ b/starlight.sh
@@ -14,7 +14,7 @@ cmake $SOURCEDIR -DCMAKE_INSTALL_PREFIX=$INSTALLROOT       \
                  -DCMAKE_BUILD_TYPE=$CMAKE_BUILD_TYPE      \
                  -DCMAKE_SKIP_RPATH=TRUE                   \
                  -DENABLE_HEPMC3=ON                        \
-                 -DHepMC2_DIR=$HEPMC3_ROOT
+                 -DHepMC3_DIR="$HEPMC3_ROOT"
 cmake --build . -- ${JOBS:+-j$JOBS}
 mkdir -p $INSTALLROOT/bin
 cp ./starlight $INSTALLROOT/bin/.
--- a/starlight.sh
+++ b/starlight.sh
@@ -1,16 +1,20 @@
 package: STARlight
 version: "%(tag_basename)s"
-tag: r313-c
+tag: main
+requires:
+  - HepMC3
 build_requires:
   - CMake
   - "GCC-Toolchain:(?!osx)"
-source: https://github.com/alisw/STARlight.git
+source: https://github.com/STARlightsim/STARlight.git
 ---
 #!/bin/bash -ex
 cmake $SOURCEDIR -DCMAKE_INSTALL_PREFIX=$INSTALLROOT       \
                  ${CMAKE_GENERATOR:+-G "$CMAKE_GENERATOR"} \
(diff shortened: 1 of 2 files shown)


jackal1-66:
@ktf the error on alidist seems unrelated to the introduction of STARlight. Could you please check?

starlight.sh
```diff
@@ -1,24 +1,27 @@
...
+source: https://github.com/STARlightsim/STARlight.git
 ---
 #!/bin/bash -ex
 cmake $SOURCEDIR -DCMAKE_INSTALL_PREFIX=$INSTALLROOT       \
                  ${CMAKE_GENERATOR:+-G "$CMAKE_GENERATOR"} \
                  -DCMAKE_BUILD_TYPE=$CMAKE_BUILD_TYPE      \
-                 -DCMAKE_SKIP_RPATH=TRUE
+                 -DCMAKE_SKIP_RPATH=TRUE                   \
+                 -DENABLE_HEPMC3=ON                        \
+                 -DHepMC2_DIR=$HEPMC3_ROOT
```
ktf, line 17: Please have a look at this.

ktf:
Fine with me. Just fix up the quotes and then I will merge.

jackal1-66:
> Fine with me. Just fix up the quotes and then I will merge.

I added the double quotes and noticed there was HepMC2 instead of 3. It was working anyway, but better upgrade that little part as well. Tested locally and it works.

starlight.sh
```diff
@@ -1,24 +1,27 @@
 package: STARlight
 version: "%(tag_basename)s"
-tag: r313-c
+tag: main
```
ktf, line 3: I just noticed. Is this a branch? Please use a tagged version.

ktf:
Thanks! Could you please use a tagged version as well?

I also noticed the recipe generates its own modulefile. It's probably better to move to use `alibuild-generate-module` so that  eventual runtime dependencies are taken into account. Grep for it in alidist to see how it's used elsewhere.

ktf:
(the last point can be done in a separate PR, if you prefer)

starlight.sh
jackal1-66: In the JIRA discussion the main developers of STARlight were saying that they "moved" from the release of tagged versions to simply updating the main branch of the software. I do agree that tagged versions would be a better solution. Should I propose it to them? They were quite open in satisfying our request and the fact that the main branch compiles is already a good starting point imho.

ktf: Just pick the hash of the current commit at the tip of the branch and use that. There is also `version` field you can use to give it a mnemonic name (e.g. the date of today). If we use the branch, there might be unexpected rebuilds, which we do not want, especially if the plan is to include this in O2sim / O2PDPSuite.

jackal1-66: Done.

ktf (approved):

jackal1-66:
> Thanks! Could you please use a tagged version as well?
> 
> I also noticed the recipe generates its own modulefile. It's probably better to move to use `alibuild-generate-module` so that eventual runtime dependencies are taken into account. Grep for it in alidist to see how it's used elsewhere.

I will move to alibuild-generate-module on a different PR. I prefere being sure that everything works as expected before moving to that (minimum changes at a time).

starlight.sh
ktf: Thanks.

ktf:
Thank you for the fixes. Let's have a last spin on the AliGenerators test and then I will merge.

jackal1-66:
build/O2/alidist-dataflow-cs8 failing is unrelated to this PR

jackal1-66:
The current alidist failures are not related to this PR.

chiarazampolli:

PR #5476, opened 2024-05-23
Add prefer_system to the golang recipe

ktf:
Add prefer_system to the golang recipe

prefer_system has priority over prefer_system_check and by default
it's not matching any architecture.


ktf:
@Barthelemy try with this...

PR #5477, opened 2024-05-23
Bump alibuild-recipe-tools to 0.2.5

ktf:


ktf:
@pzhristov once merged, this will allow you to add `--cmake` for all the packages which build with cmake and alibuild-generate-modules will add the proper path to `CMAKE_PREFIX_PATH`. I think this should fix the issues we were discussing about the OpenData builds, without having to introduce too many extra variables. 

@olantwin this might be of interest also for you guys.

olantwin:
Thanks for the heads-up, this might come in useful for our CMAKE dependencies!

Is https://github.com/alisw/alibuild-recipe-tools/compare/v0.2.3...v0.2.4#diff-ccbc5510ed63c7cb868e6ec3b8fce4855c3eabe8972ee535326c8d22b6fae980R46 correct? Shouldn't it be `--)` for the first case now?

PR #5478, opened 2024-05-24
Generate modulefiles which allow compiling O2OpenAccess when loaded

ktf:
Generate modulefiles which allow compiling O2OpenAccess when loaded


ktf:
@pzhristov this is what I meant over coffee.

PR #5485, opened 2024-05-28
Bump UCX to v1.16.0

ktf:
Bump UCX to v1.16.0


ktf:
@lkrcal any objections? This is required for me to build using GCC 13.

lkrcal:
@ktf Quite substantial releases of UCX, but in principle should be fine.

PR #5494, opened 2024-06-06
Revert "Bump UCX to v1.16.0 (#5485)"

lkrcal:
This reverts commit 8c66f48e9b971e504afeae8fa9cbf522ffbc1abc.

### DIFF
--- a/ucx.sh
+++ b/ucx.sh
@@ -1,6 +1,6 @@
 package: ucx
 version: "%(tag_basename)s"
-tag: v1.16.0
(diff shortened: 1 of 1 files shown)


lkrcal:
@ktf Unfortunately the bump of UCX had negative performance implications that were not discovered in standalone and staging tests. We were forced to revert the latest release to one with an older build of Data Distribution. We need to revert this now and in the future we can explore how to reintroduce it and mitigate the issues.

PR #5495, opened 2024-06-07
Add proper Vc support in ROOT

ktf:
Add proper Vc support in ROOT


ktf:
@Barthelemy can you check if this solves your RPM issues?

PR #5497, opened 2024-06-11
Fix ONNX ROOT_INCLUDE_PATH

sawenzel:
ONNX CMake sets include path to ${ONNX_ROOT}/include/onnxruntime. This needs to be reflected in the ROOT_INCLUDE_PATH as well.

Fixes failures in anchoredMC that appeared after the recent ONNX bump.


ktf:
I had a similar problem, indeed.

PR #5501, opened 2024-06-12
Improve defaults for xrootd

ktf:
Improve defaults for xrootd


ktf:
@costing as discussed, this will make it default for anyone building with our own xrootd and for anyone sourcing the xrootd module.

costing:
Looks good, but could we have them inherited from the environment and only be set to these values if not already defined? I want to be able to override them from outside, these values should be a good default if nothing else is already set.

adriansev:
@ktf @costing A few words of warning and info taken from alienpy:
```
   # Default value for the time after which an error is declared if it was impossible to get a response to a request.
    # N.B.!!. This is the total time for the initialization dialogue!! see https://xrootd.slac.stanford.edu/doc/xrdcl-docs/www/xrdcldocs.html#x1-580004.3.6
    if not os.getenv('XRD_REQUESTTIMEOUT'): XRD_EnvPut('RequestTimeout', int(1200))  # default 1800

    # Default value for the time after which a connection error is declared (and a recovery attempted) if there are unfulfilled requests and there is no socket activity or a registered wait timeout.
    # N.B.!!. we actually want this timeout for failure on onverloaded/unresponsive server. see https://github.com/xrootd/xrootd/issues/1597#issuecomment-1064081574
    if not os.getenv('XRD_STREAMTIMEOUT'): XRD_EnvPut('StreamTimeout', int(60))  # default 60
```

With requesttimeout set there will be NO fallback in alien.py, and even more even if only xrdcp or straight XrdCl API is used Request time is always bigger than Stream time (it includes initializations and so on ..) IMHO with RequestTimeout set so low there will be huge number of errors

costing:
This is what is used in jobs. Only users didn't have these defaults and are hit by unresponsive servers when running locally. The point of this is to set sane defaults in any running environment. That can be overwritten if needed.

As for requesttimeout, if in 60s you didn't get the bytes you have requested in one operation there is no point in waiting any longer.

ktf:
@costing there is an option to --set-if-not-defined. It requires modules 5.1 (i.e. Alma9) which will make that not work for the EPNs.

PR #5504, opened 2024-06-13
Update SHERPA + AliGen fixes

jackal1-66:
This update brings also additional compatibility with newer compilers.

PR #5505, opened 2024-06-14
Improvements to STARlight install

sawenzel:
* build STARlight as a shared library
* fix the installed header files, so that they can be included in ROOT macros without symbol shadowing (STARlight defines global C++ macro names such as `printInfo` which conflict with similar names in other packages)

With this commit, it will be possible to use the STARlight system directly in event generation.

See for instance: https://github.com/AliceO2Group/O2DPG/pull/1663

### DIFF
--- a/starlight.sh
+++ b/starlight.sh
@@ -1,6 +1,6 @@
 package: STARlight
-version: "20240714"
-tag: 9980f5d
+version: "20240617"
+tag: 196aedf
 requires:
   - HepMC3
 build_requires:
--- a/starlight.sh
+++ b/starlight.sh
@@ -14,14 +14,20 @@ cmake $SOURCEDIR -DCMAKE_INSTALL_PREFIX=$INSTALLROOT       \
                  -DCMAKE_BUILD_TYPE=$CMAKE_BUILD_TYPE      \
                  -DCMAKE_SKIP_RPATH=TRUE                   \
                  -DENABLE_HEPMC3=ON                        \
+		 -DCMAKE_EXPORT_COMPILE_COMMANDS=ON        \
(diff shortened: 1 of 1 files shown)


sawenzel:
Just for information. Needs https://github.com/STARlightsim/STARlight/pull/11 before merging (+ update of the commit/tag).

sawenzel:
@jackal1-66 f.y.i.

jackal1-66:
ciao @sawenzel, I was trying to rebuild AliGen including this fix, but I get this error: 
`AliGenerators:STARlight:0: Error during git checkout for reference repo for STARlight.` could you have a look at it? I tried recompiling with the #5504 without this PR merged and it compiled effortlessly.

starlight.sh
```diff
@@ -14,22 +14,38 @@ cmake $SOURCEDIR -DCMAKE_INSTALL_PREFIX=$INSTALLROOT       \
...
 
-#ModuleFile
-mkdir -p etc/modulefiles
-alibuild-generate-module --bin --lib > etc/modulefiles/$PKGNAME
-mkdir -p $INSTALLROOT/etc/modulefiles && rsync -a --delete etc/modulefiles/ $INSTALLROOT/etc/modulefiles
+# Modulefile
+MODULEDIR="$INSTALLROOT/etc/modulefiles"
+MODULEFILE="$MODULEDIR/$PKGNAME"
+mkdir -p "$MODULEDIR"
+alibuild-generate-module --bin --lib > $MODULEFILE
```
ktf, line 42: Given STARLight is CMake based, I think it would be better to provide `--cmake` to `alibuild-generate-module` and avoid the custom `$STARLIGHT_ROOT` variable.

chiarazampolli:

PR #5513, opened 2024-06-20
Removed not needed fix 

jackal1-66:
This fix was necessary in the past, but now it seems that it destroys things when setting CGAL library (especially). 
CGAL_ROOT is set in a correct way globally before running the rivet recipe.


jackal1-66:
alidist build failing is unrelated to the current PR

ktf:
Please, squash commits when you are ready to merge.

PR #5514, opened 2024-06-21
starlight : Specify LIBDIR = lib

sawenzel:
Otherwise this will be lib64 on some systems
and lib on others.


ktf:
Errors unrelated and trivial.

chiarazampolli:

PR #5515, opened 2024-06-22
Use our own compact repository for Clang

ktf:


ktf:
This should workaround our current issues with the clang repository taking forever to checkout.

PR #5524, opened 2024-07-01
Bump scikit-learn on older python versions

singiamtel:
Fixes the following error (tested on python 3.10)

```
  Error compiling Cython file:
  ------------------------------------------------------------
  ...
      # Max value for our rand_r replacement (near the bottom).
      # We don't use RAND_MAX because it's different across platforms and
      # particularly tiny on Windows/MSVC.
      RAND_R_MAX = 0x7FFFFFFF

  cpdef sample_without_replacement(np.int_t n_population,
                                   ^
  ------------------------------------------------------------

  sklearn/utils/_random.pxd:18:33: 'int_t' is not a type identifier
  ```

### DIFF
--- a/python-modules-list.sh
+++ b/python-modules-list.sh
@@ -62,8 +62,7 @@ env:
     seaborn == 0.11.0; python_version >= '3.9'
 
     scikit-learn == 0.20.3; python_version < '3.8'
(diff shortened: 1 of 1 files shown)


singiamtel:
Closing as [#5547](https://github.com/alisw/alidist/pull/5547) supersedes

PR #5526, opened 2024-07-01
Added cgal as rivet requirement

jackal1-66:
Without this fix the CGAL_ROOT path is wrongly expanded in myldflags pushed inside rivet-build

### DIFF
--- a/rivet.sh
+++ b/rivet.sh
@@ -13,9 +13,6 @@ requires:
   - "Python-system:(osx.*)"
 build_requires:
   - GCC-Toolchain:(?!osx)
-  - cgal
-  - GMP
-  - YODA
   - Python
 prepend_path:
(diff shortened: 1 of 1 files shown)


maireiphc (approved):

rivet.sh
```diff
@@ -6,12 +6,14 @@ requires:
   - HepMC3
   - YODA
   - fastjet
+  - cgal
```
ktf, line 9: For the record, `requires` implies `build_requires`.

jackal1-66: apparently without cgal in the build_requires GCC is not able to find the CGAL library during building

jackal1-66: Edit: I had made a typo, it works as expected locally

jackal1-66:
alidist failed CI seems unrelated to this PR.

PR #5527, opened 2024-07-01
More changes in AEGIS/MCStepLogger for the ROOT 6.32 bump

sawenzel:
this should be the final step to make simulation happy with ROOT 6.32 and TPythia6 deprecated in ROOT.


ktf (approved):

ktf:
@singiamtel this is not triggering the checks. Can you try to understand why?

ktf:
Ah, ok, nevermind, I thought the PR was against master.... Merging.

PR #5532, opened 2024-07-02
Add moderncppkafka for QC

justonedev1:


ktf:
What is this needed for? Do you really need it for the whole QC?

ktf (commented):
Please, have a look at the linter messages, they look valid to me.

justonedev1:
I will take a look at linter messages, thank you. 
It is needed for [QC-271](https://its.cern.ch/jira/browse/QC-271). Strictly speaking it is not needed but it really helps with consuming kafka from C++ justification is in the ticket.

ktf:
Ok, so it's for the actual devices, not for some infrastructure bit. Fine then.

ktf (commented):
There  is still one linter issue unresolved. Could you have a look into it?

PR #5533, opened 2024-07-03
Minimal changes to get UCX compile with GCC 13

ktf:


ktf:
@lkrcal @davidrohr ok, I took a better look and I found the two commits... In the end they are adding one header and changing one type from unsigned to the correct alias, so I guess this is basically safe.

PR #5534, opened 2024-07-03
Add python (gnu)readline package

sawenzel:
This install the python readline package. 

Currently readline support in alien.py, build locally with alibuild or published on cvmfs, is broken. Adding this python package to our installation fixes this problem (conveniently).

Tested on CC7 and Ubuntu 22.04 No other installation of libreadline-dev etc necessary.

Fixes problem described here https://its.cern.ch/jira/browse/O2-4927 fully consistently for everyone building alien.py with alibuild.

### DIFF
--- a/python-modules-list.sh
+++ b/python-modules-list.sh
@@ -98,6 +98,9 @@ env:
     dask[array,dataframe,distributed] == 2023.12.1; python_version >= '3.11'
     dask_jobqueue == 0.8.2
 
+    # readline is needed by alien.py (xjalienfs)
+    gnureadline
+
 build_requires:
   - alibuild-recipe-tools
 ---


adriansev:
@sawenzel this will not help because on linux i do not use this, see [this code](https://github.com/adriansev/jalien_py/blob/master/alienpy/tools_history.py#L13-L25)

I'm kind of against another dependency, that when will break will brake the whole usage. take into account that it took almost one year for gnureadline to be released for python3.11 
Also, i'm very reluctant to use external dependencies that replace what is already present in python. this is not a solution for the _broken_ builds of python

adriansev:
so, you are right! actually the gnureadline _package_ have the same module name as `readline` so, if 'gnureadline' package is installed, doing `import readline` will actually import the module exported by gnureadline package. Moreover, it seems that it will not overwrite the python module if present, so it would be safe to do it.
now, the issue is to synchronize this PR with the content from https://github.com/alisw/alidist/blob/master/xjalienfs.sh#L22-L27

my against points remain, but this is not my show so the deciding people should make the decision of what is saner and safer.

PR #5535, opened 2024-07-03
Bump ROOT to v6.32.02-alice1

ktf:


Barthelemy:
DO NOT MERGE

we agreed with Giulio to wait until we are back from holidays.

ktf:
@shahor02 @sawenzel @davidrohr this is now failing with 
```
./sim-challenge.log[ERROR]  - Device trd-globaltracking_TPC_FT0_ITS-TPC_CTP_: pid 18773 (exit 128)
./sim-challenge.log[INFO]    - First error: [WARN] MC-truth is not supported for FT0 recpoints currently
./sim-challenge.log[INFO]    - Last error: Backtrace complete.
```

any idea of what might be going on?

shahor02:
Not sure, it crashes when filling GPU data structures :
./sim-challenge.log[18773:trd-globaltracking_TPC_FT0_ITS-TPC_CTP_]:     /sw/slc8_x86-64/O2/5535-slc8_x86-64-local7/lib/libO2GPUWorkflowHelper.so: o2::gpu::GPUWorkflowHelper::fillIOPtr(o2::gpu::GPUTrackingInOutPointers&, o2::globaltracking::RecoContainer const&, bool, o2::gpu::GPUCalibObjectsTemplate<o2::gpu::ConstPtr> const*, o2::gpu::gpustd::bitset<32u>, o2::gpu::gpustd::bitset<32u>, o2::gpu::gpustd::bitset<32u>)

But fullCI has converged in other PRs after this one crashing, so might be related to this PR. Could you restart fullCI to check if it is reproducible?

ktf:
I wonder if it's not related to:

```
./itstpcMatch.log[16548:itstpc-track-writer]: Error in <TStreamerInfo::Build>: o2::dataformats::AbstractRef<25,5,2>, unknown type: typename std::conditional<(32> mRef
```

which we have a few lines above.

shahor02:
I've never seen this error before, apparently caused by the new root version. But then, the failing TRD reco is by far not the 1st workflow to use these structures.

ktf:
@shahor02 the issue seems to have indeed been solved by https://github.com/AliceO2Group/AliceO2/pull/13310. I would say the remaining one (missing libxandr-devel I guess), must be a newly introduced system dependency somewhere in ROOT. I will ask @singiamtel to add it to the docker container on Monday.

ktf:
@Barthelemy this is ready to be tested on your side whenever you are ready.

Barthelemy:
I tested this branch and it works for me. ROOT out of the RPMs is functional. I don't think that we should wait longer. Let's merge it if you agree @ktf @vascobarroso

PR #5538, opened 2024-07-04
Fix for RPMs for moderncppkafka.sh

Barthelemy:


moderncppkafka.sh
```diff
@@ -5,12 +5,15 @@ requires:
...
+# this is header only library, so we can just copy it                                                                                                                                                    
 mkdir -p "$INSTALLROOT"
 cp -r "$SOURCEDIR/include" "$INSTALLROOT/include"
+rm "$INSTALLROOT/include/CMakeLists.txt" # for some reason it is there                                                                                                                                   
 
-# Modulefile
+# Modulefile                                                                                                                                                                                             
 mkdir -p "etc/modulefiles"
 alibuild-generate-module --lib > "etc/modulefiles/$PKGNAME"
+mkdir -p $INSTALLROOT/etc/modulefiles && rsync -a --delete etc/modulefiles/ $INSTALLROOT/etc/modulefiles
```
justonedev1, line 18: why is the result of alibuild-generate-module copied to etc/modulefiles instead of directly into the $INSTALLROOT?

ktf: In the past, if you had an incremental build, you might have ended up with stray bits in $INSTALLROOT which the rsync would have delete. Now we can probably simply remove the folder and regenerate them always, including in the incremental recipe.

justonedev1: ah, okay thank you!

justonedev1 (approved):

moderncppkafka.sh
ktf: regenerate them with `alibuild-generate-module`, I mean.

ktf, line 18:
```suggestion
mkdir -p "$INSTALLROOT/etc/modulefiles" && rsync -a --delete etc/modulefiles/ "$INSTALLROOT/etc/modulefiles"
```

PR #5539, opened 2024-07-04
Enable back tensorflow for python3.9 and 3.10

mfasDa:


mfasDa:
For python3.9 the following combination works:
- Tensorflow 2.7.1
- Keras 2.7.0
Constrained by the version of numpy used in the 3.9 case (1.21.4) which does not allow for newer Tensorflow versions, and Keras needs to be in sync with Tensorflow.

For python3.10 the Tensorflow version 2.13 which is used in the python3.11 case should work as it is compatible with numpy 1.22-1.23.

singiamtel:
I'm not sure whether there's a good reason to keep numpy pinned, maybe bumping numpy version and having less keras/tensorflow versions would be easier

mfasDa:
@singiamtel I agree with you that I would be in favour of bumping numpy as well, because also the later versions of Tensorflow like 2.13 - which is used for python 3.11 - would support python 3.9. However I saw that this creates another set of dependency issues with other packages (uproot and scikit-learn, i.e.).

ktf:
Looks like this is now breaking ARM builds:

https://alijenkins.cern.ch/job/DailyBuilds/job/DailyO2Physics-slc9-arm/13/

PR #5540, opened 2024-07-05
Still support defaults-ali

ktf:


ktf:
@pzhristov this is what we discussed. The new ROOT cannot be used with AliRoot, due to the lack of Proof support.

PR #5541, opened 2024-07-05
Migrate to GCC 13

ktf:
Migrate to GCC 13

- Move to GCC 13.2.0
- Move to GDB 14.2
- Do not use GCC 7.3.0 as default, since now AliRoot it's built
  with its own default.
- Add support for CCACHE. While disabled by default it can be enabled
  by passing CCACHE_DISABLE=false as environment variable.

### DIFF
--- a/defaults-ali.sh
+++ b/defaults-ali.sh
@@ -24,9 +24,6 @@ overrides:
       - Vc
       - ZeroMQ
       - JAliEn-ROOT
-  GCC-Toolchain:
-    version: v12.2.0-alice1
-    tag: v12.2.0-alice1
   cgal:
     version: 4.12.2
   fastjet:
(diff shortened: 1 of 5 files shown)


davidrohr:
I just checked locally that the Geant header problem we have with new GCC versions is fixed in new Geant versions, i.e. if we have a fix and go back to new Geant reverting https://github.com/alisw/alidist/pull/5567, the GCC bump should go through.
@sawenzel : do we any ETA when we can bump Geant again?

sawenzel:
@davidrohr : https://github.com/alisw/alidist/pull/5574

davidrohr:
Great, thx!

davidrohr:
https://github.com/AliceO2Group/AliceO2/pull/13425  and https://github.com/AliceO2Group/AliceO2/pull/13426 should fix O2 compilation with this GCC

ktf:
aarch issue unrelated and it was working before. Merging this to be in time for the PbPb datataking.

PR #5546, opened 2024-07-08
fix bookkeeping module file dep in o2

Barthelemy:


vascobarroso:
@davidrohr since @ktf is away could you review this ? We need it to fix our RPM generation

PR #5560, opened 2024-07-25
build librdkafka without ssl,sasl,curl

justonedev1:
there were linking issues with QC caused by using librdkafka. I changed the recipe that librdkafka does not use ssl,sasl and curl as we don't need it now and QC is the only one who uses the library

### DIFF
--- a/librdkafka.sh
+++ b/librdkafka.sh
@@ -8,19 +8,18 @@ build_requires:
   - CMake
 source: https://github.com/edenhill/librdkafka
 ---
-#!/bin/bash -ex
-cmake -H"$SOURCEDIR"                        \
-      -B"$SOURCEDIR/_cmake_build"           \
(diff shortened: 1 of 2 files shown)


librdkafka.sh
```diff
@@ -3,20 +3,21 @@ version: "%(tag_basename)s"
 tag: v2.3.0
 requires:
   - "GCC-Toolchain:(?!osx)"
+  - curl
```
knopers8, line 6: why do you require curl if you disable it later?

justonedev1: good point, fixed

justonedev1:
@ktf could you take a look at this PR, please? It is failing while testing O2, which I don't really get as I did not touch any modules that O2 is dependent on. (there might be some other errors, I am writing this when some checks are still in process). Thank you very much for your help.

ktf:
Failure on ARM unrelated. Merging.

PR #5561, opened 2024-07-26
Correct handling of RPATH for grpc and c-ares

ktf:
Correct handling of RPATH for grpc and c-ares

For some reason c-ares and grpc set their id to `@rpath/lib...` rather
than their full path. This means that if a given dependent library has
an implicit transitive dependency on it, they will not be able to find it.

This happens in particular for bookkeeping which defines grpc as a private
dependency, breaking downstream users.


ktf:
Currently broken, fixes things on my box. Merging.

@Barthelemy the issue is that if a library uses "@rpath/libName.dylib" as library ID, then all the users of that library need to have the correct path in the LD_RPATH. I suspect in particular this does not happen because the BookkeepingAPI treats grpc as private dependency. There are probably better ways to do this in CMake directly (maybe switching to a public dependency) however I could not get it to work.

PR #5566, opened 2024-08-05
Fix starlight build

davidrohr:


chiarazampolli:
Ported to async-v1-01-03

PR #5567, opened 2024-08-06
Revert to Geant4-11.0.4 and Geant4_VMC 6-1-update1-p9

sawenzel:
Tmp poing back to older Geant4 to fix https://its.cern.ch/jira/browse/O2-5212


chiarazampolli:
Thanks @sawenzel ! slc9 test failed, but it seems unrelated. Let's see the last one.

PR #5439, opened 2024-04-25
Bump arrow 14 -> 17

aalkin:

PR #5570, opened 2024-08-09
Bump LLVM and CodeChecker to 17.0.6

davidrohr:
include #5439 and #5569

### DIFF
--- a/clang.sh
+++ b/clang.sh
@@ -1,6 +1,6 @@
 package: Clang
-version: "v15.0.7"
-tag: "llvmorg-15.0.7-alice1"
+version: "v17.0.6"
+tag: "llvmorg-17.0.6-alice1"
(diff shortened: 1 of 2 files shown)


ktf:
This works on hyperloop. I am restarting the fullCI and macos tests.

davidrohr:
FullCI was green already, when the test reran it failed for some random reason.
MacOS ran multiple times, and always failed with different reasons 🤷‍♂️.

ktf:
Ok, merging this. Let's see...

vascobarroso:
@davidrohr @ktf FYI this change is breaking our RPM generation. I have created a ticket for Barth who will be back on Wednesday (https://its.cern.ch/jira/browse/OCONF-819) but we might need to revert this in the FLP Suite release branches if we need to release new RPMs in the meantime.

PR #5572, opened 2024-08-10
Update arrow.sh

aalkin:
Move to a new tag including extra cast functions for Gandiva


aalkin:
@ktf @davidrohr the error here is not related to my changes, could it be related to the change in where clang looks for the headers now?

ktf:
Breaks, I suspect this is the relocation issue I was worried about... :-/

PR #5573, opened 2024-08-10
Fixes for ONNXRuntime for compilation with gcc 14.2

davidrohr:


onnxruntime.sh
```diff
@@ -35,8 +35,8 @@ cmake "$SOURCEDIR/cmake"
       ${PROTOBUF_ROOT:+-DProtobuf_PROTOC_EXECUTABLE=$PROTOBUF_ROOT/bin/protoc}        \
       ${RE2_ROOT:+-DRE2_INCLUDE_DIR=${RE2_ROOT}/include}                              \
       ${BOOST_ROOT:+-DBOOST_INCLUDE_DIR=${BOOST_ROOT}/include}                        \
-      -DCMAKE_CXX_FLAGS="$CXXFLAGS -Wno-unknown-warning -Wno-unknown-warning-option -Wno-error=unused-but-set-variable" \
-      -DCMAKE_C_FLAGS="$CFLAGS -Wno-unknown-warning -Wno-unknown-warning-option -Wno-error=unused-but-set-variable"
+      -DCMAKE_CXX_FLAGS="$CXXFLAGS -Wno-unknown-warning -Wno-unknown-warning-option -Wno-error=unused-but-set-variable -Wno-error=maybe-uninitialized -Wno-error=template-id-cdtor" \
```
aalkin, line 38: Shouldn't it be `-Wno-template-id-cdtor`?

davidrohr:
Hm, but for sure what I have here works and is needed for gcc 14

ktf (commented):
Seems not to work.

davidrohr:
So, they added that option only with gcc 14, beforehand it is not available...
But then with gcc 14, this will be needed...
Marking this as draft for now until we bump to gcc 14.

davidrohr:
closing this, will open PR for GCC 14.2 which will include this

PR #5575, opened 2024-08-12
Enable check for wrong signed char assumptions

ktf:
Enable check for wrong signed char assumptions


ktf:
@davidrohr as we discussed today in the corridor.

davidrohr:
Then let's try it!

PR #5579, opened 2024-08-15
arrow: use clang from llvm installation to avoid stack probing issue with llvm@18

aalkin:
@ktf @davidrohr @singiamtel 
The issue with arrow breaking at runtime is solved by pointing it to the clang from brew-provided llvm, rather than apple clang. I've tested this on my mac with llvm 18.1.8. 

The version check can be probably improved, if you have any suggestions.


ktf:
There were two issues with this:

- The `--gcc-install-dir` needs to be provided with `=<something>` and it does not support missing the =, this is patched in `-alice6`.
- The directory provided must be where crtbegin.o is located, not the toplevel clang.

PR #5592, opened 2024-08-27
[WIP] add clang arch for RISCV

noferini:
I got a maching at CNAF to test the build on RISC-V
This option is needed to compile clang with such an arch
Still in the process of building, I'll remove WIP as soon as I check other packages don't need any further action


ktf:
As discussed privately, you probably want https://github.com/riscvarchive/riscv-gcc/commit/62da3c02d78e72667f086be9410a01f8ab8e2aa8 applied to GCC-Toolchain to avoid the spurious dependency on libatomic.

PR #5594, opened 2024-08-28
Don't reset CFLAGS on python recipe

singiamtel:
I think we currently build without -O2 when using our own OpenSSL, seems like an oversight

### DIFF
--- a/python.sh
+++ b/python.sh
@@ -54,7 +54,7 @@ esac
 if [[ $OPENSSL_ROOT ]]; then
   export CPATH="$OPENSSL_ROOT/include:$OPENSSL_ROOT/include/openssl:$CPATH"
   export CPPFLAGS="-I$OPENSSL_ROOT/include -I$OPENSSL_ROOT/include/openssl $CPPFLAGS"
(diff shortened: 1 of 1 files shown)


python.sh
```diff
@@ -54,7 +54,7 @@ esac
 if [[ $OPENSSL_ROOT ]]; then
   export CPATH="$OPENSSL_ROOT/include:$OPENSSL_ROOT/include/openssl:$CPATH"
   export CPPFLAGS="-I$OPENSSL_ROOT/include -I$OPENSSL_ROOT/include/openssl $CPPFLAGS"
-  export CFLAGS="-I$OPENSSL_ROOT/include -I$OPENSSL_ROOT/include/openssl"
```
ktf, line 57: Maybe what we should do is really drop CFLAFS, since most likely CPPFLAGS is enough. Could you check?

ktf: Ok, then it's needed...

ktf:
Yes. Do we have any other Python change which we can merge at the same time (so to avoid too many rebuilds?).

PR #5596, opened 2024-08-29
Disable warnings as errors until we move to GCC 13

ktf:
Disable warnings as errors until we move to GCC 13

-Wno-error=dangling-reference is available only there.

### DIFF
--- a/o2physics.sh
+++ b/o2physics.sh
@@ -34,7 +34,7 @@ cmake "$SOURCEDIR" "-DCMAKE_INSTALL_PREFIX=$INSTALLROOT"                    \
       ${CLANG_REVISION:+-DCLANG_EXECUTABLE="$CLANG_ROOT/bin-safe/clang"}    \
       ${CLANG_REVISION:+-DLLVM_LINK_EXECUTABLE="$CLANG_ROOT/bin/llvm-link"} \
       ${LIBUV_ROOT:+-DLibUV_ROOT=$LIBUV_ROOT}                               \
-      ${ALIBUILD_O2PHYSICS_TESTS:+-DO2PHYSICS_WARNINGS_AS_ERRORS=ON}
+      ${ALIBUILD_O2PHYSICS_TESTS:+-DO2PHYSICS_WARNINGS_AS_ERRORS=OFF}
 cmake --build . -- ${JOBS+-j $JOBS} install
 
 # export compile_commands.json in (taken from o2.sh)


ktf:
As expected, enabling the warnings as errors breaks as soon as we move to a new compiler. Disabling for now until we merged GCC 13.

davidrohr:
I don't fully understand why youbwrite until we merge gcc 13. Won't we still have the wrong warning with gcc 13?

Btw, instead of fully disabling all warnings you can use gcc's #pragme diagnostic to disable selected warnings in single files for some lines.

ktf:
The issue is that the warning is new, so passing `-Wno-error=dangling-reference` only works if we are at GCC 13. I prefer not to pollute code in with pragma diagnostic, if a temporary solution in alidist brings us where we want.

vkucera:
@ktf @ddobrigk

- Was this communicated to the community (WP14+WP4 meeting, email, Mattermost)? I don't see any announcement about this.
- Why was it not reverted immediately after merging https://github.com/alisw/alidist/pull/5541 ?
- Why has it not been reverted yet? (more than a month after fixing the reason for merging it!)
- After enabling the warnings as errors now, many PRs in O2Physics will fail because of unrelated warnings accumulated over the last month and people will have to fix them even though they didn't introduce them. How can we avoid this in the future?

vkucera:
> The issue is that the warning is new, so passing `-Wno-error=dangling-reference` only works if we are at GCC 13. I prefer not to pollute code in with pragma diagnostic, if a temporary solution in alidist brings us where we want.

Isn't `-Wno-unknown-warning` exactly for these situations?

ktf:
Looks like a good idea, actually...

vkucera:
> > The issue is that the warning is new, so passing `-Wno-error=dangling-reference` only works if we are at GCC 13. I prefer not to pollute code in with pragma diagnostic, if a temporary solution in alidist brings us where we want.
> 
> Isn't `-Wno-unknown-warning` exactly for these situations?

I guess I misunderstood the GCC documentation and `-Wno-unknown-warning` is not an actual flag but a placeholder used as an example for the description of the behaviour for unrecognised warnings. It would have been clearer if it was at least written as `-Wno-<unknown-warning>`.

ktf:
At least on clang they have `-Wno-unknown-warning-option`. Maybe we could enable it for macOS only.

vkucera:
> At least on clang they have `-Wno-unknown-warning-option`. Maybe we could enable it for macOS only.

Indeed. Not optimal but better than nothing.

PR #5601, opened 2024-09-04
alienpy 1.6.3 :: fix incompatible dependency

adriansev:
* Fix behavior for missing xrootd
* Adapt async-stagger usage to the new 0.4.0 version that is backward incompatible

### DIFF
--- a/python-modules-list.sh
+++ b/python-modules-list.sh
@@ -5,7 +5,7 @@ env:
     pip == 21.3.1; python_version < '3.12'
     pip == 24.0; python_version >= '3.12'
     setuptools == 59.6.0; python_version < '3.12'
-    setuptools == 69.1.0; python_version >= '3.12'
+    setuptools == 70.0.0; python_version >= '3.12'
     wheel == 0.37.1; python_version < '3.12'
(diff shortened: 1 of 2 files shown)


costing (dismissed):

adriansev:
@ktf it would seem that macos builders need an updated pip ... also note that this 1.6.3 is needed for all new installations that will break because of incompatible dependency so you decide if pip will be first fixed and then checks restarted or just merge this and then update pip ...

adriansev:
@singiamtel my home machine is on python 3.12 (fedora 40) and i seen no problems with it .. and from what i understood macos defaults anyway to 3.12 so the pip needs to be updated anyway (i think it's on fixed version in python-modules)

adriansev:
ping @singiamtel @ktf for macOS-arm i see a timeout in o2sim_G4, and O2/fullCI/alidist seems stuck

adriansev:
@ktf @singiamtel it seems that a test is stuck...

zensanp (approved):

adriansev:
ping @ktf @singiamtel

PR #5602, opened 2024-09-05
Bump to GCC 14.2

davidrohr:
Will at least fail with current CUDA 12.6, but want to check for other failures.

### DIFF
--- a/cmake.sh
+++ b/cmake.sh
@@ -1,6 +1,6 @@
 package: CMake
 version: "%(tag_basename)s"
-tag: "v3.28.1"
+tag: "v3.31.6"
 source: https://github.com/Kitware/CMake
 requires:
   - "OpenSSL:(?!osx)"
--- a/gcc-toolchain.sh
+++ b/gcc-toolchain.sh
@@ -1,6 +1,6 @@
 package: GCC-Toolchain
 version: "%(tag_basename)s"
-tag: v13.2.0-alice1
+tag: v14.2.0-alice2
 source: https://github.com/alisw/gcc-toolchain
 prepend_path:
   "LD_LIBRARY_PATH": "$GCC_TOOLCHAIN_ROOT/lib64"
--- a/onnxruntime.sh
+++ b/onnxruntime.sh
@@ -34,8 +34,8 @@ cmake "$SOURCEDIR/cmake"                                                        
       ${PROTOBUF_ROOT:+-DProtobuf_PROTOC_EXECUTABLE=$PROTOBUF_ROOT/bin/protoc}        \
       ${RE2_ROOT:+-DRE2_INCLUDE_DIR=${RE2_ROOT}/include}                              \
       ${BOOST_ROOT:+-DBOOST_INCLUDE_DIR=${BOOST_ROOT}/include}                        \
-      -DCMAKE_CXX_FLAGS="$CXXFLAGS -Wno-unknown-warning -Wno-unknown-warning-option -Wno-error=unused-but-set-variable -Wno-error=deprecated" \
-      -DCMAKE_C_FLAGS="$CFLAGS -Wno-unknown-warning -Wno-unknown-warning-option -Wno-error=unused-but-set-variable -Wno-error=deprecated"
+      -DCMAKE_CXX_FLAGS="$CXXFLAGS -Wno-unknown-warning -Wno-unknown-warning-option -Wno-error=unused-but-set-variable -Wno-error=deprecated -Wno-error=maybe-uninitialized -Wno-error=template-id-cdtor" \
+      -DCMAKE_C_FLAGS="$CFLAGS -Wno-unknown-warning -Wno-unknown-warning-option -Wno-error=unused-but-set-variable -Wno-error=deprecated -Wno-error=maybe-uninitialized -Wno-error=template-id-cdtor"
 
 cmake --build . -- ${JOBS:+-j$JOBS} install
 
--- a/python-modules.sh
+++ b/python-modules.sh
@@ -12,10 +12,43 @@ prepend_path:
   # If we need tensorflow-metal to work on Mac during subsequent builds, we
   # must use lib/python$pyver, not lib/python here.
   PYTHONPATH: "$PYTHON_MODULES_ROOT/lib/python/site-packages"
+prefer_system: ".*"
+prefer_system_check: |
+  # if we are in a virtualenv, assume people know what they are doing
+  # and simply use the virtualenv recipe.
+  if [ -n "$VIRTUAL_ENV" ]; then
+    echo "alibuild_system_replace: virtualenv"
+    exit 0
+  fi
+  # If not, either they are using the system python or they are using our own python.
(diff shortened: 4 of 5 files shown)


davidrohr:
Currently failing due to old json-c. We need to bump json-c, but the new version needs cmake instead of autoconf, so the recipe must be adapted.

davidrohr:
Now fails in `AliAlfred/DimRpcParallal` with
```
/sw/SOURCES/DimRpcParallel/v0.1.2/v0.1.2/src/dimrpcqueue.cpp: In member function 'void DimRpcQueue::processRequests()':
/sw/SOURCES/DimRpcParallel/v0.1.2/v0.1.2/src/dimrpcqueue.cpp:59:62: error: ignoring return value of 'std::lock_guard<_Mutex>::lock_guard(mutex_type&) [with _Mutex = std::mutex; mutex_type = std::mutex]', declared with attribute 'nodiscard' [-Werror=unused-result]
   59 |                 std::lock_guard<std::mutex>(this->accessMutex);
      |                                                              ^
In file included from /sw/slc7_x86-64/GCC-Toolchain/v14.2.0-alice2-local1/include/c++/14.2.0/mutex:47,
                 from /sw/SOURCES/DimRpcParallel/v0.1.2/v0.1.2/include/DimRpcParallel/dimrpcqueue.h:5,
                 from /sw/SOURCES/DimRpcParallel/v0.1.2/v0.1.2/src/dimrpcqueue.cpp:1:
/sw/slc7_x86-64/GCC-Toolchain/v14.2.0-alice2-local1/include/c++/14.2.0/bits/std_mutex.h:249:16: note: declared here
  249 |       explicit lock_guard(mutex_type& __m) : _M_device(__m)
```
I filed a bug report here: https://its.cern.ch/jira/browse/ALF-83 .

Also, as discussed with @ktf : binutils compilation fails randomly. Probably we should downgrade to the binutils of gcc-toolchian-13.2-alice1, which was working.

davidrohr:
@singiamtel @ktf : the slc9-aarch CI fails with
```
Downlod reference files in ci_test_dir .. OK
Uploading the ci_test_dir to GRID .. Could not upload reference files in 005_cp_dir.test/test.sh
Exception encountered! it will be logged to log.txt
Please report the error and send the log file and "alien.py version" output to Adrian.Sevcenco@cern.ch
If the exception is reproductible including on lxplus, please create a detailed debug report this way:
ALIENPY_DEBUG=1 ALIENPY_DEBUG_FILE=log.txt your_command_line
Failed test!!! Exitcode == 1
```

adriansev:
@davidrohr so, for the alien.py errors in alidist-slc9-aarch64 i would need the log file to see what happened .. weird is that test 004 and 006, both cp related, worked, so, if the actual log file is not available to debug what happened (on x86_64 Alma9 seems to work without problems) then just restart the test.

adriansev:
so for xjalienfs/alien.py these tests are run https://github.com/adriansev/jalien_py/tree/master/tests
in case of failure a log.txt file will be found in the respective directory. but i suspect that is an transient error so it should be enough to restart that failed test

ktf:
@davidrohr AliRoot is now fine.

ktf:
@davidrohr do you understand the issue with CUDA and the one with xmmintr? They both seem legit, and I do not understand why we did not see them with GCC 13.

davidrohr:
For CUDA it is clear, since GCC14 is not yet supported. We have to wait for a new CUDA.
For the GPU Standalone benchmark, it is because I am using X86 intrinsics. I will just disable the build on ARM, as I do for MacOS here https://github.com/AliceO2Group/AliceO2/blob/fb8e068eff4fba325c75b3fa9c77e59db10a50a6/GPU/GPUTracking/CMakeLists.txt#L562.

davidrohr:
@ktf : Now only the FullCI remains red, will stay like this until we bump CUDA.

ktf:
Changed to WIP to avoid retesting.

davidrohr:
This is ready to go now, if it passes the CIs except for the old FullCI, which can be removed now.

davidrohr:
@ktf @singiamtel : The generators CI fails with
```
## sw/BUILD/Herwig-latest/log
--
make[1]: Entering directory `/sw/BUILD/833085d7464c68d7bc5695a8c0254dc33699a578/Herwig/src'
gengetopt < herwigopts.ggo
/bin/sh: gengetopt: command not found
make[1]: *** [herwigopts.c] Error 127
make[1]: Leaving directory `/sw/BUILD/833085d7464c68d7bc5695a8c0254dc33699a578/Herwig/src'
make: *** [all-recursive] Error 1
```
Is that known?

davidrohr:
@ktf : FullCI9 and slc9 are green. Not sure about the Generators CI? The old FullCI we can ignore. As I wrote, you can disable them know. With merging, let's please wait until FLPs create their next FLPSuite. I'd aim for merging Friday evening.

ktf:
I would say we merge on monday, then. There might be cleanups to be done on the CVMFS side which I do not want to do over the weekend. AliGenerators seems fine now.

ktf:
The way I would do it is:

- First we merge #5785, which is already cached
- We cache the byproducts of this one
- We announce it on Monday in the meeting
- We merge this one

Given there is still Quark Matter stuff going on, I do not want to end up in front of the firing squad for changing the compiler without announcing it.

Barthelemy:
I couldn't test it yet because the build is broken after https://github.com/alisw/alidist/pull/5785

ktf:
@singiamtel can we cache this PR and then merge it? Thanks.

davidrohr:
> @singiamtel can we cache this PR and then merge it? Thanks.

@ktf @singiamtel : I would recommend we do this together with bumping CMake: https://github.com/alisw/alidist/pull/5792
Both will trigger a more or less full rebuild.

singiamtel:
> > @singiamtel can we cache this PR and then merge it? Thanks.
> 
> @ktf @singiamtel : I would recommend we do this together with bumping CMake: #5792 Both will trigger a more or less full rebuild.

Should I merge the CMake PR into this one? So the hashes on the cache run match

davidrohr:
> > > @singiamtel can we cache this PR and then merge it? Thanks.
> > 
> > 
> > @ktf @singiamtel : I would recommend we do this together with bumping CMake: #5792 Both will trigger a more or less full rebuild.
> 
> Should I merge the CMake PR into this one? So the hashes on the cache run match

Fine with me

singiamtel:
Cache run ongoing @ https://alijenkins.cern.ch/job/CacheO2Package/113/

ktf:
I am not sure bumping CMake will actually trigger a rebuild, because it's a build requires. That said, fine with me as well.

singiamtel:
As discussed, I merged both #5661 and #5792 in this PR, and rebased so we have the right commits for each.

New cache run is ongoing in https://alijenkins.cern.ch/job/CacheO2Package/114/

singiamtel:
The cache build is done. This tests should be running against exactly the same code as before, I think it should be fine to merge

PR #5603, opened 2024-09-05
Bump json-c to v17.0

singiamtel:
As suggested by @davidrohr, this should bypass the reversed calloc arguments error

### DIFF
--- a/json-c.sh
+++ b/json-c.sh
@@ -8,7 +8,10 @@ build_requires:
 ---
 #!/bin/bash -e
 rsync -a --delete --exclude '**/.git' --delete-excluded $SOURCEDIR/ ./
-./cmake-configure --disable-shared --enable-static --prefix="$INSTALLROOT"
+
+mkdir build
+cd build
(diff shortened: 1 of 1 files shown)


json-c.sh
```diff
@@ -1,33 +1,19 @@
...
+  - alibuild-recipe-tools
 ---
 #!/bin/bash -e
 rsync -a --delete --exclude '**/.git' --delete-excluded $SOURCEDIR/ ./
-autoreconf -ivf
-./configure --disable-shared --enable-static --prefix="$INSTALLROOT"
+
+mkdir build
+cd build
+../cmake-configure --disable-shared --enable-static --prefix="$INSTALLROOT"
```
ktf, line 15: Shouldn't we simply use cmake? It would also allow us to get rid of the rsync above.

ktf (changes requested):
If we need to touch this, then let's just modernise the whole package.

ktf (approved):

PR #5605, opened 2024-09-06
Always ensure CLANG_EXECUTABLE exists when building arrow

singiamtel:
Currently not all paths ensure this variable is defined (it will be
empty on macs with no llvm from homebrew, for example)

### DIFF
--- a/arrow.sh
+++ b/arrow.sh
@@ -70,6 +70,7 @@ case $ARCHITECTURE in
        CLANG_EXECUTABLE="$(brew --prefix llvm)@17/bin/clang"
      fi
    fi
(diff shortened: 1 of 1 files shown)


aalkin:
llvm from homebrew is a requirement for arrow on mac, it needs the code generation libraries for its JIT. Consequently it needs to see the clang from the homebrew installation to ensure compatibility, it will fail in runtime otherwise.

aalkin:
I think we should fail, indeed. @ktf what do you suggest?

ktf:
I think we should indeed make sure that a proper LLVM + Clang installation is found. I would not make it optional.

PR #5609, opened 2024-09-10
Include AliGenerators in O2Sim

jackal1-66:
AliGenerators does not build at the moment on osx and ARM based machines, so the support for those is explicitely removed in the recipe.

### DIFF
--- a/o2sim.sh
+++ b/o2sim.sh
@@ -6,6 +6,7 @@ requires:
   - QualityControl
   - AEGIS
   - EVTGEN:(?!osx)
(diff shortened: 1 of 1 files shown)


jackal1-66:
@ktf could you have a look at this? It would be important to include AliGen as soon as possible to go forward with other tasks

ktf:
Merged to continue with the work. I would prefer a more finegrained approach though, where we get rid of AliGenerators as a project (since now O2sim serves the same purpose) and disable only the generators which are known not to work.

PR #5599, opened 2024-09-03
Do not build gprofng on ubuntu

ktf:

PR #5613, opened 2024-09-11
Disable ngprofng in binutils

davidrohr:
This is also failing sometimes with the current GCC 13.2 for me, so let's just remove it.
Then we could also use the newest binutils for GCC 14.

### DIFF
--- a/gcc-toolchain.sh
+++ b/gcc-toolchain.sh
@@ -73,6 +73,7 @@ pushd build-binutils
                         --enable-lto                           \
                         --enable-plugins                       \
                         --enable-threads                       \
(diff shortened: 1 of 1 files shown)


ktf:
I had https://github.com/alisw/alidist/pull/5599, do you see it also on centos?

davidrohr:
> I had #5599, do you see it also on centos?

Sorry, didn't see that.
Yes, it happens also on CentOS. Only on Gentoo I don't see it.

ktf:
Ok, I closed mine. Maybe wait for tonight before merging, because it will trigger a full rebuild.

PR #5614, opened 2024-09-12
Revert "Include AliGenerators in O2Sim"

ktf:
Reverts alisw/alidist#5609


ktf:
This is breaking the CI due to madgraph failing to compile. Reverting.

PR #5620, opened 2024-09-17
Bump ONNX and workaround warning

ktf:
Needed by macOS Sequoia


davidrohr:
Since there is onnxruntime 1.19 meanwhile, why don't we bump to it? To avoid that we get stuck like last time?

pzhristov:
@singiamtel I will test it today an let you know if it is compatible.

ktf:
Good enough for now, no objections to bump to the most recent version, once supported.

PR #5621, opened 2024-09-17
Introduced skimmed version of AliGenerators for O2sim integration

jackal1-66:
In the same PR an upgrade of EPOS4 has also been introduced in order to make it work with the current O2sim generation mechanism via cmd. AliGenO2 contains the basic packages of AliGenerators. More might be added in the future if the community requires them (Herwig7 is temporarily disabled due to MadGraph not being able to compile due to wget based issues).

### DIFF
--- a/aligeno2.sh
+++ b/aligeno2.sh
@@ -9,6 +9,7 @@ requires:
   - Rivet
   - lhapdf-pdfsets
   - JETSCAPE
+  - CRMC
   - EPOS4
   - STARlight
 build_requires:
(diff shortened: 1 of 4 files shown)


jackal1-66:
There was a small typo in the tag name of EPOS4, now it's fixed and the CI shouldn't fail on the git clone.

ktf:
Looks good to me.

maireiphc (approved):

sawenzel (approved):

PR #5623, opened 2024-09-17
Try to use original TPC fec utils repository

singiamtel:
Testing to see if the CI can access this repository too. This way we could get rid of maintaining the mirror

### DIFF
--- a/tpcfecutils.sh
+++ b/tpcfecutils.sh
@@ -10,7 +10,7 @@ requires:
   - LLA
 build_requires:
   - CMake
-source: https://gitlab.cern.ch/ALICEPrivateExternals/alice-tpc-fec-utils.git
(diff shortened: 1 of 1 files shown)


vascobarroso:
@singiamtel IIRC we could not use the original TPC repo because of permissions when building RPMs (it's a private repo).

vascobarroso:
@singiamtel when generating our FLP RPMs, we build the SW using aliBuild.
I remember we had access problems in the past whenever there was a new version of this package and we had to rebuild.
It's true that if both the original and our mirror repo are private, we should have the same problem. 
I don't remember what was the actual solution in the past, maybe @Barthelemy can investigate what happens if we need to rebuild this package.

Barthelemy (approved):

Barthelemy:
ok for me

PR #5624, opened 2024-09-17
Drop char to int conversion check

ktf:
Buggy and gives confusing reports. Will reenable once less noisy.


ktf:
@shahor02 as discussed.

PR #5625, opened 2024-09-17
Bump to ROOT v6-32-06-alice1

ktf:
- Make sure ROOT works on macOS 15 Sequoia (requires enabling ROOT7)
- Disable VecGeom on macOS (apparently not used)
- Include improvements to geometry handling by @sawenzel 
- Include reverts necessary to have pythia6 and proof working

### DIFF
--- a/o2.sh
+++ b/o2.sh
@@ -23,7 +23,7 @@ requires:
   - libuv
   - libjalienO2
   - cgal
-  - VecGeom
+  - "VecGeom:(?!osx.*)"
   - FFTW3
   - ONNXRuntime
   - MLModels
--- a/root.sh
+++ b/root.sh
@@ -1,6 +1,6 @@
 package: ROOT
 version: "%(tag_basename)s"
-tag: "v6-32-02-alice1"
+tag: "v6-32-04-alice1"
 source: https://github.com/alisw/root.git
 requires:
   - arrow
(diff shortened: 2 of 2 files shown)


ktf:
@aalkin @pzhristov with this I get to O2.

@aalkin, do you have a PR for the std::enable_if issue?

ktf:
@sawenzel @pzhristov this should work already for O2.

@pzhristov can you remind me what you need to add to the root.sh recipe, as discussed offline?

sawenzel:
Thanks @ktf.

@ktf @pzhristov : I am still a little worried that we bring back TPythia6 into ROOT, since there will **for sure** be duplicated symbols in AEGIS and ROOT. This may not materialize as a problem during compilation, but only during runtime. So I would like to propose to check if event generation still works ok during runtime. A suitable test would be to run `o2-sim-dpl-eventgen --nEvents 1 --generator external --configFile ${O2DPG_ROOT}/MC/config/PWGEM/ini/GeneratorLFCocktailpp.ini` in the O2sim environment (doesn't really need O2Physics).

pzhristov:
@sawenzel We can avoid the problems with the duplicated symbols, if we return AEGIS to version v1.5.3.

ktf:
Merging this since it's green. I will now revert aegis as well.

PR #5626, opened 2024-09-18
Don't fix an old GCC version since meanwhile we have generally bumped GCC

davidrohr:
@ktf @singiamtel : I assume these were leftovers, from when the default GCC was 8.x?


ktf:
I would say so. Merging.

PR #5627, opened 2024-09-19
fix the moderncppkafka recipe

Barthelemy:


knopers8 (approved):

vascobarroso:
@ktf any input on this ?

ktf:
Only what I commented. It's not clear  to me why you would need to expose the _ROOT variable to the build. Usually that means that something else is broken in the CMakeLists.

Barthelemy:
I am not 100% sure either, but at this stage I need it to be able to carry on. Many dev setups are broken.

Barthelemy:
could we merge it please ?

ktf:
Adding this variable is definitely the wrong thing to do, just like using alienv for the development setup (you should use the init.sh). You should fix your CMakeLists and / or switch to use init.sh for the dev environments.

PR #5630, opened 2024-09-20
Update root.sh

sawenzel:
Bump ROOT to ALICE specific version v6-32-04-alice2. (Needed, for instance, to get improvements in TGeoParallelWorld for ITS simulation).

### DIFF
--- a/root.sh
+++ b/root.sh
@@ -1,6 +1,6 @@
 package: ROOT
 version: "%(tag_basename)s"
-tag: "v6-32-02-alice1"
+tag: "v6-32-04-alice2"
 source: https://github.com/alisw/root.git
 requires:
   - arrow


pzhristov:
@ktf Do you know when the new tag v6-32-06 is coming? I reverted the removal of montecarlo/pythia6 and with this modification we can use v6-32-xx for AliRoot/AliPhysics. If we go this way, probably some small modification is needed in AEGIS, to be checked later.

ktf:
I think they are basically ready and just waiting to update the builders to have a binary release for macOS Sequoia.

dpiparo:
Hi @pzhristov , we froze the branch today and expect to release by early next week.

sawenzel:
Ok. We should just make sure, one way or another to have the TGeo improvements included in the next root-alidist-bump.

ktf:
Closing in favour of #5625

PR #5634, opened 2024-09-23
GCC: avoid naming conflicts

ktf:
- do not install GMP used to compile GCC. Other externals depend on their own
  GMP since we do not want to bind the lifetime of GMP upgrades to the lifetime
  of GCC ones. This was done to avoid reproducibility issues in e.g. FastJet
  and CGal.
- Remove the sim folder which conflicts with FairRoot.

### DIFF
--- a/gcc-toolchain.sh
+++ b/gcc-toolchain.sh
@@ -121,7 +121,6 @@ pushd build-gcc
                    --disable-nls
   make ${JOBS+-j $JOBS} bootstrap-lean MAKEINFO=":"
   make install MAKEINFO=":"
-  (if cd gmp || cd ../gmp; then make install MAKEINFO=":"; fi)
   hash -r
 
   # GCC creates c++, but not cc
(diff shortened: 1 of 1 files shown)


ktf:
As discussed with @Barthelemy. This fixes the issue for the single path installations where there is now conflicting gmp.h headers. AFAICT, there should not be any reason to install the GMP used to compile GCC.

Barthelemy:
thank you very much, I am testing it on the build machine

Barthelemy:
it is working

Barthelemy (approved):

Barthelemy:
could we merge it please ?

ktf:
In a bit. This triggers a mass rebuild. Is the problem with the sim folder understood?

Barthelemy:
@ktf no, you said "I am checking what brings it in." so I was waiting 😄

ktf:
@Barthelemy I could not find a clean way of removing the folder, so here it is with the rm...

Barthelemy:
thank you !

Barthelemy:
it works for us, you can merge

ktf:
Caching done, merging.

PR #5641, opened 2024-09-24
Fix broken rivet modulefile

ktf:


ktf:
I think this is what was really meant...

ktf:
@jackal1-66 can you cross check and then we merge?

jackal1-66 (approved):

jackal1-66:
New commit is working as expected

PR #5642, opened 2024-09-25
Don't hard-code the number of parallel builders for mesos

Barthelemy:
I don't think that we should hard code the number of parallel jobs for the build of mesos. 
What about `JOBS/2` ?


Barthelemy:
@ktf it seems that Adam is still the automatic reviewer here.

ktf:
You can change that in "CODEOWNERS"

PR #5654, opened 2024-10-08
Fix clang prefer_system_check

ktf:
This should make it behave correctly also on Linux with --always-prefer-system.


ktf:
@Barthelemy can you check if this allows you to work with the minimal set of changes to the RPM creation?

Barthelemy:
arrow can build with this change (OCTRL-935)

Barthelemy:
I confirm that I could build with this PR

PR #5182, opened 2023-10-13
python-modules-list: explicit list and constrains for xjalienfs depen…

adriansev:
…dencies(and their dependencies)

Hi @TimoWilken ! This is the explicit and complete list of dependency tree of xjalienfs package.

PR #5655, opened 2024-10-08
Remove readline installation from xjalienfs recipe

singiamtel:
Since https://github.com/alisw/alidist/pull/5534 we depend on gnureadline globally anyway, no point in reinstalling it in the recipe

CC @ktf @adriansev @sawenzel

### DIFF
--- a/xjalienfs.sh
+++ b/xjalienfs.sh
@@ -19,13 +19,6 @@ prepend_path:
 # works inside and outside a virtualenv, but unset VIRTUAL_ENV to make sure we
 # only depend on stuff we installed using our Python and Python-modules.
 
-# on macos try to install gnureadline and just skip if fails (alienpy can work without it)
-# macos python readline implementation is build on libedit which does not work
-[[ "$ARCHITECTURE" ==  osx_* ]] && { env -u VIRTUAL_ENV ALIBUILD=1 \
-    python3 -m pip install --force-reinstall \
-    --target="$INSTALLROOT/lib/python/site-packages" \
-    gnureadline || : ; }
-
 env -u VIRTUAL_ENV ALIBUILD=1 \
     python3 -m pip install --force-reinstall \
     --target="$INSTALLROOT/lib/python/site-packages" \


adriansev (approved):

adriansev:
@ktf @singiamtel if we're gonna put a xjalienfs dependency in python-modules-list we could be consistent and add the other dependencies too .. for reference, in a fresh venv on alma9 i got this:
```
async-stagger==0.3.1
certifi==2024.8.30
cffi==1.17.1
charset-normalizer==3.3.2
cryptography==43.0.1
idna==3.10
markdown-it-py==3.0.0
mdurl==0.1.2
pycparser==2.22
Pygments==2.18.0
pyOpenSSL==24.2.1
requests==2.32.3
rich==13.9.2
typing_extensions==4.12.2
urllib3==2.2.3
websockets==13.1
```

and on fedora 40 (which have python 3.12.6)
```
async_stagger==0.4.0.post1
certifi==2024.8.30
cffi==1.17.1
charset-normalizer==3.3.2
cryptography==43.0.1
idna==3.10
markdown-it-py==3.0.0
mdurl==0.1.2
pycparser==2.22
Pygments==2.18.0
pyOpenSSL==24.2.1
requests==2.32.3
rich==13.9.2
setuptools==75.1.0
urllib3==2.2.3
websockets==13.1
wheel==0.44.0
```

adriansev:
> I was thinking about this the other day, wouldn't it be better to have them all in https://github.com/adriansev/jalien_py/blob/master/setup.py? I think it's what we run when doing `pip install` on this recipe anyway
> 
> Also linking to the last PR on this discussion #5182

there are already there see https://github.com/adriansev/jalien_py/blob/master/setup.py#L19-L31
of course without gnureadline because from my point of view is not really required if it is a normal python (compiled with readline) ... but i could detect the darwin arch and add it as dependency and this would solve the problem...

sawenzel (approved):

adriansev:
so, this is really weird .. the xjalienfs test fail [here](https://github.com/adriansev/jalien_py/blob/master/tests/005_cp_dir.test/test.sh#L38)
and the problem is that only alidist-slc9-aarch64 have problems while all others are ok ...
this can only be investigated directly on the machine as no matter what the problem is, it needs also to be fixed and tested :)

adriansev:
FWIW on my Manjaro based ARM sbc (`Linux n2plus 6.4.3-1-MANJARO-ARM-ODROID #1 SMP PREEMPT Wed Aug 30 20:04:07 UTC 2023 aarch64 GNU/Linux`) the tests worked without problems, so at least we know that it's not an architecture problem .. xrootd was installed with pip in a venv (but it required ninja usage with `export CMAKE_GENERATOR=Ninja`)

PR #5656, opened 2024-10-09
[OCTRL-935] Remove build_requires for rsync

Barthelemy:
because we don't put it for any other recipe


Barthelemy:
@ktf could you merge this please ?

ktf:
This will rebuild ROOT, would you mind waiting until tomorrow to avoid delaying tonight tags? Also, what is wrong with the check?

Barthelemy:
it caused an issue in the RPM generation which made me question the pertinence of having a dependency on rsync in a single recipe when all recipes use rsync. 

It is not urgent, I found a way to circumvent the problem on my side

PR #5663, opened 2024-10-16
Add certificates as a direct dependency of O2

ktf:


Barthelemy:
thank you

Barthelemy:
`ERROR: AliEn-CAS should be spelt AliEn-CAs.`

PR #5666, opened 2024-10-18
xjalienfs :: temporary disable tests for slc9_aarch64

adriansev:


adriansev:
O2/alidist-cs8 failed due to git error

PR #5679, opened 2024-11-05
[O2-5506] Bump libffi and use our patched version

singiamtel:
Should bypass both the macOS automake 1.17 issues, and the cfi_startproc issue

Automake 1.17 workaround: https://github.com/libffi/libffi/issues/853#issuecomment-2306898507
cfi_startproc workaround: https://github.com/libffi/libffi/commit/8308bed5b2423878aa20d7884a99cf2e30b8daf7

Also taking the chance to modernize the modulefile generation

```diff
$ diff modulefile.old modulefile.new
9,12c9,15
< module load BASE/1.0
< # Our environment
< set LIBFFI_ROOT $::env(BASEDIR)/libffi/$version
< prepend-path LD_LIBRARY_PATH $LIBFFI_ROOT/lib
---
> if ![ is-loaded 'BASE/1.0' ] {
>  module load BASE/1.0
> }
>
> set PKG_ROOT $::env(BASEDIR)/libffi/$version
> prepend-path LD_LIBRARY_PATH $PKG_ROOT/lib
>
```
CC @ktf

### DIFF
--- a/libffi.sh
+++ b/libffi.sh
@@ -1,5 +1,5 @@
 package: libffi
-version: v3.4.4-alice1
+version: v3.2.1-alice1
 build_requires:
   - "autotools:(slc6|slc7)"
   - "GCC-Toolchain:(?!osx)"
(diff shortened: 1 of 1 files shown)


ktf:
Looks good. Thanks.

ktf:
This actually breaks because `--disable-multi-os-directory` is not doing the right thing (libraries are in lib64 rather than lib). I am patching the CVMFS installation to point to the correct path, but we should understand what is going on.

PR #5680, opened 2024-11-05
Compile starlight against dpmjet

mbroz84:
STARlight can be compiled against DPMJET, in that case a separate library is created (libDpmJetLib) that uses routines from DPMJET to simulate photo-nuclear interactions. UPC group would like to use this mode.

The libDpmJetLib is somehow not installed and remains in the build directory so I added explicit line to move it, maybe there is another solution.

There is an adjustment in STARlight itself needed to make this work. I can make the PR myself, but I will appreciate an expert opinion on issue 2.
1. The libDpmJetLib was build as static, I added option to build it as Shared. 
2. There is an issue with "FIND_LIBRARY" in the CMake module when the library path defined as a environment variable in the alidist/starlight.sh is preceded by ENV. Library is then not found when this is build via aliBuild. It definitely works in standalone installation (to see what I mean exactly have a look in the attached patch).
[STARlight.patch](https://github.com/user-attachments/files/17633765/STARlight.patch)

### DIFF
--- a/dpmjet.sh
+++ b/dpmjet.sh
@@ -22,6 +22,9 @@ cmake  $SOURCEDIR                           \
 
 make ${JOBS+-j $JOBS} install
 
+cp -r $SOURCEDIR/include $INSTALLROOT/.
+cp -r $SOURCEDIR/dpmdata $INSTALLROOT/.
+
 # Modulefile
 MODULEDIR="$INSTALLROOT/etc/modulefiles"
 MODULEFILE="$MODULEDIR/$PKGNAME"
--- a/starlight.sh
+++ b/starlight.sh
@@ -2,6 +2,7 @@ package: STARlight
 version: "20240617"
 tag: 196adefb9b840587d374b675e288b2a98fb5df0e
 requires:
+  - DPMJET
   - HepMC3
 build_requires:
   - CMake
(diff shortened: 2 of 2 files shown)


mbroz84:
It needs this: https://github.com/alisw/DPMJET/pull/7
And the attached STARlight patch

jackal1-66:
I included the DPMJET PR in my build, but I overlooked the STARlight patch you suggested. Indeed it doesn't build without that. Going to check now the suggested changes

mbroz84:
I can send the PR to STARlightsim anytime (developers are our colleagues from ALICE). I have there the library build option for the time being, let me know if anything else is needed.

jackal1-66:
I discussed with @ktf and there's an easy way to do the second point: `export DPMJET_DIR=$DPMJET_ROOT` right before the cmake is done in the starlight recipe, this way you don't need to edit FindDPMJET.cmake.
To summarize:
1) Open the PR with STARlight to for the dynamic library type of DPMJET: this way when it's merged we get directly the last version with the shared library enabled. This is, I would say, the first step so that we edit our own starlight recipe to reflect the generator, and we wait for the DPMJET PR to be merged and for the new tag
2) Add the DPMJET_DIR export in the starlight recipe
3) Fix the indentation in the recipe (please) 
4) Remove the dpmdata folder copy in the dpmjet recipe (it is done automatically already) 
5) When point 1 is fully completed we update the versions of both starlight and dpmjet in the recipes, we retest that the build works smoothly and we merge.

mbroz84:
1.
- STARlight PR is merged: https://github.com/STARlightsim/STARlight/pull/16
- I added a line in the DMPJET CmakeLists to install headers
2. 3. 4. : Done
5. I also adjusted the tags in both recipies to point to the new versions.

mbroz84:
Hi @jackal1-66, DPMJET fork was updated, we can proceed with this PR

jackal1-66:
Hi @mbroz84 waiting for https://github.com/alisw/alidist/pull/5702 to be merged, otherwise CI tests will fail.

jackal1-66 (approved):

jackal1-66:
There's a problem with the CIs right now which is unrelated to this PR. I would still wait for the tests to be successful since merging this affects O2sim.

jackal1-66:
@ktf could you please approve to start the last CI test?

jackal1-66:
This PR can be merged then @singiamtel, thank you!

mbroz84:
+async-label async-2023-pbpb-apass4

PR #5682, opened 2024-11-07
Bump DDS to 3.11

davidrohr:
@lkrcal @ktf : This is needed to run with boost 1.86.

@ktf : I see we have some custom patches on top of our current boost 1.86. Do we need to port them when bumping boost, or should we try the new version?

### DIFF
--- a/dds.sh
+++ b/dds.sh
@@ -1,6 +1,6 @@
 package: DDS
 version: "%(tag_basename)s"
-tag: "3.10"
+tag: "3.11"
 source: https://github.com/FairRootGroup/DDS
 requires:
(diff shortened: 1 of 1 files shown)


ktf:
@davidrohr The commit which I think we might still need is: 

https://github.com/alisw/boost/commit/9394fc9d7bbd80c25b7ca48dee0146beadc700e0

because it has implications on how the shared memory is mapped and that might or might not trigger the OOM when running inside a container. In principle we did raise the limits, so maybe it's not needed anymore, however we would need to carefully check, especially on the Grid.

PR #5685, opened 2024-11-11
Make sure libraries are in lib

ktf:
Make sure libraries are in lib


ktf:
Currently broken when running alien. Merging.

PR #5686, opened 2024-11-12
Add O2PDPSuite to O2Suite

ktf:
O2Suite (which is the union of FLP and PDP packages) is what is actually tested.

### DIFF
--- a/o2suite.sh
+++ b/o2suite.sh
@@ -5,8 +5,7 @@ requires:
   - coconut
   - Control-Core
   - Control-OCCPlugin
-  - O2
-  - O2Physics
+  - O2PDPSuite
   - "ReadoutCard:(slc.*x86)"
   - Readout
   - QualityControl


ktf:
@sawenzel this is the reason why O2sim was not tested. In the CI we actually use O2Suite (i.e. the union of FLP and PDP software) however O2Suite probably predates O2PDPSuite and it was missing the dependency. This will now test whatever is included as a dependency on O2PDPSuite.

As discussed, @singiamtel should implement a check which makes sure that when a new package is introduced it should belong to one of the packages we test at toplevel.

sawenzel (approved):

Barthelemy:
That currently breaks our RPM generation. 
```
Error: 
 Problem 1: conflicting requests
  - nothing provides o2-ODC >= 0.83.2-1 needed by o2-O2PDPSuite-daily_20241113_0000-1.slc8.x86_64 from @commandline
 ```
I will have a look later today

ktf:
Why are you using O2Suite for the RPM generation?

davidrohr:
I think O2Suite has always been the FLP top level Package.
You will need to disable ODC I guess.

ktf:
Why do we have O2FLPSuite, then?

ktf:
Nevermind, we don't... Then either we disable the unwanted components at RPM generation or we introduce a separate step. I do not think it makes sense to not test O2PDPSuite in the CI, in particular for alidist.

davidrohr:
Well, the FullCI tests O2PDPSuite already.
But I agree it would make sense if the normal CI would test it as well.
However, I am not sure if it is the right thing to have everything in O2Suite. Some packages like ODC should not be in there for FLPs. And disabling packages manually for RPM generation is also a bit bogus.
Why don't we create a package TopLevelCI, which includes everything we want to test?

ktf:
FullCI tests PRs in O2, not in alidist. In particular it does not test some of the generators which are included via O2sim (and this broke as while ago).

I would claim that the toplevel package IS O2Suite and that we need to introduce O2FLPSuite.sh for FLP specific purposes (like we did for O2PDPSuite).

Barthelemy:
(I am in meeting the whole morning, I will comment afterwards)

ktf:
Nevermind again. I forgot we have full CI also for alidist. Then this PR is indeed useless, since the effect we want to achieve (testing PDPSuite) is already achieved.

ktf:
@sawenzel then I am back a square 1. I see no reason why the O2sim and its dependencies should not be tested already by the current CI setup.

davidrohr:
Well, they should be tested in the FullCI, but fullci e.g. does not run ctests.
I still think it would be good to have a top level package for tests.
But IMHO it is easier to introduce a special package for the CI than for the FLP build. We don't know who else is using O2Suite.

ktf:
> Well, they should be tested in the FullCI, but fullci e.g. does not run ctests.

Then I do not understand why it did not catch the issue in O2sim...I understood it was a build issue there.

> But IMHO it is easier to introduce a special package for the CI than for the FLP build. We don't know who else is using O2Suite

Fine with me as long as we do not introduce an extra check and we reuse the one which is currently using O2Suite.

sawenzel:
> @sawenzel then I am back a square 1. I see no reason why the O2sim and its dependencies should not be tested already by the current CI setup.

I don't know but this is what happened. PR https://github.com/alisw/alidist/pull/5621 introduced AliGenO2 as required for O2sim. This should for instance build, among others, the CRMC generator. Yet, in the [logs](https://ali-ci.cern.ch/alice-build-logs/alisw/alidist/5621/99859fb74b2c23bd0c0ae33659ff805986c80d88/build_O2_fullCI_alidist/pretty.html) of the fullCI build, I do not see such trace. The CI was green, yet the AliGenO2 package did not build later on in Jenkins

PR #5699, opened 2024-11-26
Support building with master of ROOT

ktf:


ktf:
This is already the default value and the option will disappear when using the next version of ROOT.

PR #5701, opened 2024-11-28
New Graniitti generator and update of o2sim with Upcgen and Graniitti

pbuehler:


o2sim.sh
```diff
@@ -8,6 +8,8 @@ requires:
   - AliGenO2:(?!osx|.*aarch64)
   - EVTGEN:(?!osx)
   - STARlight:(?!osx)
+  - Upcgen:(?!osx)
+  - Graniitti:(?!osx)
```
sawenzel, line 12: could we add these generators to AliGenO2?

pbuehler: > could we add these generators to AliGenO2?

Hi @sawenzel  To my understanding Upcgen and Graniittti are on the same level as STARlight. That is why I added these two generators to o2sim. In any case I think that STARlight, Upcgen, and Graniitti should be at the same place. I however leave it up to you to decide where to put these. Please let me know and I will make the necessary changes.

sawenzel (dismissed):

o2sim.sh
sawenzel: That's a bit historic. We could just have all generators come from AliGenO2. But we can do this cleanup also later. It's only something cosmetic.

pbuehler: So I moved now the generators from o2sim.sh to aligeno2.sh. Hope this is ok.

sawenzel (approved):

PR #5703, opened 2024-11-29
XRootD 5.7.2

adriansev:
XRootD 5.7.2 was released and together with 5.7.1 it has some XrdCl and Python fixes
see https://github.com/xrootd/xrootd/blob/v5.7.2/docs/ReleaseNotes.txt

### DIFF
--- a/xrootd.sh
+++ b/xrootd.sh
@@ -1,6 +1,6 @@
 package: XRootD
 version: "%(tag_basename)s"
-tag: "v5.7.0"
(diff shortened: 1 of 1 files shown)


adriansev:
@ktf @singiamtel i'm not sure what error O2/fullCI/alidist have .. the full error log does not show any clear problem

adriansev:
@ktf @singiamtel the tests are green, thanks a lot!

PR #5707, opened 2024-12-05
Refresh zlib.sh

ktf:
- Bump to v1.2.13 as requested by FLP
- Generate modulefile on via alibuild-recipe-tools
- Drop support for CMS architecture strings... ;-)
- NOT switching to CMake to avoid circular dependency via OpenSSL.


ktf:
@Barthelemy this should fix your issues and do a long due cleanup.

Barthelemy:
thank you

PR #5709, opened 2024-12-09
Use dummy folder for PYTHON_ROOT

ktf:

PR #5710, opened 2024-12-09
Remove Python-system workaround

singiamtel:
The new changes to the Python recipe should be enough to properly use
the Python system when relevant

(#5660)

Also taking the chance to clean up several recipes


ktf:
@Barthelemy could you have a look. This should affect mostly the o2-dataflow builds.

Barthelemy:
Do you mean that this PR must be modified following https://github.com/alisw/alidist/pull/5709 ?

ktf:
@Barthelemy how about we merge this now? It's clearly a simplification and now that we can easily track system tools I see no advantages. I can commit to spend next week debugging, if needed on your side. In case you agree, can you give it a spin with the FLP RPM machinery before we merge it?

ktf:
Discussed offline with @Barthelemy. Let's give this a spin, it should be a clear win in terms of simplification, so we can probably get it to work in any case.

PR #5712, opened 2024-12-09
[OMON-793] Add dependency on kafka and protobuf in Monitoring

Barthelemy:


sy-c (changes requested):
Thanks for this fix. However I'm not sure if it's right to do so (see comment in JIRA).

Barthelemy:
we can close it for what I am concerned.

ktf:
Can this be closed?

vascobarroso (approved):

vascobarroso:
@sy-c looks like one of the tests is failing, could you have a look ?

sy-c:
@vascobarroso compile fix released

vascobarroso (approved):

vascobarroso:
@ktf errors look unrelated to me. If that's the case, could you merge this ? We need it for early next week. Thanks!

ktf:
I assume you only need this for the o2-dataflow default? I'd rather not have a dependency on kafka for the general case. If yes, would you mind adding the customisation to the default?

sy-c:
@ktf I moved librdkafka requires to defaults-o2-dataflow.sh. Is this the way to do it ? It seems to work on my dev setup.
I have not found in other recipes another way to have an "if" confitional on the "defaults" chosen.

ktf:
Yes, it would need to be moved to the dataflow default. Indeed we could have some "defaults dependent" specification for the dependencies... Let me see how difficult it is.

sy-c:
Thanks for checking! Optional test like for OS would indeed be nice, but the current solution for today is good enough.

ktf:
https://github.com/alisw/alibuild/pull/901 will allow you to write:

```yaml
requires:
  libkafka:defaults=o2-dataflow
```

vascobarroso (approved):

ktf:
@Barthelemy The issue in QC seems to be a genuine (yet unrelated) problem. Can you please have MCH have a look?

PR #5717, opened 2024-12-18
[OCONF-844] Remove rpm dep on libhugetlbfs

Barthelemy:


sy-c (approved):
very good, thanks !
I'll update readoutcard to avoid the need of command-line hugeadm tool.
(which I anyway manage to compile on RHEL9 from https://github.com/libhugetlbfs, if we realize we really need it)

PR #5720, opened 2024-12-19
[OCONF-848] Update sqlite.sh

Barthelemy:
This is a very very old version and it is not compatible with some stuff on RHEL9.


Barthelemy:
hmm, I did not realize that it was an alidist repo. 
@ktf when you have a moment, could you bump sqlite ? 
Tahnk you

ktf:
I propose we use #5726.

Barthelemy:
yes, thank you

PR #5727, opened 2025-01-06
Bump pip requirements on newer pythons

singiamtel:
CC @barthelemy, should fix the build with new Python versions

I got these from the current ``Python-modules-list`` recipe


ktf:
@singiamtel, do you understand why rivet breaks in python?

PR #5732, opened 2025-01-14
Forward neutron generator noon

mbroz84:
nOOn (read as noon) is a simple generator that (among other usage) interacts with STARlight and produces forward neutrons from electromagnetic dissociation of the nuclei in ultra-peripheral collisions that can be than measured by the ZDC. 
The code compiles in ROOT. 
Interface for O2DPG is also ready an can be pushed anytime.


jackal1-66:
@mbroz84 apologies for the delay, I missed the PR. I just retriggered two CIs that were failing. I'll check the output and merge it in case.

mbroz84:
@jackal1-66 Can we proceed with this?

jackal1-66 (approved):

jackal1-66:
@ktf this PR is ready to be merged. The CI was restarted by a comment typo fix PR, but it was successful already, as you can see from the history. Could you approve this?

PR #5736, opened 2025-01-16
fix for python

sy-c:


sy-c:
A previous alidist PR (https://github.com/alisw/alidist/pull/5709) broke compilation of lla and readoucard libraries with python support.
https://its.cern.ch/jira/browse/ORC-502

vascobarroso (approved):

Barthelemy:
The RPMs generation breaks with the proposed changes. The difference with the checkers here is that we use `--always-prefer-system`. 
To fix it, you can add `${PYTHON_ROOT:+-DPython3_EXECUTABLE="$(which python3)"} \` to the `alf.sh` recipe (in the cmake parameters).

sy-c:
Indeed, alf was not updated. It is strange that it worked at all before, as python was not even mentionned in the CMake args so far for this module. The grep was no enough.

vascobarroso (approved):

PR #5738, opened 2025-01-17
Bump the version for FOCAL

ktf:


ktf:
I am fixing to a specific tag, so that's easier to build. If someone wants to develop the package they should use the development package feature of aliBuild.

PR #5741, opened 2025-01-22
Revert "ORT GPU build"

ktf:
Reverts alisw/alidist#5622


ktf:
Breaks O2Physics. Reverting/

PR #5742, opened 2025-01-23
Bump to 1.9.1

ktf:
Hide more stuff from the public API.


davidrohr:
Btw, I understand the desire for short commit names, put perhaps it should be as descriptive as containing which package is bumped :).

PR #5743, opened 2025-01-27
Fix kafka dependency

ktf:


ktf:
@Barthelemy @vascobarroso we also need this, I guess..

vascobarroso:
@sy-c could you review ?

sy-c (approved):
I can't grasp the meaning, but I'm sure this is fine

ktf:
The point is that if libkafka is not available (hence LIBRDKAFKA_REVISION is not set by alibuild) we should not add the option to CMake.

knopers8:
It seems it could have introduced an issue in QC PR checks for `o2-dataflow`:
```
ninja: error: '/local/workspace/DailyBuilds/DailyO2Dataflow-slc8/daily-tags.cZi7Jrsli5/slc8_x86-64/librdkafka/v2.3.0-14/lib/librdkafka++.so', needed by 'stage/lib64/libO2Framework.so', missing and no known rule to make it
```
Somehow `librkafka++` became a dependency of `O2Framework`, is that intentional?

sy-c:
libmonitoring now depends explicitely on kafka. So whatever depends on libmonitoring also depends on kafka.

ktf:
I think it needs to either be marked as a private dependency, or we need to add the find_package also to O2.

Barthelemy:
So... who is doing what ? 
I cannot reproduce this issue on my mac nor on our jenkins under rhel9 and cs8.

PR #5749, opened 2025-01-29
grpc for extra monitoring exe

sy-c:


vascobarroso (approved):

vascobarroso:
@sy-c could you have a look at the error ? Looks related to the monitoring lib, although it seems to only affect SLC8.

sy-c:
I don't understand this issue, the error happens when building data distribution.
It complains about requiring Monitoring version 3, but having 1.18.0, and I don't know where this number comes from, it's not what is set in CMake. I will launch a test build on a rh8 node to see.

sy-c:
I can reproduce it on a test machine... trying to debug.

sy-c:
The  problem looks to be related to DataDistribution/CMakeLists.txt
https://github.com/AliceO2Group/DataDistribution/blob/4ff6e1669c96f2688ef8a433aaddb037b44bdf5a/CMakeLists.txt#L118

the error appears with:
```find_package(Monitoring 3 REQUIRED) # AliceO2::Monitoring```

CMake Error at CMakeLists.txt:118 (find_package):
  Could not find a configuration file for package "Monitoring" that is
  compatible with requested version "3".

  The following configuration files were considered but not accepted:

    /home/alibuild/test/sw/slc8_x86-64/Monitoring/dev-local1/lib/cmake/Monitoring/MonitoringConfig.cmake, version: 1.18.0

I have no clue why it reports version "1.18.0", this file states "3.18.2"
(1.18.0 is the FLPsuite version ?)

When removing the "3" requirement:
```find_package(Monitoring REQUIRED) # AliceO2::Monitoring```

everything works as expected:

 "-- Monitoring 3.18.2 found"


There were some recent changes in CMake version, could it be related ?
@ktf @lkrcal what do you think ?

**Comment by vascobarroso** [comment]
@ktf @lkrcal could you have a look at Sylvain's comment and chip in on this ? We need this pretty soon at P2 otherwise some parts of monitoring won't work when we migrate to RHEL 9.

**Comment by ktf** [comment]
I suspect the problem is that it should be:

```
find_package(Monitoring 3 REQUIRED CONFIG)
```

Can you try with it?

sy-c:
```find_package(Monitoring 3 REQUIRED CONFIG)``` does not help.
Indeed, I don't think we have a v2 anywhere, and removing the "3" works.
I'll create a PR in datadistribution then.

sy-c:
Changes proposed in DataDistribution.
@lkrcal can you please check and tag ?
https://github.com/AliceO2Group/DataDistribution/pull/186

PR #5762, opened 2025-02-07
Properly specify dependencies for ucx

singiamtel:
This way aliDoctor fails when checking

Also taking the chance to fix the lints in the recipe

### DIFF
--- a/rdma-core.sh
+++ b/rdma-core.sh
@@ -1,0 +1,9 @@
+package: rdma-core
+version: "1.0"
+system_requirement_missing: |
+  Please install rdma-core on your system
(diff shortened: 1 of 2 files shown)


rdma-core.sh
```diff
@@ -0,0 +1,9 @@
+package: rdma-core
```
ktf, line 1: This is fine. It would probably make sense to use the prefer_system mechanism and keep track of the version so that we get a rebuild if it changes.

ktf:
@lkrcal can you approve this change?

lkrcal (dismissed):

davidrohr:
We build with the FullCI builder for the EPNs, which passes the build, so should be fine. Merging.

PR #5763, opened 2025-02-10
Bump FairLogger to v2.1.0 and InfoLogger to v2.8.2

ehellbar:
Introducing a "critical" log message severity, required to distinguish critical errors for shifter operation.

### DIFF
--- a/libinfologger.sh
+++ b/libinfologger.sh
@@ -1,6 +1,6 @@
 package: libInfoLogger
 version: "%(tag_basename)s"
-tag: v2.8.1
+tag: v2.8.2
 requires:
   - boost
   - "GCC-Toolchain:(?!osx)"


ktf:
It looks like there is a genuine incompatibility between InfoLogger and the new FairLogger.

ehellbar:
ok, I will check if it is only these two `string_views` or if there is more

ktf:
Keep also in mind that c_str() is guaranteed to be 0 terminated, string_view is not.

ehellbar:
waiting for https://github.com/AliceO2Group/InfoLogger/pull/95 to also bump InfoLogger here

sy-c:
My bad, I messed up with the include file during the test... Will fix shortly

sy-c (approved):
I fixed infologger and pushed v2.8.2

The include issue escaped my attention as I did the release integration tests using InfoLogger instead of libInfoLogger (which is the one included by the O2 components).

ehellbar:
the compiler on macOS throws a warning when using enums in switch statements (in this case `fair::Severity`) and not handling all of the enumerators, which is the case here (missing the case for the new `critical` severity introduced with this bump of FairLogger). For PR checks, we use `-Werror` so we get an error here. Locally on my mac, the latest O2@dev with this alidist  compiles successfully, throwing only the warning.

ehellbar:
ping @ktf

PR #5768, opened 2025-02-17
Add kernel-devel as requirement to curl

Barthelemy:


ktf:
I actually think this is wrong. There is no need to have kernel-devel to compile curl (tested on lxplus) and introducing this dependency will prevent people from building on lxplus and alikes. Can you remind me what was the issue?

Barthelemy:
see the MM messages: https://mattermost.web.cern.ch/alice/pl/nzu1cbzhetroxb6gigday8zczo

davidrohr:
that link to mattermost doesn't work. Can you paste the reason in here?

ktf:
That only happens because the machine you are compiling with has kernel-devel installed in the first place and then you use the artifacts (the precompiled headers) on a machine which does not have it. What should be done is to hide `CCDBApi.h` from ROOT, so that the actual issue disappears.

PR #5774, opened 2025-02-27
Bump JDK to 17.0.4

zensanp:
Bumps the JDK to v17 (LTS), which will be needed for JAliEn in upcoming tags (compile target changed from 11 --> 15).

### DIFF
--- a/jdk.sh
+++ b/jdk.sh
@@ -1,5 +1,5 @@
 package: JDK
-version: "12.0.1"
+version: "17.0.14"
 build_requires:
   - curl
 prefer_system: .*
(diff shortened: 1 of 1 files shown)


costing (approved):

jdk.sh
```diff
@@ -16,9 +16,9 @@ JDK_PLATFORM=linux
 [[ $ARCHITECTURE != osx* ]] || JDK_PLATFORM=osx
 
 if [[ $JDK_PLATFORM == osx ]]; then
-  URL="https://cdn.azul.com/zulu/bin/zulu12.2.3-ca-jdk12.0.1-macosx_x64.tar.gz"
```
ktf, line 19: Do we really need JAVA on macOS? This will work only for x86. I suggest we simply disable the dependency there.

PR #5776, opened 2025-03-04
Update tbb version

BongHwi:
- Version up from `v2021.5.0` to `v2022.0.0`

### DIFF
--- a/tbb.sh
+++ b/tbb.sh
@@ -1,6 +1,6 @@
 package: TBB
-version: "v2021.5.0"
-tag: v2021.5.0
+version: "v2022.0.0"
+tag: v2022.0.0
 source: https://github.com/uxlfoundation/oneTBB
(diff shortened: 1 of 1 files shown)


ktf (approved):

ktf:
Bumping itself is fine. However it seems to break old projects.

davidrohr:
It seems to break ROOT, but ROOT works with O2. Do we use the same ROOT for O2 and for AliRoot / Generators?

davidrohr:
@ktf @pzhristov : Do you have any idea why the ROOT build for AliRoot/alidist fails, while it succeeds for O2? I thought both would use the latest ROOT 6.36.04-alice2?

davidrohr:
@ktf : This is now all green except for those that always fail. Can we merge it?

PR #5778, opened 2025-03-05
Force rebuilds whenever the CMake on the system changes

ktf:
This is needed because CMake embeds itself in the CMakeCache.txt
with the net result that if e.g. brew gets updated, you are left
with a broken setup.

### DIFF
--- a/cmake.sh
+++ b/cmake.sh
@@ -12,6 +12,11 @@ prefer_system: .*
 prefer_system_check: |
   verge() { [[  "$1" = "`echo -e "$1\n$2" | sort -V | head -n1`" ]]; }
   type cmake && verge 3.28.1 `cmake --version | sed -e 's/.* //' | cut -d. -f1,2,3`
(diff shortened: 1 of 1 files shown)


cmake.sh
```diff
@@ -12,6 +12,11 @@ prefer_system: .*
 prefer_system_check: |
   verge() { [[  "$1" = "`echo -e "$1\n$2" | sort -V | head -n1`" ]]; }
   type cmake && verge 3.28.1 `cmake --version | sed -e 's/.* //' | cut -d. -f1,2,3`
+  echo alibuild_system_replace: cmake`cmake --version | sed -e 's/.* //' | cut -d. -f1,2,3`
```
ktf, line 15:
```suggestion
  echo alibuild_system_replace: cmake$(cmake --version | sed -e 's/.* //' | cut -d. -f1,2,3)
```

ktf:
Merged by the other PR.

PR #5790, opened 2025-03-13
Bump Vc

davidrohr:


davidrohr:
@ktf @singiamtel : Here, also both MacCIs are failing with weird problems.

PR #5803, opened 2025-03-19
More fixes to correctly detect cmake from the system

ktf:


ktf:
Make it work also with `--always-prefer-system`

PR #5808, opened 2025-03-21
Bump arrow to apache-arrow-19.0.1-alice1

ktf:


ktf:
Tested on hyperloop as well. Merging.

PR #5813, opened 2025-03-27
Modernize DDS recipe

ktf:
- Fix newer versions of protobuf
- Use ninja
- Direct dependency on abseil


ktf:
Currently broken. Merging.

PR #5826, opened 2025-04-01
Do not pick up cmake 4 from the system

ktf:
Too many things break when moving to the new CMake


ktf:
@pzhristov can you try with this?

PR #5832, opened 2025-04-02
Get ROOT to compile with new macOS / XCode

ktf:


pzhristov:
@ktf We have a problem with the Clang recipe, probably it is not detected, because we have llvm@18 at the build servers.

pzhristov:
@ktf The replacement of <fp.h> by <math.h> is missing in v6-32-06-alice6, so the compilation of ROOT fails.

ktf:
As discussed with @pzhristov privately, the issue in simulation is not new, they merely added an error about it. Merging this.

PR #5836, opened 2025-04-04
Fix mesos loading some grpc internals twice

knopers8:
Mesos was linking against both grpc and grpc_unsecure, which was fine compile-time, but it was preventing mesos-slave and mesos-master to start.

Fixes O2-5908


vascobarroso:
@ktf can we merge this so we can retry our RPM generation and see if we get the CI green again ?

PR #5838, opened 2025-04-08
Make GEANT4 compile with our own expat / zlib

ktf:


ktf:
This will allow us to go back to older GEANT4 if deemed necessary, without having to backport patches to compile old zlib / expat.

pzhristov:
@ktf Please move back to version v6-6-p3 in geant4_vmc.sh

PR #5842, opened 2025-04-09
Allow using CMake4 from the system

ktf:
---
[//]: # (BEGIN SAPLING FOOTER)
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/alisw/alidist/pull/5842).
* __->__ #5842
* #5841 (2 commits)


ktf:
@pzhristov this should allow us to compile with the new CMake (assuming everything is really 3.5 compatible)

PR #5847, opened 2025-04-10
Update to EPOS4.0.3

jackal1-66:


epos4.sh
```diff
@@ -63,11 +70,11 @@ module load BASE/1.0 ROOT/$ROOT_VERSION-$ROOT_REVISION fastjet/$FASTJET_VERSION-
 # Our environment
 set EPOS4_ROOT \$::env(BASEDIR)/$PKGNAME/\$version
 setenv EPOS4_ROOT \$EPOS4_ROOT
-setenv EPO4VSN 4.0.0
+setenv EPO4VSN 4.0.3
 # Final slash is required by EPOS, please leave it be
 setenv EPO4 \$::env(EPOS4_ROOT)/epos4/
 prepend-path PATH \$::env(EPO4)bin
-setenv LIBDIR \$::env(EPO4)bin
+setenv BIN_DIR \$::env(EPO4)bin
```
ktf, line 77: Is this really needed?

ktf (approved):

sawenzel:
@ktf @singiamtel @ihrivnac : It looks like this PR caused CVMFS releases on ARM to stop working—we haven’t had one since mid-April. It seems the failing compile of AliGenO2 on ARM wasn’t caught by the CI.

Would it be possible to improve this by either extending the current ARM build (build/O2/alidist-slc9-aarch64) to something like build/O2Suite/..aarch64, or by adding a fullCI for aarch64? I think this would help a lot in catching such issues earlier. What do you think?

ktf:
yes, we can do that. @singiamtel can you take care? I think we need to stick to O2Suite until we get digitization to work on ARM, then we can even do fullCI.

ktf:
We need to patch our streamers and I have not yet managed to understand how.

ktf:
(plus it needs a new version of ROOT).

singiamtel:
> yes, we can do that. @singiamtel can you take care? I think we need to stick to O2Suite until we get digitization to work on ARM, then we can even do fullCI.

I don't follow, the ARM CI is already testing O2Suite. EPOS4 is not included though, it's only in O2sim via AliGenO2. 

Should we add EPOS4 (or AliGenO2) to O2Suite?

ktf:
O2PDPSuite.

sawenzel:
ah .. I thought O2Suite actually == O2PDPSuite.

ktf:
IIRC, O2Suite is the full thing for sync, including the FLP stuff, while O2PDPSuite is what includes the generators...

PR #5849, opened 2025-04-11
enable building shared libraries in protobuf recipe

knopers8:
This fixes mesos which was silently falling back into static libs, probably causing double runtimes in `mesos-fetcher` (see https://its.cern.ch/jira/browse/OCTRL-1005).

Thanks to @ktf for the suggestion.


knopers8:
Let's see if it breaks something else though. I am also going to test it on the FLP RPM machine.

knopers8:
@ktf do you have any idea why is this breaking arrow?
```
/sw/BUILD/3b0e0ac0f42dbb23524001651fbbc668c474b799/arrow/src_tmp/cpp/src/gandiva/precompiled/arithmetic_ops.cc:18:10: fatal error: 'cmath' file not found
   18 | #include <cmath>
      |          ^~~~~~~
1 error generated.
```

PR #5851, opened 2025-04-14
Attempt at keeping track of the detected GPU subsystem

ktf:


ktf:
@davidrohr this is what we discussed this morning. It's not complete, but you should get the idea.

PR #5855, opened 2025-04-14
Add gpu-system package to detect GPU features, use it in ONNXRuntime and O2

davidrohr:
Supersedes 5851


davidrohr:
@ktf @singiamtel : Could you explain what this error means:
```
Error: prefer_system_replacement_specs..*.version: substitution variable 'key' is invalid [ali:schema]
```

PR #5861, opened 2025-04-24
Bump mesos to 1.11.0-alice4

knopers8:
It includes a fix for mesos-fetcher, needed due to an issue introduced by the recent protobuf bump to v29.3


vascobarroso:
Merging because CI is green, we rely on this for the FLP RPMs and AFAIK we are the only users of Mesos.

PR #5868, opened 2025-05-05
Fix linter issue in recipe

singiamtel:


singiamtel:
@ktf is this okay to merge? It should be the only thing left to get the CI to green

PR #5870, opened 2025-05-09
bump aegis to 1.5.7

sawenzel:


sawenzel:
@jackal1-66 : There is a failure in the full-system-test related to a QED thing. Probably related to the dev in AEGIS. Could you please take a look?

PR #5875, opened 2025-05-14
Fix location of c runtime

ktf:


ktf:
Currently broken in any case. Merging.

PR #5882, opened 2025-05-15
alienpy 1.6.6

adriansev:
* fix for running shell commands (with `!`) in alienpy shell mode and batch files

### DIFF
--- a/xjalienfs.sh
+++ b/xjalienfs.sh
@@ -1,6 +1,6 @@
 package: xjalienfs
 version: "%(tag_basename)s"
-tag: "1.6.5"
+tag: "1.6.6"
 source: https://gitlab.cern.ch/jalien/xjalienfs.git
 requires:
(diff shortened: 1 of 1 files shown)


costing (approved):

adriansev:
for O2/alidist-dataflow-cs8 :
```
[3668:TST-MNLTest54090-proxy]: [15:35:05][STATE] Exiting FairMQ state machine
[3668:TST-MNLTest54090-proxy]: [15:35:05][FATAL] error while setting up workflow in o2-qc: 1 of the binding channels could not initialize. Initial configuration incomplete.
[3668:TST-MNLTest54090-proxy]: terminate called after throwing an instance of 'o2::framework::RuntimeErrorRef'
[3668:TST-MNLTest54090-proxy]: *** Program crashed (Aborted)
```

PR #5886, opened 2025-05-20
Update aegis.sh

sawenzel:


ktf:
@singiamtel it looks like there is still some stale branches. Could you have a look?

PR #5888, opened 2025-05-21
Allow compiling only the parts you care about

ktf:
E.g.

O2PHYSICS_COMPONENTS=Tutorials alibuild build --defaults o2 O2Physics

Will only compile the tutorial. Changing O2PHYSICS_COMPONENTS to something else will also trigger a rebuild if https://github.com/alisw/alibuild/pull/927 is used.

### DIFF
--- a/o2physics.sh
+++ b/o2physics.sh
@@ -13,8 +13,17 @@ build_requires:
   - ninja
   - alibuild-recipe-tools
 source: https://github.com/AliceO2Group/O2Physics
+track_env:
+  O2PHYSICS_COMPONENTS: "echo ${O2PHYSICS_COMPONENTS:-install}"
 incremental_recipe: |
-  cmake --build . -- ${JOBS:+-j$JOBS} install
+  if [ "X$O2PHYSICS_COMPONENTS" = "Xinstall" ]; then
+      cmake --build . -- ${JOBS+-j $JOBS} install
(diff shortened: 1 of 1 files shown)


ktf:
@aalkin this is what we discussed, together with  https://github.com/alisw/alibuild/pull/927 it should allow people to only compile what they need (and rebuild in case they change their mind). We can think of some syntactic sugar like:

```
alibuild build --defaults o2 O2Physics#Tutorials
```

which expands to correctly. In a future  update we could limit what we compile in O2, however at the moment we export everything in O2Targets and then CMake Complains if (say) I only try to build the framework and then O2Physics.


@ddobrigk We should probably integrate this feature for the next tutorial, so that people rebuild only a minimal part of O2Physics.

o2physics.sh
```diff
@@ -13,8 +13,17 @@ build_requires:
   - ninja
   - alibuild-recipe-tools
 source: https://github.com/AliceO2Group/O2Physics
+track_env:
```
ktf, line 16: @singiamtel can you add it to the linter, please?

singiamtel: Will do. Could you give me write access to the repo? The infra admins only had read permissions there for some reason

ktf: done

PR #5894, opened 2025-05-26
Force Full System Test environment

ktf:
Without this in the CI the timeout for CCDB queries is 1 second rather than 15.


davidrohr:
fine with me

ktf:
Actually, this is not needed anymore. 1s is more than enough for local tests, the issue was some misconfiguration on the CCDB side.

PR #5910, opened 2025-06-05
Bump ROOT to v6-32-06-alice8

ktf:
Includes backport of fix for large zip files https://github.com/root-project/root/pull/18958.


ktf:
@singiamtel it looks like we still have a cs8 test, without the builder. Could you please take care of it? Thanks.

PR #5922, opened 2025-06-12
Bump FairMQ to std::pmr supporting version

ktf:


ktf:
@singiamtel why are we still testing CS8 here?

PR #5925, opened 2025-06-16
Revert to CodeChecker V18.1.2

ktf:
Picking up clang from our own build apparently silently breaks.


ktf:
Currently broken, merging.

PR #5928, opened 2025-06-17
Drop CS8 checks from Clean Action

ktf:


ktf:
@singiamtel please review. Are we really still building CS8 tests?

PR #5929, opened 2025-06-17
o2.sh: Add env variable to disable GPU event display

davidrohr:


ktf:
I would suggest adding a track_env for all those, so that it triggers a recompilation if they change.

Something like:

```
track_env:
  GPUCA_BUILD_EVENT_DISPLAY: "echo ${GPUCA_BUILD_EVENT_DISPLAY}"
```

PR #5932, opened 2025-06-20
Bump ODC to 0.85.0

lkrcal:


lkrcal:
@ktf I see there are some CMake changes in the fork you made. But there is a branch `alisw/fix-new-protobuf` and a diverging no-branch tag `0.84.2-alice2` that was actually used in alidist. Can you let me know what CMake changes exactly you want? Then we should upstream them directly.

PR #5934, opened 2025-06-23
Bump to onnxruntime to 1.22.0

ktf:
Provides nicer API for amending models on the fly.
We need to mask a few more warnings.


ktf:
@ChSonnabend can you check this works for you as well?

ChSonnabend:
Yep, can confirm, that works

PR #5936, opened 2025-06-24
Fix KFParticle module

singiamtel:
Regression in #5927. Currently breaks daily build


kfparticle.sh
```diff
@@ -29,6 +29,6 @@ mkdir -p "$MODULEDIR"
 cat > "$MODULEFILE" <<EoF
 $(alibuild-generate-module --bin --lib)
 # Our environment
-setenv KFPARTICLE_ROOT \$KFPARTICLE_ROOT
```
ktf, line 32: It should have been `\$PKG_ROOT`, indeed.

PR #5937, opened 2025-06-26
Unpin pip

singiamtel:
Older versions of pip are currently failing with this error:

```
ImportError: cannot import name 'RequirementInformation' from 'pip._vendor.resolvelib.structs'
```

It seems to be a compatibility issue between pip 25 and older.

As far as I can tell, the pip package was only pinned to follow the
convention set by other packages


Barthelemy:
thank you !

Barthelemy (approved):

PR #5941, opened 2025-07-07
fix C23 compatibility in GMP

musinsky:
Temporary fix C23 (since gcc 15) compatibility in GMP (gmp-6.2.1).
Note that even the latest version of GMP (gmp-6.3.0) needs this fix.


musinsky:
Now it should work also with older compilers.

ktf:
@jackal1-66 can you have a look at the YODA issue?

PR #5942, opened 2025-07-09
Set HDF5_DIR as environment variable during YODA build

jackal1-66:
h5cc sets the include directory in yoda-config which is used later to set cppflags in Rivet. the dailytag is wrongly taken when building in docker/CI, so the folder must be correctly set at building time. 
Not clear why this was working before. An update somewhere must have had an effect here (no updates since a month on Rivet, hdf5, Yoda and GMP).

### DIFF
--- a/yoda.sh
+++ b/yoda.sh
@@ -13,6 +13,8 @@ build_requires:
   - Python
 prepend_path:
   PYTHONPATH: $YODA_ROOT/lib/python/site-packages
+env:
+  HDF5_DIR: "$HDF5_ROOT"
 ---
 #!/bin/bash -e
 rsync -a --exclude='**/.git' --delete --delete-excluded "$SOURCEDIR"/ ./


jackal1-66:
From what I saw in the source code of YODA the main issue is that HDF5_CPPFLAGS is replaced in yoda-config with the output of `h5cc --show` at build time. I tested the --with-h5df option for Rivet and it does not seem to be recognised. However hdf5 is automatically pulled to build as an additional compatibility package. @ktf do you have any idea why everything seemed to work fine before? I'm a bit puzzled by this...

ktf:
Can't we simply set H5DF_DIR inside the rivet recipe?

ktf:
Can't we simply set HDF5_DIR in the rivet recipe?

jackal1-66:
Unfortunately this won't work. Because the include path is generated from the yoda-config executable:
```
  YODACPPFLAGS=`$YODACONFIG --cppflags`
  YODALIBS=`$YODACONFIG --libs`
  AC_MSG_RESULT([$YODA_PYTHONPATH])
fi
AC_SUBST(YODALIBS)
AC_SUBST(YODACPPFLAGS)
AC_SUBST(YODA_PYTHONPATH)

## HDF5
AC_ARG_ENABLE([h5], [AS_HELP_STRING([--disable-h5],[disable H5 extension  @<:@default=build@:>@])], [], [with_h5=yes])
if test "x$with_h5" = "xyes"; then
  oldCPPFLAGS=$CPPFLAGS
  CPPFLAGS="$CPPFLAGS $YODACPPFLAGS"
```

jackal1-66:
Closing. Fixed with https://github.com/alisw/alidist/pull/5948 and https://github.com/alisw/alidist/pull/5953

PR #5948, opened 2025-07-10
Fix YODA issue with HDF5

singiamtel:
Fixes daily builds

### DIFF
--- a/yoda.sh
+++ b/yoda.sh
@@ -21,7 +21,7 @@ rsync -a --exclude='**/.git' --delete --delete-excluded "$SOURCEDIR"/ ./
 
 export PYTHON=$(type -p python3)
 
-./configure --disable-silent-rules --enable-root --prefix="$INSTALLROOT"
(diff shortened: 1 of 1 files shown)


yoda.sh
```diff
@@ -21,7 +21,7 @@ rsync -a --exclude='**/.git' --delete --delete-excluded "$SOURCEDIR"/ ./
 
 export PYTHON=$(type -p python3)
 
-./configure --disable-silent-rules --enable-root --prefix="$INSTALLROOT"
+./configure --disable-silent-rules --enable-root --prefix="$INSTALLROOT" --with-hdf5="$HDF5_ROOT"
```
ktf, line 24: Needs to handle the case hdf5 comes from the system, no?

jackal1-66:
Hello! wouldn't this be equivalent to add HDF5_DIR in the env? Both should retrigger a rebuild I believe...

PR #5952, opened 2025-07-14
Add nlohmann/json in the ROOT_INCLUDE_PATH

fmazzasc:
nlohmann/json package is required for some generators in O2DPG, hence it needs to be available in the ROOT include path


nlohmann_json.sh
```diff
@@ -25,3 +25,6 @@ MODULEDIR="$INSTALLROOT/etc/modulefiles"
 MODULEFILE="$MODULEDIR/$PKGNAME"
 mkdir -p "$MODULEDIR"
 alibuild-generate-module --bin --lib > "$MODULEFILE"
+cat << EOF >> $MODULEFILE
```
ktf, line 28:
```suggestion
cat << EOF >> "$MODULEFILE"
```

PR #5955, opened 2025-07-16
Move gpu-system to be a requires

ktf:
Already built packages do not get rebuild if the build_requires changes, so this special package needs to be a requires.


davidrohr:
But then we need to chane it also for o2 to be requires?

davidrohr:
Done here: https://github.com/alisw/alidist/pull/5956

PR #5960, opened 2025-07-21
Allow incremental build of Geant4

ktf:


ktf:
@jackal1-66 FYI

sawenzel:
This PR unfortunately broke the MC productions. The modulefile is copied to the installation before the dataset variables were included. I'll prepare a fix.

PR #5961, opened 2025-07-24
XRootD 5.8.4

adriansev:
[Release notes](https://github.com/xrootd/xrootd/blob/v5.8.4/docs/ReleaseNotes.txt)
Detailed [commit list vs 5.8.3](https://github.com/xrootd/xrootd/compare/v5.8.3...v5.8.4)

important fixes for client side operations

### DIFF
--- a/xrootd.sh
+++ b/xrootd.sh
@@ -1,6 +1,6 @@
 package: XRootD
 version: "%(tag_basename)s"
-tag: "v5.8.3"
+tag: "v5.8.4"
 source: https://github.com/xrootd/xrootd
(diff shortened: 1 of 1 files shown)


adriansev:
Hi @ktf @singiamtel ! could you merge this? it seems that everything is all right

costing:
Indeed, we were waiting for this version to fix the jobs stuck in infinite loops accessing data files or CCDB objects. Not a blocker though. It can wait until you are back, also because the entire stack will need to be rebuilt so it will take additional effort.

PR #5977, opened 2025-08-15
Revert "Bump FairMQ to std::pmr supporting version"

ktf:
Reverts alisw/alidist#5922


ktf:
@vascobarroso FYI, I have already reverted the associated code in O2.

PR #5988, opened 2025-08-27
Bump to clang 20

ktf:


ktf:
Finally. @davidrohr could you check for GPU regressions and then we merge?

PR #5997, opened 2025-09-15
[WIP] Upgrade and include Herwig on AliGenO2

jackal1-66:


aligeno2.sh
```diff
@@ -17,6 +17,7 @@ requires:
   - Upcgen:(?!osx)
   - Graniitti:(?!osx)
   - nOOn:(?!osx)
+  - Herwig
```
sawenzel, line 20: we should check if this compiles on ARM and or osx

jackal1-66:
CI yesterday was green, something must have retriggered it, but failure seems unrelated to this PR

jackal1-66:
Failures seem unrelated to this PR

sawenzel:
@ktf: The full-system-test appears to be failing with a problem in ROOT: 
```
./hmpreco.log[14625:HMP-Clusterization]:     /sw/slc9_x86-64/ROOT/v6-36-04-alice2-local2/lib/libHist.so.6.36: TVirtualFitter::Fitter(TObject*, int)
./hmpreco.log[14625:HMP-Clusterization]:     /sw/slc9_x86-64/O2/5997-slc9_x86-64-local1/lib/libO2DataFormatsHMP.so: o2::hmpid::Cluster::solve(std::vector<o2::hmpid::Cluster, std::allocator<o2::hmpid::Cluster> >*, float*, bool)
```
Could be related to the recent ROOT bump and something to watch out for more closely.

jackal1-66:
The problem with O2Suite in macOS instead seems to be related simply to the `batch-test` which goes in timeout. 
Both of these failures are unrelated to this PR

PR #5998, opened 2025-09-15
Bump o2codechecker to v20.1.7-alice1

ktf:


ktf:
Issue on macOS unrelated (@singiamtel I think there is an old certificate on one of the build machines).

PR #6009, opened 2025-10-05
Compile librdkafka on macOS Tahoe

ktf:
./configure is broken due to depending on an obsolete ctags flag.


ktf:
Failed attempt. Closing and retrying when the new OS comes.

PR #6133, opened 2026-02-13
Do not attempt creating pip symlinks when pip is taken from system

knopers8:
https://github.com/alisw/alidist/pull/5583 makes sure to avoid building new pip if there is one available in the system. However, when attempting to create symlinks, the script would still look for a pip coming from alibuild and fail here:
```
  PIP_BIN=$(for X in pip*; do echo "$X"; done | grep -E '^pip[0-9]+\.[0-9]+$' | head -n1)
```

This commit allows the installation to continue by avoiding attempts to symlink custom pip when it was purposefully not built.

This helps with my build, not sure if it's going to break something else though.

### DIFF
--- a/python.sh
+++ b/python.sh
@@ -105,13 +105,15 @@ pushd "$INSTALLROOT/bin"
   sed -i.deleteme -e "1 s|^#!${INSTALLROOT}/bin/\(.*\)$|#!/usr/bin/env \1|" * || true
   rm -f *.deleteme
   PYTHON_BIN=$(for X in python*; do echo "$X"; done | grep -E '^python[0-9]+\.[0-9]+$' | head -n1)
-  PIP_BIN=$(for X in pip*; do echo "$X"; done | grep -E '^pip[0-9]+\.[0-9]+$' | head -n1)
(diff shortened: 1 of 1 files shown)


knopers8:
@ktf , @singiamtel would this fix be of interest to you?

singiamtel:
The idea of https://github.com/alisw/alidist/pull/5583 was to not use the system pip at all and force python to install it's own pip, which seems contrary to this PRs purpose? I don't know if this PR will lead to broken installations after.

Adding @vkucera to the thread.

knopers8:
Ah, then I probably misunderstood the intention of there. If it is as you are saying, then disregard my PR please.

As far as I remember, in my case there was no pip coming from alibuild though, so apparently something was stilling missing. I can have another look next time I pull alidist.

knopers8:
For the record, #5583 fails to work for me because my `pip` is not in a `python` directory, but in `~/.local/bin/` (which, as I learned, is installed with o2-flp-setup) and another one in `/usr/bin/`.

I'm not sure what to propose as a fix though...

PR #6135, opened 2026-02-17
Bump CMake to 4.2.3 and boost to 1.90

davidrohr:


pzhristov (dismissed):
It seems the current version can be merged.

treelite.sh
```diff
@@ -1,28 +1,41 @@
...
 
 cmake src                                   \
   ${CMAKE_GENERATOR:+-G "$CMAKE_GENERATOR"} \
+  -DCMAKE_POLICY_VERSION_MINIMUM=3.27       \
   -DCMAKE_INSTALL_PREFIX="$INSTALLROOT"     \
   -DUSE_OPENMP=OFF
 
 cmake --build . -- ${JOBS:+-j$JOBS} install
 
+[[ -d $INSTALLROOT/lib64 ]] && [[ ! -d $INSTALLROOT/lib ]] && ln -sf ${INSTALLROOT}/lib64 $INSTALLROOT/lib
```
vkucera, line 37: The symlink `lib -> lib64` is not relocated which is currently breaking the AliPhysics CI builds:
```
lib -> /sw/INSTALLROOT/6f8bcece55616f5b23671770a825fc8e1249b05f/slc7_x86-64/treelite/df6a892b7faf532aea13bacc0f30b8a4743f82cf-local1/lib64
```

vkucera, line 37:
```suggestion
[[ -d $INSTALLROOT/lib64 ]] && [[ ! -d $INSTALLROOT/lib ]] && ln -sf lib64 $INSTALLROOT/lib
```
I guess a relative link should fix it.

ktf: There is actually a way to tell CMake to use lib rather than lib64. I would use that. @singiamtel can you have a look?

davidrohr: I only vaguely remember, but I think I tried the CMake libdir setting, and it failed for some reason, that is why I put the symlink. We can of course try to fix it the "correct" way, but if that causes issues, I would just create a relative symlink.

vkucera: Let's see if https://github.com/alisw/alidist/pull/6213 fixes it.

PR #6177, opened 2026-03-11
Arrow: bump to 25.0.0; thrift: bump to 0.22; xsimd: bump to 14.2

aalkin:
* Bump version of thrift to current (0.22)
* Move thrift building to cmake instead of configure
* Add thrift as a dependency to Arrow, so that Parquet is enabled 
* Bump arrow version to current (24.0.0)

### DIFF
--- a/arrow.sh
+++ b/arrow.sh
@@ -1,6 +1,6 @@
 package: arrow
-version: "v23.0.1-alice"
-tag: apache-arrow-23.0.1-alice1
+version: "v24.0.0-alice"
+tag: apache-arrow-24.0.0-alice1
 source: https://github.com/alisw/arrow.git
 requires:
   - boost
--- a/arrow.sh
+++ b/arrow.sh
@@ -10,7 +10,6 @@ requires:
   - utf8proc
   - OpenSSL:(?!osx)
   - xsimd
-  - thrift
 license: Apache-2.0
 build_requires:
   - zlib
--- a/arrow.sh
+++ b/arrow.sh
@@ -10,6 +10,7 @@ requires:
   - utf8proc
   - OpenSSL:(?!osx)
   - xsimd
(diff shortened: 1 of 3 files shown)


aalkin:
Aparently this will require some small changes to O2, as usage of compute functions now requires a call to `arrow::compute::Initialize()`.

arrow.sh
```diff
@@ -1,6 +1,6 @@
 package: arrow
-version: "v20.0.0-alice1"
-tag: apache-arrow-20.0.0-alice1
+version: "v23.0.1-alice"
```
aalkin:
```suggestion
version: "v24.0.0-alice"
```

arrow.sh
```diff
@@ -1,6 +1,6 @@
...
+tag: apache-arrow-23.0.1-alice1
```
aalkin:
```suggestion
tag: apache-arrow-24.0.0-alice1
```

aalkin:
@davidrohr I've bumped arrow to the latest 25.0.0 and integrated our changes. Everything works on mac, could you also check on linux?

davidrohr:
OK, I'll give it a try. Can you also pull in the changes from @ktf 's PR with the fixes for LLVM symbols, so that we have everything in one PR?

arrow.sh
```diff
@@ -10,6 +10,7 @@ requires:
   - utf8proc
   - OpenSSL:(?!osx)
   - xsimd
+  - thrift
```
aalkin:
```suggestion
```

ktf (approved):

ktf:
Finally working everywhere. Also tested on hyperloop. I will merge once I get to a desktop.

ktf:
Merged so that tomorrow daily will have it.

PR #6192, opened 2026-03-27
Bump ROOT to v6-36-10-alice1

ktf:
Supports macOS 26.4 and new XCode

### DIFF
--- a/root.sh
+++ b/root.sh
@@ -1,6 +1,6 @@
 package: ROOT
 version: "%(tag_basename)s"
-tag: "v6-36-04-alice9"
+tag: "v6-36-10-alice1"
 source: https://github.com/alisw/root.git
 license: LGPLv2.1
 requires:


pzhristov (dismissed):

Barthelemy:
@ktf it fails on our build machine in onnxruntime: 
```
2026-04-27@14:26:35:DEBUG:O2Suite:ONNXRuntime:0: CMake Error at /home/alibuild/sw/slc9_x86-64/CMake/v4.1.4-local1/share/cmake-4.1/Modules/FetchContent.cmake:2106 (message):
2026-04-27@14:26:35:DEBUG:O2Suite:ONNXRuntime:0:   FETCHCONTENT_FULLY_DISCONNECTED is set to true, which requires the source
2026-04-27@14:26:35:DEBUG:O2Suite:ONNXRuntime:0:   directory for dependency mp11 to already be populated.  This generally
2026-04-27@14:26:35:DEBUG:O2Suite:ONNXRuntime:0:   means it must not be set to true the first time CMake is run in a build
2026-04-27@14:26:35:DEBUG:O2Suite:ONNXRuntime:0:   directory.  The following source directory should already be populated, but
2026-04-27@14:26:35:DEBUG:O2Suite:ONNXRuntime:0:   it doesn't exist:
2026-04-27@14:26:35:DEBUG:O2Suite:ONNXRuntime:0: 
2026-04-27@14:26:35:DEBUG:O2Suite:ONNXRuntime:0:     /home/alibuild/sw/BUILD/d137fdbb0b8083bbc417547d2df825bd2ab85663/ONNXRuntime/_deps/mp11-src
```
I added your 3 commits on top of `flp-suite-v1.80.0` in a new branch called `flp-suite-v1.80.0-new-root`. 

[log.txt](https://github.com/user-attachments/files/27130236/log.txt)

ktf:
@Barthelemy sorry for missing your comment. I have fixed the extra, unintended commits.

Barthelemy:
@ktf It works for us but it must not go to P2 though. We need at least to change the default for o2-dataflow. 

@ehellbar Is there some other file to edit to make sure that the old ROOT is still used in case we release for P2 ?

ktf:
Apologies, I understood the release was already cut and that you were only cherry picking?

ehellbar:
@Barthelemy @ktf we froze the software, so I would assume we would only cherry-pick in the alidist/flp-suite-v1.81.0 branch as well? @vascobarroso should confirm. 

But then I don't know what we want to do after PbPb. @Barthelemy Should we keep the old ROOT for some time also then?

Barthelemy:
@ktf @ehellbar We can cherry-pick on top of the current branches. @vascobarroso do you agree ? Then it is easier. 

I don't see a reason to stick with the old ROOT after PbPb.

vascobarroso:
Indeed we can keep `alidist/flp-suite-v1.81.0` for production and cherry pick there if needed. So all good to merge this

PR #6193, opened 2026-03-27
Bump ONNXRuntime to 1.29.0

ktf:


davidrohr:
Did you meanwhile try a newer version like 1.29?

ktf:
I am doing multiple things at the same time:

- Testing the new aliBuild remote store
- Testing my security proxy to better hide secrets 
- Testing el10
- Cleaning up dependencies and bumping them as needed.

so yes, this will eventually have the latest ONNXRuntime.

ktf:
@davidrohr @ChSonnabend this seems to pass the full CI. How do we validate this with actual GPUs? Can you take care of it?

ChSonnabend:
@ktf I will test it as soon as I can and let you know

> @davidrohr @ChSonnabend this seems to pass the full CI. How do we validate this with actual GPUs? Can you take care of it?

knopers8:
For the record, the FLP Suite CI pipeline, which runs a few workflows through AliECS, passed with the current version of the branch.

ktf:
Ok thanks. CCDB issues in non production environments understood. Merging since it improves detection of a bunch of stuff. I will deal with the fallout if needed.

ChSonnabend:
@ktf Did you check if anything runs? The current implementation with ROCm execution providers is not supported and MiGraphX is not supported / tested properly on EPNs yet. So why was this commit merged now?

ktf:
It was merged in order to get the FLP validated mesos aligned to the bumped ONNXRuntime dependencies. Dealing with the updated executor is in our hands so it should converge quickly.

PR #6213, opened 2026-04-20
Make sure we install in lib, not lib64

ktf:
---
[//]: # (BEGIN SAPLING FOOTER)
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/alisw/alidist/pull/6213).
* #6192
* #6193
* __->__ #6213


ktf:
@vkucera this is what I meant.

vkucera (approved):

PR #6215, opened 2026-04-22
Attempt at silencing clang-diagnostic-vla-cxx-extension

ktf:
It's a feature, not a bug

### DIFF
--- a/o2checkcode.sh
+++ b/o2checkcode.sh
@@ -6,6 +6,7 @@ requires:
 license: GPL-3.0
 build_requires:
   - CMake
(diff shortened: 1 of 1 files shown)


vkucera:
Why not adding the list directly in `.clang-tidy`?

ktf:
does that work?

ktf:
Fine with me if that simplifies the recipe. Can you do a PR with what you mean?

vkucera:
> Fine with me if that simplifies the recipe. Can you do a PR with what you mean?

I will test and make a PR if it works.

PR #6218, opened 2026-04-23
[OCONF-910] Fix mesos.sh for RHEL10

Barthelemy:


mesos.sh
```diff
@@ -45,7 +46,8 @@ sed -i.bak -e's/c++11/c++20/' ../configure
     --with-re2=${RE2_ROOT} \
     --with-grpc=${GRPC_ROOT} \
     --with-glog=${GLOG_ROOT} \
-    --with-rapidjson=${RAPIDJSON_ROOT}
+    --with-rapidjson=${RAPIDJSON_ROOT} \
+    --with-zlib=/lib64
```
ktf: shouldn't you use `${ZLIB_ROOT}/lib`?

Barthelemy: indeed thanks

mesos.sh
```diff
@@ -33,6 +33,7 @@ export CPPFLAGS="-I${ABSEIL_ROOT}/include"
 export CFLAGS="-I${ABSEIL_ROOT}/include"
 
 rsync -av --delete --exclude="**/.git" $SOURCEDIR/ .
+sed -i 's/\[deflate, gzread, gzwrite, inflate\]/[gzread]/' configure.ac # fix the wrong syntax
```
knopers8, line 36: to be super-correct, we could update our mesos mirror, but given the planned deprecation, i think it's fine.

knopers8 (approved):

mconcas:
Hello @singiamtel @ktf !
Are the CI errors genuine ones or we can proceed with this?

Thanks and cheers!

PR #6220, opened 2026-04-23
Add RECC configuration for O2Physics

vkucera:


ktf:
thinking about it, the recc configuration should probably not be part of the distribution. since it might vary depending on the deployment, especially once we get also the scheduler in place.

o2physics.sh
```diff
@@ -13,9 +13,12 @@ build_requires:
   - CMake
   - ninja
   - alibuild-recipe-tools
+  - "RECC:(?!osx)"
```
ktf:
```suggestion
```

ktf:
looks like this actually breaks staging?

PR #6221, opened 2026-04-25
Improve o2checkcode

vkucera:
- Allow auto-detect of `.clang-tidy`. (Report if found.)
- Remove explicit list of Clang-Tidy checks.
  - Configuration is taken from `.clang-tidy` if found. (Requires https://github.com/AliceO2Group/AliceO2/pull/15321)
  - Additional check configuration from `O2_CHECKER_CHECKS` is appended via the `-checks` command line option. (Report the provided value.)
- Print enabled checks before running instead of all available.
- Don't turn warnings into errors. Report both.
- Fix unbound `ALIBUILD_BASE_HASH`.

FIXME: `git diff` can fail silently, leading to an early exit due to an empty list of changed files. This is currently happening in the alidist fullCI check.

EDIT: [Exit gracefully if git diff fails.](https://github.com/alisw/alidist/pull/6221/commits/896478d5e652b225d19006a458a2a5f1d0359259)

EDIT: [Report errors and warnings separately. Fail on errors only.](https://github.com/alisw/alidist/pull/6221/commits/2a891cc80e811e0e6f70c112be68e804c8da014c)

### DIFF
--- a/o2checkcode.sh
+++ b/o2checkcode.sh
@@ -25,7 +25,7 @@ O2_SRC=$(python3 -c 'import json, os; print(os.path.commonpath([x["file"] for x 
 # We have something to compare our working directory to (ALIBUILD_BASE_HASH). We check only the
 # changed files (including the untracked ones) if the list of relevant files that changed is up to
 # 50 entries long
-if [[ $ALIBUILD_BASE_HASH ]]; then
+if [[ -n "${ALIBUILD_BASE_HASH:-}" ]]; then
   pushd "$O2_SRC"
     [[ -d .git ]]
     ( git diff --name-only $ALIBUILD_BASE_HASH${ALIBUILD_HEAD_HASH:+...$ALIBUILD_HEAD_HASH} || true ; git ls-files --others --exclude-standard ) | ( grep -E '\.cxx$|\.h$' || true ) | sort -u > $BUILDDIR/changed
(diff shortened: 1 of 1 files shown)


o2checkcode.sh
```diff
@@ -46,44 +46,24 @@ fi
 ThinCompilationsDatabase.py -exclude-files '(?:.*G\_\_.*\.cxx|.*\.pb.cc|.*\_amalgamated\..*)' ${O2_CHECKCODE_CHANGEDFILES:+-use-files ${O2_CHECKCODE_CHANGEDFILES}}
 cp thinned_compile_commands.json compile_commands.json
 
-# List of explicitely enabled C++ checks (make sure they are all green)
-CHECKS="${O2_CHECKER_CHECKS:--*\
-,modernize-avoid-bind\
```
sawenzel, line 51: where is the default list of O2_CHECKER_CHECKS now maintained? I think such default list would be useful.

vkucera: See the description.

sawenzel: My question is maybe more, why is it a good thing to delete the default list of checks from the recipe?
Initially the recipe was meant to be fully configured for ALICE and just typing `aliBuild build o2checkcode` should be enough. Now, it seems, you need to inject more information from the outside.

vkucera: Typing `aliBuild build o2checkcode` is still enough, plus the configuration is stored at a standard place which makes it reusable and modifiable independently of `o2checkcode`, alidist and aliBuild.

PR #6222, opened 2026-05-01
Copy full system tests logs to the artifacts folder

ktf:


ktf:
Seems to work beatifully after https://github.com/alisw/ali-bot/pull/1685 was merged. Now all the logs from the full system check are available even if the build is successful.

PR #6240, opened 2026-06-12
Fix FastJet and LHAPDF path forwarding and Pythia8 interface

jackal1-66:
FastJet and LHAPDF loading was broken because the /local/DailyBuild paths were set before in CVMFS. This patch fixes them by replacing the configuration on-the-fly while loading the modules, creating the ~/.mg5 folder and the patched version inside of it (unless another file is already available). Pythia8 interface build has been fixed by removing an unnecessary space making the installation crash.


jackal1-66:
macOS unrelated, rest is green. PR can now be merged @ktf

PR #6246, opened 2026-06-17
Update jalien-root.sh

costing:
Avoid including unneeded headers in the public interface


zensanp (approved):

ktf:
Thanks for the quick reaction. I am running on Hyperloop now and I will let you know the effect.

ktf:
Tested on hyperloop, works. It does not (yet) have any effect to the memory usage of the JAliEn-ROOT plugin. Merging as incremental update.

PR #6252, opened 2026-06-19
Create default dummy conf file

jackal1-66:
madgraph/__init__.py checks for the existence of the mg5_configuration.txt file in the MADGRAPH_ROOT folder when the executable is launched. If this is not found it creates a default one, however CVMFS is a read-only file system, so mg5_aMC was crashing. This PR creates the dummy file at build time bypassing the problem.


jackal1-66:
@ktf the failures are unrelated to the PR, this can be merged. @vkucera what's exactly the problem with the failing CIs?

vkucera:
> @ktf the failures are unrelated to the PR, this can be merged. @vkucera what's exactly the problem with the failing CIs?

@jackal1-66 Unrelated. These two checks always fail at the moment.

PR #6257, opened 2026-06-23
Fix binary blob in full system test output

ktf:


ktf:
Currently broken. Merging.

@vkucera

PR #6262, opened 2026-06-24
o2-full-system-test: Compress perf script output

vkucera:


o2-full-system-test.sh
```diff
@@ -64,7 +61,8 @@ find $BUILDDIR/full-system-test-sim -name '*.log' -exec cp {} $BUILDDIR/../artif
 if [[ -n "$USE_PERF" && -f perf.data ]]; then
   echo "O2-full-system-test: generating perf text report (perf script -i perf.data)"
   perf script -i perf.data > perf.script.txt 2>/dev/null || echo "O2-full-system-test: 'perf script' failed" >&2
-  cp perf.script.txt $BUILDDIR/../artifacts/ 2>/dev/null || true
+  mkdir -p $BUILDDIR/../artifacts/perf
+  tar cfz $BUILDDIR/../artifacts/perf/perf.script.tgz perf.script.txt 2>/dev/null || true
```
ktf: why tarring a single file?

ktf (changes requested):

o2-full-system-test.sh
```diff
@@ -65,6 +65,7 @@ if [[ -n "$USE_PERF" && -f perf.data ]]; then
   echo "O2-full-system-test: generating perf text report (perf script -i perf.data)"
   perf script -i perf.data > perf.script.txt 2>/dev/null || echo "O2-full-system-test: 'perf script' failed" >&2
   cp perf.script.txt $BUILDDIR/../artifacts/ 2>/dev/null || true
+  gzip $BUILDDIR/../artifacts/perf.script.txt
```
ktf: can you compress it directly while copying it? Something like `cat perf.script.txt | gzip -c > $BUILDDIR/../artifacts/perf.script.txt.gz`.

ktf: actually, even better, we can gzip it directly into the destination by adding the pipe in the `perf script` line.

ktf (commented):
see inline comments

o2-full-system-test.sh
vkucera: Done

o2-full-system-test.sh
```diff
@@ -64,7 +64,7 @@ find $BUILDDIR/full-system-test-sim -name '*.log' -exec cp {} $BUILDDIR/../artif
 if [[ -n "$USE_PERF" && -f perf.data ]]; then
   echo "O2-full-system-test: generating perf text report (perf script -i perf.data)"
   perf script -i perf.data > perf.script.txt 2>/dev/null || echo "O2-full-system-test: 'perf script' failed" >&2
```
ktf: I guess you can simply do it here, no?

ktf (commented):
see the other inline comment. just do the gzipping inline.

ktf (dismissed):

o2-full-system-test.sh
vkucera: Done

PR #6263, opened 2026-06-24
KFParticle bump to alice10

ktf:


ktf:
Simply drops unneeded libraries. Tested on hyperloop. Merging.

PR #6264, opened 2026-06-29
Add recipe for O2 GPU CI tests

fweig:
Add new recipes for O2 GPU CI tests.

Structure follows the full CI recipe:
- Added a meta recipe for GPU tests
- Added a test recipe that runs `o2-gpu-standalone-benchmark`

Test script currently only runs `o2-gpu-standalone-benchmark --noEvents` for available architectures to test that GPUs are present in container.

### DIFF
--- a/o2-gpu-deterministic-test.sh
+++ b/o2-gpu-deterministic-test.sh
@@ -70,30 +70,7 @@ cmake -DCMAKE_INSTALL_PREFIX=../install "$O2_SOURCEDIR/GPU/GPUTracking/Standalon
 cmake --build . --target install -- ${JOBS:+-j $JOBS}
 
 for BACKEND in "${GPU_BACKENDS[@]}"; do
-  case ${BACKEND^^} in
-    CUDA)
-      LIB=../install/libGPUTrackingCUDA.so
-      GPU_TYPE=CUDA
-      ;;
-    HIP|ROCM)
-      LIB=../install/libGPUTrackingHIP.so
-      GPU_TYPE=HIP
-      ;;
-    OCL|OPENCL)
-      LIB=../install/libGPUTrackingOCL.so
-      GPU_TYPE=OCL
-      ;;
-    *)
-      echo "O2-GPU-deterministic-test: unsupported backend '$BACKEND'. Use CUDA, HIP, or OCL." >&2
-      exit 1
-      ;;
-  esac
-
-  if [[ ! -e $LIB ]]; then
-    echo "O2-GPU-deterministic-test: missing $LIB" >&2
-    exit 1
-  fi
-  ../install/ca --noEvents -g --gpuType "$GPU_TYPE"
+  ../install/ca --noEvents -g --gpuType "${BACKEND^^}"
 done
 
 popd
--- a/o2-gpu-deterministic-test.sh
+++ b/o2-gpu-deterministic-test.sh
@@ -1,0 +1,120 @@
+package: O2-GPU-deterministic-test
+version: "1.0"
+requires:
+  - Vc
+  - boost
+  - fmt
+  - ms_gsl
+  - TBB
+  - ROOT
+  - ONNXRuntime
+  - GLFW
+  - gpu-system
+build_requires:
+  - CMake
+  - Clang
+  - ninja
+license: GPL-3.0
+force_rebuild: true
+---
+#!/bin/bash -e
+
+if [[ -n ${GPU_SYSTEM_ROOT:-} && -f $GPU_SYSTEM_ROOT/etc/gpu-features-available.sh ]]; then
+  source $GPU_SYSTEM_ROOT/etc/gpu-features-available.sh
+fi
+
+if [[ -n ${O2GPUCI_BACKENDS:-} ]]; then
+  read -r -a GPU_BACKENDS <<< "${O2GPUCI_BACKENDS//,/ }"
+else
+  GPU_BACKENDS=()
+  [[ ${O2_GPU_CUDA_AVAILABLE:-0} == 1 ]] && GPU_BACKENDS+=(CUDA)
+  [[ ${O2_GPU_ROCM_AVAILABLE:-0} == 1 ]] && GPU_BACKENDS+=(HIP)
+fi
+
+if [[ ${#GPU_BACKENDS[@]} == 0 ]]; then
+  echo "O2-GPU-deterministic-test: no GPU backend selected or detected." >&2
+  echo "Set O2GPUCI_BACKENDS='CUDA,HIP' in CI to require both production GPU backends." >&2
+  exit 1
+fi
+
+# HACK to find O2 sources without depending on O2 as a dependency (and potentially building all of it as a consequence)
+O2_SOURCEDIR=${O2GPUCI_O2_SOURCEDIR:-}
+for SOURCE_CANDIDATE in "$WORK_DIR/../O2" "$ALIBUILD_CONFIG_DIR/../AliceO2"; do
+  if [[ -z $O2_SOURCEDIR && -f $SOURCE_CANDIDATE/GPU/GPUTracking/Standalone/CMakeLists.txt ]]; then
+    O2_SOURCEDIR=$SOURCE_CANDIDATE
+  fi
+done
+if [[ -z $O2_SOURCEDIR && -d $WORK_DIR/SOURCES/O2 ]]; then
+  O2_SOURCEDIR=$(find "$WORK_DIR/SOURCES/O2" -mindepth 2 -maxdepth 2 -type d \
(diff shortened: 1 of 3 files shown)


o2-gpu-test.sh
```diff
@@ -0,0 +1,51 @@
...
+
+rm -Rf $BUILDDIR/gpu-test
+mkdir $BUILDDIR/gpu-test
+pushd $BUILDDIR/gpu-test
+
+if [[ -n ${O2GPUCI_BACKENDS:-} ]]; then
+  read -r -a GPU_BACKENDS <<< "${O2GPUCI_BACKENDS//,/ }"
+else
+  GPU_BACKENDS=()
+  [[ ${O2_GPU_CUDA_AVAILABLE:-0} == 1 ]] && GPU_BACKENDS+=(CUDA)
```
davidrohr, line 25: We could add also OpenCL here, at least without processing events, but to check that it can start it up.

o2-gpu-test.sh
```diff
@@ -0,0 +1,51 @@
...
+fi
+
+if [[ ${#GPU_BACKENDS[@]} == 0 ]]; then
+  echo "O2-GPU-test: no GPU backend selected or detected." >&2
+  exit 1
+fi
+
+for BACKEND in "${GPU_BACKENDS[@]}"; do
+  o2-gpu-standalone-benchmark --noEvents -g --gpuType "${BACKEND^^}"
+done
```
davidrohr, line 36: I was thinking a bit. We can use the o2-gpu-standalone-benchmark for performance tests.

But we also need a build with deterministic mode to compare CPU and GPU output.
I think to get that, we should:
- Use https://github.com/alisw/alidist/blob/b16586e8f1c2e81b1e2c5515b9c71bc125b712ed/o2.sh#L287
- Change that to `$ALIBUILD_O2_FORCE_GPU == "ci"` (The ci string did not exist, when we set that up).
- Add a third CI build with DETERMINISTIC mode enabled to `GPU/GPUTracking/Standalone/cmake/build.sh` if GPUCA_STANDALONE_CI=1
- Change o2.sh and / or build.sh to actually install the standalone builds. Or at least if we add a determinisstic build as last build, we need to install at least that one,  don't care for the others for now., to $O2_ROOT/share/GPU/standalone_builds/deterministic_build 
- The standalone builds use RPATH for the GPU libraries: https://github.com/AliceO2Group/AliceO2/blob/cd36608b9783132c889097648b6f2ad44afb5416/GPU/GPUTracking/Standalone/CMakeLists.txt#L21
  - This needs to be relocated, we probably need to talk to Giulio how to do that, but we already do it for the normal installation anyway. The point is that this RPATH is needed, to avoid that standalone libraries and O2 libraries from LD_LIBRARY_PATH can collide, and also in case we want to create multiple standalone builds later. So I think we should keep it.

davidrohr (commented):
Nice, convenient that this is so straight forward. So for the actual tests, I would like to test at least 2 things:

- Run the o2-gou-standalone-benchmark on a Pb-Pb reference dataset on CPU and GPU, and compare the processing time to a reference time, and give a warning if we get 5% slower than the reference.
- Run the standalone compilation in deterministic mode on CPU, GPU, and GPU with RTC on a reference data set (can be smaller than the PbPb data set, that we use for performance), and check that the dump files are identical. (In the long run, we should add a feature that we do not write huge dump files but just create an SHA512 hash of the file on the fly.

With the latest version, you can use `./ca --recreateTrivialCalibObjects` to create default calibration objects for the current software version. I.e. we need only the event and settings dump file from the data, and we can create the rest on the fly.

2 detailed comments inline.

o2-gpu-test.sh
fweig: Do we need PoCL for that? Or is the OpenCL runtime shipped with ROCm or CUDA enough to start the binary?

o2-gpu-test.sh
fweig: I've added a new recipe that builds the standalone binary in deterministic mode. It's kind of hacky because I'm compiling the binary in the test script and I have to guess the location of the O2 sources. But this avoids the RPATH issue, because it has it's own installation folder.

o2-gpu-test.sh
davidrohr: If we don't compile for a device, any 3.x OpenCL runtime should be good enough.

o2-gpu-test.sh
davidrohr: > I've added a new recipe that builds the standalone binary in deterministic mode. It's kind of hacky because I'm compiling the binary in the test script and I have to guess the location of the O2 sources. But this avoids the RPATH issue, because it has it's own installation folder.

But this will be faulure-prone. From where do you take the O2 sources? From the O2 build? But it is not guaranteed that the O2 build happens before, it can for instance download a precompiled O2 tarball.
I think we should do it properly and use patchelf to fix the RPATH.

davidrohr: I actually thought about this again. Because the standalone test builds are only done in the CI, they will not be done for the tarballs we create, so actually O2 tarballs will not contain the standalone builds, even if we install them to the share folder in o2.sh.
Thus indeed we need a way to access the correct O2 sources. Perhaps, what you have done is the easiest solution for now. We should discuss with @ktf when he is back.

o2gpuci.sh
```diff
@@ -0,0 +1,39 @@
...
+  - O2-GPU-deterministic-test:(.*x86-64)
+license: GPL-3.0
+valid_defaults:
+  - o2
+  - o2-epn
+  - ali
+---
+#!/bin/bash -ex
+
+MODULEFILE_DEPS=
```
ktf: alibuild-generate-module should be able do generate the file just fine.

o2-gpu-test.sh
ktf: As discussed, let's assume the sources are there for now. We will have some way of depending on sources at some point soon and then we can change.

o2-gpu-deterministic-test.sh
```diff
@@ -0,0 +1,120 @@
...
+    exit 1
+  fi
+  ../install/ca --noEvents -g --gpuType "$GPU_TYPE"
+done
+
+popd
+rm -Rf "$BUILDDIR/gpu-standalone-test"
+
+# Dummy modulefile
+MODULEFILE_DEPS=
```
ktf: again, I guess you can just use alibuild-generate-module, no?

o2-gpu-deterministic-test.sh
```diff
@@ -0,0 +1,120 @@
...
+fi
+
+rm -Rf "$BUILDDIR/gpu-standalone-test"
+mkdir -p "$BUILDDIR/gpu-standalone-test/build" "$BUILDDIR/gpu-standalone-test/install"
+pushd "$BUILDDIR/gpu-standalone-test/build"
+
+cp "$O2_SOURCEDIR/GPU/GPUTracking/Standalone/cmake/config.cmake" .
+cat >> config.cmake <<'EoF'
+set(GPUCA_DETERMINISTIC_MODE GPU)
+set(GPUCA_BUILD_EVENT_DISPLAY 0)
```
davidrohr, line 64: For the NVIDIA Blackwell, you should also set `set(GPUCA_DETERMINISTIC_NO_FTZ 1)`

o2-gpu-deterministic-test.sh
```diff
@@ -0,0 +1,120 @@
...
+EoF
+
+cmake -DCMAKE_INSTALL_PREFIX=../install "$O2_SOURCEDIR/GPU/GPUTracking/Standalone"
+cmake --build . --target install -- ${JOBS:+-j $JOBS}
+
+for BACKEND in "${GPU_BACKENDS[@]}"; do
+  case ${BACKEND^^} in
+    CUDA)
+      LIB=../install/libGPUTrackingCUDA.so
```
davidrohr: I think you do not need to set the LIB variable, but you can derive that below on-the-fly from GPU_TYPE

davidrohr (commented):
Looks good in general, just 2 more minor comments from my side.
Please fix all the comments, then we can merge it.

o2-gpu-test.sh
fweig: I tried running with `--noEvents` with both OpenCL runtimes that we have in the container, but both raise an error. Didn't investigate further. I would postpone this for now, until we have the CI running.

o2gpuci.sh
fweig: Fixed. Using `alibuild-generate-module` for both recipes now.

o2-gpu-deterministic-test.sh
fweig: Fixed. Added the flag.

o2-gpu-deterministic-test.sh
fweig: Leftover that I forgot to cleanup. I've removed the variable.

PR #6266, opened 2026-06-30
[ORT] TensorRT build working on pdp-dev00 (RTX5080)

ChSonnabend:


ChSonnabend:
@ktf do you think it makes sense to have this feature / merge this PR?

onnxruntime.sh
```diff
@@ -106,7 +106,7 @@ cmake "cmake"
       -DCMAKE_BUILD_TYPE=Release                                                                            \
       -DCMAKE_INSTALL_LIBDIR=lib                                                                            \
       -DPython_EXECUTABLE="$(python3 -c 'import sys; print(sys.executable)')"                               \
-      -DFETCHCONTENT_FULLY_DISCONNECTED=ON                                                                  \
+      -DFETCHCONTENT_FULLY_DISCONNECTED=OFF                                                                 \
```
davidrohr: Generally I am in favor of getting the TensorRT build fixed, but I think we do not want to disable FETCHCONTENT_FULLY_DISCONNECTED.
Why is that needed?

ChSonnabend: Sorry forgot to adjust the source dirs. Now it should work without the remote fetching.

davidrohr:
Fine to merge for me now. However, the CI seems to be generally broken. I'll wait for the CI to become fixed before merging, to make sure we don't break anything accidentally.

PR #6270, opened 2026-07-07
Add CODEOWNERS for ACTS team

njacazio:


njacazio:
Ciao @ktf we would like to review the ACTS bumps since this is transparent to the rest

PR #6272, opened 2026-07-08
ACTSO2: Restore source URL to SSH in actso2.sh

njacazio:
Updated the source URL back from HTTPS to SSH for actso2, reverting https://github.com/alisw/alidist/pull/6205
The init via ssh is seamless if the ssh key is configured in gitlab


ktf:
why????

PR #6274, opened 2026-07-12
Fix MadGraph: Pythia8 shower chain and conf relocation

jackal1-66:
Multiple fixes applied: 
- madgraph.sh &rarr;  1) fix to HepMC2 path inside Pythia8 MakeFile.inc; 2) No more dummy configurations and runtime patching. Everything is done using the relative ./, which is understood by MG5.
- lhapdf.sh &rarr; I noticed `/local/workspace/etc` was set in the symlink which was clearly not right and it was causing `Failed to access python version of LHAPDF`. While digging I saw that the recipe created lib/python with `find $PWD ... ln -nfs {} python`, so with an absolute symlink frozen to the build-time INSTALLROOT. The applied fix creates a relative symlink (python -> python3.10), valid at any install prefix (tested locally and with an alice-docker fresh install).
- pythia.sh &rarr; Pythia8 is now built with zlib. Before gzipped hepmc files could not be read, making the MG5 Pythia8 interface hang indefinitely.  
@ktf let me know what you think... If we're caching by node on the GRID this solution should also work as the configuration will be the same for all the jobs in a node if they can access the /tmp folder, but please correct me if I'm wrong.

### DIFF
--- a/lhapdf.sh
+++ b/lhapdf.sh
@@ -45,9 +45,9 @@ pushd "$INSTALLROOT"
   elif [[ -d lib && ! -d lib64 ]]; then
     ln -nfs lib lib64
   fi
-  # Uniform Python library path
+  # Uniform Python library path (relative symlink so it survives relocation)
   pushd lib
-    find $PWD -name "python3*" -exec ln -nfs {} python \;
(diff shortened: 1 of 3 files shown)


jackal1-66:
@ktf this should now work and be compatible with the GRID nodes as well, could you have a look?

PR #6279, opened 2026-07-22
Software packages for CAD support

sawenzel:
Make OpenCascade and it's python binding PythonOCC part of the ALICE software stack, for instance used
for rapid prototyping and simulation of detector geometries.

For now, it is not part of the O2sim meta-package and not automaticically published on CVMFS. But with this recipe, users can conveniently install necessary software for local development.

### DIFF
--- a/occt.sh
+++ b/occt.sh
@@ -1,0 +1,23 @@
+package: OCCT
+version: "v7.9.3"
+tag: V7_9_3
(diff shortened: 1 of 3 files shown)


sawenzel:
@njacazio

ktf:
Please make sure to add the new files to the CODEOWNERS.

PR #6282, opened 2026-07-24
Add keyring for build verification

ktf:
In the future, we will use this to guarantee that tarballs were actually
produced with the declared recipe.


ktf:
Not tested.

PR #6284, opened 2026-07-31
Add TGeo2VecGeom recipe and wire it into O2 as an optional VecGeom ma…

sawenzel:
…terial-scan backend

Needed to use VecGeom for navigation in O2

### DIFF
--- a/o2.sh
+++ b/o2.sh
@@ -24,6 +24,7 @@ requires:
   - libjalienO2
   - cgal
   - "VecGeom:(?!osx.*)"
+  - "TGeo2VecGeom:(?!osx.*)"
   - FFTW3
   - ONNXRuntime
   - nlohmann_json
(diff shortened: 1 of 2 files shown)


mhemmer-cern:
New PRs in O2Physics now see this error:
```
CMake Error in Detectors/Base/CMakeLists.txt:
  Imported target "TGeo2VecGeom::TGeo2VecGeom" includes non-existent path

    "/local/workspace/DailyBuilds/DailyO2Physics-slc9/daily-tags.qD5m7JNmUl/slc9_x86-64/ROOT/v6-36-10-alice2-2/include"

  in its INTERFACE_INCLUDE_DIRECTORIES.  Possible reasons include:

  * The path was deleted, renamed, or moved to another location.

  * An install or uninstall procedure did not complete successfully.
```

There is also an issue with today's tag on hyper loop that might be connected to this PR. Please have a look:
https://its.cern.ch/jira/browse/O2-7119

ktf:
@sawenzel This breaks the CI because its CMake config file caches the location of the relocated ROOT which might or might not be correct when rebuilding O2. We had similar problems already in VGM or GEANT4. Can it be fixed or reverted?

PR #6285, opened 2026-08-03
Temporarily take out TpcFecUtils to avoid CI problem

sawenzel:
CI currently down due to access problems with TpcFecUtils. This should bring CI back to working and unblock many PRs.


vkucera (approved):

vkucera:
So, according to @ktf , this package is used in the online reconstruction and O2Suite is supposed to test everything. So it's probably better to keep it.

PR #6286, opened 2026-08-05
TGeo2VecGeom: fix CMake relocation problem, avoid hardcoded libdir

sawenzel:
Bumps TGeo2VecGeom to v0.1.1 (https://gitlab.cern.ch/VecGeom/tgeo2vecgeom/-/merge_requests/3), which stops baking the build machine's ROOT/VecGeom include paths into `TGeo2VecGeomTargets.cmake`. aliBuild only relocates a package's own INSTALLROOT, so O2 builds taking TGeo2VecGeom from the build cache failed with `includes non-existent path .../ROOT/v6-36-10-alice2-2/include`.

Recipe side: a guard against the same class of bug, `CMAKE_INSTALL_LIBDIR=lib` (el9 installs to lib64, which `--lib` does not put on LD_LIBRARY_PATH), `CMAKE_PREFIX_PATH` instead of a hardcoded `lib64` path, and `--cmake` on the modulefile.

### DIFF
--- a/tgeo2vecgeom.sh
+++ b/tgeo2vecgeom.sh
@@ -1,6 +1,6 @@
 package: TGeo2VecGeom
 version: "%(tag_basename)s"
-tag: v0.1.0
+tag: v0.1.1
 source: https://gitlab.cern.ch/VecGeom/tgeo2vecgeom.git
 requires:
   - "GCC-Toolchain:(?!osx)"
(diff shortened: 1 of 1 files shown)


sawenzel:
Linux builds green. Merging since currently a blocker for all CI.

vkucera:
> Linux builds green. Merging since currently a blocker for all CI.

`build/O2/alidist-slc9-x86` has not finished but let's be optimistic. 🤞🏻

PR #6303, opened 2026-08-24
IgProf: add profiling modulefiles

ktf:
IgProfCPU or IgProfMemory can be added to your JDL to enable profiling


ktf:
Package not actually tested. Merging.

PR #6323, opened 2026-09-07
Keep old number of threads when compiling CUDA

ktf:


ktf:
Verified to fix the issue. Merging.

PR #6325, opened 2026-09-07
Fixing EPN build issues for cmake and gcc-toolchain

ChSonnabend:
I encountered build issues of the gcc toolchain and CMake when building on epn000. This fixed it so far


ChSonnabend:
@ktf does this seem reasonable to you?

ktf:
No. :) There is `ALIBUILD_BUILD_WORK_DIR` exactly for broken filesystems... ;-) Check c84d38658376. If it does not work, we should fix that.

PR #6326, opened 2026-09-08
Encoding of gpu architecture using . instead of @

ChSonnabend:
Since 6eafd9bc the gpu-system version string encodes the GPU architecture as
`arch@<arch>@`, e.g.

    gpu-system/rocm_6.3.42134_arch@gfx906@_home_F5XXA5BPOJXWG3IK-miopen-local2

On hosts running Environment Modules 5 (e.g. the EPN nodes, Modules 5.3.0),
`advanced_version_spec` is enabled by default and `@` is parsed as a version
specifier. `alienv enter O2PDPSuite/...` then fails with

    ERROR: Unable to locate a modulefile for 'gpu-system/rocm_6.3.42134_arch@gfx906@_home_F5XXA5BPOJXWG3IK-miopen-local2'

and every dependent module (ONNXRuntime, O2, O2PDPSuite) fails to load.
Modules 4.x is not affected, which is why this went unnoticed.

This PR replaces the `@` delimiter with `.`, which Modules 5 accepts, already
occurs in version strings, and cannot clash with the `#` / `_` used inside the
architecture list. The parser that reconstructs `O2_GPU_*_AVAILABLE_ARCH` in
`gpu-features-available.sh` is updated accordingly. The `_home_` regexes used
`[^@]*` as terminator; they now stop at `-` or `_`, which is exact for base32.
`~`, `+` and `=` were also tested and are rejected by Modules 5.

Note: this changes the gpu-system version string, so gpu-system and its
dependents get rebuilt.

Tested on epn000 (Modules 5.3.0): the new module name loads, and the regexes
reproduce `gfx906`, `86;90-real` and `/opt/rocm` from sample version strings.

### DIFF
--- a/gpu-system.sh
+++ b/gpu-system.sh
@@ -42,8 +42,8 @@ prefer_system_check: |
 
     if [[ ${ALIBUILD_O2_FORCE_GPU} == "build" ]]; then
       GPU_FEATURES=build-from-alidist
-      [[ -n $ALIBUILD_O2_FORCE_GPU_CUDA_ARCH ]] && add_feature _ cudaarch@$(sed -e "s/;/#/g" -e "s/-/_/g" <<< "${ALIBUILD_O2_FORCE_GPU_CUDA_ARCH}")@
-      [[ -n $ALIBUILD_O2_FORCE_GPU_HIP_ARCH ]] && add_feature _ rocmarch@$(sed -e "s/;/#/g" -e "s/-/_/g" <<< "${ALIBUILD_O2_FORCE_GPU_HIP_ARCH}")@
+      [[ -n $ALIBUILD_O2_FORCE_GPU_CUDA_ARCH ]] && add_feature _ cudaarch.$(sed -e "s/;/#/g" -e "s/-/_/g" <<< "${ALIBUILD_O2_FORCE_GPU_CUDA_ARCH}").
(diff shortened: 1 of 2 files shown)


ChSonnabend:
@davidrohr did you ever face this issue? Could this interfere with some reading of the string in O2?

ktf:
@autumn-mck and @jgrosseo did on hyperloop (and the rest of the collaboration). I am not sure if having the dots is a good idea though.

ChSonnabend:
> @autumn-mck and @jgrosseo did on hyperloop (and the rest of the collaboration). I am not sure if having the dots is a good idea though.

Which delimeter would you prefer? Maybe three underscores: ___?

ktf:
Whatever works with the RPM generating scripts.

davidrohr:
I think we should get rid of encoding this stuff in the version. Wr can just use _ in the version, but for extracting it, we use the tracked env featzre as already discussed. I can give that a try tomorrow.

davidrohr:
> @davidrohr did you ever face this issue? Could this interfere with some reading of the string in O2?

And yes, of course this can collide with parsing of the string, in case something in the architecture version contains a . :).

davidrohr:
Tentative solution inm #6327

PR #6328, opened 2026-09-10
ORT CI tests

ChSonnabend:
This PR provides ONNXRuntime Ci tests for CPU and GPU execution with different execution providers. Additionally a local runner script is provided for local execution


ktf (changes requested):
It looks like most of this belongs to O2. Please move it there.

ChSonnabend:
> It looks like most of this belongs to O2. Please move it there.

The point of this PR is to have a test to see when a new ORT version is used whether it still builds AND runs. Otherwise these tests have to be redone manually all the time. Alidist is the place where the version is changed, therefore I put it here.

ktf:
O2 is tested for each alidist PR. Just put it there, please.

Ciao,
Giulio
________________________________

PR #6329, opened 2026-09-14
ORT CI tests

ChSonnabend:
This relates to https://github.com/AliceO2Group/AliceO2/pull/15792


ChSonnabend:
@ktf I think this script is still required to pickup the execution providers for the O2 execution. What do you think?

o2fullci.sh
```diff
@@ -7,6 +7,7 @@ requires:
   - O2-full-system-test
   - O2-sim-challenge-test
   - O2-RTC-test:(.*x86-64)
+  - O2GPUCI:(.*x86-64)
```
davidrohr: O2FullC shall not run the GPU CI!
The GPU CI is supposed to have a GPU available to actually run code, which is not available on all FullCI builders!

o2-onnxruntime-inference-test.sh
```diff
@@ -0,0 +1,92 @@
+package: O2-ONNXRuntime-inference-test
+version: "1.0"
+requires:
+  - ONNXRuntime
+  - gpu-system
```
davidrohr: ONNXRuntime already requires gpu-system. No need to require gpu-system explicitly.

o2-onnxruntime-inference-test.sh
```diff
@@ -0,0 +1,92 @@
...
+  [[ ${O2_GPU_ROCM_AVAILABLE:-0} == 1 ]] && GPU_BACKENDS+=(HIP)
+fi
+
+if [[ ${#GPU_BACKENDS[@]} == 0 ]]; then
+  echo "O2-ONNXRuntime-inference-test: no GPU backend selected or detected." >&2
+  echo "Set O2GPUCI_BACKENDS='CUDA,HIP' in CI to require both production GPU backends." >&2
+  exit 1
+fi
+
+# HACK to find O2 sources without depending on O2 as a dependency (and potentially building all of it as a consequence)
```
davidrohr: Why do we do this hack here, in a separate alidist recipe. Couldn*t this test just be part of o2.sh, if it needs O2?

davidrohr (changes requested):

PR #6333, opened 2026-09-16
Add dittomc

njacazio:
@sawenzel


sawenzel:
Can you make if part of O2Generators? Then it will also be CI checked here.

PR #6336, opened 2026-09-17
Remove Hepmc3 abs path for ROOT libs

njacazio:
When loading ACTS without compiling hepmc3 as dev package one gets a failure:

```
CMake Error in Examples/Io/HepMC3/CMakeLists.txt: Imported target "HepMC3::rootIO" includes non-existent path "/local/workspace/DailyBuilds/DailyO2Physics-slc9/daily-tags.UZZI2biIZB/slc9_x86-64/ROOT/v6-36-10-alice4-1/include" in its INTERFACE_INCLUDE_DIRECTORIES. Possible reasons include: 
* The path was deleted, renamed, or moved to another location. * An install or uninstall procedure did not complete successfully. 
* The installation package was faulty and references files it does not provide. -- Generating done (0.5s) 
CMake Generate step failed. Build files cannot be regenerated correctly
```


jackal1-66:
This should not affect sherpa, rivet and Photos. So it is good for me

